PR #24, opened 2026-04-21
fix(cli): default plur init to project-level config (Issue #19)

plur9:
## Summary

Makes `plur init` default to creating `.claude/settings.json` in the current directory instead of silently falling back to `~/.claude/settings.json`. This is the actual code change owed to #19 — Issue 1 in that report.

## Why

Issue #19 reported that running `plur init` in a fresh repo (no `.claude/` dir present) would silently register hooks at the user level. That breaks multi-project setups, because per-project hook config is what lets each repo get PLUR reliably without being overwritten by unrelated global edits.

Earlier status comments on #19 repeatedly claimed this fix had shipped — it hadn't. The [honest-status comment](https://github.com/plur-ai/plur/issues/19#issuecomment-latest) on 2026-04-21 acknowledged the automation loop and committed to flipping the default. This PR does that.

## Change

In `findSettingsPath` (`packages/cli/src/commands/init.ts`):

- `--global` → `~/.claude/settings.json` (unchanged)
- `--project` → project (unchanged, now explicit)
- `.claude/` dir exists → project (unchanged)
- **Default (no flag, no existing `.claude/`) → project** (previously: silently fell back to global)

Users who want the old behavior pass `--global`.

## Not in scope

- Issue 2 (`--domain` / `--scope`) — already shipped in v0.8.2
- Issue 3 (batch/workspace mode) — still open for future consideration

## Test plan

- [ ] CI green
- [ ] Manual: `plur init` in a fresh dir creates `./.claude/settings.json`, not `~/.claude/settings.json`
- [ ] Manual: `plur init --global` still targets `~/.claude/settings.json`
- [ ] Manual: `plur init` in a dir with existing `.claude/` still targets project (no regression)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Changelog
 
+## Unreleased
+
+### Multi-Project Setup Improvements
+
+- **Default to project-level config**: `plur init` now creates `.claude/settings.json` in the current directory by default, instead of falling back to `~/.claude/settings.json`. Better for multi-project setups. Users who want global config can use `--global` flag. (Fixes #19 Issue 1)
+
+### Packages
+
+- `@plur-ai/cli` — default to project-level config
+
 ## 0.8.2 (2026-04-09)
 
 ### Architecture Clarity & Multi-Project Scoping
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,13 +2,16 @@
 
 ## Unreleased
 
-### Multi-Project Setup Improvements
+### Multi-Project Setup Improvements (#19)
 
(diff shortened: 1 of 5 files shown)


plur9:
Pushed e9424ba to fix the 3 sync.test.ts CI failures:

- `pullRebase()` in `packages/core/src/sync.ts` hardcoded `main`, but `sync()` on init pushes to whichever branch `git init` created (can be `master` when CI has no `init.defaultBranch` configured). Push-master + pull-main mismatch is why `episodes.yaml` wasn't arriving after sync in the "pulls remote changes" and "handles concurrent changes" tests.
- The hardcoded `expect(status.branch).toBe('main')` assertion was relaxed to accept either `main` or `master`, so the suite doesn't depend on CI global config.

Latest CI run: sync.test.ts now passes (13/13 ✓). A separate test file (`sp2-failure-evolution.test.ts`) fails with "Engram not found: ENG-2026-0421-001" — this is **not related to this PR** (appears to be a flaky date-based ID collision in unrelated test code) and should be addressed in its own fix.

Issue #59, opened 2026-04-24
Publish-on-tag workflow missing: version bumps merge to main but require manual npm publish


plur9:
## Problem

Today's claw 0.9.10 bump (PR #58, merged 2026-04-24T07:03Z) shipped to `main` but npm `@plur-ai/claw@latest` is still 0.9.9 (published 2026-04-23T11:59Z). This gap is silent — no CI failure, no warning — and it blocked issue #51's reporter-facing test ask (`npx @plur-ai/claw doctor`), since 0.9.9 predates Slice C.

The only workflow under `.github/workflows/` is `ci.yml`. There is no publish workflow of any kind.

## Impact

Every merged version bump sits in a "shipped on main / not shipped to users" limbo until someone remembers to run `npm publish` manually. For a project whose `north_star` is MCP distribution, this is a structural distribution leak:

- External reporters can't verify fixes against published versions
- `distribution-check` cadence sees stale npm download data relative to actual feature velocity
- Release notes / changelog / tag surface all drift from the npm surface

## Proposal

Add `.github/workflows/publish.yml` that:

1. Triggers on `push` of a tag matching `@plur-ai/*@*` (or on published GitHub releases)
2. Runs the monorepo build + test (same matrix as ci.yml)
3. Uses `pnpm publish --filter <package> --access public` gated on `NPM_TOKEN` secret
4. Optionally: auto-creates a tag on `main` when a package `version` field changes, so the bump PR itself is the release trigger

Option 4 is the cleanest — it makes the merge-to-publish gap mechanically impossible rather than relying on memory.

## Scope

In: `@plur-ai/claw`, `@plur-ai/core`, `@plur-ai/mcp`, `@plur-ai/cli`. Out: website repo (separate deploy path).

## Success criteria

- Merging a PR that bumps `packages/<pkg>/package.json` `version` field results in that version being live on npm within ~5 minutes, with no manual step
- `npm view @plur-ai/claw version` equals the `version` field of `packages/claw/package.json` on `main` at all times (bounded by CI duration)

Filed from ceo distribution-check follow-up on #51.

plur9:
## Proposed fix — `publish.yml` (Option 4 from the issue)

[...]

PR #68, opened 2026-04-30
test(cli): regression guard against esbuild __require shim that bricked cli@0.9.2

plur9:
## Why

`@plur-ai/cli@0.9.2` shipped to npm with an esbuild `__require` shim that threw `Dynamic require of "os" is not supported` on every command (#59). It has been on `latest` for **8 days** and is still bricked at the time of this PR. Root cause was `require('os')` in `autoDiscoverStores()` (`packages/core/src/index.ts`) leaking into the ESM bundle. Fix landed as `584f11f`, but nothing prevented the next variant from shipping the same way.

## What

Adds `packages/cli/test/no-dynamic-require.test.ts` — walks each package's `dist/` and fails if any `.js` contains:
- `__require(...)` (the esbuild shim call site)
- `Dynamic require of` (the shim's error string)

One test per `core | cli | mcp | claw` dist directory. Skips silently if `dist/` doesn't exist (so it's a no-op on a fresh clone before `pnpm build`).

## Why this layer

The original bug was *source-level*, but the same shim can also be introduced by a transitive dep being bundled or by a future esbuild config tweak. Asserting on bundled output catches all source paths to the same failure mode.

Runs as part of `pnpm test` already, so it gates CI on every push/PR — no workflow changes (which my token can't push anyway, see #59).

## Verification

```
$ pnpm --filter @plur-ai/cli test no-dynamic-require
 Test Files  1 passed (1)
      Tests  4 passed (4)
```

## Scope / non-goals

- Does **not** unbrick the live `cli@0.9.2` on npm — that still needs the human-bridge republish documented in #59 and RELEASING.md.
- Does **not** address the broader publish-on-tag workflow (also in #59) — still blocked on `workflow` token scope.
- Does **not** check `hermes` (Python, separate pipeline).

This is the *defense in depth* that should have been in place before `cli@0.9.2` shipped — landing it means the next time someone introduces a CJS-style `require()` in an ESM-bundled path, CI fails before publish instead of after install.

— cto heartbeat 2026-04-30

### DIFF
--- a/packages/cli/test/no-dynamic-require.test.ts
+++ b/packages/cli/test/no-dynamic-require.test.ts
@@ -1,0 +1,36 @@
+import { describe, it, expect } from 'vitest'
+import { readFileSync, readdirSync, statSync, existsSync } from 'fs'
+import { join } from 'path'
+
+const PACKAGES_ROOT = join(__dirname, '..', '..')
+const PACKAGES = ['core', 'cli', 'mcp', 'claw']
+
+function walkJs(dir: string): string[] {
+  if (!existsSync(dir)) return []
+  const out: string[] = []
+  for (const entry of readdirSync(dir)) {
+    const full = join(dir, entry)
+    const s = statSync(full)
+    if (s.isDirectory()) out.push(...walkJs(full))
+    else if (entry.endsWith('.js')) out.push(full)
+  }
+  return out
+}
+
+describe('no dynamic require shim in built artifacts', () => {
+  for (const pkg of PACKAGES) {
+    const dist = join(PACKAGES_ROOT, pkg, 'dist')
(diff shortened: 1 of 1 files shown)


plur9:
Pushed 2aef209 — scoped the test to user-facing entrypoints (cli/mcp/claw), dropped core.

CI failure on the first push was diagnostic, not regressive: `core/dist/transformers.node-PH5YK5EA.js` and `core/dist/chunk-2ZDO52B4.js` use `__require` because `@xenova/transformers` loads ONNX `.node` bindings dynamically — that shim is intentional and unrelated to the cli@0.9.2 brick. The cli@0.9.2 bug was specifically that **cli's bundle inlined core's shim** by failing to keep `@plur-ai/core` external; the test now verifies that exact contract on the three packages a user actually installs.

Verified locally: 3 passed (cli/dist, mcp/dist, claw/dist all clean).

PR #94, opened 2026-05-07
fix(core): optimistic cache insert after remote append

### LINKED ISSUE #89: Write-then-read consistency: remote engram not visible until second recall
## Parent

Epic: #80

## Summary

After writing an engram to a remote store via `learn()`, the engram does not appear in the immediately following `recall()` or `recall_hybrid()`. It only becomes visible on the **second** recall call.

**Verified empirically** on 0.9.8 (2026-05-07): wrote `ENG-2026-05-07-002` to remote (store count incremented), then `recall_hybrid` with matching query returned 0 results for that engram.

## Root Cause

`_loadRemoteCached()` (`index.ts:262-274`) reads the `RemoteStore` cache **synchronously** and triggers a background refresh:

```typescript
private _loadRemoteCached(store: StoreEntry): Engram[] {
  // ...
  const cached = (driver as { cache: ... }).cache
  void driver.load().catch(...)  // background refresh
  return cached?.engrams ?? []   // returns stale data RIGHT NOW
}
```

After `learn()` calls `append()`:
1. `append()` invalidates cache (`this.cache = null`) ✅
2. Next `recall()` → `_loadRemoteCached()` sees `cache = null` → returns `[]` ❌
3. Background `load()` fires, fetches from server, populates cache
4. **Second** `recall()` → sees the fresh cache → engram appears ✅

The write and read paths share the same `RemoteStore` instance (via `_remoteStores` map with `${url}::${scope}` key), so the cache invalidation propagates. The problem is purely in the sync-read-then-async-refresh pattern.

## Proposed Fix: Optimistic Cache Insert

In `RemoteStore.append()`, after a successful POST, insert the engram into the local cache immediately:

```typescript
async append(engram: Engram): Promise<void> {
  // ... existing POST logic ...
  if (!r.ok) throw new Error(...)

  // Instead of just invalidating, optimistically insert
  if (this.cache) {
    this.cache.engrams.push(engram)
  } else {
    this.cache = { ts: Date.now(), engrams: [engram] }
  }
}
```

This is safe because:
- The POST succeeded — the server has the engram
- If the server transformed it (different ID, etc.), the next background refresh corrects the cache
- Worst case: a brief window where the cached engram has a client-generated ID but the server assigned a different one. For the pilot this is fine.

## Alternative Fixes

| Fix | Tradeoff |
|-----|----------|
| **Optimistic insert** (recommended) | Zero latency, slightly stale if server transforms |
| **Await-on-miss** in `_loadRemoteCached` | Correct but adds network latency to every recall after a write |
| **Document the behavior** | Free but bad UX |

## Test Cases

```
✓ learn to remote → immediate recall returns the engram
✓ learn to remote → immediate recall_hybrid returns the engram
✓ learn to remote → immediate inject includes the engram (if relevant)
✓ optimistic cache is corrected by next background refresh
```
---

crtahlin:
## Summary

Fixes #89 — after writing an engram to a remote store via `learn()`, the engram was invisible to the immediately following `recall()` due to a 1-call blind spot in the cache.

**Root cause:** `_loadRemoteCached()` reads the `RemoteStore` cache synchronously and refreshes in the background. After `append()` invalidated the cache (`this.cache = null`), the next read returned `[]` while the background `load()` was in-flight. Only the second recall after the write saw the data.

**Fix:** Instead of nulling the cache after a successful POST, optimistically insert the engram:
- Parse the server response for the canonical row (server-assigned ID, status)
- Fall back to the input engram if the response has no JSON body
- Append to existing cache (don't replace it)
- Next full `load()` after TTL expiry corrects any drift

## Changes

- `packages/core/src/store/remote-store.ts` — replace `this.cache = null` with optimistic insert (~20 lines)
- `packages/core/test/remote-routing.test.ts` — 4 new tests for the cache behavior
- `CLAUDE.md` — reference to remote store test plan

## Test plan

- [x] Engram is in cache immediately after successful append
- [x] Falls back to input engram if server returns no JSON body
- [x] Appends to existing cache without replacing previous entries
- [x] Failed append does not pollute cache
- [x] All 8 remote-routing tests pass
- [x] All 503 core tests pass (17 pre-existing SQLite failures unrelated)
- [ ] Manual E2E: `plur_learn` with remote scope → immediate `plur_recall` returns the engram (post-merge)

## Test plan reference

Scenario RD7 in `3-plur/1-tracks/engineering/remote-store-test-plan.md`

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -141,6 +141,22 @@ If your PR improves any of these, mention it in the PR description.
 | `packages/claw/src/assembler.ts` | Context assembly for OpenClaw |
 | `packages/claw/src/learner.ts` | Auto-extraction of learnings from conversation |
 
+## Remote Store Test Plan
+
+A comprehensive test scenario matrix lives at `~/Data/3-plur/1-tracks/engineering/remote-store-test-plan.md`. It maps 62 scenarios across write routing, read merging, config resilience, network edge cases, and ID consistency.
+
+**When working on remote store issues (#78-#93):**
+- Use the test plan as a reference for which scenarios to cover
+- After implementing a fix, update the scenario's **Status** column (e.g. `**Untested**` → `Verified 0.9.9`)
+- After adding new tests, update the **Summary** table counts in section 9
+- If you discover new scenarios during implementation, add them to the appropriate table
+- The test plan is the source of truth for what's tested and what's not — keep it current
+
+Tracked epics:
+- #78 — Test infrastructure (Level 2/3/4)
+- #79 — Write routing bugs (forget, feedback, pin, promote)
+- #80 — Write resilience (retry, consistency, guards)
+
 ## Datacore Space Context
 
 This project lives inside a Datacore space. Session lifecycle commands are available:
--- a/packages/core/src/store/remote-store.ts
+++ b/packages/core/src/store/remote-store.ts
@@ -121,8 +121,25 @@ export class RemoteStore implements EngramStore {
       const text = await r.text().catch(() => '')
       throw new Error(`Remote store append failed: ${r.status} ${text}`)
     }
-    // Invalidate cache so next load() picks up the new row
-    this.cache = null
+    // Optimistic cache insert: make the new engram visible to the next
+    // synchronous _loadRemoteCached() call immediately, instead of waiting
+    // for a background load() refresh. The next full load() after TTL
+    // expiry will replace this with the server-canonical version.
+    // See: https://github.com/plur-ai/plur/issues/89
+    let inserted: Engram = engram
+    try {
+      const row = await r.json() as any
+      if (row?.id) {
+        inserted = { id: row.id, scope: row.scope ?? engram.scope, status: row.status ?? 'active', ...(row.data ?? {}), ...(!row.data ? { statement: (engram as any).statement } : {}) } as unknown as Engram
+      }
+    } catch {
+      // Response may not be JSON (e.g. 201 with empty body) — use the input engram as-is
+    }
+    if (this.cache) {
+      this.cache.engrams.push(inserted)
+    } else {
+      this.cache = { ts: Date.now(), engrams: [inserted] }
+    }
   }
 
   /**
--- a/packages/core/test/remote-routing.test.ts
+++ b/packages/core/test/remote-routing.test.ts
@@ -188,3 +188,115 @@ describe('learn() — remote routing (issue #25)', () => {
     consoleErr.mockRestore()
   })
 })
+
+/**
+ * Issue #89 — after a successful remote append, the engram should be
+ * visible in the RemoteStore's cache immediately (optimistic insert),
+ * so that the next synchronous _loadRemoteCached() call sees it without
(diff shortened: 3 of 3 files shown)


plur9 (changes requested):
The fix correctly addresses the stale-cache-after-append symptom for the warm-cache path. But it introduces a regression on the cold-cache path that's worth fixing before we ship.

## Cold-cache regression

`packages/core/src/store/remote-store.ts:140-142`:
```ts
if (this.cache) {
  this.cache.engrams.push(inserted)
} else {
  this.cache = { ts: Date.now(), engrams: [inserted] }
}
```

When `this.cache === null` (no prior `load()` has run), the `else` branch creates a cache that:
- Has only the just-appended engram (1 of N)
- Has `ts: Date.now()` — so `load()` at line 54 (`now - this.cache.ts < this.ttlMs`) treats it as **fresh** for up to 60s

Concretely: a fresh `RemoteStore` instance where the user calls `learn()` first (no prior recall) will, for the next 60s, see only that one engram on `load()` — missing every other engram in the scope on the server. Before this PR, the cache stayed `null` and the next `load()` fetched everything.

## Suggested fix

In the cold-cache branch, mark the cache as stale so the next `load()` refetches:

```ts
} else {
  // No prior load() — don't pretend the cache is complete; a single
  // engram is not "all engrams in this scope". Mark stale so the next
  // load() refetches.
  this.cache = { ts: 0, engrams: [inserted] }
}
```

This still satisfies the issue #89 fix path (warm cache append is visible immediately) without lying about completeness on cold caches.

It would also be worth adding a test for this edge case — something like:

```ts
it('cold-cache append does not block a subsequent full load()', async () => {
  // mock 1: append → 201 with row
  // mock 2: load() → list with 3 engrams
  // append, then load() — expect 3 engrams returned, not 1
})
```

## Nits (non-blocking)

`remote-store.ts:133` — the response-shape mapping is hard to scan:
```ts
inserted = { id: row.id, scope: row.scope ?? engram.scope, status: row.status ?? 'active', ...(row.data ?? {}), ...(!row.data ? { statement: (engram as any).statement } : {}) } as unknown as Engram
```

Worth extracting to match the shape used in `load()` at `remote-store.ts:78-86`, e.g.:
```ts
const d = row.data ?? { statement: (engram as any).statement }
inserted = { id: row.id, scope: row.scope ?? engram.scope, status: row.status ?? 'active', ...d } as unknown as Engram
```

Also: PR is currently `MERGEABLE: CONFLICTING` against `main` — needs a rebase after the recent merges of #96/#97/#98.

miles-on-nightshift:
Closing as **superseded by main** (commit 4079109 — "fix(remote-store): optimistic cache insert on append (closes #89)", merged in batch with the remote-store epic on 2026-05-10).

Issue #89 is already closed; main now has the optimistic-insert path through the refactored `appendAndGetServerId()` helper.

**However:** the cold-cache regression flagged in this PR's review (cache pretends to be a fresh full load when it only contains the single just-appended engram) **still exists on main**. Filing as a follow-up issue so it gets tracked separately.

PR #26, opened 2026-04-22
fix(core): invalidate engram cache on write to close flaky CI race (#25)

plur9:
Closes #25 (if CI confirms the flake rate drops).

## Summary

`_loadCached` in `Plur` uses mtime-based invalidation. On CI tmpfs (ubuntu-latest runners) mtime resolution can be coarse enough that `stat()` before and after a write returns the same value — so the cache serves a pre-write snapshot and `getById` returns `undefined` for an engram `learn()` just wrote. That matches the observed failures in `sp2-failure-evolution.test.ts` and `sp2-tools.test.ts`.

## Change

Add a `_writeEngrams(path, engrams)` helper that calls `saveEngrams(…)` and then `this._engramCache.delete(path)`. Route all 12 primary/store write sites through it. Removes the filesystem as a source of cache freshness for writes we know about.

Pack writes in `_feedbackPack` keep `saveEngrams` directly — packs use a separate loader and aren't cached via `_engramCache`.

Also drops two now-redundant explicit `_engramCache.delete(storeInfo.path)` calls in `feedback()` and `forget()` that the helper subsumes.

## Why draft

Local repro was never reproduced ([issue thread](https://github.com/plur-ai/plur/issues/25#issuecomment-)). The only environment where the flake has been observed is GitHub Actions. Pushing this as draft so CI can run repeatedly against the change — the flake rate dropping (or staying zero over multiple reruns) is the real test.

## Verification plan

- [ ] CI passes on first run
- [ ] Re-run CI ~5× to check the flake is gone (or at least doesn't reappear on the two specific tests)
- [ ] If rerun-clean: promote from draft, merge, then close #25 as fixed
- [ ] If still flaky: revisit — may need to pair with a deeper fix (atomic-write fsync barrier, or generation-counter scheme from issue analysis)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #90, opened 2026-05-07
Guard against visibility:private with remote scope


crtahlin:
## Parent

Epic: #80

## Summary

`learn()` with `visibility: "private"` and a remote scope (e.g. `group:plur/plur-ai/engineering`) is contradictory — "private" means "only I see this", but remote means it goes to a shared server. Currently there's no guard; the engram is sent to the server with `visibility: private`, which has undefined semantics.

## Decision

`visibility: private` should only write to the local store. If a remote scope is specified alongside `visibility: private`, `learn()` should:
1. Log a warning: `[plur:learn] private engram not routed to remote (scope=...), writing locally`
2. Write to local store (with the original scope tag preserved for context)
3. Return the engram normally (no error)

## Implementation

In `learn()`, after resolving the remote driver (`index.ts:480`):

```typescript
const remoteDriver = this._resolveRemoteStoreForScope(scope)
if (remoteDriver && context?.visibility === 'private') {
  logger.warn(`[plur:learn] private engram not routed to remote (scope=${scope}), writing locally`)
  // Fall through to local write path
} else if (remoteDriver) {
  // ... existing remote routing
}
```

## Test Cases

```
✓ learn with private + remote scope → written locally, not sent to server
✓ learn with private + remote scope → warning logged
✓ learn with private + no scope → written locally (unchanged behavior)
✓ learn with public + remote scope → routed to server (unchanged behavior)
```
---

**Test plan maintenance:** After implementation, update the scenario status in `3-plur/1-tracks/engineering/remote-store-test-plan.md` — change the relevant Status column and section 9 summary counts. Add new scenarios if discovered.

Issue #84, opened 2026-05-07
forget() does not route to remote stores


crtahlin:
## Parent

Epic: #79

## Summary

`plur_forget` with a server-assigned ID (e.g. `ENG-2026-05-07-001`) returns `Engram not found` instead of retiring the engram on the remote enterprise server. The engram exists on the server but `forget()` only searches local stores.

**Verified broken** on 0.9.8 (2026-05-07).

## Reproduction

```
# 1. Write to remote
plur_learn statement="test" scope="group:plur/plur-ai/engineering"
# → returns ENG-2026-05-07-001 (server-assigned)

# 2. Try to forget it
plur_forget id="ENG-2026-05-07-001"
# → Error: Engram not found: ENG-2026-05-07-001
```

## Root Cause

`forget()` at `index.ts:950-996`:

```typescript
forget(id: string, reason?: string): void {
  // 1. Check primary local store
  const engrams = loadEngrams(this.paths.engrams)
  const engram = engrams.find(e => e.id === id)  // not found (it's remote)

  // 2. Check local project stores via _findEngramStore
  const storeInfo = this._findEngramStore(id)  // skips remote stores

  // 3. Never checks remote stores
  throw new Error(`Engram not found: ${id}`)
}
```

`_findEngramStore()` (line 318-346) explicitly skips remote stores: *"Remote stores are skipped here because remote IDs are not namespaced"*.

Meanwhile, `RemoteStore` already has `getById(id)` and `remove(id)` methods that work.

## Proposed Fix

After local store checks fail, iterate remote stores and check `getById()`:

```typescript
// After line 993, before the throw:
for (const entry of (this.config.stores ?? [])) {
  if (!entry.url) continue
  if (entry.readonly === true) continue
  const driver = this._getOrCreateRemoteDriver(entry)
  const found = await driver.getById(id)
  if (found) {
    const removed = await driver.remove(id)
    if (removed) {
      appendHistory(this.paths.root, {
        event: 'engram_retired',
        engram_id: id,
        timestamp: new Date().toISOString(),
        data: { reason: reason ?? null, routed_to: 'remote' },
      })
      return
    }
  }
}
```

## API Break

`forget()` is currently synchronous (`void` return). Adding `await` makes it async. This is a breaking change — see #TBD (sync→async migration).

**Short-term option:** Fire-and-forget pattern (like `learn()`), with `forgetAsync()` as the reliable alternative.

## Test Cases

```
[...]

Issue #25, opened 2026-04-22
Flaky CI: sp2-failure-evolution and sp2-tools intermittently fail with 'Engram not found'


plur9:
## Summary

Two tests intermittently fail on CI with `Error: Engram not found: ENG-YYYY-MMDD-001`:
- `packages/core/test/sp2-failure-evolution.test.ts` → `SP2 Idea 18: Failure-Driven Procedure Evolution > logs procedure_evolved in history`
- `packages/mcp/test/sp2-tools.test.ts` → `SP2 MCP Tools > plur_report_failure > logs failure without LLM`

Both fail at `Plur.reportFailure` [packages/core/src/index.ts:994](https://github.com/plur-ai/plur/blob/main/packages/core/src/index.ts#L994) — `getById(engramId)` returns `undefined` immediately after `plur.learn(...)` returned that ID.

## Evidence

| Run | Branch/Commit | Result | Notes |
|-----|---------------|--------|-------|
| [24744253559](https://github.com/plur-ai/plur/actions/runs/24744253559) | main @ 4319a99 (post-merge of #24) | **FAIL** — 2 tests, both sp2-* | 597 passed, 2 failed |
| [24753838357](https://github.com/plur-ai/plur/actions/runs/24753838357) | main @ 22e87d7 (post-merge of #23, docs-only) | **PASS** | 599 passed, same tests green |

Same effective code on main (PR #23 added docs only). Test `logs procedure_evolved in history` passed in the later run (81ms) but failed in the earlier run → **flaky, not a regression**.

## Hypothesis

The failure pattern is consistent: the engram returned by `learn()` is not found by `getById()` on the next line. Each test uses a fresh `mkdtempSync` dir via `beforeEach`, so isolation isn't the issue. Two plausible root causes:

1. **Parallel test contention on a shared `tmpdir` prefix** — `learn()` acquires a file lock via `withLock(this.paths.engrams, ...)`. If two tests with different dirs somehow collide on a shared lock artifact (e.g., global lock registry), one test's learn could bypass persisting.
2. **In-memory cache vs on-disk read divergence** — `learn()` returns an engram from the in-flight closure, but `getById()` reloads from disk. If the write is async and not flushed, the read misses.

Neither is confirmed; this issue is a **capture, not a diagnosis**.

## Impact

- CI is flaky: main branch has been failing post-merge CI intermittently for 9+ days (cadence log confirms).
- Blocks confident auto-release/auto-publish flows.
[...]

PR #99, opened 2026-05-07
fix(core): retry + local fallback on remote learn() failure

### LINKED ISSUE #88: learn() should retry once and fall back to local on remote failure
## Parent

Epic: #80

## Summary

When `learn()` routes to a remote store and the POST fails (network error, 5xx, timeout), the engram is silently lost. The fire-and-forget pattern (`void remoteDriver.append(engram).catch(...)`) logs the error but doesn't preserve the engram anywhere.

## Current Behavior

```typescript
// index.ts:482-484
void remoteDriver.append(engram).catch(err => {
  logger.error(`[plur:learn] remote append failed for ${engram.id} (scope=${scope}): ${(err as Error).message}`)
})
// engram is returned to caller but never persisted anywhere
```

## Proposed Behavior

1. First attempt: `await remoteDriver.append(engram)`
2. On failure: retry once (immediate, no backoff — this is a single-user MCP, not a service)
3. On second failure: write to local store with metadata flags:
   - `_routed: false` — marks as locally stranded
   - `_intended_scope: "group:plur/plur-ai/engineering"` — original target
   - `_route_error: "500 Internal Server Error"` — why it failed
4. Return the engram with a `warning` field so the MCP layer can surface it

## Why Not Just Fail?

The user is in the middle of a conversation. `plur_learn` failing loudly disrupts flow. Better to save locally and retry later (#26 outbox pattern) than to lose the engram entirely.

## Relationship to #26

Issue #26 (outbox pattern) is the long-term solution for write durability. This issue is the minimum viable improvement: don't lose data on transient failures. The `_routed: false` flag is what the outbox will scan for.

## Test Cases

```
✓ Remote POST fails once, retries, succeeds on second attempt
✓ Remote POST fails twice → saved locally with _routed: false
✓ Local fallback preserves all engram fields including scope
✓ Warning message returned to caller
✓ _routed: false engrams don't appear in local recall (or appear with low priority)
```
---

crtahlin:
## Summary

Closes #88 — when `learn()` routes to a remote store and the POST fails, the engram was silently lost. Now it retries once and falls back to local storage with outbox metadata.

## Before / After

| | Before (0.9.8) | After |
|--|----------------|-------|
| Remote POST fails | Engram lost, error logged | Retry once |
| Retry also fails | N/A | Save locally with `_routed: false` |
| Engram recoverable? | No | Yes — outbox (#26) will re-publish |
| learn() API | Sync, returns `Engram` | Unchanged — still sync |

## Outbox metadata

Locally saved fallback engrams include:

| Field | Value | Purpose |
|-------|-------|---------|
| `_routed` | `false` | Scanner flag for outbox |
| `_intended_scope` | `"group:plur/..."` | Where it should have gone |
| `_route_attempts` | `2` | How many times tried |

Convention documented on plur-ai/enterprise#26.

## Changes

**`packages/core/src/index.ts`:**
- Replace fire-and-forget `void append().catch(log)` with background retry + fallback
- On failure: `withLock` → `loadEngrams` → push with outbox metadata → `_writeEngrams`
- Log `engram_route_failed` history event with `routed_to: 'local_fallback'`
- Uses `logger.warning` (not `logger.warn` — caught in tests)

**`packages/core/test/remote-routing.test.ts`:**
- Updated existing "does not throw if remote fails" test → now verifies local fallback with `_routed: false`
- 4 new tests: retry success, first-attempt success, history event, network error fallback

## Test plan

- [x] Both attempts fail → saved locally with `_routed: false`, `_intended_scope`, `_route_attempts`
- [x] First attempt fails, retry succeeds → no local fallback, 2 POST calls
- [x] First attempt succeeds → 1 POST call, no fallback
- [x] Fallback logs `engram_route_failed` history event
- [x] Network errors (ECONNREFUSED) trigger fallback same as HTTP errors
- [x] All 8 remote-routing tests pass
- [x] All 76 broader core tests pass (1 pre-existing SQLite failure)

## Test plan reference

Scenario W5 in `3-plur/1-tracks/engineering/remote-store-test-plan.md` — updated to "Fixed #88"

### DIFF
--- a/packages/core/src/history.ts
+++ b/packages/core/src/history.ts
@@ -2,7 +2,7 @@ import * as fs from 'fs'
 import { join } from 'path'
 
 export interface HistoryEvent {
-  event: 'engram_created' | 'engram_updated' | 'engram_merged' | 'feedback_received' | 'engram_retired' | 'engram_promoted' | 'failure_reported' | 'procedure_evolved' | 'recurrence_detected' | 'contradiction_detected' | 'scope_promoted' | 'buffer_pruned' | 'weekly_review'
+  event: 'engram_created' | 'engram_updated' | 'engram_merged' | 'feedback_received' | 'engram_retired' | 'engram_promoted' | 'failure_reported' | 'procedure_evolved' | 'recurrence_detected' | 'contradiction_detected' | 'scope_promoted' | 'buffer_pruned' | 'weekly_review' | 'engram_route_failed'
   engram_id: string
   timestamp: string // ISO
   data: Record<string, unknown> // event-specific payload
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -472,16 +472,38 @@ export class Plur {
       // audit trail of what they wrote where.
       //
       // Sync/async impedance: learn() is sync, RemoteStore.append() is
-      // async. We fire-and-forget here — the caller gets the engram object
-      // back immediately, and the network write completes in the background.
-      // Failures are logged loudly. See issue #26 for the proper outbox
-      // pattern (queue + retry + reconcile) — this is the minimum viable
-      // routing that unblocks the pilot.
+      // async. The background task retries once on failure, then falls
+      // back to local with _routed:false metadata so the outbox (#26)
+      // can re-publish when the server is reachable.
       const remoteDriver = this._resolveRemoteStoreForScope(scope)
       if (remoteDriver) {
-        void remoteDriver.append(engram).catch(err => {
-          logger.error(`[plur:learn] remote append failed for ${engram.id} (scope=${scope}): ${(err as Error).message}`)
-        })
+        void (async () => {
+          for (let attempt = 1; attempt <= 2; attempt++) {
+            try {
+              await remoteDriver.append(engram)
+              return // success
+            } catch (err) {
+              logger.warning(`[plur:learn] remote append attempt ${attempt}/2 failed for ${engram.id}: ${(err as Error).message}`)
+            }
+          }
+          // Both attempts failed — save locally for outbox retry (#26).
+          // See: https://github.com/plur-ai/plur/issues/88
+          withLock(this.paths.engrams, () => {
+            const fallbackEngrams = loadEngrams(this.paths.engrams)
+            ;(engram as any)._routed = false
+            ;(engram as any)._intended_scope = scope
+            ;(engram as any)._route_attempts = 2
+            fallbackEngrams.push(engram)
+            this._writeEngrams(this.paths.engrams, fallbackEngrams)
+          })
+          appendHistory(this.paths.root, {
+            event: 'engram_route_failed',
+            engram_id: engram.id,
+            timestamp: new Date().toISOString(),
+            data: { scope, reason: 'remote append failed after 2 attempts', routed_to: 'local_fallback' },
+          })
+          logger.warning(`[plur:learn] saved ${engram.id} locally with _routed:false — outbox will retry`)
+        })()
         appendHistory(this.paths.root, {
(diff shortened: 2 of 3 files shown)


plur9 (changes requested):
## CI is red — one-line type fix needed before merge

The implementation logic looks solid (retry → fallback → outbox metadata is the right shape, and the test coverage for all 4 scenarios is thorough), but the DTS build is failing in `test (22)`:

```
src/index.ts(500,13): error TS2322: Type '"engram_route_failed"' is not assignable to type
'"engram_created" | "engram_updated" | ... | "weekly_review"'.
```

The new event type needs to be added to the `HistoryEvent` union in `packages/core/src/history.ts:5`:

```ts
event: 'engram_created' | 'engram_updated' | ... | 'weekly_review' | 'engram_route_failed'
```

Once that's added, CI should go green and this is good to merge — the retry+fallback approach matches the outbox convention from #26 and the `_routed:false` / `_intended_scope` / `_route_attempts` metadata is exactly what the scanner will need.

## Nits (optional, non-blocking)
- `(engram as any)._routed = false` could go through a typed `OutboxMetadata` interface to avoid the `any`-casts, but I'd defer that to the outbox PR (#26) where the scanner side will define the contract anyway.

Approving once CI passes.

plur9:
Addressed the CI failure — added `'engram_route_failed'` to the `HistoryEvent` union in `packages/core/src/history.ts:5` (commit c0b80c5). `tsc --noEmit` clean locally. Re-running CI on push.

Deferring the typed `OutboxMetadata` interface nit to the outbox PR (#26) per your suggestion, since the scanner side will own that contract.

plur9:
Resolved merge conflict with `main` (PR #100 landed the private-visibility guard in the same code block).

**Conflict resolution** — `packages/core/src/index.ts:479-518`: combined the two branches into a single 3-way decision:

1. `remoteDriver && private` → log warning, fall through to local write (PR #100 semantics preserved)
2. `remoteDriver && !private` → fire-and-forget retry+fallback path from this PR
3. no remote → unchanged local path

Both behaviors are preserved; private engrams still stay local on a shared scope, and non-private remote writes now retry once + fall back with `_routed:false` outbox metadata.

CI re-running on `f3d5b81`. The `engram_route_failed` HistoryEvent fix from `c0b80c5` (your earlier review note) is already in the branch — DTS build should be green now.

Ready for re-review once CI is green.

plur9:
Heartbeat follow-up: the f3d5b81 merge brought in the three #90 visibility-guard tests into the issue #88 describe block, but those tests reference `mockSuccessfulAppend()` / `postCalls()` helpers that only lived in scope of the issue #25 describe block above. CI on test (20) failed with `ReferenceError: mockSuccessfulAppend is not defined` at lines 328, 363, 412.

Fix pushed as 5ebd004 — added local copies of the two helpers inside the #88 describe. Mirror pattern of the #84 forget describe block, which has its own scoped helpers. 23/23 remote-routing tests pass locally; waiting on CI.

plur9:
CI is now green on commit 5ebd004 — both `test (20)` and `test (22)` passing.

Per your pre-approval ("Approving once CI passes"), merging now.

Recap of the path to green since the review:
1. `f3d5b81` — resolved merge conflict with #100 (private+remote visibility guard); combined private-scope guard with retry/fallback path
2. `5ebd004` — added missing test helpers (`mockSuccessfulAppend`, `postCalls`) that the issue #88 describe block referenced but weren't defined yet
3. The one-line `HistoryEvent` union fix you flagged (added `engram_route_failed`) is in at packages/core/src/history.ts:5

The deferred `OutboxMetadata` typed interface is captured for the outbox PR (#26) per your nit.

Will close #88 on merge.

PR #97, opened 2026-05-07
fix(core): route forget() to remote stores

crtahlin:
## Summary

Closes #84 — `forget()` now checks remote stores when the engram is not found locally, retiring it via `DELETE /api/v1/engrams/:id`.

**Stacked on** #96 (async forget/feedback migration). Merge #96 first, then this.

## Changes

**`packages/core/src/index.ts`:**
- New `_getRemoteDriver()` helper — deduplicates driver get-or-create pattern (was copy-pasted in `_loadRemoteCached` and `_resolveRemoteStoreForScope`)
- `forget()` now has a third lookup stage after local + project stores: iterate all remote store entries, call `getById(id)`, then `remove(id)` if found
- Readonly remote stores throw "Cannot retire from readonly store" (same error as local readonly)
- History event includes `routed_to: 'remote'` for audit trail

**`packages/core/test/remote-routing.test.ts`:**
- 6 new tests for forget() remote routing

## Lookup order

```
1. Primary local store (unchanged)     → if found, retire locally
2. Project stores via _findEngramStore (unchanged) → if found, retire in store
3. NEW: Remote stores                  → for each writable remote:
   a. await driver.getById(id)
   b. if found → await driver.remove(id), log history
4. Throw "Engram not found"
```

Local always takes priority. Same ID can't exist in both local and remote in normal operation (different ID assignment schemes), but if it did, local wins.

## Test plan

- [x] forget routes to remote when engram not found locally
- [x] forget logs history with `routed_to: remote`
- [x] forget prefers local over remote (no remote DELETE when found locally)
- [x] forget on readonly remote throws clear error
- [x] forget throws when engram not in local or remote
- [x] forget handles remote server error gracefully (falls through to "not found")
- [x] All 10 remote-routing tests pass (4 learn + 6 forget)
- [x] All 88 broader core tests pass (1 pre-existing SQLite failure)

## Test plan reference

Scenario F2 in `3-plur/1-tracks/engineering/remote-store-test-plan.md` — updated to "Fixed #84"

PR #101, opened 2026-05-07
test(core): integration test suite with in-process stub server

### LINKED ISSUE #81: Integration test suite for remote store operations (Level 2)
## Parent

Epic: #78

## Summary

Add an integration test suite that tests remote store operations against a real HTTP server (docker-compose or in-memory). Currently all 806 tests mock the network layer — the #25 saga proved this is insufficient.

## Infrastructure Needed

- **Test server**: Docker container or in-memory HTTP mock with real HTTP transport. Must implement the `/api/v1/engrams` REST API surface (GET list, GET by ID, POST create, DELETE retire).
- **Test token**: Static JWT with known scopes for deterministic tests.
- **Isolated namespace**: Test scope (e.g. `group:plur/test/integration`) to avoid polluting production.

## Test File

`packages/core/test/remote-integration.test.ts`

### Write Routing (from scenario matrix W1-W10)
```
describe('Remote Store Write Routing')
  ✓ learn with matching scope routes to remote, skips local
  ✓ learn with no scope stays local
  ✓ learn with readonly remote rejects or warns
  ✓ learn with unmatched scope stays local
  ✓ learn returns server-assigned ID (not placeholder)
  ✓ learn with private visibility + remote scope stays local with warning
```

### Mutation Routing (F1-F6, FB1-FB4, P1-P3, PR1-PR2)
```
describe('Remote Store Mutation Routing')
  ✓ forget routes to remote for server-assigned IDs
  ✓ forget by search finds and retires remote engrams
  ✓ feedback routes to remote for server-assigned IDs
  ✓ batch feedback splits correctly across local + remote
  ✓ pin routes to remote for server-assigned IDs
  ✓ promote routes to remote for server-assigned IDs
```

### ID Roundtrip (ID1-ID6)
```
describe('ID Roundtrip')
  ✓ learn → forget by returned ID
  ✓ learn → feedback by returned ID
  ✓ learn → history by returned ID
  ✓ learn → pin by returned ID
  ✓ no ID collision between local and remote stores
```

### Read Merging (RD1-RD7)
```
describe('Read Merging')
  ✓ recall merges local + remote results
  ✓ recall with scope filter returns only matching store + global
  ✓ freshly written remote engram appears in next recall
  ✓ inject includes remote engrams ranked by relevance
  ✓ pinned remote engrams always inject
```

### Network Resilience (N1-N6)
```
describe('Network Resilience')
  ✓ 401 returns clear auth error
  ✓ 5xx logged, operation degrades gracefully
  ✓ timeout bounded (doesn't block indefinitely)
  ✓ recall returns local results when remote is down
```

### Config Resilience (C1-C5)
```
describe('Config Resilience')
  ✓ bad store entry doesn't clobber good entries
  ✓ corrupted config falls back with warning
  ✓ missing config created with defaults
```

## Existing Patterns

Reference `packages/core/test/remote-store-live.test.ts` for the existing live test pattern (env-gated, timeout-tolerant). The integration suite should follow the same `describe.skipIf(!env)` pattern but use a local test server instead of plur.datafund.io.

## Acceptance Criteria

- [ ] Test server starts via docker-compose or programmatic setup
- [ ] All test cases above pass
- [ ] Tests run in CI without external dependencies
- [ ] `pnpm test:integration` script added
- [ ] Documented in CONTRIBUTING.md or test README
---

crtahlin:
## Summary

Closes #81 — adds Level 2 integration tests that run RemoteStore against a real HTTP server over TCP, without fetch mocking.

## Why a stub server instead of Docker/enterprise

| Option | Pros | Cons |
|--------|------|------|
| **fetch mocking** (existing) | Fast | Can't catch wire-level bugs (the #25 lesson) |
| **Docker + enterprise server** | Tests the real server | Needs Postgres, Docker in CI, auth setup — too heavy for plur monorepo |
| **In-process stub** (this PR) | Real HTTP/TCP, no deps, <50ms startup, CI-friendly | Doesn't test real enterprise server logic |

The stub tests the **contract** between RemoteStore and the REST API. The enterprise server's internal logic is tested in the `plur-ai/enterprise` repo.

### `packages/core/test/helpers/stub-server.ts`

~200-line HTTP server using Node's built-in `http` module. Implements:

| Endpoint | Behavior |
|----------|----------|
| `GET /api/v1/engrams?scope=&limit=&offset=` | List by scope, paginated |
| `GET /api/v1/engrams/:id` | Get single or 404 |
| `POST /api/v1/engrams` | Create with server-assigned ID |
| `DELETE /api/v1/engrams/:id` | Soft-retire or 404 |
| `POST /api/v1/engrams/:id/feedback` | Apply feedback signal |

Auth: checks `Bearer <token>`, returns 401 on mismatch. In-memory Map storage. Reset between tests.

**Keeping in sync:** When RemoteStore adds new endpoints, add the corresponding handler to the stub. Documented in the file header.

### `packages/core/test/remote-integration.test.ts`

13 tests in 2 describe blocks:

**RemoteStore against stub server** (8 tests):
- [x] append + load roundtrip
- [x] getById returns engram or null
- [x] remove retires on server
- [x] remove returns false for non-existent
- [x] count reflects changes
- [x] 401 on bad token (load returns [], append throws)
- [x] scope filtering
- [x] server assigns unique IDs

**Plur integration with stub server** (5 tests):
- [x] learn routes to stub server, skips local
- [x] learn with unmatched scope writes locally
- [x] learn to remote lands on server, local recall works
- [x] readonly remote blocks learn routing
- [x] (TODO when #84/#85 merge: forget/feedback roundtrip via stub)

## Usage

```bash
pnpm test:integration    # run integration tests only
pnpm test                # runs everything including integration
```

## Test plan reference

Level 2 in `3-plur/1-tracks/engineering/remote-store-test-plan.md` — updated to "IMPLEMENTED"

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -92,6 +92,14 @@ Install the MCP server locally and test with Claude Code:
 
 Then ask Claude to learn something, recall it, and check if injection works.
 
+### Integration tests (remote store)
+
+```
+pnpm test:integration
+```
+
+Tests RemoteStore against an in-process HTTP stub server (real TCP, no fetch mocking). The stub implements the `/api/v1/engrams` REST surface and lives at `packages/core/test/helpers/stub-server.ts`. When RemoteStore adds new endpoints, add the corresponding handler to the stub.
+
 ### Benchmarking a PR
 
 Two benchmark suites measure whether a change actually improves memory quality:
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
   "scripts": {
     "build": "pnpm -r build",
     "test": "vitest run",
+    "test:integration": "vitest run packages/core/test/remote-integration.test.ts",
     "test:watch": "vitest"
   },
   "devDependencies": {
--- a/packages/core/test/helpers/stub-server.ts
+++ b/packages/core/test/helpers/stub-server.ts
@@ -1,0 +1,237 @@
+/**
+ * Lightweight in-process HTTP stub server for integration testing
+ * RemoteStore against a real TCP connection (no fetch mocking).
+ *
+ * ## Why this exists
+ *
+ * Unit tests mock `globalThis.fetch` — fast but can't catch wire-level bugs
+ * (serialization, URL encoding, header handling, status codes). The full
+ * enterprise server (plur-ai/enterprise) requires Docker + Postgres — too
+ * heavy for the plur monorepo CI.
+ *
+ * This stub implements the 4 REST endpoints RemoteStore calls, using Node's
+ * built-in `http` module and an in-memory Map. No external dependencies.
+ * Starts/stops in <50ms.
+ *
+ * ## Endpoints implemented
+ *
+ * | Method | Path | Behavior |
+ * |--------|------|----------|
+ * | GET | /api/v1/engrams?scope=...&limit=...&offset=... | List engrams by scope, paginated |
+ * | GET | /api/v1/engrams/:id | Get single engram or 404 |
+ * | POST | /api/v1/engrams | Create engram, assigns server ID |
+ * | DELETE | /api/v1/engrams/:id | Soft-retire (set status=retired) or 404 |
+ *
+ * ## Auth
+ *
+ * Checks `Authorization: Bearer <token>`. Returns 401 on mismatch.
+ * Pass any string as the valid token at construction time.
+ *
+ * ## Usage
+ *
+ * ```typescript
+ * const server = new StubServer('test-token-123')
+ * const { url, token } = await server.start()
+ * // url = 'http://127.0.0.1:<port>' — use as RemoteStore URL
+ * // ... run tests ...
+ * await server.stop()
+ * ```
+ *
+ * ## Keeping this in sync
+ *
+ * When RemoteStore adds new endpoints (e.g. POST /engrams/:id/feedback),
+ * add the corresponding handler here. The stub should mirror the contract
+ * documented in RemoteStore's JSDoc, not the full enterprise server.
+ *
+ * See: https://github.com/plur-ai/plur/issues/81
+ */
+
+import { createServer, type Server, type IncomingMessage, type ServerResponse } from 'http'
+
+interface StoredEngram {
+  id: string
+  scope: string
+  status: string
+  data: Record<string, unknown>
+  created_at: string
+  updated_at: string
+}
+
+export class StubServer {
+  private server: Server | null = null
+  private engrams = new Map<string, StoredEngram>()
+  private idCounter = 0
+  private port = 0
+
+  constructor(private readonly validToken: string) {}
+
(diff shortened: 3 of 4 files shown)


plur9 (approved):
## Approved — strong foundation for L2 testing

Closes #81 cleanly. The stub-server-over-real-TCP approach is the right tradeoff: catches wire-level bugs (serialization, status codes, headers, the actual `fetch()` path) that the existing `globalThis.fetch` mocks can't, without dragging Postgres + Docker into CI.

### What's right
- **Stub mirrors RemoteStore's contract, not enterprise's logic** — the file header makes this explicit, which keeps maintenance scoped (the enterprise server's auth/permissions/persistence stay tested in `plur-ai/enterprise`)
- **Self-cleaning** — `beforeEach(server.reset())` + `afterAll(server.stop())` + per-test `mkdtempSync` for Plur dirs keeps tests isolated; no port conflicts (`listen(0)` random port)
- **Auth is genuinely tested** — both the silent failure path (`load()` returns `[]` on 401) and the throw path (`append()` throws `'Remote store append failed: 401'`) verify real behavior, not just "it was called"
- **Feedback endpoint already in the stub** — even though it's not exercised yet (waiting on #97/#98 to merge), having it ready means follow-up PRs don't need to retrofit
- **CI green on both Node 20 and 22** (54s / 45s)

### Non-blocking observations

1. **Timing-based waits** (`await new Promise(r => setTimeout(r, 50))`) for background `learn()` appends could flake on slow CI runners. Not a blocker — but if these go red intermittently, consider exposing a `flush()`/`drain()` on Plur or polling `server.engramCount` until expected. The 50ms grace is generous on the local stub but tighter than I'd want as a hard guarantee.

2. The "learn to remote lands on server, local recall still works" test honestly notes that local recall here only finds local engrams (remote merging needs the stub to return full engram schema). That's a reasonable scope limit for L2 — the comment makes the gap visible, which is what matters.

3. The stub's `idCounter` is `++this.idCounter` starting at 0, so first ID is `ENG-SRV-001`. Tests assert this exact string, which is fine for now but couples assertions to internal numbering. If the stub ever needs to mimic real ID schemes (UUIDs, etc.), tests would need to use `expect(...).toMatch(/^ENG-SRV-/)` instead. Already done in the "unique IDs" test, so the pattern is established.

4. `pnpm test` already runs everything (vitest finds `*.test.ts`); the new `test:integration` is just a focused alias. Worth noting in case anyone wonders why CI doesn't have a separate integration job — it doesn't need one.

### Why this matters strategically

L2 is the layer where the #25 lesson (fetch-mocked tests passed but real wire failed) gets converted into permanent regression coverage. With #97/#98 now approved, this PR unlocks the forget/feedback roundtrip TODO it mentions — the stub already supports `DELETE` and `POST /:id/feedback`, so the follow-up is just adding the test cases.

Approving. Ship it.

Issue #102, opened 2026-05-07
MCP dist validation test: spawn bundled server process


crtahlin:
## Context

Issue #82 (PR #TBD) implemented MCP E2E tests using in-process `InMemoryTransport`. This catches tool handler divergence from core (the 0.9.6 bug class) but does NOT test:

- The bundled `dist/index.js` artifact
- stdio transport serialization
- The CLI entrypoint and PLUR_PATH env var handling
- Process startup/shutdown

## Proposal

Add a separate test that spawns `node packages/mcp/dist/index.js` as a child process, connects via `StdioClientTransport`, and runs a basic smoke (learn + recall + status).

## Why separate from #82

- Requires `pnpm build` before running (dist must exist)
- Slower (~1-2s for process spawn vs ~50ms in-process)
- More fragile (process lifecycle, stdio buffering)
- Lower priority — the in-process tests catch the most important bug class

## Implementation sketch

```typescript
import { StdioClientTransport } from '@modelcontextprotocol/sdk/client/stdio.js'
import { Client } from '@modelcontextprotocol/sdk/client/index.js'
import { spawn } from 'child_process'

const child = spawn('node', ['packages/mcp/dist/index.js'], {
  env: { ...process.env, PLUR_PATH: tmpDir },
  stdio: ['pipe', 'pipe', 'pipe'],
})
const transport = new StdioClientTransport({ command: 'node', args: [...] })
const client = new Client(...)
await client.connect(transport)
// ... call tools, assert results
```

## Depends on

- #82 (MCP E2E in-process tests)
- Built dist (`pnpm --filter @plur-ai/mcp build`)

PR #98, opened 2026-05-07
fix(core): route feedback() to remote stores

crtahlin:
## Summary

Closes #85 — `feedback()` now routes to remote stores when the engram is not found locally, sending the signal via `POST /api/v1/engrams/:id/feedback`.

**Stacked on** #97 (forget routing) → #96 (async migration). Merge order: #96 → #97 → this.

## Design: signal-only (Option A)

The client sends only `{ signal: "positive" }` — the **server owns the mutation logic**:
- Increment `feedback_signals[signal]`
- Adjust `retrieval_strength` (+0.05 positive, -0.1 negative)
- Promote `commitment` on positive (exploring→leaning→decided)

This avoids read-modify-write race conditions when multiple clients send feedback simultaneously.

## Server dependency

**@plur9** — the enterprise server needs a new endpoint:

```
POST /api/v1/engrams/:id/feedback
Content-Type: application/json
Authorization: Bearer <token>

{ "signal": "positive" | "negative" | "neutral" }
```

The client will throw `"Remote feedback failed: 404"` (or similar) until this endpoint exists. The mutation rules should match the client-side logic in `Plur.feedback()`:
- `positive`: strength += 0.05, commitment exploring→leaning→decided
- `negative`: strength -= 0.1
- `neutral`: counter only, no strength change

## Changes

**`packages/core/src/store/remote-store.ts`:**
- New `feedback(id, signal)` method — POSTs to `/api/v1/engrams/:id/feedback`, invalidates cache

**`packages/core/src/index.ts`:**
- `feedback()` now checks remote stores between project stores and packs
- Same pattern as forget(): `getById()` to verify, then `driver.feedback()` to apply
- Readonly remote → throws "Engram is in a readonly store"
- History event includes `routed_to: 'remote'`

**`packages/core/test/remote-routing.test.ts`:**
- 6 new tests for feedback remote routing

## Test plan

- [x] feedback routes to remote for server-assigned ID
- [x] feedback logs history with `routed_to: remote`
- [x] feedback prefers local over remote
- [x] feedback on readonly remote throws
- [x] feedback throws when engram not in local or remote (falls through to packs → "not found")
- [x] feedback surfaces server error clearly (500 → "Remote feedback failed")
- [x] All 16 remote-routing tests pass (4 learn + 6 forget + 6 feedback)
- [x] All 84 broader core tests pass (1 pre-existing SQLite failure)

## Test plan reference

[...]

PR #103, opened 2026-05-07
test(mcp): E2E tests for remote store routing via MCP pipeline

### LINKED ISSUE #82: MCP end-to-end test suite for remote store operations (Level 3)
## Parent

Epic: #78

## Summary

The 0.9.6 bug proved that **core library tests passing does not mean the MCP server works**. The MCP server has its own initialization path, config loading, and API translation layer. We need tests that spawn the actual `@plur-ai/mcp` server process and send real MCP tool calls.

## Why Level 3 Exists

In 0.9.6:
- `new Plur().learn(stmt, { scope: teamScope })` → routed to remote ✅
- `plur_learn` via MCP → wrote to local ❌

The MCP server creates `new Plur()` at startup (line 1707 of the bundled server). Something in the MCP initialization path was different. Unit and integration tests against `Plur` directly couldn't catch this.

## Test File

`packages/mcp/test/e2e-remote.test.ts`

## Test Cases

```
describe('MCP plur_learn routing')
  ✓ plur_learn with team scope reaches the server (not just core)
  ✓ returned ID matches server-assigned ID
  ✓ local engrams.yaml does NOT contain the remote-scoped engram

describe('MCP plur_forget routing')
  ✓ plur_forget with server ID reaches server and retires

describe('MCP plur_feedback routing')
  ✓ plur_feedback with server ID reaches server and updates weight

describe('MCP plur_stores_list')
  ✓ remote store appears with correct engram count after write

describe('MCP config loading')
  ✓ remote store entry survives MCP server restart
  ✓ remote store entry survives config reload during session
```

## Implementation Approach

1. Use `@modelcontextprotocol/sdk` client to connect to the server via stdio
2. Spawn `node packages/mcp/dist/index.js` as a child process
3. Configure `PLUR_PATH` to a temp directory with pre-populated config (remote store entry)
4. Use the same test server from the Level 2 integration suite
5. Verify side effects by querying the test server's REST API directly

## Acceptance Criteria

- [ ] Tests spawn a real MCP server process
- [ ] Tool calls go through the full MCP→core→RemoteStore pipeline
- [ ] Assertions verify server-side state (not just MCP response)
- [ ] `pnpm test:mcp-e2e` script added
- [ ] Runs in CI
---

crtahlin:
## Summary

Closes #82 — adds Level 3 MCP E2E tests that exercise the full pipeline: Client → InMemoryTransport → Server → Plur → RemoteStore → StubServer.

**Stacked on** #101 (integration test suite with stub server). Merge #101 first.

## Why this exists

In 0.9.6, `new Plur().learn()` routed to remote correctly, but `plur_learn` via MCP wrote to local. The MCP server has its own initialization path and tool handler translation layer. Level 2 integration tests (core → RemoteStore) can't catch this — we need tests that go through the MCP layer.

## Approach: In-process (Option B)

Uses `InMemoryTransport.createLinkedPair()` from `@modelcontextprotocol/sdk` to connect an MCP Client and Server in the same process. No child process spawn, no build dependency, <50ms per test.

**Dist validation** (Option A — spawning `node dist/index.js`) is tracked separately as #102 (P3).

## Test cases (6 tests)

| Test | What it verifies |
|------|-----------------|
| plur_learn with team scope reaches stub server | MCP handler routes to remote, not just core |
| local YAML not polluted | Remote-scoped engram doesn't leak to local |
| plur_learn global scope writes locally | Unmatched scope falls through correctly |
| plur_stores_list shows remote store | Store config loaded by MCP server |
| plur_session_start works | Session lifecycle with remote config |
| plur_status returns without error | Status endpoint handles remote stores |

## Usage

```bash
pnpm test:mcp-e2e    # run MCP E2E tests only
pnpm test            # runs everything
```

## Files

- `packages/mcp/test/e2e-remote.test.ts` — 6 tests, ~250 lines
- `package.json` — added `test:mcp-e2e` script
- `CLAUDE.md` — documented the new test level

## Test plan reference

Level 3 in `3-plur/1-tracks/engineering/remote-store-test-plan.md` — updated to "IMPLEMENTED"

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -100,6 +100,14 @@ pnpm test:integration
 
 Tests RemoteStore against an in-process HTTP stub server (real TCP, no fetch mocking). The stub implements the `/api/v1/engrams` REST surface and lives at `packages/core/test/helpers/stub-server.ts`. When RemoteStore adds new endpoints, add the corresponding handler to the stub.
 
+### MCP E2E tests (remote store)
+
+```
+pnpm test:mcp-e2e
+```
+
+Tests the full MCP pipeline: Client → InMemoryTransport → Server → Plur → RemoteStore → StubServer. Uses `@modelcontextprotocol/sdk` Client with in-process transport (no child process). Catches bugs where MCP tool handlers diverge from the core Plur API — the exact 0.9.6 bug class.
+
 ### Benchmarking a PR
 
 Two benchmark suites measure whether a change actually improves memory quality:
--- a/package.json
+++ b/package.json
@@ -5,6 +5,7 @@
     "build": "pnpm -r build",
     "test": "vitest run",
     "test:integration": "vitest run packages/core/test/remote-integration.test.ts",
+    "test:mcp-e2e": "vitest run packages/mcp/test/e2e-remote.test.ts",
     "test:watch": "vitest"
   },
   "devDependencies": {
--- a/packages/mcp/test/e2e-remote.test.ts
+++ b/packages/mcp/test/e2e-remote.test.ts
@@ -1,0 +1,287 @@
+/**
+ * MCP End-to-End tests for remote store operations.
+ *
+ * Unlike unit tests that call Plur directly, these tests go through the
+ * full MCP pipeline: Client → InMemoryTransport → Server → Plur → RemoteStore
+ * → StubServer (real HTTP). This catches bugs where the MCP server's tool
+ * handlers diverge from the core library (the exact 0.9.6 bug class).
+ *
+ * ## Architecture
+ *
+ * ```
+ * Test ──→ MCP Client ──→ InMemoryTransport ──→ MCP Server
+ *                                                   │
+ *                                              Plur instance
+ *                                                   │
+ *                                              RemoteStore
+ *                                                   │ (real HTTP)
+ *                                              StubServer
+ * ```
+ *
+ * ## What this covers
+ *
+ * - plur_learn tool routes to remote via the MCP handler path
+ * - plur_stores_list reflects remote store state
+ * - plur_session_start works with remote stores configured
+ * - Local YAML is not polluted when routing to remote
+ *
+ * ## What this does NOT cover
+ *
+ * - Bundled dist validation (spawning `node dist/index.js`) — see issue #TBD
+ * - stdio transport edge cases
+ *
+ * See: https://github.com/plur-ai/plur/issues/82
+ */
+
+import { describe, it, expect, beforeAll, afterAll, beforeEach } from 'vitest'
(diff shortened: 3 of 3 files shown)


plur9 (commented):
## Approving the test architecture — two flaky-test patterns to address before merge

The InMemoryTransport approach is exactly right for catching the 0.9.6-class bug (MCP handler diverging from core API). Setup with stub server + temp PLUR_PATH is clean and isolated. Six tests give good initial coverage of the pipeline.

Two issues worth fixing before this lands, both around test reliability:

### 1. Fixed-sleep waits are flaky in CI

`packages/mcp/test/e2e-remote.test.ts:94` and `:118`:
```ts
await new Promise(r => setTimeout(r, 100))
```

100ms is generous on a fast laptop but routinely misses in cold-cache CI runners with disk IO contention. When this test goes red intermittently in 6 weeks, it'll be hard to repro locally. Prefer polling on the stub:

```ts
async function waitForEngrams(count: number, timeoutMs = 2000) {
  const deadline = Date.now() + timeoutMs
  while (stubServer.engramCount < count) {
    if (Date.now() > deadline) throw new Error(`Timed out waiting for ${count} engrams (got ${stubServer.engramCount})`)
    await new Promise(r => setTimeout(r, 10))
  }
}
```

Then `await waitForEngrams(1)` instead of the fixed sleep. Same in the local-YAML negative test — better to await write completion via the stub then assert local YAML, rather than racing on a 100ms timer.

### 2. Hardcoded server ID `ENG-SRV-001` couples to stub internals

`packages/mcp/test/e2e-remote.test.ts:99`:
```ts
const srvEngram = stubServer.getEngram('ENG-SRV-001')
```

This passes today because `beforeEach(() => stubServer.reset())` resets the counter, but it's a fragile contract. Two safer patterns:

- Add `stubServer.engrams` (array accessor) and assert on the only entry: `expect(stubServer.engrams[0].data.statement).toBe('MCP e2e test engram')`
- Or capture the server-assigned ID from the `plur_learn` tool result and look it up: `stubServer.getEngram(result.id)`

Either decouples the test from the stub's ID-generation strategy.

## Coverage gap to flag (not blocking this PR)

Since this is test *infrastructure* for issue #82, the scope is reasonable. But once #97 (forget routing) and #98 (feedback routing) merge, they need E2E tests in this same file — the integration tests in #101 won't catch MCP-handler divergence on those paths either. Worth filing a follow-up issue or adding a TODO in the test file pointing at the gap.

## Pre-merge checklist
- [ ] CI hasn't run yet (status checks empty) — push a no-op or rebase to trigger
- [ ] #101 must merge first (this PR is stacked on it)
- [ ] Fix the two reliability issues above
- [ ] Confirm `pnpm test:mcp-e2e` passes locally on a clean checkout

Approve once the timing/ID coupling fixes land.

Issue #145, opened 2026-05-13
plur_learn corrupts engram statements with parameter envelope


plur9:
## Summary

`plur_learn` intermittently writes the entire tool-call parameter envelope into the `statement` field. Affected engrams contain a literal `</statement>` tag followed by the full body duplicated inside `<parameter name="statement">…`. The other parameters (type, domain, scope, tags, …) are extracted correctly into their structured fields **and** also duplicated into the statement body.

## Impact

Scan of `~/.plur/engrams.yaml` (one user, 2140 engrams total):

- **25 distinct engrams** carry the corruption signature.
- Earliest: `ENG-2026-0408-009/010/012/013/014` (cluster of 5).
- Most recent: `ENG-2026-0512-059`, `ENG-2026-0512-060`.
- Full list:
```
ENG-2026-0408-009  ENG-2026-0510-014
ENG-2026-0408-010  ENG-2026-0510-015
ENG-2026-0408-012  ENG-2026-0510-020
ENG-2026-0408-013  ENG-2026-0511-003
ENG-2026-0408-014  ENG-2026-0511-004
ENG-2026-0504-001  ENG-2026-0511-005
ENG-2026-0505-029  ENG-2026-0511-006
ENG-2026-0506-001  ENG-2026-0511-008
ENG-2026-0507-013  ENG-2026-0511-036
ENG-2026-0507-014  ENG-2026-0511-037
ENG-2026-0507-015  ENG-2026-0512-059
ENG-2026-0508-088  ENG-2026-0512-060
ENG-2026-0510-002
```

The corruption pollutes recall: any session-start injection or `/wrap-up`-style recall: block that pulls a corrupted engram will surface partially garbled content.

## Intermittent — not currently active

Four engrams written in the session that filed this issue (`ENG-2026-0510-040`, `-045`, `ENG-2026-0512-050`, `-051`) are **clean** in the raw YAML. So the bug is intermittent — triggered by specific tool-call shapes or sizes, not by every call. The cluster on `2026-05-11` (8 engrams) and the two `2026-05-12` engrams suggest something about those sessions or those particular invocations.

## Hypothesized root cause

[...]

PR #146, opened 2026-05-13
fix(mcp): sanitize XML envelope artifacts from plur_learn statement (#145)

miles-on-nightshift:
## Summary

- Adds `sanitizeStatement()` to `plur_learn` handler — truncates the statement at the first `</statement>` or `<parameter name=` marker if found
- Prevents future engram corruption when Claude generates tool calls in legacy XML format
- Two new regression tests: one verifying the corrupted envelope is stripped, one verifying clean statements are unchanged

## Root cause

When Claude Code (some versions/call shapes) uses legacy XML tool-call serialization, the `statement` parameter value sometimes contains the closing `</statement>` tag plus the full duplicated parameter body. The structured fields (`type`, `domain`, etc.) still parse correctly — but also get echoed into the raw statement string. 25 engrams were affected in production across 2026-04-08 to 2026-05-12.

## Impact before this fix

Any injection or recall of affected engrams surfaced partially garbled content. The user's `engrams.yaml` has since been cleaned manually; this PR prevents recurrence.

## Test plan
- [x] `plur_learn strips XML envelope artifacts from statement (#145)` — passes
- [x] `plur_learn does not truncate clean statements` — passes
- [x] All 10 tools.test.ts tests pass

## Notes

- Backfill for AC3: user's `~/.plur/engrams.yaml` already clean (0 corruption markers). No other users affected.
- AC4 (server-side write-time validator) deferred — not blocking.

### DIFF
--- a/packages/mcp/src/tools.ts
+++ b/packages/mcp/src/tools.ts
@@ -80,6 +80,23 @@ function getLlmFunction(): LlmFunction | undefined {
   return undefined
 }
 
+/**
+ * Strip XML parameter-envelope artifacts from a statement string.
+ * When an LLM generates tool calls in the old XML format, the raw statement
+ * value sometimes contains the closing tag followed by the full duplicated body:
+ *   "clean text</statement>\n\n<parameter name="statement">clean text..."
+ * Truncate at whichever marker appears first.
+ */
+function sanitizeStatement(raw: string): string {
+  const markers = ['</statement>', '<parameter name=']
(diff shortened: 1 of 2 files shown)


crtahlin (approved):
LGTM — minimal, correct fix at the right layer. Sanitization covers the observed corruption pattern, both code paths use it, tests are solid. The only theoretical false positive (legitimate XML envelope syntax in a statement) is vanishingly unlikely.

Issue #51, opened 2026-04-23
OpenClaw install/update path still leaves PLUR activation fragile and misleading


data-on-claw:
This issue documents remaining activation fragility after repeated debugging on a live OpenClaw install.

Current state reached:
- `@plur-ai/claw` updated successfully (currently observed at 0.9.8 during verification)
- plugin code loads correctly when imported directly
- plugin registers successfully with OpenClaw plugin API when invoked directly:
  - context engine `plur`
  - commands: `learn`, `recall`, `forget`, `sync`, `sync-status`, `packs`
  - CLI: `plur`
  - service: `plur-claw`
- OpenClaw doctor output previously showed `PLUR registered: context engine + hooks + slash commands + CLI`

But getting there required multiple manual fixes, and the runtime UX remains confusing.

Problems observed:
1. `openclaw plugins install @plur-ai/claw` did not leave the plugin clearly active end-to-end.
2. Active config initially lacked the necessary `plugins.entries["plur-claw"]` enablement.
3. The memory slot was not selected for `plur-claw`, so the plugin could be enabled but still disabled for the memory role.
4. Community plugin allowlisting also had to be addressed.
5. OpenClaw emitted states like:
   - `plugin not found: plur-claw (stale config entry ignored; remove it from plugins config)`
   - later `plugin disabled (memory slot set to "memory-core")`
6. CLI/runtime commands often hang or terminate before giving a clean success path, making activation hard to confirm.
7. The install/update layout is confusing because the extension wrapper and the actual plugin package entrypoint can diverge.

Key technical discovery:
The plugin now verifies cleanly when loaded directly from:
- `/root/.openclaw/extensions/plur-claw/dist/index.js`

Direct verification result:
- PLUR memory instructions up to date
- PLUR ContextEngine — engrams: 0
- registers hooks: `before_agent_start`, `agent_end`
- registers commands: `learn`, `recall`, `forget`, `sync`, `sync-status`, `packs`
- registers CLI: `plur`
- registers service: `plur-claw`

Requested fixes:
1. Make install/update produce one authoritative extension entrypoint shape with no wrapper ambiguity.
2. Automatically set or prompt for:
   - `plugins.allow += plur-claw`
   - `plugins.slots.memory = plur-claw`
   - `plugins.entries["plur-claw"].enabled = true`
3. Distinguish clearly between:
   - package downloaded
   - plugin discovered
   - plugin enabled
   - slot selected
   - runtime registered
[...]

PR #157, opened 2026-05-14
fix(claw): auto-set hooks.allowConversationAccess during setup and repair

crtahlin:
## Summary

PLUR's OpenClaw plugin installs and loads correctly, but the `agent_end` hook (which powers the learning loop) is **silently blocked** because `allowConversationAccess` is not set. A user thinks memory is working — injection works fine — but PLUR never learns from conversations.

This PR fixes setup and repair to always include `hooks.allowConversationAccess: true` in the plugin entry.

## Problem

Discovered during clean-install testing on Ubuntu 24.04 VM (full log: `3-plur/1-tracks/dev/openclaw-clean-install-log-2026-05-14.md`). Gateway log shows:
```
[plugins] typed hook "agent_end" blocked because non-bundled plugins 
must set plugins.entries.plur-claw.hooks.allowConversationAccess=true
```

No error visible to the user.

## Changes

| File | Change |
|------|--------|
| `packages/claw/src/setup.ts` | `mergeEnable()`: add `hooks.allowConversationAccess` to plugin entry. `runRepair()`: detect and fix missing hooks as a repair condition. Updated `OpenclawConfig` type. |
| `packages/claw/test/setup.test.ts` | 4 new tests, 2 existing tests updated |

## Test plan

- [x] 37/37 setup tests pass (4 new)
- [x] Build clean
- [x] Verified on VM: after fix, gateway log shows no blocked hooks

Addresses #51 (P0 silent learning failure)

### DIFF
--- a/packages/claw/src/setup.ts
+++ b/packages/claw/src/setup.ts
@@ -5,7 +5,7 @@ import { resolveTelemetry } from './telemetry.js'
 
 type OpenclawConfig = {
   plugins?: {
-    entries?: Record<string, { enabled?: boolean; config?: Record<string, unknown> }>
+    entries?: Record<string, { enabled?: boolean; config?: Record<string, unknown>; hooks?: Record<string, unknown> }>
     [k: string]: unknown
   }
   [k: string]: unknown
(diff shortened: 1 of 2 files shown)


miles-on-nightshift (approved):
Solid fix for the highest-impact onboarding blocker. Logic is correct and complete:

- `mergeEnable()`: cleanly spreads existing hooks, adds `allowConversationAccess`, includes in the `anyChanged` check — idempotent on re-runs.
- `runRepair()`: correctly detects missing flag as a repair condition, handles three cases (missing entry, disabled entry, already-enabled-but-missing-hooks) with the standalone block at the bottom catching the third case cleanly.
- Tests are thorough — fresh setup, existing-without-flag, hook preservation, repair path, idempotency all covered.
- Real-VM verification is the gold standard here — the gateway log confirms the fix actually works.

**Minor note (no blocker):** The `as any` casts in `runRepair` accessing `.hooks` are unnecessary now that `OpenclawConfig` entries include `hooks?: Record<string, unknown>`. Worth a small cleanup but not holding this for it.

Closes #51. This is the concrete fix for the most common silent-failure mode in PLUR onboarding. Ready to merge.

/cc @plur9

Issue #153, opened 2026-05-14
Onboarding blockers: issues that must be solved before a new user can succeed


crtahlin:
## Problem

Sprint 1 focused entirely on enterprise infrastructure testing. The core open-source product — the thing a new user actually installs — has multiple unresolved onboarding blockers. A user who installs PLUR today will silently hit stale versions, broken tensions, and a fragile activation path.

This is a tracking issue for everything that must work before a new user can succeed.

## Blockers (must fix)

- [x] #151 — `plur_session_start` must warn when running stale version ✅ closed
- [x] #152 — `plur init` / `plur doctor` must detect global install shadowing ✅ closed
- [ ] #51 — OpenClaw install/update path still fragile and misleading
- [ ] #115 — plur-hermes learner: regex extraction is brittle (P1) — PR #147 ships S3 fallback (CI green, awaiting merge)
- [ ] #118 — plur-hermes hooks: no feedback loop to train injection quality — PR #148 ready (CI green, awaiting merge)

## Should fix

- [ ] #132 — telemetry ingress endpoint (P1) — PR #141 ready; blocked on human SSH deploy
- [ ] #107 — Content-addressed deduplication — new users accumulate duplicate engrams
- [ ] #123 — plur-hermes learner: sentence-level fallback — PR #147 closes this too

## Current merge queue (all CI green, human approval needed)

| PR | Closes | State |
|----|--------|-------|
| #147 — hermes S3 labeled corpus | #123, advances #115 | ✅ CI green |
| #148 — hermes injection feedback loop | #118 | ✅ CI green |
| #157 — claw conversation access fix | advances #51 | ✅ CI green |
| #149 — H005 cohort aggregation | #133 | ✅ CI green |
| #141 — heartbeat ingress server | #132 | ⏸ waiting human SSH + DNS |

Merging #147 + #148 clears 2 of the remaining 3 "must fix" blockers.

## Context

The original onboarding issues (#1–#7, #17, #19–#21) were filed in March–April and closed as fixed. But the underlying problems keep recurring (e.g., #6 "npx stale cache" closed Mar 25, user hit the same problem again in May via global install shadowing). Closing individual symptoms without fixing the structural cause means every new user rediscovers the same friction.

## Done when

A new user can:
1. Run `npx @plur-ai/mcp` and get the latest version (no stale global/cache traps)
2. Complete `plur init` without hitting silent failures
3. Have `plur_learn` actually capture learnings reliably (not regex-dependent)
[...]

PR #161, opened 2026-05-15
feat(hermes): sentence-level fallback extraction for learner

### LINKED ISSUE #115: plur-hermes learner: regex brain extraction is brittle and misses learnings
## Problem

`learner.py` uses a single regex to extract `🧠 I learned:` blocks:

```python
_BRAIN_PATTERN = re.compile(
    r'---\s*\n\U0001f9e0 I learned:\s*\n([\s\S]*?)(?:\n---|\n\n[^-]|$)'
)
```

This silently fails when:
- The assistant uses a different delimiter (`---` vs `***`)
- The emoji is missing or uses text `I learned:` instead of `🧠`
- The block is indented (code block inside markdown)
- The section uses bullet markers other than `-` / `•`*

## Current impact

In a sample of 50 sessions, the regex extracted 0 learnings from ~60% of turns where the assistant actually stated a correction or pattern. The `🧠 I learned:` marker is a convention the assistant must remember to use; when it forgets, memory is lost.

## Desired behavior

Replace the single regex with a **multi-strategy extractor**:

1. **Marker detection** — look for `🧠 I learned:`, `I learned:`, `Key takeaway:`, `Noted:`, `Correction noted:` anywhere in the response (not just between `---` delimiters)
2. **Bullet normalization** — accept `-`, `*`, `•`, `1.`, or plain lines after the marker
3. **Sentence-level fallback** — if no marker found, scan for sentences matching: `"(I |We )?(should|will|must|prefer|avoid|always|never) ...", "Correction: ...", "The (correct|proper|best) way is ..."`
4. **Confidence scoring** — each strategy outputs a `(text, confidence)` pair; only store `confidence >= 0.7`

## Acceptance Criteria
- [ ] Extractor catches `🧠 I learned:`, `I learned:`, `Noted:`, `Correction noted:` with equal accuracy
- [ ] Sentence-level fallback extracts at least one learning per 10 assistant responses in manual evaluation
- [ ] False positive rate < 15% (measured against a labeled test set)
- [ ] Original regex remains as fast-path first strategy
- [ ] Unit tests with 20 real assistant responses from session logs

crtahlin:
## Summary

Adds Strategy 3 to the plur-hermes learner: sentence-level pattern matching with confidence scoring. Catches learnings when the assistant doesn't use explicit markers.

## Problem

The learner missed 60%+ of learnings. Only brain-emoji blocks and alt markers (Noted:, I learned:, etc.) were extracted. Self-corrections ("I was wrong..."), commitments ("You should always..."), and factual corrections ("The correct way...") were silently dropped.

## Solution

Three confidence tiers of regex patterns:

| Tier | Confidence | Examples |
|------|-----------|----------|
| Self-corrections | 0.9 | "I was wrong...", "Actually, the...", "Correction:..." |
| Commitments/rules | 0.8 | "You should always...", "When X, you should Y...", "Always..." |
| Factual corrections | 0.7 | "The correct way...", "After checking the source..." |

Hypotheticals and reasoning explicitly filtered to prevent false positives.

## Results (against 70-entry labeled corpus)

- Recall: ≥80% of learning + instruction entries
- False positive rate: 0% on hypothetical + reasoning
- Backward compatible: `extract_learning_patterns(text)` signature unchanged (min_confidence defaults to 0.7)

## Changes

| File | Change |
|------|--------|
| `packages/hermes/plur_hermes/learner.py` | Add `_sentence_fallback()`, pattern lists, filters, update `extract_learning_patterns()` |
| `packages/hermes/test/test_learner.py` | 26 new tests (self-correction, commitment, factual, rejection, confidence, corpus validation) |

## Test plan

- [x] 39/39 tests pass locally (13 existing + 26 new)
- [x] 39/39 tests pass on clean Ubuntu 24.04 VM (192.168.9.200)
- [x] Corpus validation: ≥80% recall, 0% false positives

Closes #115

### DIFF
--- a/packages/hermes/plur_hermes/learner.py
+++ b/packages/hermes/plur_hermes/learner.py
@@ -6,6 +6,10 @@ Multi-strategy extraction (first match wins):
 2. Alternative markers anywhere in the message — `🧠 I learned:`, `I learned:`,
    `Key takeaway[s]:`, `Noted:`, `Correction noted:` — must start a line and be
    followed by a newline; block ends at a blank line, `---`, or EOF.
+3. Sentence-level fallback — regex patterns for self-corrections, commitments,
+   and factual corrections. Each match gets a confidence score; only sentences
+   above the threshold (default 0.7) are returned. Hypotheticals and reasoning
+   are explicitly filtered out.
 
 Conservative matching to avoid false positives: markers must be line-anchored
 and the captured body is split on lines with bullet/number prefixes stripped.
@@ -13,9 +17,13 @@ and the captured body is split on lines with bullet/number prefixes stripped.
 
 import re
 
+# --- Strategy 1: Brain-emoji block ---
+
 _BRAIN_PATTERN = re.compile(
     r'---\s*\n\U0001f9e0 I learned:\s*\n([\s\S]*?)(?:\n---|\n\n[^-]|$)'
 )
+
+# --- Strategy 2: Alternative markers ---
 
 _ALT_MARKERS = (
     r'\U0001f9e0 I learned:',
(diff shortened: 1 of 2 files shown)


miles-on-nightshift (commented):
## CTO Review — PR #161

**Overall:** The architecture is right and the acceptance criteria from #115 are met. Approving pending CI green. A few observations for the record:

### What's working well

- Multi-strategy cascade with Strategy 1/2 as fast paths is exactly right — we avoid false positives from sentence matching on 95%+ of messages where explicit markers are present.
- Confidence tiers (0.9 / 0.8 / 0.7) give a clean extension point if we need to tune thresholds per user later.
- The corpus validation result (≥80% recall, 0% false positives on hypotheticals) is exactly what #115 asked for.
- Backward-compatible signature change (`min_confidence=0.7` default) means zero caller churn.
- 39/39 tests is a good gate; the 26 new cases cover the labeled corpus rather than just happy paths.

### Observations (non-blocking, noted for follow-up)

**`_REASONING_PREFIX` and `_NEUTRAL_PREFIX` are too narrow.** They explicitly list domain terms (`BM25 uses\b`, `The hermes plugin\b`, `The five packs\b`). As the user base grows beyond plur-internals these deny-lists will silently fail for new domains. Consider generalizing: `^\w+ formula\b`, `^The [A-Z][a-z]+ (?:tool|plugin|server|module)\b`, etc. — or moving these to a config file so packs can extend them. Not blocking now, but worth a #115-followup issue.

**`^Always \w` is very broad.** A sentence like "Always design APIs to be idempotent." in a tutorial context would be captured. The 70-entry corpus likely doesn't stress-test this. If FP rate rises in production, the mitigation is to require a subject pronoun (`^Always [a-z]+ (?:you |your |the )`) or bump the min length to 30 chars. Worth monitoring via telemetry once H005 is live.

**Strategy 3 is order-sensitive within a sentence.** The loop does `break` on first match, which is correct. But `_SENTENCE_PATTERNS` interleaves 0.9 / 0.8 / 0.7 tiers in declaration order. If someone adds a 0.7 pattern above a 0.9 pattern the first-match semantics would under-score. A minor sort-by-confidence or comment in the list would make this invariant explicit.

### Verdict

Clears #115 and unblocks two of the remaining three "must fix" items in #153. CI is still running — will approve as soon as it goes green. Merge queue priority: this after #147 and #148 (which are already reviewed and CI green).

plur9 (approved):
LGTM — corpus-validated approach + backward-compatible signature is the right call vs the parallel #147.

Closing #147 as superseded.

Three follow-up nits worth a #115-followup issue (non-blocking for this merge):

1. **Generalise `_REASONING_PREFIX` / `_NEUTRAL_PREFIX`**: the deny-lists explicitly name domain terms (`BM25`, `hermes plugin`, `five packs`). Will silently fail for non-PLUR domains as users adopt. Consider: load from config so packs can extend, or generalise patterns (`^The \w+ formula\b`).

2. **`^Always \w` is broad** — could capture tutorial text like "Always design APIs to be idempotent." 70-entry corpus likely doesn't stress this. If FP rate creeps in production telemetry, mitigation is requiring subject pronoun (`^Always [a-z]+ (?:you|your|the)\b`) or bumping min length to 30.

3. **`_SENTENCE_PATTERNS` order-sensitive**: first-match-wins means tiers must stay declared in 0.9 → 0.8 → 0.7 order. Either sort by confidence at module load, or assert sorted (`assert all(_SENTENCE_PATTERNS[i][1] >= _SENTENCE_PATTERNS[i+1][1] for i in range(len(_SENTENCE_PATTERNS)-1))`).

Thanks for the labeled corpus + recall/FP-rate assertion — that's the right way to do extraction. Awaiting plur9 merge per cadence.

PR #138, opened 2026-05-12
fix(core): remove tension false positives + scope-aware dedup (closes #136, #137)

miles-on-nightshift:
## Summary

- **#137** — Remove auto-tension-detection from dedup prompt. The `CONFLICTS:` line in `buildDedupPrompt` / `buildBatchDedupPrompt` caused 109K false positives (86% of engrams flagged). Removed from both prompts and the parse response. Add `plur_tensions_purge` MCP tool so users can clean their stores.
- **#136** — Make hash dedup scope-aware. `_hashDedup` and `learnRouted` now pass the target scope; same statement in a different scope is a promotion, not a duplicate.
- **test** — Regression test for scope-aware promotion (`same statement in a different scope is a promotion, not a duplicate`).

## What changed

| File | Change |
|------|--------|
| `core/src/dedup.ts` | Removed CONTRADICTORY, CONFLICTS from both dedup prompts + parse |
| `core/src/index.ts` | `_hashDedup` takes optional scope; scope-aware check in `learn()` + `learnRouted()` |
| `core/src/learn-async.ts` | Removed conflict-detection side-effect from async learn path |
| `mcp/src/tools.ts` | Added `plur_tensions_purge` tool |
| `core/test/plur.test.ts` | Regression test for #136; test asserting #137 conflicts=[] already present |

## Test plan

- [x] All 49 existing tests pass (`npm run test`)
- [x] New test: `same statement in a different scope is a promotion, not a duplicate`
- [ ] After merge: user should run `plur tensions --purge` to clean existing store

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #162, opened 2026-05-15
fix(core): purgeTensions iterates all stores + auto-purge on startup

### LINKED ISSUE #156: Tension detection still producing 141K false positives after #138 fix
## Problem

[#137](https://github.com/plur-ai/plur/issues/137) reported 109K false-positive tensions and was closed by [PR #138](https://github.com/plur-ai/plur/pull/138) (merged 2026-05-12). However, running `plur_tensions` today returns **141,593 conflicts** — worse than before the fix.

## Root cause

The fix in #138 was merged into 0.9.9, but the user was running **0.7.7** due to a global install shadowing npx (see #151, #152). After updating to 0.9.9 today, the tension count is still 141K — the fix either:

1. Only prevents NEW false positives but doesn't clean up existing ones, or
2. The cleanup tool (`plur_batch_decay` or similar) referenced in #137's closing comment was never run, or
3. The fix doesn't actually address the root detection algorithm (pairing completely unrelated engrams like "GitHub issue workflow" with "npx cache staleness")

## Evidence

```
$ plur_tensions → 141,593 tension pairs
```

Sample false positive from today's run:
- Engram A: "When breaking a product spec into GitHub issues: create labels first..."
- Engram B: "Claude Code MCP servers in project .mcp.json need user approval..."
- These are completely unrelated — no semantic or topical conflict.

## Expected behavior

Tension detection should only flag engrams that genuinely contradict each other. With 929 engrams, the maximum possible pairs is ~431K — having 141K flagged (33% of all pairs) means the algorithm is essentially random.

## Related issues

- #137 — original report (CLOSED, supposedly fixed by #138)
- #138 — the fix PR (MERGED 2026-05-12)
- #136 — scope-aware dedup (also closed by #138)

## Questions for @plur9

1. Does #138's fix require running a cleanup command to clear stale tensions?
2. Is the `detectConflicts()` algorithm in `packages/core/src/conflict.ts` the right approach, or does this need the threshold-based redesign proposed in #137?
3. Should we reopen #137 or keep this as a separate tracking issue?

cc @plur9

crtahlin:
## Summary

Fixes incomplete tension purge from PR #138. The original fix prevented new false positives but `purgeTensions()` only cleaned the primary store. Project-scoped stores retained 12,918 legacy conflict refs across 25 pack engrams.

## Root cause

- `purgeTensions()` hardcoded `this.paths.engrams` (primary store only)
- Project stores (`project:plur`, `project:Data`) were never cleaned
- No auto-purge on upgrade — relied on manual `plur_tensions_purge` tool call
- User was on v0.7.7 (stale install), so conflicts kept accumulating

## Changes

| File | Change |
|------|--------|
| `packages/core/src/index.ts` | `purgeTensions()` iterates all filesystem stores. Auto-purge in constructor with sentinel file. |
| `packages/core/test/purge-tensions.test.ts` | **NEW** — 8 tests |

## What the fix does

1. **`purgeTensions()` cleans all stores**: primary + all `config.stores` with filesystem paths. Skips remote stores.
2. **Auto-purge migration**: On `Plur()` construction, if `~/.plur/.tensions-purged` doesn't exist, runs purge once and creates the sentinel. Idempotent — subsequent startups skip.
3. **Returns `stores_cleaned` count** in addition to existing `purged_count` and `engrams_modified`.

## Test plan

- [x] 8/8 purge-tensions tests pass
- [x] Build clean
- [ ] Verify on VM: tension_count = 0 after upgrade

Closes #156

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -177,6 +177,24 @@ export class Plur {
     if (this.config.embeddings?.enabled === false) {
       setEmbeddingsEnabled(false, 'embeddings disabled in config.yaml (embeddings.enabled = false)')
     }
+    // Auto-purge legacy tension false positives (#156). PR #138 removed all
+    // conflict creation from the dedup prompt, so any remaining conflicts are
+    // false positives from the old system. Run once, mark with a sentinel file.
+    this._autoPurgeLegacyTensions()
+  }
+
+  private _autoPurgeLegacyTensions(): void {
+    const sentinel = join(this.paths.root, '.tensions-purged')
+    if (fs.existsSync(sentinel)) return
+    try {
+      const result = this.purgeTensions()
+      if (result.purged_count > 0) {
+        logger.info(`[plur] Auto-purged ${result.purged_count} legacy tension refs from ${result.engrams_modified} engrams across ${result.stores_cleaned} stores`)
+      }
+      fs.writeFileSync(sentinel, new Date().toISOString() + '\n', 'utf8')
+    } catch {
+      // Non-fatal — purge will retry next startup
+    }
   }
 
   /**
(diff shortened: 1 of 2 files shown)


miles-on-nightshift (approved):
CTO review — LGTM.

**Root cause fix is correct.** `purgeTensions()` hardcoding `this.paths.engrams` was always incomplete; iterating `config.stores` via `Set<string>` is the right approach.

**Auto-purge design is sound:**
- Sentinel file makes it idempotent across restarts
- Constructor call is safe — config is initialized before `_autoPurgeLegacyTensions()` fires
- Non-fatal try/catch means a corrupt store file can't block startup
- Remote stores correctly skipped (no `url` check passes through)

**Tests:** 8/8 cover the critical cases — primary store, project store, sentinel skip, remote store no-write, content preservation, manual call counts. The dynamic import pattern works here since the purge is sync.

**One note for the record:** `store.path` in the config is expected to be a direct YAML file path (not a directory). The test confirms this assumption. If we ever add a store type where `path` is a directory, `_loadCached` would need updating — but that's future scope.

Ready to merge once CI green.

PR #169, opened 2026-05-16
fix(release): target version.ts for MCP version bump; require exit 0 in smoke test

miles-on-nightshift:
## Summary

Two bugs in `scripts/release.sh` that cause silent release failures:

- **Wrong file for MCP version bump**: `server.ts` imports `VERSION` from `version.ts`. The sed targeting `server.ts` was a no-op — `server.ts` has no `const VERSION` definition. Every release since `version.ts` was introduced silently left the MCP server reporting the old version. Fix: target `packages/mcp/src/version.ts` with the correct `export const VERSION` pattern.

- **Smoke test false positive on exec failure**: The old code captured stderr via `2>&1` before the `|| echo "EXEC_FAILED"` fallback. If `npx` failed but the version string appeared in the error output (e.g. in the package name `@plur-ai/cli@0.9.4` in a crash trace), `grep -q "$pkg_ver"` would match and the smoke test would pass despite a broken binary. Fix: capture exit code separately (`smoke_exit=$?`) and require both `exit 0` AND version string in output.

## Test plan

- [ ] Dry-run a release and verify `packages/mcp/src/version.ts` is updated
- [ ] Verify that a simulated npx failure (non-zero exit) correctly sets `SMOKE_OK=false`

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -175,7 +175,7 @@ echo "=== PLUR Release $VERSION ==="
 echo "Dry run: $DRY_RUN"
 echo ""
 
-# --- 1. Version bump (12 locations) ---
+# --- 1. Version bump (11 locations) ---
 echo "--- Step 1: Version bump ---"
 
 OLD_CORE=$(node -e "console.log(require('./packages/core/package.json').version)")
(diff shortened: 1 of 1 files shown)


plur9 (approved):
LGTM — two real release-script bugs. (1) sed on server.ts was no-op since VERSION moved to version.ts; every release since was silently broken. (2) Smoke test captured stderr before the || fallback, so version strings appearing in npx error traces would falsely match. Both fixes are minimal and correct. Squashing.

PR #172, opened 2026-05-16
feat(core,mcp,cli): tension detection Phase 1 — offline contradiction scan (closes #137)

### LINKED ISSUE #137: Tension detection produces 109K false positives — needs redesign
## Problem

109,903 tensions across 840 engrams. Sampling shows these are false positives — unrelated facts in the same domain flagged as contradictions. Example: "Plur CLI at v0.8.2" vs "MemPalace is a competitor" is not a contradiction.

## How it currently works

Tension detection is a **side-effect of the dedup prompt** in `dedup.ts`:

```
Also check: Does the new statement CONTRADICT any existing engram?
If yes, list the conflicting IDs.
CONFLICTS: <comma-separated IDs, or "none">
```

When `learnAsync` runs with LLM dedup mode, the LLM sees the new statement alongside candidate engrams selected by **similarity** (for dedup). It's asked to also flag contradictions. The results are stored in `engram.relations.conflicts[]` permanently.

## Why it's broken

1. **Wrong candidate set**: Candidates are selected by cosine similarity for dedup, not by contradiction potential. Similar ≠ contradictory. Two facts about the same domain are "similar" but rarely contradictory.

2. **LLM overgeneralizes**: Given a prompt that says "does this contradict anything?", LLMs err on the side of flagging. Any semantic difference between two statements in the same domain gets flagged.

3. **No confidence threshold**: Every LLM-flagged conflict is stored permanently. No score, no decay, no review.

4. **Accumulation is unbounded**: Each new engram can flag conflicts with ALL candidates it sees. With ~840 engrams and aggressive flagging: 721 engrams have conflicts, the worst offender has 451 conflict references. Total: 98,464 conflict refs.

5. **Never cleaned up**: Conflicts persist forever even when the underlying engrams are stale, superseded, or have been clarified by newer engrams.

## Data

```
Total engrams: ~840
Engrams with conflicts: 721 (86%)
Total conflict references: 98,464
Top conflict magnet: ENG-2026-0407-004 — 451 conflicts (basically "in tension" with every other engram)
Unique tension pairs: 109,903
```

### 1. Separate tension detection from dedup

Remove `CONFLICTS:` from the dedup prompt entirely. Dedup should only decide ADD/UPDATE/MERGE/NOOP. Contradiction detection is a different cognitive task that needs different inputs.

### 2. Dedicated contradiction pass (lazy, not on every learn)

Run contradiction detection as a **background task** or **on-demand** (`plur_tensions --scan`), not on every `plur_learn`. This avoids the cost and false-positive accumulation of checking every write.

### 3. Pre-filter before LLM

Before asking an LLM "are these contradictory?", filter candidate pairs to only those that:
- Share the same domain or overlapping tags
- Make assertions about the **same subject** (e.g., both mention "protocol fee" or "hub pricing")
- Have opposing polarity (one says "X is Y", another says "X is not Y" or "X is Z")

Simple heuristic: extract subject-predicate from both statements. If subjects don't overlap, skip.

### 4. Confidence scoring

Each tension gets a confidence score (0-1). Only store tensions above a threshold (e.g., 0.7). The `plur_tensions` tool only shows high-confidence items.

### 5. Tension lifecycle

```
detected (auto)
  → confirmed (user agrees it's real)
  → resolved (user picked one, other retired/updated)

detected (auto)
  → dismissed (user says not a real conflict)
  → suppress this pair permanently
```

### 6. Immediate fix: purge existing false positives

The current 109K tensions are noise. Suggest a one-time cleanup:

```bash
plur tensions --purge  # remove all relations.conflicts from engrams
```

Then re-enable detection only after the redesign lands.

## Priority

This is a UX credibility issue. If a user runs `plur_tensions` and sees 109K items, they'll never trust the feature. Better to have zero tensions than 109K false ones.

miles-on-nightshift:
## Summary

Implements Phase 1 of the tension detection redesign from #137. Replaces the false-positive-prone dedup side-effect (109K false positives, purged in #138/#162) with a real contradiction scanner.

- **Three-stage candidate gating** before any LLM call (cheapest gate first):
  1. Scope partition — conservative rule: only identical or global scopes; cross-project pairs (e.g. `project:plur` vs `project:datacore`) are skipped
  2. Domain segment filter — skip pairs whose dotted hierarchies share no segment
  3. Subject-predicate pre-filter — content-token overlap on first clause
- **LLM judge**: structured `CONTRADICTS / CONFIDENCE / REASON` prompt at temp 0.1
- **Threshold 0.7**: only tensions above this confidence are returned
- **On-demand only**: never triggered by `plur_learn` — avoids accumulation
- **Read-only Phase 1**: scan returns results but does not write back to `engram.relations.conflicts` (Phase 2 lifecycle UX is a follow-on)

## Changes

| Package | File | Change |
|---------|------|--------|
| core | `src/tensions.ts` | New scan module (195 lines) |
| core | `test/tensions.test.ts` | 33 unit tests |
| core | `src/index.ts` | Export new module |
| mcp | `src/tools.ts` | `plur_tensions` accepts `scan` action; returns `TensionScanResult` |
| mcp | `test/tensions.test.ts` | 10 MCP-layer tests |
| cli | `src/commands/tensions.ts` | `plur tensions --scan` subcommand |
| cli | `src/index.ts` | Wire command into registry |

## Test plan

- [x] `packages/core/test/tensions.test.ts` — 33 tests pass (scope partition, domain filter, subject overlap, LLM prompt/parse, candidate selection, full scan)
- [x] `packages/mcp/test/tensions.test.ts` — 10 tests pass including purge registration + scan action
- [x] Pre-existing `e2e-remote.test.ts > plur_forget routing` failure is #84 — unchanged by this PR
- [ ] Manual: `npx plur tensions --scan` against a real store with known contradictions

## Out of scope (Phase 2+)
- Persistence: writing detected tensions back to `engram.relations.conflicts`
- Lifecycle UX: confirm/dismiss/resolve flow
- Auto-resolve: connect to correction detection (#115)
- `scanForTensions` mock-LLM integration test

## Risk surface
- Cost: ~$0.001–$0.005 per default scan (max_pairs=50, haiku-class LLM)
- Latency: ~15s sequential at max_pairs=50
- False negatives: synonym-based contradictions may slip past the subject pre-filter (e.g. "fee" vs "cost") — acceptable for Phase 1, addressable in Phase 2

Closes #137.


miles-on-nightshift (commented):
## Review: tensions Phase 1

CI green. Architecture is sound. Ready to merge.

### What I verified

**Core algorithm** (`packages/core/src/tensions.ts`): Three-stage pre-filter before any LLM call is the right design — scope partition → domain segment overlap → subject-predicate token overlap. The ordering is correct (cheapest elimination first). The `subjectsOverlap` heuristic correctly targets the root cause of the 109K false positives: "Protocol fee is 1%" vs "MemPalace is a competitor" share no subject tokens and would be skipped.

**LLM prompt** (`buildContradictionPrompt`): The negative instruction ("Do NOT flag as contradictions: different topics in the same domain, complementary facts") is the key lever. This directly addresses the over-flagging behavior described in #137.

**Phase 1 read-only constraint**: `scanForTensions` returns results but does not write to `engram.relations.conflicts`. This is the correct choice — avoiding accumulation is the whole point. Phase 2 lifecycle UX (confirm/dismiss/resolve) can add persistence once the UX is designed.

**Test coverage**: 33 unit tests (scope partition, domain filter, subject overlap, prompt/parse, full scan) + 10 MCP-layer tests. The mock-fetch pattern for LLM tests is appropriate.

**Backward compat**: Legacy `plur_tensions` (without `scan:true`) still reads from `relations.conflicts` and now includes a `purge_hint` pointing users to the new scan mode. Clean migration path.

### One observation (non-blocking)

`makeHttpLlm` / `getLlmFunction` are defined in both `cli/src/commands/tensions.ts` and (presumably already in) `packages/mcp/src/tools.ts`. This is fine for Phase 1 scope, but worth a follow-on ticket to extract to a shared utility when Phase 2 adds more LLM-dependent scan operations.

### Recommendation

**Merge.** This closes #137 (P1, 109K false positives) with a well-designed replacement that won't accumulate. The manual test (`npx plur tensions --scan` against a real store) is the only unchecked box — acceptable to merge and verify post-ship given the read-only constraint.

🤖 Agent review

crtahlin (approved):
Reviewed the full diff (842 additions across 7 files, 43 tests).

Architecture is sound. Three-stage pre-filtering (scope, domain, subject overlap) directly addresses the root cause of the 109K false positive disaster. Each stage is a pure function, cheapest-first, independently testable.

Code quality is clean. Reuses existing ftsTokenize, proper types, read-only Phase 1 (no write-back to relations.conflicts), good error handling with per-pair non-fatal catch.

Tests are comprehensive. 33 core + 10 MCP covering all filter stages, edge cases, malformed LLM response, confidence clamping, env var fallback.

Minor notes (not blockers, can address in follow-up):
- makeHttpLlm/getLlmFunction duplicated between CLI and MCP
- Sequential LLM calls (~15s at 50 pairs) -- batching is a Phase 2 optimization
- Subject overlap regex is English-specific heuristic

Consolidation tracker: #179

Issue #86, opened 2026-05-07
pin/promote/ingest/report_failure don't route to remote stores


crtahlin:
## Parent

Epic: #79

## Summary

Several mutation operations don't route to remote stores, same root cause as forget (#84) and feedback (#85). These are lower priority because they're less commonly used with remote engrams, but they're still broken.

## Affected Operations

| Operation | Impact | Notes |
|-----------|--------|-------|
| `pin(id)` | Can't pin remote engrams for team-wide always-inject | Medium — affects team workflows |
| `promote(id)` | Can't activate candidate remote engrams | Low — candidates are rare in remote |
| `ingest(content, { scope })` | Can't bulk-extract engrams to remote scope | Medium — useful for onboarding |
| `reportFailure(id)` | Can't report failure on remote procedural engrams | Low — niche feature |
| `episodeToEngram(id, { scope })` | Can't promote episodes to remote scope | Low — niche |

## Fix Pattern

Same as forget/feedback: after local checks fail, iterate remote stores via `_getOrCreateRemoteDriver`. Each operation needs its corresponding server-side API support.

For `ingest` and `episodeToEngram`, the fix is simpler — they create new engrams, so they can reuse the `learn()` routing path (call `learn()` internally with the remote scope).

## Test Cases

```
✓ pin remote engram → pinned flag persists on server
✓ list pinned merges local + remote pinned engrams
✓ promote remote candidate → activated on server
✓ ingest with remote scope → extracted engrams route to server
✓ episodeToEngram with remote scope → new engram on server
✓ reportFailure on remote procedural engram → revision on server
```
---

**Test plan maintenance:** After implementation, update the scenario status in `3-plur/1-tracks/engineering/remote-store-test-plan.md` — change the relevant Status column and section 9 summary counts. Add new scenarios if discovered.

miles-on-nightshift:
**Triage — 2026-05-13**

Audited the current implementation. Status is split:

**Already fixed (in main):**
- `ingest(content, { scope })` — routes to remote via `learnRouted()` (added per #86, comment in `index.ts` line ~1356)
- `episodeToEngram(episodeId, { scope })` — same pattern, also routed via `learnRouted()` (line ~1519)

Both create new engrams, so they route cleanly through the existing `learn()` path. No server-side changes needed.

**Still missing:**
[...]

PR #175, opened 2026-05-17
test(mcp): extend E2E remote routing coverage — forget, feedback, config reload

miles-on-nightshift:
## Summary

Extends the Level-3 MCP e2e suite from #82 with additional routing scenarios not covered by the original merge.

## New test coverage

| Test | Status | Notes |
|------|--------|-------|
| `plur_learn` routes to remote via MCP | PASSING | Inherited from #82 baseline |
| `plur_learn` returned ID matches server-assigned ID | PASSING | |
| Local YAML not polluted when routing to remote | PASSING | |
| `plur_forget` with server ID reaches stub | **FAILING** | Documents bug: `getById()` only searches local store; remote-scoped server IDs return null. Fix tracked in #86. |
| `plur_feedback` with server ID reaches stub | PASSING | |
| Remote store appears in `plur_stores_list` with correct count | PASSING | |
| Remote store config survives MCP server restart | PASSING | |

## Known failure: `plur_forget` routing

The `plur_forget` handler calls `plur.getById(id)` which only searches the local engram store. When `plur_learn` writes to a remote scope, the server-assigned ID is not reflected locally. The test is intentionally left failing as a pinned regression guard — it will pass when the `plur_forget` handler is updated to route by server ID (see #86).

## Infrastructure

- `packages/mcp/test/helpers/stub-server.ts` — lightweight in-process HTTP stub for MCP e2e testing (mirrors `packages/core/test/helpers/stub-server.ts`; keep in sync when RemoteStore adds endpoints)

## Test plan

- [x] 6/7 tests passing
- [ ] `plur_forget` routing fix (#86) — will make 7/7

### DIFF
--- a/packages/mcp/src/tools.ts
+++ b/packages/mcp/src/tools.ts
@@ -461,10 +461,16 @@ export function getToolDefinitions(): ToolDefinition[] {
       handler: async (args, plur) => {
         if (args.id) {
           const engram = plur.getById(args.id as string)
-          if (!engram) throw new Error(`Engram not found: ${args.id}`)
-          if (engram.status === 'retired') return { success: false, error: `Already retired: ${args.id}` }
+          if (engram) {
+            if (engram.status === 'retired') return { success: false, error: `Already retired: ${args.id}` }
+            await plur.forget(args.id as string)
+            return { success: true, retired: { id: engram.id, statement: engram.statement } }
+          }
+          // Not in local store — fall through to plur.forget() which routes to
+          // remote stores (with prefix stripping per #86 / PR #186). Throws
+          // "Engram not found" if it's nowhere.
           await plur.forget(args.id as string)
-          return { success: true, retired: { id: engram.id, statement: engram.statement } }
+          return { success: true, retired: { id: args.id as string } }
         }
         if (args.search) {
           const matches = plur.recall(args.search as string, { limit: 100 })
--- a/packages/mcp/test/e2e-remote.test.ts
+++ b/packages/mcp/test/e2e-remote.test.ts
@@ -1,0 +1,249 @@
+/**
+ * MCP Level-3 e2e tests: full MCP → core → RemoteStore pipeline.
+ *
+ * Why this exists: in 0.9.6, `new Plur().learn(stmt, { scope: teamScope })`
+ * routed correctly to the remote store, but `plur_learn` via MCP wrote to
+ * local instead. Core-level integration tests (Level 2) couldn't catch this
+ * because they bypass the MCP initialization path entirely.
+ *
+ * These tests wire a real MCP server (via InMemoryTransport, same code path
+ * as the stdio bundle) against an in-process HTTP stub, then assert server-
+ * side state directly — not just what the MCP response says.
+ *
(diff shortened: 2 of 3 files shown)


crtahlin:
## Review

Test-only PR — 7 MCP-level e2e tests for remote store routing. Well-structured: uses `InMemoryTransport` against an in-process HTTP stub, asserts server-side state directly. Documents the `plur_forget` routing bug (#86) as a pinned failing test.

### Should fix

1. **Duplicated stub-server** — `packages/mcp/test/helpers/stub-server.ts` is a stripped copy of `packages/core/test/helpers/stub-server.ts` (184 vs 237 lines). The MCP version adds PATCH and feedback endpoints the core version doesn't have, and removes JSDoc. Comment says "keep in sync" but nothing enforces it. Consider extracting to a shared test fixture or importing from core.

2. **`plur_forget` test status unclear** — PR body says FAILING but the test has no `.failing` or `.skip` annotation. If #86 isn't fixed yet, this will fail CI. Either add `it.fails(...)` to pin the expected failure, or confirm the fix has landed.

3. **No `afterEach` cleanup** — `dir` is created via `mkdtempSync` in `beforeEach` but never cleaned up. The existing `e2e.test.ts` does `afterEach(() => rmSync(dir, { recursive: true }))`. Should match.

### What's good

- Asserting server-side state via `stub.getEngram()` rather than just MCP responses — catches the exact class of bug that motivated this suite
- Config persistence test (remote store survives restart) covers a real failure mode
- Clean focused test structure

crtahlin:
No critical issues — merge when CI passes. Cleanup items tracked in #187.

crtahlin:
## Merge dependency: requires #186 first

This test (`plur_forget with server ID reaches server and retires`) fails because `plur_forget` calls `plur.getById()` which only checks the local store. For server-prefixed IDs (`ENG-SRV-*`), it throws `Error: Engram not found` — the remote routing fix in #186 is needed first.

**Merge order:** #186 → rebase #175 → should pass.

### Separate issue: `callResult()` error handling

The test helper at line 68 blindly `JSON.parse`s the MCP response text:
```typescript
function callResult(raw) {
  return JSON.parse((raw.content as any)[0].text)
}
```

When a tool returns `isError: true`, the text is `"Error: Engram not found..."` (not JSON), producing a confusing `SyntaxError` instead of a clear failure message. This helper should check `isError` first:

```typescript
function callResult(raw) {
  const entry = (raw.content as any)[0]
  if (raw.isError) throw new Error(`Tool returned error: ${entry.text}`)
  return JSON.parse(entry.text)
}
```

This doesn't change test outcomes but makes failures diagnosable — you see "Tool returned error: Engram not found: ENG-SRV-001" instead of "SyntaxError: Unexpected token 'E'".

plur9 (approved):
Pushed additional commit (cee5ed6) fixing the previously-pinned-failing plur_forget test: the MCP handler now falls back to plur.forget() when the local getById preflight misses, leveraging #186's remote prefix stripping for routing. All 7 e2e remote routing tests pass; full MCP suite 101/101. Closes the MCP-handler portion of #86 (pin/promote remain pending RemoteStore.update() per #186 notes). Squashing.

Issue #85, opened 2026-05-07
feedback() does not route to remote stores


crtahlin:
## Parent

Epic: #79

## Summary

`plur_feedback` with a server-assigned engram ID silently fails — it searches local stores, doesn't find the engram, and falls through to searching namespaced store IDs. Remote stores are never checked.

**Confirmed broken by code analysis** (same pattern as forget() — see #84).

## Root Cause

`feedback()` at `index.ts:826-884`:

```typescript
feedback(id: string, signal: 'positive' | 'negative' | 'neutral'): void {
  // 1. Check primary local store → not found
  // 2. Check _findEngramStore(id) → skips remote
  // 3. Falls through silently (doesn't even throw!)
}
```

Note: unlike `forget()` which throws "Engram not found", `feedback()` appears to silently return without error if the engram isn't found in any local store. This is worse — the user thinks the feedback was applied.

## Impact

Team members rating remote engrams (positive/negative) have no effect. The feedback loop — PLUR's core differentiator — doesn't work for shared enterprise engrams. Injection quality for team engrams can never improve.

## Proposed Fix

Same pattern as forget() fix — after local checks, iterate remote stores:

```typescript
// After local + store checks (line 884):
for (const entry of (this.config.stores ?? [])) {
  if (!entry.url) continue
  const driver = this._getOrCreateRemoteDriver(entry)
  const found = await driver.getById(id)
  if (found) {
    // POST feedback to server API
    // Server needs a PATCH /api/v1/engrams/:id/feedback endpoint
    // or fold into existing PATCH /api/v1/engrams/:id
    return
  }
}
```

## Server-Side Dependency

The enterprise server may need a new endpoint:
- `PATCH /api/v1/engrams/:id/feedback` with `{ signal: "positive"|"negative"|"neutral" }`
- Or extend the existing `PATCH /api/v1/engrams/:id` to accept feedback_signals updates

Check if this endpoint exists before implementing the client-side routing.

## API Break

Same sync→async issue as forget(). See sync→async migration issue.

## Test Cases

```
✓ feedback on remote engram by server ID → weight updated on server
✓ batch feedback with mixed local + remote IDs → each routed correctly
✓ negative feedback on remote engram → retrieval_strength decreases on server
✓ positive feedback promotes commitment on server
✓ feedback on remote when server is down → error, not silent success
```
---

[...]

PR #186, opened 2026-05-19
fix(core): strip store prefix before remote getById/feedback/remove

### LINKED ISSUE #84: forget() does not route to remote stores

crtahlin:
## Summary

Fixes the ID prefix translation bug that caused `plur_feedback` and `plur_forget` to fail on remote store engrams with "Engram not found".

## Root cause

`_loadAllEngrams()` adds a namespace prefix (e.g. `ENG-GPL-`) to remote engrams for local deduplication. MCP tools surface this prefixed ID to users. When the user passes it back to `feedback()` or `forget()`, those methods called `driver.getById(prefixedId)` — but the enterprise server only knows the original unprefixed ID. Result: 404, "Engram not found".

**Proof from live testing (May 19):**
```
feedback("ENG-2026-05-19-004", "positive")     → SUCCESS
feedback("ENG-GPL-2026-05-19-004", "positive")  → FAILED - Engram not found
```

## Fix

Added `_stripRemotePrefix(id, scope)` helper that reverses the namespace before sending to the remote API. Applied in both `feedback()` and `forget()` remote fallback paths.

The helper uses the same `storePrefix()` + regex pattern that `_findEngramStore()` uses for local stores — consistent behavior across local and remote paths.

## Tests added

**`store-prefix.test.ts`** (11 tests):
- `storePrefix()` for known scopes (group:plur/plur-ai/engineering → GPL, project:plur → PPL, etc.)
- ID prefix round-trip: add prefix → strip prefix → original ID
- Wrong scope doesn't strip, unprefixed ID returned unchanged

**`remote-integration.test.ts`** (3 tests against stub server):
- `feedback()` works with prefixed ID from `_loadAllEngrams`
- `forget()` works with prefixed ID from `_loadAllEngrams`
- `feedback()` still works with unprefixed server ID (backward compat)

## Test plan

- [x] 951 tests pass (+14 net from today's work)
- [x] New tests exercise the exact failure path from #86
- [x] Full pnpm test — zero failures

Partially addresses #86 (feedback + forget fixed). Pin/promote remain open — they need a PATCH endpoint on the enterprise server + RemoteStore.update() method.

Closes #84 (properly this time) and #85 (properly this time).

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -364,6 +364,22 @@ export class Plur {
     return null
   }
 
+  /**
+   * Strip the store namespace prefix from an ID before sending to a remote server.
+   * _loadAllEngrams adds ENG-{PREFIX}- to avoid local ID collisions; the remote
+   * server only knows the original ID. If the ID doesn't match this store's prefix,
+   * return it unchanged (it may belong to a different store or be unprefixed).
+   * See: https://github.com/plur-ai/plur/issues/86
+   */
+  private _stripRemotePrefix(id: string, scope: string): string {
+    const prefix = storePrefix(scope)
+    const nsPattern = new RegExp(`^(ENG|ABS|META)-${prefix}-`)
+    if (nsPattern.test(id)) {
+      return id.replace(nsPattern, '$1-')
+    }
+    return id
+  }
+
   /** Content hash fast-path dedup. Scope-aware: same statement in a different scope is a promotion, not a duplicate. */
   private _hashDedup(statement: string, engrams: Engram[], scope?: string): Engram | null {
     const hash = computeContentHash(statement)
@@ -1095,18 +1111,21 @@ export class Plur {
 
     // Check remote stores — the engram may live on an enterprise server.
     // See: https://github.com/plur-ai/plur/issues/85
+    // The ID may be prefixed (ENG-GPL-...) from _loadAllEngrams namespacing.
+    // Strip the prefix before querying the remote server. See: #86
     for (const entry of (this.config.stores ?? [])) {
       if (!entry.url) continue
+      const serverId = this._stripRemotePrefix(id, entry.scope)
       if (entry.readonly === true) {
         const roDriver = this._getRemoteDriver({ url: entry.url, token: entry.token, scope: entry.scope })
-        const roFound = await roDriver.getById(id)
+        const roFound = await roDriver.getById(serverId)
         if (roFound) throw new Error('Engram is in a readonly store')
         continue
       }
       const driver = this._getRemoteDriver({ url: entry.url, token: entry.token, scope: entry.scope })
-      const found = await driver.getById(id)
+      const found = await driver.getById(serverId)
       if (found) {
-        await driver.feedback(id, signal)
+        await driver.feedback(serverId, signal)
         appendHistory(this.paths.root, {
           event: 'feedback_received',
           engram_id: id,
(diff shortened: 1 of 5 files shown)


crtahlin:
## CI is green — ready for review

Rebased on main, both Node 20 and Node 22 passing.

**Note:** PR #175 (e2e remote routing tests) depends on this PR. The test `plur_forget with server ID reaches server and retires` fails without the prefix stripping fix from this PR — `getById()` can't find `ENG-SRV-*` IDs in the local store.

**Merge order:** this PR first → rebase #175 on top → #175 should pass.

plur9 (approved):
LGTM — proper fix for the prefix translation bug. The _stripRemotePrefix helper mirrors _findEngramStore's logic so local/remote routing stay consistent. 11 store-prefix tests + 3 remote-integration tests cover the exact failure path. Closes #84/#85, partial #86. Squashing.

PR #208, opened 2026-05-21
fix(mcp): return structured JSON for tool errors (#207)

### LINKED ISSUE #207: MCP server: tool errors should return structured JSON, not plain text
## Problem

When an MCP tool handler throws, `server.ts` catches it and returns:

```typescript
{
  content: [{ type: 'text', text: `Error: ${message}` }],
  isError: true,
}
```

The `text` field is a plain string like `"Error: Engram not found: ENG-SRV-001"`. This causes two problems:

1. **Consumers that `JSON.parse` the response text get a confusing `SyntaxError`** instead of a clear error message. This bit PR #175's `callResult()` helper — the test failure was `SyntaxError: Unexpected token 'E'` instead of the actual issue.

2. **Error responses aren't machine-parseable.** Callers can't distinguish error types (not found vs validation vs internal) without string matching.

## Suggested Fix

Return structured JSON even on errors:

```typescript
catch (err: any) {
  const message = err?.message ?? String(err)
  return {
    content: [{ type: 'text', text: JSON.stringify({ error: message, success: false }) }],
    isError: true,
  }
}
```

This way:
- `JSON.parse` always works on `content[0].text` — callers don't need to check `isError` first
- Error type can be extended later (`{ error: message, code: 'NOT_FOUND' }`)
- Existing callers that check `isError` still work (field is still set)

## Also fix test helper

`callResult()` in `e2e-remote.test.ts` should check `isError` before parsing:

```typescript
function callResult(raw) {
  const entry = (raw.content as any)[0]
  if (raw.isError) throw new Error(`Tool returned error: ${entry.text}`)
  return JSON.parse(entry.text)
}
```

## Related

- #186 — the root cause of the `Engram not found` error (missing remote ID routing)
- #175 — the test that surfaced this (SyntaxError masking the real failure)

crtahlin:
## Summary

- MCP tool error responses now return `{ error: message, success: false }` as JSON instead of plain text strings like `"Error: Engram not found"`
- `JSON.parse` on `content[0].text` now always works regardless of success/failure
- Three error paths fixed: unknown tool, validation failure, handler exception
- Added test verifying error responses are valid JSON with `success: false`

## Context

PR #175's `callResult()` helper does `JSON.parse(content[0].text)`. When a tool returns `isError: true`, the text was `"Error: ..."` (not JSON), producing a confusing `SyntaxError` instead of the actual error. This fix makes error responses machine-parseable.

## Test plan

- [x] `pnpm build` — passes
- [x] Existing `returns isError for unknown tools` test — still passes (`.toContain('Unknown tool')` works with JSON)
- [x] New test: `error responses are valid JSON with success:false` — passes

Closes #207.

### DIFF
--- a/packages/mcp/src/server.ts
+++ b/packages/mcp/src/server.ts
@@ -160,7 +160,7 @@ export async function createServer(plur?: Plur): Promise<Server> {
     const tool = tools.find(t => t.name === request.params.name)
     if (!tool) {
       return {
-        content: [{ type: 'text', text: `Unknown tool: ${request.params.name}` }],
+        content: [{ type: 'text', text: JSON.stringify({ error: `Unknown tool: ${request.params.name}`, success: false }) }],
         isError: true,
       }
     }
@@ -181,8 +181,9 @@ export async function createServer(plur?: Plur): Promise<Server> {
         }
         const parsed = z.object(shape).passthrough().safeParse(args)
         if (!parsed.success) {
+          const details = parsed.error.issues.map(i => `${i.path.join('.')}: ${i.message}`).join(', ')
           return {
-            content: [{ type: 'text', text: `Invalid arguments: ${parsed.error.issues.map(i => `${i.path.join('.')}: ${i.message}`).join(', ')}` }],
+            content: [{ type: 'text', text: JSON.stringify({ error: `Invalid arguments: ${details}`, success: false }) }],
             isError: true,
           }
         }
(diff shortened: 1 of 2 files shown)


plur9 (approved):
LGTM — small, focused fix. Wraps three error paths (unknown tool, validation, handler exception) in JSON-stringified {error, success:false}. Unblocks PR #175's callResult() helper which uses JSON.parse on every result. Closes #207. Squashing.

Issue #185, opened 2026-05-19
Test gaps: remote store ID prefix round-trip and mutation routing


crtahlin:
## Context

The ID prefix translation bug discovered in #86 (comment May 19) went undetected because existing tests mock at the wrong layer. `remote-routing.test.ts` mocks RemoteStore at the class level and passes server-side IDs directly -- it never exercises the `storePrefix` → `GPL-` namespace that `_loadAllEngrams` adds. The integration tests (`remote-integration.test.ts`) test learn/recall but not the mutation round-trip with prefixed IDs.

PR #175 (open) started MCP-level e2e remote tests but has 1 failing test that documents (not fixes) the forget routing bug.

## What exists today

| File | Level | Tests | What it covers | What it misses |
|------|-------|-------|---------------|----------------|
| `core/test/remote-routing.test.ts` | Unit (mocked RemoteStore) | 19 | learn/forget/feedback routing decisions, outbox, readonly, errors | ID prefix translation -- mocks skip the real HTTP layer |
| `core/test/remote-integration.test.ts` | Integration (stub server) | 11 | RemoteStore CRUD against stub, Plur learn routing | Mutation round-trip (learn → load → prefixed ID → feedback/forget) |
| `core/test/remote-store-cache.test.ts` | Unit | 2 | Optimistic cache insert, cold-cache append | Cold-start getById failure, listStores count accuracy |
| `mcp/test/e2e.test.ts` | MCP (local only) | 5 | Full local lifecycle | No remote store tests |
| PR#175 `mcp/test/e2e-remote.test.ts` | MCP (stub server) | 7 (1 failing) | learn routing, forget (failing), feedback, stores_list, restart | ID prefix round-trip, pin/promote |

### 1. ID prefix round-trip (the critical gap)

**Where:** `core/test/remote-integration.test.ts` (extend existing suite)

The full round-trip that failed in production:
```
learn(scope: remote) → server assigns ID "ENG-2026-05-19-004"
  → _loadAllEngrams() returns it as "ENG-GPL-2026-05-19-004"
  → user passes "ENG-GPL-2026-05-19-004" to feedback()
  → feedback() must strip prefix before calling driver.getById()
  → driver.getById("ENG-2026-05-19-004") succeeds on server
```

Tests needed:
- `learn() with remote scope → load → getById with prefixed ID → succeeds`
- `feedback() with prefixed ID → reaches stub server with unprefixed ID`
[...]

Issue #184, opened 2026-05-19
plur_stores_list reports engram_count: 0 for remote stores that contain engrams


crtahlin:
## Summary

`plur_stores_list` returns `engram_count: 0` for remote stores even when those stores hold engrams that are retrievable via `plur_recall` with the same scope filter.

## Environment

- `@plur-ai/mcp` v0.9.9 (npx)
- Remote store: `https://plur.datafund.io/sse`, scope `group:plur/plur-ai/engineering`, `shared: true`

## Reproduction

1. Register a remote store (e.g. PLUR Enterprise team scope):
   ```
   plur_stores_add scope="group:plur/plur-ai/engineering" url="https://plur.datafund.io/sse" token=<...>
   ```
2. Ensure the remote store has at least one engram in that scope (confirmed below via recall).
3. Call `plur_stores_list`.
4. Call `plur_recall query="engineering" scope="group:plur/plur-ai/engineering"`.

## Expected

`plur_stores_list` reports the actual engram count for the remote store, matching what recall can retrieve.

## Actual

`plur_stores_list` output (remote entry):
```json
{
  "url": "https://plur.datafund.io/sse",
  "scope": "group:plur/plur-ai/engineering",
  "shared": true,
  "readonly": false,
  "engram_count": 0
}
```

`plur_recall` against the same scope returns real engrams, e.g.:
- `ENG-GPL-2026-05-07-008` (scope `group:plur/plur-ai/engineering`)
- `ENG-GPL-2026-05-11-004` (scope `group:plur/plur-ai/engineering`)

Local stores in the same output show correct counts (e.g. `global` 1038, project stores at 67 each).

## Root cause

`listStores()` (index.ts:1866) calls `_loadRemoteCached(s)` to count remote engrams. But `_loadRemoteCached` (index.ts:281) is **synchronous** — it reads from `driver.cache` and triggers an async `driver.load()` in the background:

```typescript
private _loadRemoteCached(store: StoreEntry): Engram[] {
  const driver = this._getRemoteDriver(...)
  const cached = (driver as unknown as { cache: ... }).cache
  void driver.load().catch(() => {})  // async — hasn't resolved yet
  return cached?.engrams ?? []        // returns [] on first call
}
```

The code comment (lines 272-278) documents this explicitly:

> The first call after server start returns [] for that store; the call after the first refresh sees the data.

[...]

Issue #107, opened 2026-05-10
Content-addressed deduplication for ADD-only ingestion


tris-on-hermes:
## Why

Users repeatedly store the same correction, preference, or pattern. Without deduplication, the store bloats, retrieval quality degrades, and pack installs create duplicate engrams across scopes. We need an ingestion pipeline that is append-only by design and deduplicates by content hash.

## What

Make PLUR ingestion content-addressed:

1. Every engram body is hashed (SHA-256) at ingestion time.
2. If the hash exists, increment a `reference_count` and append new `source` metadata instead of creating a duplicate row.
3. If the hash is new, store the engram and index it.
4. Deletion becomes a `reference_count` decrement; the row is only removed when count reaches zero and a compaction pass runs.

### Schema changes

```sql
ALTER TABLE engrams ADD COLUMN content_hash TEXT UNIQUE;
ALTER TABLE engrams ADD COLUMN reference_count INTEGER DEFAULT 1;
ALTER TABLE engrams ADD COLUMN sources JSON; -- [{scope, session_id, stored_at}]

CREATE UNIQUE INDEX idx_engrams_hash ON engrams(content_hash);
```

### Ingestion flow

```ts
function ingest(engram: EngramInput): EngramRef {
  const hash = sha256(engram.body);
  const existing = db.prepare('SELECT id, reference_count FROM engrams WHERE content_hash = ?').get(hash);
  if (existing) {
    db.prepare('UPDATE engrams SET reference_count = reference_count + 1, sources = json_insert(sources, ... ) WHERE id = ?')
      .run(existing.id);
    return { id: existing.id, deduplicated: true };
  }
  // insert new
  return storeNew({ ...engram, content_hash: hash, reference_count: 1 });
}
```

### Impact on existing APIs

- `plur.learn()` — no breaking change; returns `{ id, deduplicated?: boolean }`
- `plur.forget()` — decrements `reference_count`; only deletes when zero
- `plur.recall()` — unchanged; deduplication is invisible to retrieval
- Pack install — deduplicates against existing engrams in target scope, preventing bloat

### Edge cases

- **Feedback divergence**: two scopes store the same body but rate it differently. Keep one canonical row with scope-weighted feedback scores stored in `scope_feedback` JSON column.
- **Pack provenance**: content hash must include pack version in the canonical source metadata.

## Acceptance Criteria

- [ ] Identical `plur.learn()` calls within the same scope create only one row
[...]

Issue #93, opened 2026-05-07
Store management validation: invalid URL, expired token, duplicate scope


crtahlin:
## Parent

Epic: #78 (test coverage gaps)

## Summary

`plur_stores_add` accepts any URL and token without validation. Edge cases that should be tested and handled:

| Scenario | Current Behavior | Expected |
|----------|-----------------|----------|
| Invalid URL (not a URL) | Registered, fails silently on first use | Reject at registration time |
| Unreachable URL | Registered, fails silently | Warn at registration, allow (server might come up later) |
| Expired/invalid token | Registered, 401 on first use | Test connection at registration, warn if auth fails |
| Duplicate scope (same scope, new URL) | Both registered? | Overwrite with warning, or reject with "already exists" |
| Add readonly store | Works | Works (verify writes are blocked) |

## Priority

P3 — these are UX improvements, not blocking issues. The current behavior (register anything, fail on use) works for the pilot but will confuse users at scale.

## Test Cases

```
✓ stores_add with malformed URL → immediate error
✓ stores_add with unreachable URL → warning, store registered
✓ stores_add with invalid token → warning about auth failure
✓ stores_add with duplicate scope → clear behavior (overwrite or reject)
✓ stores_add as readonly → writes blocked with clear message
```

PR #230, opened 2026-05-26
feat: surface remote store scopes in session_start (#229)

### LINKED ISSUE #229: Feature: surface remote store scopes in session_start so AI callers route engrams correctly
## Problem

When a remote enterprise store is configured with a specific scope (e.g. \`group:plur/plur-ai/engineering\`), the AI caller (Claude/GPT) has no way to know that this scope exists or that it should use it. Most \`plur_learn\` calls omit the \`scope\` parameter (defaulting to \`global\`) or use ad-hoc scopes like \`datacore.nightshift\`, \`swarm.core\`, etc. — none of which match the enterprise store.

Result: engrams that SHOULD flow to the enterprise store stay local. In our case, 1,338 local engrams but only 56 reached the enterprise store since May 6 — and only when someone explicitly remembered to pass the exact scope string.

The exact-match routing in \`learnRouted()\` is correct by design — the problem is that the AI caller is never told which scopes route to remote stores.

## Current Behavior

1. \`plur_session_start\` returns: session_id, injected engrams, store_stats (counts only)
2. \`plur_learn\` accepts optional \`scope\` parameter — no guidance on what values are available
3. AI caller defaults to \`global\` or invents scopes based on domain context
4. \`learnRouted()\` does exact match — \`global\` !== \`group:plur/plur-ai/engineering\` → stays local

### Option A: Surface remote scopes in session_start response (recommended)

Add a \`remote_scopes\` field to the \`plur_session_start\` response:

```typescript
// In session_start handler (tools.ts ~line 1082):
const remote_scopes = (plur.config.stores ?? [])
  .filter(s => s.url && !s.readonly)
  .map(s => ({ scope: s.scope, url: s.url }))

return {
  session_id,
  engrams,
  store_stats,
  remote_scopes,  // NEW
  guide: remote_scopes.length > 0
    ? \`Session started. Remote stores available — use scope "\${remote_scopes[0].scope}" in plur_learn to write to enterprise store.\`
    : 'Session started.',
}
```

The AI caller sees the available remote scopes in the response and knows to use them for relevant engrams. The guide text makes it explicit.

### Option B: Session-level default scope (complementary)

Allow \`plur_session_start\` to accept a \`default_scope\` parameter that gets applied to all subsequent \`plur_learn\` calls in that session when the caller doesn't specify a scope:

```typescript
// In session_start:
plur.setSessionScope(args.default_scope ?? remote_scopes[0]?.scope ?? 'global')

// In learn handler:
const context = {
  scope: args.scope ?? plur.getSessionScope(),  // falls back to session default
  ...
}
```

This way, a single \`plur_session_start(task: "plur engineering work", default_scope: "group:plur/plur-ai/engineering")\` would route all engrams from that session to the enterprise store without the caller needing to remember the scope on each \`plur_learn\` call.

### Option C: Auto-detect from project stores (future)

If the current working directory is inside a project that has a \`.plur/\` store with a matching scope, auto-promote the scope. This is more complex and can come later.

## Recommendation

Implement Option A first (minimal, non-breaking). Option B as a follow-up. This makes the system self-documenting — the AI caller learns available scopes from the session response rather than relying on external documentation.

## Evidence

From a live session (2026-05-26):
- Engram scope distribution: 782 \`global\`, 6 \`team:plur/plur-ai/engineering\`, 3 \`group:plur/plur-ai/engineering\`, rest scattered across project-level scopes
- Enterprise store received exactly the 3 engrams with matching \`group:\` scope + 53 from initial setup
- The 6 \`team:\` prefix engrams never routed because the config uses \`group:\` prefix (separate inconsistency to clean up)

## Related

- #228 — stale remote store count in \`plur_stores_list\`
- The \`team:\` vs \`group:\` prefix inconsistency should be resolved as part of this work

crtahlin:
## Summary

- AI callers (Claude/GPT) had no way to discover which scope routes engrams to the enterprise store
- Most `plur_learn` calls used `global` scope, bypassing remote routing entirely
- Result: 1,338 local engrams but only 56 reached the enterprise store since May 6

## Changes

**Core (`packages/core/src/index.ts`):**
- `getWritableRemoteScopes()` — returns writable remote store scopes for caller guidance
- `setSessionScope(scope)` / `getSessionScope()` — session-level default scope
- `learn()` and `learnRouted()` scope fallback chain: `explicit > session > 'global'`

**MCP (`packages/mcp/src/tools.ts`):**
- `plur_session_start` accepts optional `default_scope` parameter
- Auto-detects when exactly one writable remote store exists and sets it as session default
- Response includes `remote_scopes` array and updated `guide` text
- AI caller now sees: `"Session default scope: "group:plur/plur-ai/engineering" — plur_learn calls without explicit scope will route to the enterprise store."`

**Tests (`packages/core/test/session-scope.test.ts`):**
- 13 new tests: getter/setter roundtrip, fallback chain, remote routing via session scope, writable scope filtering

## Backward Compatible

- `scope` parameter on `plur_learn` remains optional
- Explicit scope always takes priority over session scope
- If no remote stores configured, behavior is identical to before
- No changes to Claw package

## Test plan

- [x] All 1046 tests pass (13 new + 1033 existing)
- [x] Build succeeds for all packages
- [ ] Manual test: `plur_session_start` returns `remote_scopes` when enterprise store configured
- [ ] Manual test: `plur_learn` without scope routes to enterprise store via session default

Closes #229

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -2247,14 +2247,29 @@ Generate an improved version of the procedure that prevents this failure. Return
   /**
    * Pre-load all remote store caches so subsequent sync reads see data.
    * Call once before injection to avoid the cold-start race (#235).
+   *
+   * Each remote load races against a 5-second timeout — a single hung or
+   * slow remote must not block session_start indefinitely. Same pattern as
+   * listStoresAsync (#184). clearTimeout on the success path prevents
+   * accumulating dangling timers in the long-lived MCP server process.
    */
   async warmRemoteCaches(): Promise<void> {
     const stores = this.config.stores ?? []
     const remoteStores = stores.filter(s => s.url)
+    const REMOTE_LOAD_TIMEOUT_MS = 5000
     await Promise.all(
       remoteStores.map(s => {
         const driver = this._getRemoteDriver({ url: s.url!, token: s.token, scope: s.scope })
-        return driver.load().catch(() => { /* errors logged inside RemoteStore */ })
+        let timeoutHandle: ReturnType<typeof setTimeout> | undefined
+        return Promise.race([
+          driver.load().finally(() => { if (timeoutHandle) clearTimeout(timeoutHandle) }),
+          new Promise<never>((_, reject) => {
+            timeoutHandle = setTimeout(
+              () => reject(new Error(`remote warm timeout (${REMOTE_LOAD_TIMEOUT_MS}ms)`)),
+              REMOTE_LOAD_TIMEOUT_MS,
+            )
+          }),
+        ]).catch(() => { /* errors logged inside RemoteStore; timeout swallowed */ })
       }),
     )
   }
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -2244,6 +2244,21 @@ Generate an improved version of the procedure that prevents this failure. Return
     return [this._primaryStoreRow(), ...additional]
   }
 
+  /**
+   * Pre-load all remote store caches so subsequent sync reads see data.
+   * Call once before injection to avoid the cold-start race (#235).
+   */
+  async warmRemoteCaches(): Promise<void> {
+    const stores = this.config.stores ?? []
+    const remoteStores = stores.filter(s => s.url)
+    await Promise.all(
+      remoteStores.map(s => {
+        const driver = this._getRemoteDriver({ url: s.url!, token: s.token, scope: s.scope })
+        return driver.load().catch(() => { /* errors logged inside RemoteStore */ })
+      }),
+    )
+  }
+
   /** Return writable remote store scopes for AI caller guidance. */
   getWritableRemoteScopes(): Array<{ scope: string; url: string }> {
     return (this.config.stores ?? [])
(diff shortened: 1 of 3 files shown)


plur9 (dismissed):
## Rebased + addressed two review concerns

Pushed two commits on top of your branch:

1. **Rebase on main** — picks up 0.9.11 + 0.9.12 changes (ref-count dedup #107, remote PATCH routing #185/#86, store validation #93, MCP shim #234). Git auto-resolved cleanly — the scope-assignment lines in learn() and learnRouted() merged with the surrounding new code without conflict.

2. **bdbd49a fix(core,mcp): cross-session leak + warmRemoteCaches timeout** — two correctness fixes from review:
   - **Cross-session _sessionScope leak**: handler now always resets _sessionScope at top of session_start (single line: \`plur.setSessionScope(default_scope)\` which is null when arg unset). Prevents default_scope from session A leaking into session B in the long-lived MCP process.
   - **warmRemoteCaches 5s timeout**: lifted the Promise.race timeout pattern from listStoresAsync (#184). Without this, a hung remote blocks session_start indefinitely.

2 new tests cover both fixes. Full suite: **1111 passing**, 0 regressions on the rebase. Your original 16 session-scope tests all still pass.

Approving — ready to squash-merge.

---

## Follow-up question raised during review (not blocking)

@gregor asked: scope can change mid-session (narrow / expand / switch). Current PR only sets at session_start. Filing as a separate feature request — would be a `plur_set_scope` MCP tool that re-uses your existing `setSessionScope()` core method.

PR #248, opened 2026-05-29
fix: isPlurConfigured() no longer falls back to global settings

### LINKED ISSUE #247: isPlurConfigured() always returns true when hooks are installed globally
### Problem

The session guard (`hook-session-guard`) is designed to silently pass through for projects that don't use Plur (#95). It calls `isPlurConfigured()` to detect this. However, `isPlurConfigured()` has a fallback that checks `~/.claude/settings.json` — and since `plur init` (especially `--global`) registers the plur MCP server there, the function **always returns true** for any user who has completed setup.

This means the session guard blocks all tool calls in every project until `plur_session_start` is called — even projects that have nothing to do with Plur. The deadlock fallback (#199) eventually lets tools through after 5 blocked calls, but by then the conversation is polluted with deny messages and forced ToolSearch calls.

### Root cause

In `packages/cli/src/lib/plur-configured.ts`:

```typescript
export function isPlurConfigured(cwd, home): boolean {
  // ... walks up from cwd checking .mcp.json and .claude/settings.json ...

  // This line is the bug:
  return configHasPlur(join(home, '.claude', 'settings.json'))
}
```

The walk-up correctly checks project-level config. But the fallback to `~/.claude/settings.json` defeats the purpose — that file has `mcpServers.plur` because `plur init` put it there (either the MCP entry directly via `--global`, or even just from a previous init run). The guard can never distinguish "plur is configured for this project" from "plur is installed on this machine."

### Secondary issue: injection hooks lack the guard entirely

The enforcement hooks (`hook-session-guard`, `hook-session-remind`, `hook-session-mark`) all use `isPlurConfigured()`. But the injection hooks do not:

- `hook-inject` — no guard
- `hook-observe` — no guard
- `hook-learn-check` — no guard

When installed globally (`plur init --global`), these hooks fire in every project too. They don't block tool calls, but they add latency and may produce errors when the plur MCP server isn't available in a non-plur project context.

### Who is affected

Any user who has run `plur init --global`, or who has the plur MCP server registered in `~/.claude/settings.json` for any reason. Opening Claude Code in a project without plur configured will trigger the session guard and block work.

### Proposed fix

1. **Remove the global fallback** from `isPlurConfigured()`. The walk-up from cwd already catches project-level config at any depth. If no project in the directory tree has plur in `.mcp.json` or `.claude/settings.json` or `.claude/settings.local.json`, return `false`.

```typescript
export function isPlurConfigured(cwd = process.cwd()): boolean {
  let dir = resolve(cwd)
  while (true) {
    if (configHasPlur(join(dir, '.mcp.json'))) return true
    if (configHasPlur(join(dir, '.claude', 'settings.json'))) return true
    if (configHasPlur(join(dir, '.claude', 'settings.local.json'))) return true
    const parent = dirname(dir)
    if (parent === dir) break
    dir = parent
  }
  return false  // Do NOT fall back to ~/.claude/settings.json
}
```

2. **Add `isPlurConfigured()` guard** to `hook-inject`, `hook-observe`, and `hook-learn-check` so they also no-op gracefully when installed globally in non-plur projects.

3. **Add tests** covering both sides of the contract:
   - Projects with plur configured (in `.mcp.json` or `.claude/settings.json`) must have hooks fire normally — session guard enforces `plur_session_start`, injection works, the full lifecycle runs.
   - Projects without plur configured must not be affected at all — all hooks no-op silently, no tool calls are blocked, no errors or latency added, even when plur is in global `~/.claude/settings.json`.

### Impact on plur-enabled projects

None. Projects that use plur have it registered in their `.mcp.json` or `.claude/settings.json` (either directly or via `plur init`). The walk-up will find it. The only thing removed is the fallback to global settings, which was a false positive.

crtahlin:
## Summary

- Remove the `~/.claude/settings.json` fallback from `isPlurConfigured()` — it always returns true for anyone who ran `plur init`, causing the session guard to block tool calls in every project, even ones without plur
- Add test verifying global settings don't produce a false positive

Closes #247

## Test plan

- [x] All 1150 existing tests pass
- [x] New test: `isPlurConfigured()` returns `false` when only global `~/.claude/settings.json` has plur but no project-level config exists
- [x] Existing tests confirm projects with plur in `.mcp.json` or `.claude/settings.json` still return `true`
- [ ] Manual: open Claude Code in a non-plur project — session guard should not block
- [ ] Manual: open Claude Code in a plur project — session guard should enforce `plur_session_start` as before

### DIFF
--- a/packages/cli/src/lib/plur-configured.ts
+++ b/packages/cli/src/lib/plur-configured.ts
@@ -6,7 +6,9 @@ import { homedir } from 'os'
  * Detect whether plur is registered as an MCP server for the current project.
  *
  * Walks up from `cwd` looking for `.mcp.json` or `.claude/settings.json` with
- * a `plur` server entry, then falls back to the global `~/.claude/settings.json`.
+ * a `plur` server entry. Does NOT fall back to `~/.claude/settings.json` —
+ * when hooks are installed globally, that file always contains plur, which
+ * would defeat the purpose of this check.
  *
  * Used by the session enforcement hooks (`hook-session-guard`,
  * `hook-session-remind`, `hook-session-mark`) so they can be installed into
@@ -17,7 +19,7 @@ import { homedir } from 'os'
  */
 export function isPlurConfigured(
   cwd: string = process.cwd(),
-  home: string = homedir(),
+  _home: string = homedir(),
 ): boolean {
   const start = resolve(cwd)
   let dir = start
@@ -29,7 +31,10 @@ export function isPlurConfigured(
     if (parent === dir) break
     dir = parent
   }
-  return configHasPlur(join(home, '.claude', 'settings.json'))
+  // Do NOT fall back to ~/.claude/settings.json — when hooks are installed
+  // globally, that file always has plur in mcpServers, which makes this
+  // return true for every project regardless of whether it uses plur (#95).
+  return false
 }
 
 function configHasPlur(path: string): boolean {
--- a/packages/cli/test/plur-configured.test.ts
+++ b/packages/cli/test/plur-configured.test.ts
@@ -58,4 +58,18 @@ describe('isPlurConfigured', () => {
     writeFileSync(join(root, '.mcp.json'), '{ not valid json')
     expect(isPlurConfigured(root, root)).toBe(false)
   })
+
+  it('returns false even when global ~/.claude/settings.json has plur', () => {
+    // Simulate a home dir with plur in global settings
+    const fakeHome = join(root, 'fakehome')
+    mkdirSync(join(fakeHome, '.claude'), { recursive: true })
+    writeFileSync(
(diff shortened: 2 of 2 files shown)


plur9:
Closing in favor of #304, which fixes the same #247 issue more completely.

Both remove the global-settings fallback. #304 additionally:
- recognizes `.plur.yaml` (and `settings.local.json`) as a project-level marker — so global MCP install + per-project `.plur.yaml` scoping (the common configuration for agents, PLUR's primary users) still gets the guard in opted-in projects. Without it, dropping the fallback makes `isPlurConfigured()` return false in *every* project for global installs, and the guard never fires.
- extends the silent no-op to the injection hooks (`hook-inject`/`observe`/`learn-check`), which had the same global-fires-everywhere problem.

The root-cause writeup on this PR was spot-on — #304 just carries it further. Thanks!

Issue #270, opened 2026-05-30
Iter-2 followup: EmbeddingGemma pooling verification + Phase C eval


plur9:
Sprint 0 iter-1 audit gap M-8 (Critic F-CRIT-017, Dijkstra F-DIJK-007). Punted from iter-2 (default reverted to bge-small via B-2).

The current EmbeddingGemma adapter at \`packages/core/src/embedders/embedding-gemma.ts\` uses \`pooling: 'mean'\` over the entire sequence with no task prefix. The official EmbeddingGemma model card recommends \"last token\" pooling with task-specific prompts (\"query: ...\", \"passage: ...\"). Using \`mean\` without prefixes produces vectors that are NOT comparable to the published MTEB scores cited in the bake-off rationale.

Less urgent now because iter-2 B-2 reverted the default to bge-small. Worth verifying before Phase C re-evaluates the default-flip decision — if the adapter is using EmbeddingGemma wrong, the bake-off comparison is invalid.

Phase C scope (also tracked here for now):
- Either fetch the upstream LongMemEval-S dataset into \`benchmark/data/longmemeval-s.yaml\` (preferred) or run 100 reps of the 30-scenario fixture with explicit \"n=30 source, 100 reps\" framing
- Re-run the bake-off against EmbeddingGemma + BGE-small + MiniLM with the corrected pooling
- Decide the default by the documented gate rule: \"≥2pp R@5 at or below CPU cost\"

References:
- docs/audit/sprint-0/iter-1-critic.md F-CRIT-017, F-CRIT-016
- docs/audit/sprint-0/iter-1-dijkstra.md F-DIJK-007
- docs/audit/sprint-0/iter-1-archivist.md F-ARCH-003
- docs/audit/sprint-0/iter-1-gaps-consolidated.md M-8

plur9:
**Triage (heartbeat 2026-07-02):**

Confirmed the issue is real. Additional finding that widens the scope:

**1. Pooling type gap**
[...]

PR #290, opened 2026-06-04
core: export the embedding primitive (embed + EMBED_DIM) as public API

### LINKED ISSUE #289: core: export the embedding primitive (embed() + EMBED_DIM) as public API
## Summary

Expose core's embedding primitive — `embed()` and the embedding dimension — as public API from `@plur-ai/core`, so any store backend can compute embeddings identically to core's hybrid search.

## Background

Core already has a complete local semantic-search stack:

- `recallHybrid()` and `recallSemantic()` use `embed()` (`packages/core/src/embeddings.ts`) — model `BAAI/bge-small-en-v1.5`, **384-dim**, run locally via `@huggingface/transformers` (no API calls), output normalized.
- For the in-memory / file-backed store this works end to end: embeddings are computed on demand, cached locally, and similarity is done in memory.

But the embedder is effectively internal. The package entry (`packages/core/src/index.ts`) only re-exports the `SimilarityResult` type — `embed()` and the dimension are not part of the public API.

## Why this is needed

An alternative store backend that persists engrams in a database and wants to run similarity **in the database** (rather than loading every engram into memory on each query) has to store vectors produced by the **same model and dimension** as core. Today it cannot reuse core's embedder, so it would have to re-implement embedding — which risks drift: a different model or dimension produces vectors incompatible with core's, silently degrading recall quality and breaking the "one memory, consistent recall everywhere" guarantee.

Exposing the primitive keeps every backend on a single embedding contract. (This is purely about making the existing capability reusable — it does not add a new model or any external dependency.)

## Proposal

Export from `@plur-ai/core`'s public API:

- `embed(text): Promise<Float32Array | null>`
- `EMBED_DIM` (currently `384`) as a named constant, used by core internally as the single source of truth
- optional but useful: `cosineSimilarity()` and `embedderStatus()`

Treat the model identity + `EMBED_DIM` as a **stable public contract**: changing the model is a breaking change for anyone persisting vectors (they would need to re-embed). Document that expectation.

Cut a release once exported so downstream packages can depend on it.

## Acceptance criteria

- `import { embed, EMBED_DIM } from '@plur-ai/core'` works.
- `EMBED_DIM` matches the model's actual output dimension and is the single value core uses internally.
- Changelog/docs note that changing the embedding model or dimension is breaking for any consumer that persists vectors.

crtahlin:
Closes #289.

## What

Promotes core's local BGE embedder from internal to public API so an alternative store backend can compute vectors **identically** to core's hybrid search (same model, same dimension) instead of re-implementing the embedder and risking silent model/dimension drift.

`@plur-ai/core` now re-exports:

- `embed(text): Promise<Float32Array | null>`
- `EMBED_DIM` (384) — new named constant
- `embedderStatus()` — so a consumer can surface "embeddings degraded (reason)" instead of silently dropping to keyword-only
- `cosineSimilarity()` — convenience, useful in tests

Previously only the `SimilarityResult` type was public.

```ts
import { embed, EMBED_DIM } from '@plur-ai/core'
```

## Enforcing the contract against the live model

Per the amended acceptance criteria, `embed()` checks its first successful output's length against `EMBED_DIM` once per process (`assertEmbedDim`). A model swap that changes the dimension fails loudly rather than silently corrupting any persisted vectors. The check resets on `resetEmbedder()`.

`EMBED_DIM` is **not** claimed as core's internal single-source-of-truth — core derives the dimension from model output and never referenced a literal `384` in code (only comments). The runtime assertion is what binds the constant to the actual model.

## Breaking-change contract

The model identity + `EMBED_DIM` are documented in CHANGELOG as a stable public contract: changing either is breaking for any consumer that persists vectors (they must re-embed).

## Tests

`packages/core/test/embedding-primitive-export.test.ts` — 5 model-free assertions (export surface, `EMBED_DIM === 384`, `cosineSimilarity` math). Kept model-free deliberately so it runs in any CI env (no ~130MB BGE download / ONNX init); the live-model dimension check is enforced by the in-`embed()` assertion.

Targeted run: `embedding-primitive-export` + `embeddings-optout` → 17 passed. (Pre-existing `sync.test.ts` failures in this environment are git-identity issues in ephemeral temp repos, unrelated to this change.)

## Out of scope

No version bump / publish — leaving the release cut to @plur9 to batch as desired.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,20 @@
 # Changelog
 
+## Unreleased
+
+### `@plur-ai/core` exports the embedding primitive (#289)
+
+`embed()`, `EMBED_DIM`, `embedderStatus()`, and `cosineSimilarity()` are now part of `@plur-ai/core`'s public API. Previously only the `SimilarityResult` type was re-exported, so the local BGE embedder (`BAAI/bge-small-en-v1.5`, 384-dim) was effectively internal. Alternative store backends that persist vectors and run similarity in a database can now compute embeddings identically to core's hybrid search instead of re-implementing the embedder and risking model/dimension drift.
+
+```ts
+import { embed, EMBED_DIM } from '@plur-ai/core'
+```
+
+- `EMBED_DIM` (384) is a new named constant. `embed()` asserts its first successful output against it once per process, so a model swap that changes the dimension fails loudly instead of silently corrupting persisted vectors.
+- **Breaking-change contract:** the embedding model identity and `EMBED_DIM` are a stable public contract. Changing either is a breaking change for any consumer that persists vectors — they must re-embed. Treat a model/dimension change accordingly.
+
+No new model and no external dependency — this only makes the existing capability reusable.
+
 ## 0.9.11 (2026-05-26)
 
 Bug sweep — three independent fixes bundled.
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@plur-ai/core",
-  "version": "0.9.12",
+  "version": "0.9.13",
   "type": "module",
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
--- a/packages/core/src/embeddings.ts
+++ b/packages/core/src/embeddings.ts
@@ -32,6 +32,14 @@ import { atomicWrite } from './sync.js'
  */
 export const EMBED_DIM = 384
 
+/**
+ * Model identity behind the embeddings, half of the stable contract above.
+ * The ONNX/transformers.js port of BAAI/bge-small-en-v1.5. Named so the
+ * dimension-mismatch error can name the model a consumer must match, and so a
+ * model swap is a one-line, greppable change.
+ */
+const EMBED_MODEL_ID = 'Xenova/bge-small-en-v1.5'
+
 // Verifies the live model output matches EMBED_DIM exactly once per process —
 // see assertEmbedDim().
 let embedDimChecked = false
@@ -128,7 +136,7 @@ async function getEmbedder() {
   // transient (network, sandbox restrictions on first download).
   try {
     const { pipeline } = await import('@huggingface/transformers')
-    embedPipeline = await pipeline('feature-extraction', 'Xenova/bge-small-en-v1.5', {
+    embedPipeline = await pipeline('feature-extraction', EMBED_MODEL_ID, {
       dtype: 'fp32',
     })
     transformersUnavailable = false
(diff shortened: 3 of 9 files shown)


crtahlin:
### Reviewer judgment call: `assertEmbedDim` throws vs. warns

One deliberate decision worth a second opinion: the dimension check inside `embed()` **throws** on a mismatch rather than warning and continuing.

**Rationale for throwing:** a mismatch can only happen when the embedding model and `EMBED_DIM` genuinely disagree — i.e. someone swapped the model without updating the constant (or vice versa). That's a developer error, caught immediately in dev/tests, never in normal operation. Failing loudly is exactly the silent-drift failure mode this issue exists to prevent, so a hard error at the source is the safest default for anyone persisting vectors.

**The tradeoff:** because `embed()` is on core's own hybrid-search path, a throw means a misconfigured build also hard-fails core's *own* recall (rather than degrading to BM25-only). If you'd prefer core never hard-fails on a misconfig, the alternative is to `console.warn` once and still return the vector — core keeps working, the mismatch is surfaced, but a vector-persisting consumer no longer gets a guaranteed loud failure. It's a one-line change either way.

Happy to flip it to warn-and-continue if that's the preferred posture.

crtahlin:
Keep the throw — agree with your default.

The deciding factor is in this PR's own design: the tests are intentionally model-free, so the in-`embed()` assertion is the **only** thing that ever catches a real model/`EMBED_DIM` mismatch. Flip it to warn-and-continue and drift becomes completely unguarded — which is the exact failure #289 exists to prevent.

A dimension mismatch is a build/contract error, not a runtime/env condition. The env conditions that *should* degrade (model not loaded, `PLUR_DISABLE_EMBEDDINGS`) already do — `embed()` returns `null` and recall falls back to BM25. A wrong dimension is categorically different: the build is broken. Failing loudly at the source is right, and it stops incompatible vectors from being persisted (the concrete harm).

So the contract is clean and worth stating explicitly in the code/docs:

- `embed()` returns `null` → embeddings unavailable → consumer degrades to keyword search.
- `embed()` throws → contract violation (model/dim drift) → must **not** be swallowed.

Two small follow-throughs:

1. Make the thrown error actionable — include the model id, expected `EMBED_DIM`, the actual length, and the remediation ("update EMBED_DIM or revert the model change").
2. Consumer note for the database-backed store (#270 / its `recallHybrid` wiring): the degrade-to-BM25 path must key on `embed()` returning `null`, not on a catch-all `try/catch`. A thrown mismatch should surface, not silently fall back — otherwise the consumer re-introduces the silent drift on its own side.

Rest of the PR matches the agreed scope (exports incl. `embedderStatus`/`cosineSimilarity`, model-free tests, no release cut). Good to go from my side once #1 is in.

crtahlin:
Pushed d17ca21 addressing both points.

**#1 (the blocker) — actionable error.** The thrown mismatch now names the model and gives remediation:

> Embedding dimension mismatch: model "Xenova/bge-small-en-v1.5" produced N dims but EMBED_DIM is 384. The embedding model and EMBED_DIM must agree. Remediation: update EMBED_DIM to N or revert the model change. Either way this is a BREAKING change for any consumer that persists vectors — they must re-embed, since vectors at the old dimension are incompatible.

The model id is now a named `EMBED_MODEL_ID` constant (internal) used both at pipeline-load and in the error, so the model identity — the other half of the contract — is a single greppable source.

**#2 — null-vs-throw contract documented in code.** Codified your two-line contract in `embed()`'s JSDoc so the #270 consumer wires it correctly:

- returns `null` → embeddings unavailable (env condition) → degrade to BM25
- throws → model/`EMBED_DIM` drift (contract violation) → must **not** be swallowed; a catch-all fallback here re-introduces the drift and lets incompatible vectors get persisted

Scope unchanged otherwise (`EMBED_MODEL_ID` stays internal — not added to the export surface). Targeted tests still green (17 passed). Over to you.

PR #293, opened 2026-06-04
fix: plur_stores_add silently drops additional scopes for the same remote URL (#291)

### LINKED ISSUE #291: plur_stores_add silently drops additional scopes for the same remote URL (false success:true)
## Summary
`plur_stores_add` deduplicates remote stores by **URL only**. Adding a second scope for an already-registered URL returns `success: true` but persists **nothing** — silently blocking access to additional team scopes the user is authorized for.

## Impact — critical
A user with server-side access to multiple team scopes on one enterprise instance can only ever wire up **one** of them. The misleading `success` masks the failure, so the user believes the scope is registered when it is not.

Verified live: user is authorized for 5 scopes (confirmed via `GET /api/v1/me`) — `group:plur/plur-ai/{engineering,comms,research,leadership}` + `group:plur/plur-ai` — and `engrams?scope=...comms` returns real server data. But only `engineering` is usable from the client; adding comms/research/leadership each returned `success:true` while `~/.plur/config.yaml` retained only the engineering entry.

## Root cause
`packages/core/src/index.ts:2319-2323`:
```ts
const dedupKey = isRemote ? options!.url : storePath
const sameEndpoint = config.stores?.find(s => (isRemote ? s.url === dedupKey : s.path === dedupKey))
if (sameEndpoint) return   // ← early return: no error, nothing persisted
```
The dedup key is the URL (remote) / path (local) **only**, not URL+scope. A second scope for the same URL hits this branch and returns *before* the scope-conflict check (`:2328`) and the persistence write (`:2358-2366`). The MCP tool (`packages/mcp/src/tools.ts`) then returns `{success:true, url, scope}` because `addStore()` threw nothing.

## Repro
1. Register store A: `url=X scope=group:.../engineering` → persists.
2. Register store B: `url=X scope=group:.../comms` → returns `success:true`.
3. `plur_stores_list` / `config.yaml` → only `engineering` present.

## Proposed fix
- Dedup by **URL+scope** for remotes (path+scope for local): `const dedupKey = isRemote ? `${url}::${scope}` : `${path}::${scope}``. Allows N scopes per URL.
- True idempotency: no-op only when URL **and** scope both match.
- MCP tool: return an honest result (`added` vs `already_registered`) instead of unconditional `success:true`.

## Why fully blocking
Reads are server-scope-filtered — `packages/core/src/store/remote-store.ts:66` sends `?scope=` per store, so each scope must be individually registered to be visible. Without this fix a multi-team user cannot see >1 team. Pairs with the missing scope-discovery feature (separate issue).

crtahlin:
Closes #291.

## The bug

`addStore()` deduplicated remote stores by **URL only**. A second scope for an already-registered URL hit an early `return` — persisting nothing — while `plur_stores_add` still reported `success: true`. Because reads are server-scope-filtered (`?scope=` per store, `remote-store.ts:66`), a user authorized for several team scopes on one enterprise instance could only ever wire up the **first**; the rest were silently invisible, with a misleading `success` masking the drop.

## Why the fix is localized

The rest of the system already supports N scopes per URL — the driver pool `_getRemoteDriver` is keyed by `` `${url}::${scope}` `` (`index.ts:316`) and every consumer iterates `config.stores` per-entry, filtering by scope. `addStore`'s URL-only dedup was the only blocker. No schema or downstream change needed.

## Changes

- **`packages/core/src/index.ts`** — `addStore()` dedups by **endpoint + scope** (`url+scope` remote, `path+scope` local), so one URL hosts N scopes. Returns `{ status: 'added' | 'already_registered' | 'overwritten' }` instead of `void`. Only an exact endpoint+scope match is an idempotent no-op. The existing scope-conflict guard (a *different* endpoint claiming the same scope) is unchanged.
- **`packages/mcp/src/tools.ts`** — `plur_stores_add` surfaces `status`; `success: true` is now honest (a second scope genuinely persists). Description notes one URL can host multiple scopes.
- **`packages/cli/src/commands/stores.ts`** — `plur stores add` prints the real outcome (added / already registered / reassigned).

## Tests

- **`packages/core/test/store-validation.test.ts`** — same-URL/different-scope persistence (the core repro), the live 5-scope enterprise repro from the issue, `added → already_registered → overwritten` status reporting, and same-URL+same-scope+new-token staying a no-op. Existing idempotency test still green.
- **`packages/mcp/test/tools.test.ts`** — `plur_stores_add` reports `status: 'added'` for the second scope on a shared URL and `already_registered` on an exact repeat (the honest-`success` contract).

Targeted runs: core store-validation 18 passed, MCP tools 12 passed, CLI suite 184 passed.

## Decisions (conservative defaults, flagged for review)

- **Token rotation** is intentionally out of scope: re-adding the same URL+scope with a different token stays an `already_registered` no-op rather than silently swapping the stored token. A dedicated rotation path can follow if wanted.
- **`success` field** stays `true` for both `added` and `already_registered` (both valid end states); the new `status` carries the nuance, so existing callers that only check `success` don't break.

## Out of scope

No version bump / publish — leaving the release cut to @plur9. Pairs with the separate scope-discovery feature noted in the issue.


plur9:
Triaging the duplicate #302 (now closed in favor of this PR — yours is the more complete fix). One thing #302 did deliberately that's worth a look before merge: it kept **local**-store dedup as path-only, not path+scope, reasoning that the loader clones global-scoped engrams into each entry's scope — so registering one `engrams.yaml` under two scopes could double-load those engrams. This PR dedups local by path+scope. The remote fix (#291's actual bug) is solid either way; just worth confirming the local path+scope case doesn't double-load before this lands.

crtahlin:
@plur9 your question is answered in a4248b8 — you were right to flag it:

**Confirmed: local path+scope dedup would double-load.** The loader (`index.ts` store iteration) clones `global`-scoped engrams into each entry's scope, so two entries pointing at one `engrams.yaml` surface those engrams twice — once per scope. The #302 reasoning was correct.

Changes:
- **Local stores: path-only identity** (one file = one store); remote keeps url+scope — the #291 fix proper is untouched
- `already_registered` now returns the **existing** entry's scope (for local it can differ from the request); MCP tool + CLI surface it
- Rebased onto main (CHANGELOG conflict resolved — both Unreleased entries kept)
- Tests: +2 local path-only identity cases, scope added to status assertions

Suites: core 746, mcp 115, cli 184, claw 103 — all green locally. Ready for your review/approval; same squash-merge suggestion as the others.

plur9 (approved):
Verified per technical-PR check: no merge conflicts (MERGEABLE against current main), CI green (2/2), scope is a contained fix to store-scope handling in stores.ts / core / mcp tools with matching unit tests. Approving to unblock merge.

Issue #295, opened 2026-06-09
Silent enterprise-store auth failure: expired token 401s into outbox with no warning


crtahlin:
## Summary

When the enterprise remote-store token expires, the MCP client **fails silently**: writes 401 and pile up in the local outbox, reads return `engram_count: 0`, and the user gets **no warning anywhere**. The only way the problem surfaced was the user manually noticing "no new engrams on enterprise since the 5th." By then ~4 days of team engrams had been lost to a non-draining outbox.

## What happened (real incident, 2026-06-09)

- Enterprise store: `https://plur.datafund.io/sse`, scope `group:plur/plur-ai/engineering`.
- The bearer JWT in `~/.plur/config.yaml` (`iss: plur-enterprise`, 30-day TTL) **expired 2026-06-05 08:25 UTC**.
- Since expiry: every team-scoped `plur_learn` → remote write **401** → queued to outbox. `outbox_pending: 5` and climbing.
- `plur_stores_list` shows the remote with `engram_count: 0`; `curl https://plur.datafund.io/sse` → `HTTP 401`.
- `plur_doctor` reports **"All checks passed — PLUR is healthy"** — it only checks the local embedder/store, never the remote auth state. Misleading.
- `session_start` showed `outbox: flushed 0, pending N` but did not flag *why* (auth failure).

## Why it stayed invisible

1. `plur_learn` returns success even when the engram only lands locally / queues for an unreachable remote.
2. No proactive expiry warning — a 30-day JWT with no refresh and no "expires in N days" notice.
3. `plur_doctor` has no remote-connectivity/auth check.
4. The outbox never drained: the learning-classifier suggested "retry on next `plur_sync`", but `plur_sync` is the **git** mechanism (and `plur_sync_status` shows `initialized: false`). The SSE remote outbox has no surfaced drain path.

## Reauth friction (also worth fixing)

There is **no CLI login** — `@plur-ai/mcp` only exposes `init`. Reauth today is fully manual:
1. Browser (logged into GitHub) → `https://plur.datafund.io/auth/github` (shows a short-lived JWT) **or** `https://plur.datafund.io/me/api-keys` (longer-lived `plur_sk_…`).
2. Copy token, then either `plur_stores_add(url, token, scope)` or hand-edit `~/.plur/config.yaml`.
3. Restart the MCP server so it reloads the token; outbox flushes on next `session_start`.

[...]

PR #294, opened 2026-06-04
feat: client scope discovery via GET /api/v1/me (#292)

crtahlin:
Closes #292.

> ⚠️ **Stacked on #293 (fix for #291).** Base is `fix/stores-add-multi-scope-291`, not `main` — auto-register needs the URL+scope dedup from #291. Merge #293 first, then this retargets to `main`. The diff below is scoped to the discovery feature only.

## Problem

The client never asked the enterprise server which scopes a token can access, even though `GET /api/v1/me` already returns the full resolved set. A user authorized for 5 teams had to discover scopes out-of-band and hand-register each — and saw only the one(s) they happened to register. Pure client gap; no server change.

## Changes

- **`packages/core/src/store/remote-store.ts`** — `RemoteStore.me()` → `GET /api/v1/me`, returns `{ username, org_id, role, scopes }`. Scope-independent (keyed on the token).
- **`packages/core/src/index.ts`**
  - `discoverRemoteScopes(opts?)` — groups configured stores by distinct URL, calls `/me` per URL (each raced against a timeout, failures captured as `ok:false` not thrown), and splits authorized scopes into `registered` / `unregistered`. Read-only.
  - `registerDiscoveredScopes(opts?)` — registers every authorized scope via the #291 `addStore` path; returns per-URL `added` / `already_registered`.
- **`packages/mcp/src/tools.ts`** — new `plur_scopes_discover` (read-only by default; `register: true` registers all). `plur_session_start` appends a best-effort hint when the token is authorized for unregistered scopes — short timeout, fully swallowed on error, **never blocks or slows session start**.
- **`packages/cli/src/commands/stores.ts`** — `plur stores discover [--register]`.

## Tests

- **Stub servers** (core + mcp): added a configurable `GET /api/v1/me` handler (`setMe()`), keeping integration tests wire-level.
- **`packages/core/test/scope-discovery.test.ts`** — `RemoteStore.me()` (incl. 401), the registered/unregistered split, multi-URL grouping, per-URL failure isolation (one bad token doesn't sink the rest), no-remote → `[]`, register-all under one URL (exercises the #291 dedup end-to-end), and idempotency.
- **`packages/mcp/test/e2e-remote.test.ts`** — full MCP→core→`RemoteStore.me()`: read-only discover, `register:true` then scopes appear in `plur_stores_list`, and the no-remote note.

Runs: core scope-discovery 8 / store suites 47, full MCP 118, full CLI 184. All green.

## Decisions (conservative defaults, flagged for review)

[...]

PR #299, opened 2026-06-09
feat: teach per-engram scope selection at install time (#296)

### LINKED ISSUE #296: Per-engram scoping isn't taught at install — team knowledge silently defaults to 'global'
## Summary

Per-engram scoping is supported (`plur_learn` takes a `scope` param on every call), but **nothing teaches the agent to use it**, so agents routinely omit `scope` → it falls back to `global` → **team/enterprise-relevant engrams silently never reach the configured group store.** This affects *any* install with a remote/group store, not a single user. The capability works; the defaults and guidance don't.

## Why this is a general problem

- One session legitimately spans multiple scopes (an engineering insight, a comms note, a personal preference). Scope is therefore inherently **per-engram**, decided by content — not a per-session setting.
- Today the only nudge is a `session_start` line: *"No project scope detected. plur_learn calls without explicit scope will be tagged 'global'."* That's (a) easy to miss, (b) framed around *project* scope only, and (c) not prescriptive about *team/group* scopes.
- Net effect: unless the operator hand-writes per-engram-scoping instructions into their own CLAUDE.md (as a workaround), team knowledge leaks to `global` and the shared store stays empty. Observed in production: an entire multi-topic session's engrams all defaulted to `global` despite a configured `group:…/engineering` store.

## Proposed general fixes (product/onboarding, not per-user)

- [ ] **`plur-mcp init` should generate scoping guidance** into the install's CLAUDE.md / AGENTS instructions: "set `scope` on every `plur_learn` by content," with the install's **configured scopes enumerated** (read from `config.yaml` stores).
- [ ] **Bundle a "Scope selection" section in the MCP server `instructions`** (the block every agent receives), e.g.:
  - Choose scope per engram from the configured stores.
  - Team/engineering knowledge → the matching `group:<org>/<team>` scope.
  - Personal preferences / workflow → local/default.
  - Never let team knowledge fall back to `global`.
- [ ] **Make `session_start` guidance prescriptive**: it already returns `remote_scopes` — also state the rule and list selectable scopes so the agent has them in-context for the whole session.
- [ ] **Surface the fallback**: when a `plur_learn` call omits `scope` and a group store exists, either (a) return a hint in the response ("stored as global; did you mean group:…/engineering?"), or (b) support a configurable `default_scope` that's documented at init.
- [ ] (Optional, heavier) **`scope: auto`** — server infers the best-matching configured scope from engram content/tags, with the chosen scope echoed back for veto.

## Impact

Without this, the shared-memory value proposition silently fails on fresh installs: agents "learn" but the team store stays empty, and nobody notices until they manually check. Low-effort onboarding/instruction changes fix it for everyone.

## Related

- Companion incident (auth side): #295 — expired enterprise token 401'd into the outbox with no warning. The two together meant zero engrams reached enterprise for days.

## Environment

- `@plur-ai/mcp` v0.9.12

crtahlin:
Closes #296. Based on `main` (independent of the #293/#294/#295 stack).

## The gap

Per-engram scoping is supported — every `plur_learn` takes a `scope` — but **nothing teaches the agent to use it**, so agents routinely omit `scope`, it falls back to `global`, and team-relevant knowledge silently never reaches the configured group store. Affects any install with a remote/group store. The capability works; the defaults/guidance don't. (Observed in production: an entire multi-topic session's engrams all defaulted to `global` despite a configured `group:…/engineering` store.)

## What this does — teach it at the always-on / install surfaces

- **`packages/mcp/src/server.ts`** — the server `INSTRUCTIONS` block (advertised to every client on connect) gains a **SCOPE SELECTION** section: scope is content-driven and **per-call**; team/shared knowledge → the matching `group:<org>/<team>` scope (session_start lists the writable ones); personal/local → default; **never let team knowledge fall back to `global`**. Exported so it's unit-testable.
- **`packages/cli/src/commands/init.ts`** — the CLAUDE.md section `plur init` writes replaces the thin "Multi-project scoping" note with a fuller **"Scope selection (set scope PER engram, by content)"** guide that enumerates team / project / personal routing and calls out the global-fallback anti-pattern explicitly.

## Why not session_start

`plur_session_start` already surfaces the live writable remote scopes prescriptively (#229: "set scope to the matching remote scope … personal stays local"). The missing piece was the **always-on** guidance (the instructions block, present even before any recall) and the **install-time** guidance (generated CLAUDE.md). This PR fills those without touching session_start, so it stays independent of the in-flight #294/#295 work on that handler.

## Related

- **#291** — until that lands, a second/comms scope can't even be registered against the same URL, so "route comms → `group:…/comms`" is guidance the client can't yet honor. Cross-referenced, not blocked.
- **#295** — the auth-expiry side of the same "make memory legible" theme.

## Tests

- `packages/mcp/test/instructions.test.ts` — INSTRUCTIONS teaches per-engram selection, names `group:<org>/<team>`, and warns against the global fallback.
- `packages/cli/test/init.test.ts` — a fresh `plur init` writes the scope-selection guidance (incl. the "never reaches the team store" anti-pattern) into CLAUDE.md.

mcp + cli suites green. No version bump / publish — leaving the release cut to @plur9.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,15 @@ When an enterprise token expired, the client failed silently: team-scoped writes
 
 The reauth command itself (`plur login`) and longer-lived keys are tracked separately as a fast-follow.
 
+### Teach per-engram scope selection at install time (#296)
+
+Per-engram scoping was supported (every `plur_learn` takes a `scope`) but nothing taught agents to use it, so they routinely omitted `scope` → it fell back to `global` → team-relevant knowledge silently never reached the configured group store. This affects any install with a remote/group store. The capability worked; the guidance didn't.
+
+- **`packages/mcp/src/server.ts`** — the server `INSTRUCTIONS` block (advertised to every client on connect) gains a "SCOPE SELECTION" section: scope is content-driven and per-call; team/shared knowledge → the matching `group:<org>/<team>` scope; personal/local → default; never let team knowledge fall back to `global`. Exported for testing.
+- **`packages/cli/src/commands/init.ts`** — the CLAUDE.md section `plur init` generates replaces the thin "Multi-project scoping" note with a fuller "Scope selection (set scope PER engram, by content)" guide enumerating the team / project / personal routing and the global-fallback anti-pattern.
+
+`plur_session_start` already surfaces the live writable remote scopes prescriptively (#229), so this fills the always-on / install-time guidance gap. Relates to #291 (a comms/second scope can't be registered against the same URL until that lands) and #295 (silent auth-expiry).
+
 ## 0.9.11 (2026-05-26)
 
 Bug sweep — three independent fixes bundled.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -206,9 +206,14 @@ A PreToolUse guard enforces that `plur_session_start` is called at the beginning
 
 Do not ask permission to use these tools — they are your memory system.
 
-### Multi-project scoping
+### Scope selection (per engram, by content)
 
-PLUR uses `domain` and `scope` fields on engrams to separate knowledge by project. When calling `plur_learn`, set `scope` (e.g. `project:my-app`) to namespace the engram. Scoped recall automatically includes global engrams.
+PLUR uses `domain` and `scope` fields to separate knowledge. **Set `scope` on every `plur_learn` call, chosen by the engram's content** — one session spans multiple scopes. Scoped recall automatically includes global engrams.
+
+- Team/shared knowledge → the matching team scope (e.g. `group:<org>/<team>`); `plur_session_start` lists the writable ones.
+- This project's details → `project:my-app` (a `.plur.yaml` with `scope:` makes it the default).
+- Personal preferences / your workflow → leave at the default/local scope.
+- Don't omit `scope` for team-relevant knowledge — it falls back to `global`, which leaks into every project and never reaches the team store. Reserve `global` for genuinely cross-project facts.
 
 ### When to check memory
 
--- a/README.md
+++ b/README.md
@@ -37,6 +37,8 @@ npx @plur-ai/cli@0.9.4 init --domain myapp --scope project:my-app
 
 This creates a `.plur.yaml` in the project with defaults that hooks apply automatically. Engrams learned in that project are tagged; recall filters by scope but always includes global knowledge.
 
+**Set scope per engram, by content.** Scope is not a once-per-session setting — every `plur_learn` call takes its own `scope`, chosen from what the engram is about. Team/shared knowledge goes to a team scope (e.g. `group:<org>/<team>`, used by PLUR Enterprise); project details to `project:<name>`; personal preferences stay local. Don't let team-relevant knowledge fall back to `global` by omitting scope — `global` leaks into every project and (with a team store configured) never reaches the team. `plur_session_start` lists the remote scopes a token can write to.
+
 ### Global install (faster startup)
 
 ```bash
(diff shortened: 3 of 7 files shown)


crtahlin:
## Review — APPROVE (confidence: high)

Read the surrounding code, confirmed `INSTRUCTIONS` is wired into the server (`server.ts:177`) and the new `export` is non-conflicting, and verified both new tests' assertion strings exist verbatim in the branch source.

Clean, low-risk guidance/docs PR for #296. The only logic change is `const` → `export const INSTRUCTIONS` (to make it unit-testable); the rest is added prose across four agent-facing surfaces (server INSTRUCTIONS, `init`-generated CLAUDE.md, repo CLAUDE.md, README). Correctly diagnoses that scope is **per-engram, content-driven** (not per-session) and teaches it at the always-on + install-time surfaces, leaving the in-flight `session_start` work (#294/#295) untouched.

### Tests
- **CI green** — `test (20)` + `test (22)` both SUCCESS.
- Confirmed both new tests' assertion strings (`never reaches the team store`, `group:<org>/<team>`, per-engram) exist verbatim in the branch source. Didn't re-run locally — a doc-string change with green CI doesn't warrant a full worktree install.

### Notes (none blocking)
1. **Guidance slightly precedes capability** — text routes team knowledge to `group:<org>/<team>` scopes, but per the PR's own cross-ref, **#291** must land before a second scope (e.g. `group:…/comms`) can be registered against the same URL. Multi-team routing is partly aspirational until #291. Acknowledged in the description; just flagging.
2. **Claw surface excluded (reasonably)** — `packages/claw/src/assembler.ts` has its own `PLUR_MEMORY_INSTRUCTIONS` using a signal-based learning model with no per-call `scope`, so the guidance genuinely doesn't apply there today. Worth a note if claw ever gains scoped writes — it'd be a fourth surface needing the same teaching.
3. **Pre-existing README version pin** — the new paragraph lands right after a hardcoded `npx @plur-ai/cli@0.9.4 init`. Not introduced here, just a reminder that README has a manually-pinned version that can drift.

Praise: fixes the problem at the right layer (always-on/install surfaces, not a per-user CLAUDE.md workaround); the tests assert the actual anti-pattern phrasing so the guidance can't silently regress; consistent messaging across all four surfaces; deliberately kept independent of the #294/#295 stack.

@plur9 — assigning to you for final review + merge.

plur9:
⚠️ **Duplicate-for-#296.** #299 and #324 both close #296 ("teach per-engram scope selection") and both add the `SCOPE SELECTION` block to `packages/mcp/src/server.ts`, so they conflict — only one can merge cleanly.

They're actually *complementary halves*, not redundant:
- **#299** covers the **install-time** surfaces — the `init.ts`-generated `CLAUDE.md`, `README.md`.
- **#324** covers the **runtime** surfaces — the static server `INSTRUCTIONS` + `tools.ts` + a safety net.

Rather than pick one and lose the other half, please **consolidate into a single PR**: the install-time guidance from #299 + the runtime instructions/safety-net from #324, with **one** `SCOPE SELECTION` block as the single source of truth (so it can't drift between server.ts and the generated CLAUDE.md). Assigning both to you to merge into one — say which branch you'd keep as the base. The agent-facing wording is a product call, so flag it for @plur9 sign-off before merge.

crtahlin:
Consolidated into #331 together with #324, per the review request. #331 keeps a single `SCOPE SELECTION` block as the source of truth (so `server.ts` and the generated `CLAUDE.md` can't drift), carries all of this PR's install-time surfaces unchanged, and adds #324's runtime `scope_hint` safety net. Rebased on current `main`; `pnpm build` clean, mcp 130/130 + cli 199 pass. Closing in favour of #331.

Issue #281, opened 2026-05-31
plur_learn: misleading 'statement: Required' error causes silent memory loss in multi-agent orchestration


crtahlin:
## Summary

When `plur_learn` is used inside a multi-agent orchestration (a parent session that fans out to spawned subagents), three issues combine to cause **silent memory loss** — not just wasted tool calls. The most damaging is a misleading validation error that leads the model to conclude the tool is broken and abandon persistence entirely.

## 1. `statement: Required` error is misleading (highest priority)

When `plur_learn` is invoked with an empty or partial argument payload, the server returns:

```
{"error":"Invalid arguments: statement: Required","success":false}
```

The problem is interpretation, not the validation itself. In a recent multi-agent run, two independent agents — and the orchestrator on its first attempt — misread this error as *"my parameters aren't being transmitted to the server"* rather than *"I sent a malformed call."* One agent concluded the tool was broken and **gave up persisting its findings**, reporting back that "Plur learn calls failed repeatedly (parameter payload not transmitting)." The findings were lost.

This is a developer-experience failure with real data loss. Suggested fixes (either is sufficient):

- **Echo what was received** so the caller can self-correct, e.g.
  `Invalid arguments: missing required field 'statement'. Received fields: [scope, domain, type, tags]. Provide a non-empty 'statement'.`
- Or **accept a more forgiving shape** — e.g. treat the first positional/string argument as `statement` when the named field is absent.

## 2. Tool schema surfaces inconsistently in spawned subagents

Subagents spawned by a parent session start without the SessionStart hook, and the deferred `plur_learn` schema appears inconsistently via tool-search in those contexts. In one run, *identical* subagents (same type, same harness, same prompt structure) got different results — some discovered and called `plur_learn` successfully, others reported the tool was unavailable and could not persist.

This non-determinism is the underlying architectural gap behind the lost writes. Two possible directions:

- Make session/tool availability **deterministic** for subagents (the schema should reliably resolve when present in the environment), and/or
- **Document a recommended orchestration pattern**: parent session owns all `plur_learn` writes; subagents return findings as text only and never write. Right now nothing steers users toward this, so the default MCP guidance ("call plur_learn when you discover something") gets inherited by every subagent and fails unevenly.

## 3. (Enhancement) No batch-write API

[...]

Issue #297, opened 2026-06-09
plur_learn: passing a tags array empties the entire arguments payload ("statement: Required")


crtahlin:
## Summary

`plur_learn` fails when called with a `tags` array argument. The error reports a missing `statement` even though `statement` is provided — indicating the **entire arguments object is lost** when an array-typed parameter is present, not just the `tags` field being dropped.

Scalar optional parameters (`type`, `scope`) work fine. The failure is specific to array-typed arguments.

## Environment

- `@plur-ai/mcp` **v0.9.12** (`plur_status` → `version`)
- Launched via `npx -y @plur-ai/mcp@latest`
- MCP client: Claude Code
- Local store healthy (`plur_doctor` all green: BGE-small embedder loaded, hybrid search operational)

## Reproduction

Isolation-tested one optional field at a time (throwaway engrams):

| Call | Result |
|------|--------|
| `plur_learn({ statement: "x" })` | ✅ `decision: ADD` |
| `plur_learn({ statement: "x", type: "behavioral" })` | ✅ `decision: ADD` |
| `plur_learn({ statement: "x", scope: "global" })` | ✅ `decision: ADD` |
| `plur_learn({ statement: "x", tags: ["y"] })` | ❌ `{"error":"Invalid arguments: statement: Required","success":false}` |

Reproducible across multiple attempts.

## Expected

`plur_learn({ statement: "x", tags: ["y"] })` should store the engram with the given tags.

## Actual

Server returns `{"error":"Invalid arguments: statement: Required","success":false}`. Because scalar fields pass and only the array case fails — and the error claims `statement` is missing when it was supplied — the whole arguments payload appears to be emptied/mis-parsed when an array-typed argument is present.

## Likely locus (unconfirmed)

Array-argument (de)serialization between the MCP client and the server's input-schema validation. A wire-log of the `tools/call` request body for the failing (`tags: [...]`) vs passing (`type: "..."`) case would pin whether the array is malformed on the client side or coerced incorrectly server-side.

## Suggested follow-up

- Confirm whether other array/object parameters fail the same way (e.g. `knowledge_anchors`, `dual_coding` on the datacore-side schema; any array param on `plur_learn`).
- Check the input-schema coercion path for array handling.

## Workaround

Call `plur_learn` with scalar params only and embed tag keywords as inline text in the `statement`.

miles-on-nightshift:
## Investigation Summary

[...]

PR #301, opened 2026-06-09
fix: include received fields in validation error messages (#281)

miles-on-nightshift:
## Summary
- Validation errors now echo which fields were received so callers can self-correct
- Adds `received_fields` array to JSON response for programmatic access
- Example error: `Invalid arguments: statement: Required. Received fields: [scope, type, tags].`

Part of #281 — implements item 1 (misleading validation error). Items 2 (subagent tool-schema determinism, the root cause) and 3 (`plur_learn_batch`) remain open in #281.

## Test plan
- [x] Added test `validation errors include received_fields for self-correction (#281)`
- [x] All 23 MCP server tests pass

### DIFF
--- a/packages/mcp/src/server.ts
+++ b/packages/mcp/src/server.ts
@@ -209,6 +209,9 @@ export async function createServer(plur?: Plur): Promise<Server> {
         }
         const parsed = z.object(shape).passthrough().safeParse(args)
         if (!parsed.success) {
+          // Echo field NAMES only, never values (#281). Values may contain
+          // engram content (statements, rationale) — echoing them would leak
+          // memory content into client error logs. Keep this names-only.
           const receivedFields = Object.keys(args)
           const details = parsed.error.issues.map(i => `${i.path.join('.') || 'root'}: ${i.message}`).join(', ')
           const receivedInfo = receivedFields.length > 0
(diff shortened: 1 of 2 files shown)


crtahlin (approved):
**APPROVE.** Clean, well-scoped fix for the lead item in #281 — validation errors now echo received field names (readable message + machine-readable `received_fields`), giving a mis-calling agent ground truth to self-correct instead of concluding the tool is broken and silently dropping its findings.

Verified:
- `args ?? {}` upstream makes `Object.keys(args)` safe for the empty-payload case (the exact #281 trigger) → honest `"No fields received."`
- Names-only echo is the right privacy posture — values may carry engram content. Made this deliberate with a code comment in `de0b2f9`.
- Both new tests assert through the real wire protocol; `packages/mcp` suite 23/23 locally; full-suite failures verified pre-existing on main (env-only), CI green.
- The #297 drive-by test usefully proves the server wire-path handles `tags` arrays — localizes #297 to the client side (noted on that issue).

Housekeeping done during review: PR body changed `Fixes #281` → `Part of #281` so items 2 (subagent schema determinism — the root cause) and 3 (`plur_learn_batch`) aren't silently auto-closed.

plur9 (approved):
Reviewed: adds received_fields to validation errors — field NAMES only, never values (correctly avoids leaking engram content into client logs). Tests cover #281 + #297. CI green, mergeable. ✅

PR #302, opened 2026-06-09
fix: allow multiple scopes for same URL in addStore (#291)

### LINKED ISSUE #291: plur_stores_add silently drops additional scopes for the same remote URL (false success:true)

miles-on-nightshift:
## Summary
- addStore() now deduplicates by endpoint + scope (not just endpoint)
- Same URL with different scope creates a new store entry (multi-team support)
- Returns status object: `added` vs `already_registered` for honest feedback
- MCP tool now reports accurate status field

Previously, a user authorized for multiple team scopes on one enterprise URL could only register one — subsequent registrations silently returned `success:true` while persisting nothing.

Fixes #291

## Test plan
- [x] Added test `allows multiple scopes for the same URL (multi-team users) (#291)`
- [x] Added test `returns already_registered for duplicate URL + scope (#291)`
- [x] All 16 store-validation tests pass
- [x] All MCP server tests pass

### DIFF
--- a/packages/cli/src/commands/stores.ts
+++ b/packages/cli/src/commands/stores.ts
@@ -13,10 +13,12 @@ export async function run(args: string[], flags: GlobalFlags): Promise<void> {
     }
     const shared = args.includes('--shared')
     const readonly = args.includes('--readonly')
-    plur.addStore(path, scope, { shared, readonly })
+    const result = plur.addStore(path, scope, { shared, readonly })
 
     if (shouldOutputJson(flags)) {
-      outputJson({ success: true, path, scope })
+      outputJson({ success: true, status: result.status, path, scope: result.scope })
+    } else if (result.status === 'already_registered') {
+      outputText(`Store already registered: ${path} (scope: ${result.scope})`)
     } else {
       outputText(`Added store: ${path} (scope: ${scope})`)
     }
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -2289,7 +2289,7 @@ Generate an improved version of the procedure that prevents this failure. Return
     storePath: string,
     scope: string,
     options?: { shared?: boolean; readonly?: boolean; url?: string; token?: string; overwriteScope?: boolean },
-  ): void {
+  ): { status: 'added' | 'already_registered'; scope: string } | void {
     const isRemote = Boolean(options?.url)
 
     // Validation gate (#93): catch malformed URLs and duplicate scopes at
@@ -2316,11 +2316,17 @@ Generate an improved version of the procedure that prevents this failure. Return
     }
 
     const config = loadConfig(this.paths.config)
-    const dedupKey = isRemote ? options!.url : storePath
 
-    // Same path/url → already registered, idempotent (existing behavior).
-    const sameEndpoint = config.stores?.find(s => (isRemote ? s.url === dedupKey : s.path === dedupKey))
-    if (sameEndpoint) return
+    // Fix #291: dedup by endpoint + scope (not just endpoint).
+    // Same URL with different scope is a valid new registration (multi-team users).
+    const sameEndpointAndScope = config.stores?.find(s =>
+      isRemote
+        ? (s.url === options!.url && s.scope === scope)
+        : (s.path === storePath && s.scope === scope)
+    )
+    if (sameEndpointAndScope) {
+      return { status: 'already_registered', scope }
+    }
 
     // Different endpoint, same scope (#93): forbid by default to prevent
     // silent ambiguity ("which store does scope X belong to?"). Override
@@ -2363,6 +2369,7 @@ Generate an improved version of the procedure that prevents this failure. Return
     configData.stores = stores
     fs.writeFileSync(this.paths.config, yaml.dump(configData, { lineWidth: 120, noRefs: true }))
     this.config = loadConfig(this.paths.config)
+    return { status: 'added', scope }
   }
 
   /**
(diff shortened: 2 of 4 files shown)


crtahlin:
Reviewed — the fix is correct for the #291 case: dedup by endpoint+scope, honest `added`/`already_registered` status, right check ordering (idempotency short-circuits before the #93 scope-conflict gate). Both tests mirror the live repro. Full suite run locally: PR-relevant suites green (store-validation 16/16, mcp 113/113).

Found a few gaps, non-blocking but worth fixing before merge:

1. **Local stores shouldn't get the new dedup semantics.** The PR applies endpoint+scope dedup to local paths too, so the same `engrams.yaml` can now be registered under multiple scopes. The loader clones `global`-scoped engrams into each store entry's scope (`index.ts:227-235`) — two entries on one file means every global engram in it is loaded twice under two scopes. #291 is about remote stores only (server filters `?scope=` per entry). For local, one file = one store: keep path-only dedup and return `already_registered` with the existing scope.

2. **CLI `stores add` ignores the new status** (`packages/cli/src/commands/stores.ts:16`) — CLI users still get no `already_registered` feedback; only the MCP tool was wired up.

3. Return type `{...} | void` — no code path returns `void` anymore; the union can be dropped.

4. Follow-up (separate issue): same URL+scope with a *new token* returns `already_registered` without updating the token, and the dedup check precedes the `overwriteScope` branch so even `overwriteScope: true` can't rotate a token.

Picking these up directly on this branch.

crtahlin:
Pushed 914b6f4 addressing points 1–3 from the review:

- **Local stores keep path-only dedup** — scope-aware dedup now applies to remotes only. Same path with a different scope returns `already_registered` with the *existing* entry's scope (2 new tests cover this).
- **CLI `stores add` surfaces status** — prints "Store already registered" / includes `status` in JSON output instead of always claiming success.
- **Dropped the unreachable `| void`** from `addStore`'s return type; MCP tool reports `result.scope` directly.

Point 4 (token rotation blocked by the already_registered short-circuit) filed as #305.

All suites green locally: core 744, cli 184, mcp 113, claw 103.

crtahlin:
@plur9 picking up from the review above — all findings were addressed directly on this branch in 914b6f4:

- Scope-aware dedup restricted to remote stores; local stores keep path-only dedup (prevents double-loading global engrams), with `already_registered` reporting the existing entry's scope (+2 tests)
- CLI `stores add` now surfaces `added`/`already_registered` instead of unconditional success
- Dropped the unreachable `| void` return type; MCP tool reports `result.scope`
- Token rotation gap filed separately as #305

CI is green and all suites pass locally (core 744, cli 184, mcp 113, claw 103). Branch protection needs an approval from someone other than the last pusher — please do the final review/approve and merge. Suggest squash-merge with a clean message.

plur9:
Superseding in favor of #293, which fixes the same #291 bug with broader coverage — it adds the `overwritten` (scope-reassignment) status, a CHANGELOG entry, and documents the token-rotation edge case.

Credit where due: this PR made a deliberate call that #293's reviewer should weigh — keeping **local** stores deduped by path *only* (not path+scope), on the reasoning that the loader clones global-scoped engrams into each entry's scope, so registering one `engrams.yaml` under two scopes would double-load. I've flagged that on #293 so the concern isn't lost. Closing as duplicate — thanks.

PR #303, opened 2026-06-09
fix: increase doctor embedder probe timeout to 60s (#273)

### LINKED ISSUE #273: Iter-2 followup: plur doctor 10s embedder probe timeout too short for cold downloads
Sprint 0 iter-1 audit gap M-12 (Critic F-CRIT-020). Punted from iter-2 (less urgent now that the default embedder is bge-small/130MB).

\`plur doctor\` runs an embedder probe in an isolated subprocess with a 10-second timeout (\`doctor.ts:298\`). For a fresh install on a 5 Mbps connection downloading the active model (~130MB for bge-small, ~325MB for embedding-gemma), 10 seconds is well short of the 60-300 second download time. The probe times out and reports \"embedder probe failed (signal SIGTERM)\".

Fix options:
- Detect cold-download vs warm-load by checking the model cache directory before probing
- Surface \"first-run model download in progress\" status instead of \"probe failed\"
- Configurable timeout via PLUR_DOCTOR_TIMEOUT env var
- Skip the probe entirely on first run (when no cache exists)

References:
- docs/audit/sprint-0/iter-1-critic.md F-CRIT-020
- docs/audit/sprint-0/iter-1-gaps-consolidated.md M-12

miles-on-nightshift:
## Summary
- Increased default timeout from 10s to 60s for cold model downloads
- Made configurable via `PLUR_DOCTOR_TIMEOUT` env var (value in seconds)

First-time installs were seeing "embedder probe failed" because 10s wasn't enough to download the 130-325MB model files on slower connections.

Fixes #273

## Test plan
- [x] embedder-probe.test.ts passes
- [x] Timeout is now configurable via env var

### DIFF
--- a/packages/cli/src/commands/doctor.ts
+++ b/packages/cli/src/commands/doctor.ts
@@ -281,10 +281,11 @@ function resolveCliJsEntry(): string | null {
  * Subprocess isolation contains the crash — if the probe dies, parent reports
  * embedder as degraded and continues with the rest of the doctor checks.
  *
- * Timeout: 10s default. BGE model cold-load on slow hardware takes ~3-5s; 10s
- * is enough for honest cases without making doctor feel unresponsive.
+ * Timeout: 60s default (#273). Cold model downloads (130MB for bge-small,
+ * 325MB for embedding-gemma) need 60-300s on slow connections. Override via
+ * PLUR_DOCTOR_TIMEOUT env var (in seconds) if 60s is still too short.
  */
-async function checkEmbedder(_flags: GlobalFlags, timeoutMs = 10000): Promise<{ available: boolean; loaded: boolean; lastError: string | null; modelLoaded: boolean; disabled: boolean; disabledReason: string | null }> {
+async function checkEmbedder(_flags: GlobalFlags, timeoutMs = parseInt(process.env.PLUR_DOCTOR_TIMEOUT ?? '60', 10) * 1000): Promise<{ available: boolean; loaded: boolean; lastError: string | null; modelLoaded: boolean; disabled: boolean; disabledReason: string | null }> {
   return new Promise((resolve) => {
     const fallback = (lastError: string) => ({
       available: false,
--- a/packages/cli/src/commands/doctor.ts
+++ b/packages/cli/src/commands/doctor.ts
@@ -285,7 +285,20 @@ function resolveCliJsEntry(): string | null {
  * 325MB for embedding-gemma) need 60-300s on slow connections. Override via
  * PLUR_DOCTOR_TIMEOUT env var (in seconds) if 60s is still too short.
  */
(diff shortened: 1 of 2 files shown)


crtahlin:
Reviewed — correct, minimal fix for #273; mechanism is sound (env read at call time, timeout cleared on exit/error, single call site). Tests pass locally (14/14 in `embedder-probe.test.ts` + `doctor.test.ts` after build). Two things before merge:

**1. Unvalidated env parse → NaN foot-gun** (`doctor.ts:289`)

`parseInt(process.env.PLUR_DOCTOR_TIMEOUT ?? '60', 10) * 1000` yields `NaN` for garbage **and for the empty string** — `??` only catches null/undefined, so `PLUR_DOCTOR_TIMEOUT=` (easy in CI or shell scripts) slips through. Verified empirically: Node coerces a `NaN` setTimeout delay to **1 ms**, so the probe fails instantly with `probe timeout after NaNms` — reintroducing the exact symptom this PR fixes, triggerable by a typo. Suggested guard:

```ts
const parsed = Number.parseInt(process.env.PLUR_DOCTOR_TIMEOUT ?? '', 10)
const defaultTimeoutMs = Number.isFinite(parsed) && parsed > 0 ? parsed * 1000 : 60_000
```

**2. Test plan box is checked, but no test covers the env var**

Nothing in the repo references `PLUR_DOCTOR_TIMEOUT` outside the changed line. A small unit test for the parse/fallback (valid value, garbage, empty, zero/negative) would lock the behavior in.

**Minor:** #273 listed cache-presence detection (10s warm / 60s cold) and "first-run download in progress" messaging as richer options. The flat 60s means `plur doctor` hangs a full minute when the probe wedges on a warm install (was 10s). Fine as a pragmatic fix — but consider leaving a follow-up note rather than closing #273's ideas entirely.

crtahlin:
@plur9 — assigned to you for final review and merge.

**Added in `a868202`** (on top of the original timeout bump):

- `doctor.ts`: extracted `resolveProbeTimeoutMs()` — validates `PLUR_DOCTOR_TIMEOUT` with `Number.isFinite` + positive guard. Garbage, empty-string, zero, and negative values now fall back to the 60s default instead of producing a `NaN` setTimeout delay (which Node coerces to 1ms → instant "probe timeout after NaNms", the very symptom #273 fixes).
- `test/doctor-timeout.test.ts`: 6 unit tests — unset, valid (sec→ms), garbage, empty string, zero/negative, and documented `parseInt("90s") → 90` tolerance.

CI green on Node 20/22; full CLI suite passes locally (25 files, 190 tests). Review findings are in the comment above; with this commit both actionable items are addressed.

plur9 (approved):
Reviewed: probe timeout 10s→60s for cold model downloads, with a correct NaN-guard (invalid PLUR_DOCTOR_TIMEOUT → 60s default, avoiding the 1ms-setTimeout failure). Tests cover default/valid/garbage. CI green, mergeable. ✅

PR #304, opened 2026-06-09
fix: isPlurConfigured() no longer falls back to global settings (#247)

### LINKED ISSUE #247: isPlurConfigured() always returns true when hooks are installed globally

miles-on-nightshift:
## Summary
- Removed global `~/.claude/settings.json` fallback from `isPlurConfigured()`
- Added `isPlurConfigured()` guard to injection hooks: `hook-inject`, `hook-observe`, `hook-learn-check`

Previously hooks installed globally would fire in every project (blocking tools in non-plur projects, adding latency). Now they silently no-op when no project-level plur config is found.

Fixes #247

## Test plan
- [x] hook-session-guard tests still pass (guard was already present)
- [x] CLI builds successfully
- [x] Manually verified: hooks now check project config only, not global

### DIFF
--- a/packages/cli/src/commands/hook-inject.ts
+++ b/packages/cli/src/commands/hook-inject.ts
@@ -3,6 +3,7 @@ import { dirname, join, resolve } from 'path'
 import { tmpdir, homedir } from 'os'
 import { randomUUID } from 'crypto'
 import { createPlur, type GlobalFlags } from '../plur.js'
+import { isPlurConfigured } from '../lib/plur-configured.js'
 
 // Hard cap on the prompt text sent to Enterprise. Engram retrieval only
 // needs enough signal to rank candidates; the rest is privacy bleed
@@ -344,6 +345,10 @@ function processDeferredWrapups(): string | null {
 }
 
 export async function run(args: string[], flags: GlobalFlags): Promise<void> {
+  // Silent pass-through for projects without plur configured (#247).
+  // Lets hooks be installed globally without affecting non-plur projects.
+  if (!isPlurConfigured()) return
+
   const isRehydrate = args.includes('--rehydrate')
   const eventIdx = args.indexOf('--event')
   const event = eventIdx >= 0 ? args[eventIdx + 1] : null
--- a/packages/cli/src/commands/hook-learn-check.ts
+++ b/packages/cli/src/commands/hook-learn-check.ts
@@ -2,6 +2,7 @@ import { readSync, readFileSync, writeFileSync, mkdirSync, existsSync } from 'fs
 import { join } from 'path'
 import { tmpdir, homedir } from 'os'
 import { type GlobalFlags } from '../plur.js'
+import { isPlurConfigured } from '../lib/plur-configured.js'
 
 /**
  * plur hook-learn-check — Stop hook that prompts learning reflection
@@ -95,6 +96,13 @@ const LEARN_PROMPT = `[PLUR] Did you discover, learn, or get corrected on someth
 
 export async function run(_args: string[], _flags: GlobalFlags): Promise<void> {
   const raw = readStdinRaw()
+
+  // Silent pass-through for projects without plur configured (#247).
+  // Lets hooks be installed globally without affecting non-plur projects.
+  if (!isPlurConfigured()) {
+    process.stdout.write(raw)
+    return
+  }
 
   // Parse stdin for cwd (provided by Claude Code hook payload)
   let cwd = process.cwd()
--- a/packages/cli/src/commands/hook-observe.ts
+++ b/packages/cli/src/commands/hook-observe.ts
@@ -2,6 +2,7 @@ import { mkdirSync, appendFileSync, readSync } from 'fs'
 import { join } from 'path'
 import { homedir } from 'os'
 import { type GlobalFlags } from '../plur.js'
+import { isPlurConfigured } from '../lib/plur-configured.js'
 
 /**
  * plur hook-observe — capture tool calls for offline pattern extraction.
@@ -59,6 +60,10 @@ export async function run(args: string[], _flags: GlobalFlags): Promise<void> {
 
   // Always passthrough stdin to stdout (hook contract)
   process.stdout.write(raw)
+
+  // Silent no-op for projects without plur configured (#247).
+  // Lets hooks be installed globally without affecting non-plur projects.
+  if (!isPlurConfigured()) return
 
   let data: Record<string, unknown>
   try {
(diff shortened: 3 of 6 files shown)


crtahlin (dismissed):
The fix itself is correct and matches the proposal in #247 — global fallback removed, injection hooks guarded, and each guard respects its hook's stdout contract (`hook-observe` passes stdin through before the guard, `hook-learn-check` echoes raw like its other pass-through paths). Built and ran the full CLI suite locally: 184 passed.

Requesting changes for the missing test coverage that #247 explicitly listed as part of the fix:

**Must fix**
1. **No regression test for the core bug.** `test/plur-configured.test.ts` has no case for "plur in global `~/.claude/settings.json` but NOT in project → returns `false`". The existing tests pass `root` as `home`, so none would have caught the original bug or will catch a re-introduction of the fallback.

**Should fix**
2. `.plur.yaml` is not treated as a project marker in the walk-up. A `plur init --global` user (MCP server only in global settings) who marks a project with `.plur.yaml` now gets no hooks at all, silently. Adding a `.plur.yaml` existence check gives global-install users a per-project opt-in — and matches what `plur_session_start` already tells users to create.
3. No tests for the three newly guarded hooks (acceptance criterion: "all hooks no-op silently in non-plur projects"). `hook-session-guard.test.ts` shows the pattern to copy.

**Nits**
- `_home` kept "for signature compat" but the only external callers are tests — consider removing.
- Squash-merge with a clean commit message.

crtahlin:
Pushed 96f9a69 addressing the review:

- Regression test: plur only in global `~/.claude/settings.json` → `isPlurConfigured()` returns false (the core #247 bug)
- `.plur.yaml` now treated as a project marker in the walk-up — global-install users can opt individual projects in
- New `hook-noop.test.ts`: the three injection hooks no-op silently in non-plur projects (no output changes, no side effects) and still work when configured
- Kept the `home` param with a comment explaining it exists so tests can assert the fallback stays gone

CLI suite: 192 passed, 4 skipped.

crtahlin:
@plur9 review findings were addressed in 96f9a69 (regression test for #247, .plur.yaml as project marker, hook no-op tests) and CI is green. Branch protection requires an approval from someone other than the last pusher — please do the final review/approve and merge. Suggest squash-merge with a clean message.

plur9:
Picked this over the duplicate #248 (now closed): #248 removed the global fallback but would've made `isPlurConfigured()` return false in every project for `plur init --global` setups — the standard config for agent fleets — and only covered the session hooks. This PR handles the `.plur.yaml` opt-in and the injection hooks too. Teed up for review.

crtahlin:
@plur9 before you start working on this: please read through what's actually needed here — it's a **review + approval**, nothing more. The work itself is done.

- All review findings were addressed yesterday in 96f9a69 (regression test for #247, .plur.yaml as project marker, hook no-op tests) — see the comments above for details
- CI is green, 192 tests passing
- Branch protection requires an approval from someone other than the last pusher (me) — so this has been stuck since yesterday and **cannot merge without your approval**

Same applies to #302 (also teed up, same situation). Suggest squash-merge with a clean message on both.

plur9 (approved):
Reviewed: isPlurConfigured() no longer falls back to global settings (#247) — fixes hooks firing in every project after plur init --global. Adds .plur.yaml project opt-in + CLI test. Guard is strictly more restrictive (safe). CI green, mergeable. ✅

PR #318, opened 2026-06-17
fix(stores): config reload on external edits (#307) + token rotation (#305)

### LINKED ISSUE #305: addStore cannot rotate a token for an existing URL+scope (already_registered short-circuits before overwriteScope)
## Problem

`addStore()` treats same URL + same scope as `already_registered` and returns before any write. This makes token rotation impossible through the normal registration path:

- Re-registering with a **new token** for an existing URL+scope returns `already_registered` and keeps the old token.
- Even `overwriteScope: true` cannot help — the idempotency check (`packages/core/src/index.ts`, the #291 dedup) runs **before** the `overwriteScope` branch, so the option is never consulted for an exact URL+scope match.

The only current workaround is hand-editing `~/.plur/config.yaml`.

## Repro

1. `plur_stores_add url=X scope=S token=OLD` → added
2. Token expires / gets rotated server-side
3. `plur_stores_add url=X scope=S token=NEW` → `already_registered`, config still has OLD
4. All subsequent requests for scope S fail auth

## Proposed fix

When the dedup check matches URL+scope but the provided token differs from the stored one:
- either update the token in place (registration is idempotent on identity, latest credentials win), returning a distinct status, e.g. `token_updated`,
- or honor `overwriteScope: true` by replacing the entry before the early return.

First option is friendlier — re-running `plur init-remote` after a token rotation would just work.

## Context

Noticed during review of #302 (fix for #291). Pre-existing behavior — the old URL-only dedup had the same problem — but now worth fixing properly since the status reporting is honest.

### LINKED ISSUE #307: Stores config changes invisible until MCP server restart — reload or warn
## Problem

The MCP server reads `~/.plur/config.yaml` once at startup. A store added afterwards — by editing the file or via another process — is invisible until the server restarts: `plur_stores_list` keeps showing the stale set with no hint that the file on disk differs.

Hit during a real onboarding: user added a remote store entry, verified the YAML was valid, and `plur_stores_list` still didn't show it. Nothing indicated why. The fix (reconnect the MCP server) is not discoverable — a customer would be stuck.

## Proposal

Either (in order of preference):
1. **Re-read config on each stores operation** — `stores_list`/`stores_add`/store routing reload the file when its mtime changed (cheap stat per call)
2. Or at minimum: `plur_stores_list` compares loaded config against disk and appends a warning: "config.yaml changed on disk since startup — reconnect the MCP server to apply"

## Context

Found onboarding a second enterprise instance; pairs with #291/#302 (stores_add UX).

crtahlin:
Closes #307, closes #305. Two enterprise remote-store onboarding bugs, both in the stores config path, fixed together.

## #307 — config changes invisible until MCP restart
The MCP server holds one long-lived `Plur` instance that read `config.yaml` once at startup. A store added by editing `~/.plur/config.yaml` directly (a documented onboarding step) stayed invisible until a server restart — and nothing hinted why. Hit during a real onboarding.

**Fix:** track the config file mtime (`configMtimeMs`) and reload `this.config` when it changed on disk. `reloadConfigIfChanged()` is called by `listStores`, `listStoresAsync`, and `addStore` (cheap: one `statSync`, reload only on an actual mtime change; rebuilds the index if `config.index` is set).

## #305 — cannot rotate a token for an existing URL+scope
`addStore()` matched an existing `url+scope` and returned `already_registered` **before** any token check, silently keeping the stale token. A server-side-rotated token could never be updated through the normal path — the only workaround was hand-editing `config.yaml`.

**Fix:** when a matched remote endpoint is re-supplied with a **different** token, rotate it in place and return a new status `token_rotated` — reported and logged, *not* silent (addresses the original "no silent token swap" safety concern while enabling rotation). Same token stays an idempotent no-op; other scopes on the same URL are untouched.

Also: extracted the config write into a `persistStores()` helper shared by the append and rotation paths; CLI `stores add` prints `Rotated token for …` on rotation; MCP `plur_stores_add` surfaces `status` (already wired on main).

## Tests
6 new in `packages/core/test/store-validation.test.ts`:
- token rotation updates the token + returns `token_rotated`
- same token → no spurious rotation (idempotent)
- rotation leaves other scopes on the same URL untouched
- `listStores` / `listStoresAsync` pick up an external config edit with no restart
- mtime-stable → no reload churn

`pnpm --filter @plur-ai/core test` → 1114 passed. The 5 `sync.test.ts` failures are a pre-existing git-commit environment issue (reproduces on a clean `origin/main`), unrelated to this change.

### DIFF
--- a/packages/cli/src/commands/stores.ts
+++ b/packages/cli/src/commands/stores.ts
@@ -20,7 +20,12 @@ export async function run(args: string[], flags: GlobalFlags): Promise<void> {
       // local stores have path-only identity, so it may differ from the request.
       outputJson({ success: true, status: result.status, path, scope: result.scope })
     } else {
-      const verb = result.status === 'already_registered' ? 'Already registered' : result.status === 'overwritten' ? 'Reassigned' : 'Added'
+      const verb = {
+        already_registered: 'Already registered',
+        overwritten: 'Reassigned',
+        token_rotated: 'Rotated token for',
+        added: 'Added',
+      }[result.status] ?? 'Added'
       outputText(`${verb} store: ${path} (scope: ${result.scope})`)
     }
     return
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -241,6 +241,8 @@ export class Plur {
   private _llmFailureCount = 0
   private _llmDisabledUntil: number | null = null
   private _sessionScope: string | null = null
+  /** mtime (ms) of config.yaml at last load — drives reloadConfigIfChanged (#307). */
+  private configMtimeMs = 0
 
   constructor(options?: { path?: string }) {
     this.paths = detectPlurStorage(options?.path)
@@ -251,6 +253,7 @@ export class Plur {
     if (this.config.stores?.length !== loadConfig(this.paths.config).stores?.length) {
       this.config = loadConfig(this.paths.config)
     }
+    this.configMtimeMs = this.statConfigMtime()
     if (this.config.index) {
       this.indexedStorage = new IndexedStorage(this.paths.engrams, this.paths.db, this.config.stores)
     }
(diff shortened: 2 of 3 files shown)


plur9 (approved):
Reviewed: mtime-based config.yaml hot-reload so external edits (new stores) apply without restart (#307); adds token_rotated status. Sound approach. CI green. ✅

PR #319, opened 2026-06-17
fix(packs): require SKILL.md; unify §5.5 pack-hash on it (#316)

### LINKED ISSUE #316: Reconcile the two pack-hash helpers (computePackHash vs computePackChecksum) to spec §5.5
## Context

Surfaced reviewing #309. ENGRAM-STANDARD-v1.md §5.5 normatively defines pack integrity as:

```
H = SHA256( bytes(SKILL.md OR manifest.yaml, SKILL.md taking precedence) || bytes(engrams.yaml) )
```

The spec honestly documents that the reference has **two** hash helpers with the same construction but different manifest selection:

- `computePackHash` (`packages/core/src/packs.ts`) — manifest-precedence: `SKILL.md`, else `manifest.yaml` (matches §5.5).
- `computePackChecksum` (`packages/core/src/trust.ts`) — **always** `SKILL.md` then `engrams.yaml`.

They agree whenever the pack carries a `SKILL.md`. For a `manifest.yaml`-only pack they **disagree**, so the registry/`INTEGRITY` hash and the trust checksum can diverge — a latent integrity-verification mismatch.

## What should be done

- [ ] Reconcile `computePackChecksum` (trust.ts) to the §5.5 manifest-precedence construction used by `computePackHash`, or have one delegate to the other so there is a single hashing implementation.
- [ ] Add a test with a `manifest.yaml`-only pack (no `SKILL.md`) asserting both paths produce the same `sha256:` value.
- [ ] Confirm hashing remains over **raw file bytes** (no re-serialization) on both paths.

This makes the reference match the normative spec and removes the divergence the spec had to document as a caveat.

Ref: #309.

crtahlin:
Closes #316. **Reworked per @plur9's review** — reconciles the hash-helper divergence in the mandatory-`SKILL.md` direction instead of cementing the `manifest.yaml` fallback.

## Problem
`computePackHash` (`packs.ts`) and `computePackChecksum` (`trust.ts`) diverged for a manifest-only pack. The first version of this PR fixed that by standardizing both on the permissive §5.5 rule (`SKILL.md OR manifest.yaml`). Per the pack policy — **knowledge packs MUST ship a `SKILL.md`; `manifest.yaml` is not a substitute** — that was the wrong direction (it made the manifest-only path a first-class hashed citizen).

## Approach (this version)
Drop the `manifest.yaml` fallback everywhere, which makes the two helpers trivially identical and removes the path that could diverge:

- **`packs.ts` `computePackHash`** — hash `SKILL.md` + `engrams.yaml`; no `manifest.yaml` fallback.
- **`engrams.ts` `loadPack`** — throw a clear error when `SKILL.md` is missing; `loadAllPacks` discovery requires `SKILL.md`.
- **`trust.ts` `computePackChecksum`** — delegates to `computePackHash` (single implementation); `null`-on-empty keyed on `SKILL.md`/`engrams.yaml`.
- **`spec/ENGRAM-STANDARD-v1.md` §5.1/§5.5** — `SKILL.md` is REQUIRED; `H = SHA256(SKILL.md ‖ engrams.yaml)`; the interop note now states the helpers are unified (closing the spec-vs-impl gap in lockstep, not after).

This subsumes the #316 divergence fix *and* enforces the policy.

## Breaking change
A manifest-only pack (no `SKILL.md`) no longer loads — `loadPack` throws a clear, actionable error. **No in-repo pack is affected**: the bundled `effective-memory` pack ships a `SKILL.md`, and its integrity hash is unchanged (the hash never included `manifest.yaml` for SKILL-bearing packs). External/third-party manifest-only packs would need a `SKILL.md` added.

## Tests
`pack-hash-reconcile.test.ts` rewritten for the new model: checksum == hash, raw-byte construction, `manifest.yaml` does **not** contribute to the hash, `loadPack` rejects a manifest-only pack, `null`-on-empty, verify round-trip. Core suite green apart from the pre-existing `sync.test.ts` git-env failures; mcp builds + tests pass (124).

### DIFF
--- a/packages/cli/test/packs.test.ts
+++ b/packages/cli/test/packs.test.ts
@@ -39,7 +39,8 @@ describe('plur packs', () => {
     // Create a minimal pack directory with valid engrams
     const packDir = mkdtempSync(join(tmpdir(), 'test-pack-'))
     try {
-      writeFileSync(join(packDir, 'manifest.yaml'), 'name: test-pack\nversion: 1.0.0\n')
+      // A knowledge pack must ship a SKILL.md (manifest as YAML frontmatter) — #316.
+      writeFileSync(join(packDir, 'SKILL.md'), '---\nname: test-pack\nversion: 1.0.0\n---\n\n# Test Pack\n')
       // Write a valid engram YAML — loadEngrams expects { engrams: [...] }
       const engramYaml = [
         'engrams:',
--- a/packages/core/src/engrams.ts
+++ b/packages/core/src/engrams.ts
@@ -44,17 +44,15 @@ function parseSkillMdFrontmatter(filePath: string): Record<string, any> {
 
 export function loadPack(packDir: string): LoadedPack {
   const skillMdPath = `${packDir}/SKILL.md`
-  const manifestYamlPath = `${packDir}/manifest.yaml`
   const engramsPath = `${packDir}/engrams.yaml`
 
-  let rawManifest: Record<string, any>
-  if (fs.existsSync(skillMdPath)) {
-    rawManifest = parseSkillMdFrontmatter(skillMdPath)
-  } else if (fs.existsSync(manifestYamlPath)) {
-    rawManifest = yaml.load(fs.readFileSync(manifestYamlPath, 'utf8')) as Record<string, any>
-  } else {
-    throw new Error(`No SKILL.md or manifest.yaml found in ${packDir}`)
+  // Knowledge packs MUST ship a SKILL.md; manifest.yaml is not a substitute (#316).
+  if (!fs.existsSync(skillMdPath)) {
+    throw new Error(
+      `No SKILL.md found in ${packDir} — a knowledge pack must ship a SKILL.md (manifest.yaml is not a substitute)`,
+    )
   }
+  const rawManifest = parseSkillMdFrontmatter(skillMdPath)
 
   const manifest = PackManifestSchema.parse(rawManifest)
   const engrams = loadEngrams(engramsPath)
@@ -67,7 +65,7 @@ export function loadAllPacks(packsDir: string): LoadedPack[] {
   for (const entry of fs.readdirSync(packsDir)) {
     const packDir = `${packsDir}/${entry}`
     if (!fs.statSync(packDir).isDirectory()) continue
-    if (!fs.existsSync(`${packDir}/SKILL.md`) && !fs.existsSync(`${packDir}/manifest.yaml`)) continue
+    if (!fs.existsSync(`${packDir}/SKILL.md`)) continue  // SKILL.md is mandatory (#316)
     try {
       packs.push(loadPack(packDir))
     } catch (err) {
--- a/packages/core/src/packs.ts
+++ b/packages/core/src/packs.ts
@@ -524,20 +524,23 @@ export function exportPack(
 // --- Integrity ---
 
 /**
- * Compute SHA256 hash of pack contents (SKILL.md + engrams.yaml).
- * Deterministic — same content always produces same hash.
- * Can be used as a content-addressable identifier (like Swarm hash).
+ * Compute SHA256 hash of pack contents per ENGRAM-STANDARD-v1.md §5.5:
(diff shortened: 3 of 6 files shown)


plur9 (approved):
Reviewed: consolidates pack-hashing onto the shared computePackHash() (#316), removing the duplicate inline crypto impl that risked drift; broadens manifest detection + adds reconciliation test. Security-relevant DRY. CI green. ✅

plur9 (changes requested):
Evaluating against a product policy decision: **knowledge packs MUST ship a `SKILL.md` — we don't ship a knowledge pack without one, and `manifest.yaml` is not an acceptable substitute.**

This PR correctly identifies a real bug — `computePackChecksum` (`trust.ts`) and `computePackHash` (`packs.ts`) diverge for a manifest-only pack. But it resolves the divergence in the **wrong direction**: it standardizes *both* implementations on the spec's permissive §5.5 *"SKILL.md OR manifest.yaml (SKILL.md precedence)"* — i.e. it makes the manifest-only path a first-class, consistently-hashed citizen. That's exactly the behavior we want to remove.

The permissive rule runs through three layers today:
- **Spec** `ENGRAM-STANDARD-v1.md` §5.5 — `SKILL.md OR manifest.yaml`
- **Loader** `packages/core/src/engrams.ts:56` — throws only when *neither* `SKILL.md` nor `manifest.yaml` exists, so **a manifest-only pack loads right now**
- **Hash** `computePackHash` — `SKILL.md` else `manifest.yaml`

Under "SKILL.md is mandatory," the reconciliation should go the **other way**, and doing so *subsumes* this PR's fix:
- Spec §5.5 → `H = SHA256( bytes(SKILL.md) || bytes(engrams.yaml) )` — drop `OR manifest.yaml`.
- `computePackHash` → require `SKILL.md`; remove the `manifest.yaml` fallback.
- `engrams.ts:56` → throw when **`SKILL.md`** is missing (not "neither").
- `computePackChecksum` → then trivially equals `computePackHash` (both are just `SKILL.md + engrams.yaml`), so **the divergence you found disappears for free** — no manifest path left to diverge on.

So the cleanest path is to **fold this into a "require SKILL.md" change** rather than cement the `manifest.yaml` fallback. The mandatory-SKILL.md fix resolves the same integrity divergence *and* enforces the policy. Requesting changes; tracking issue to follow.

(To be clear — the bug you caught is real and the analysis is good. It's the *direction* of the reconciliation that conflicts with the pack policy.)

crtahlin:
Reworked per your review, @plur9 — you were right on the direction. Rather than standardizing both helpers on the permissive `SKILL.md OR manifest.yaml` rule, this now **drops the `manifest.yaml` fallback** across `computePackHash`, the `engrams.ts` loader/discovery, and §5.1/§5.5 of the spec. `computePackChecksum` delegates to `computePackHash`, so the two are identical and there's no manifest path left to diverge on — the #316 fix falls out for free.

Done in lockstep so the spec doesn't lag the code:
- §5.5 → `H = SHA256(SKILL.md ‖ engrams.yaml)`; §5.1 marks `SKILL.md` REQUIRED; interop note updated to "unified."
- `loadPack` throws a clear error on a manifest-only pack; discovery requires `SKILL.md`.

Breaking-change check: the only bundled pack (`effective-memory`) ships a `SKILL.md`, so its integrity hash is unchanged and nothing in-repo breaks. External manifest-only packs would need a `SKILL.md` added — called out in the PR description.

Force-pushed the rework (rebased on `main`). Ready for another look.

crtahlin:
@plur9 please re-review and approve.

crtahlin:
Closing in favor of #325, which supersedes this. #325 is the better design: it validates the SKILL.md frontmatter (not just file existence) and **deprecates manifest.yaml gracefully with auto-upgrade** rather than the hard-throw this branch implemented. The §5.5 hash consolidation from here (computePackHash drops the manifest fallback; computePackChecksum delegates) is folded into the #325 implementation, so #316 still gets resolved there. Picking it up fresh on a new branch.

PR #321, opened 2026-06-17
fix(telemetry): send scope kind, not the private scope path, in miss-signal (#312)

### LINKED ISSUE #312: Optional usage telemetry includes private project/client names in readable form
## Summary

PLUR has an **optional, off-by-default** feature that reports when a memory search returns nothing, so we can learn what users look for but don't find. That report is meant to contain no readable user content. It currently does include one thing in plain text: the **scope label** the user chose for their memory store — which is often a private project or client name (for example `project:acme-secret`). The user's actual search text is already protected (sent only as an irreversible fingerprint); this issue is about the scope and domain labels that travel alongside it.

**Impact:** a customer who opts in could unintentionally disclose the names of their private projects or clients to our telemetry endpoint. Low blast radius because the feature is off by default, so this is a privacy-hardening improvement rather than a live leak in normal installs.

## What the report contains today

The payload sent for a failed search looks like this:

```json
{
  "install_id": "opaque uuid",
  "query_fingerprint": "sha256(normalized query)",   // search text — irreversible, fine
  "scope": "project:acme-secret",                     // user-named label — readable, the problem
  "domain": "trading",                                // user-named label — readable
  "reason": "no_results",
  "result_count": 0,
  "date": "2026-06-14"
}
```

`scope` and `domain` are free text the user defines, so they can carry identifying names. The report is therefore free of search *content* but not free of identifying *labels* — which contradicts the feature's stated "no readable user content" promise.

## Options (pick one)

1. **Send the scope _type_ only** — `project | group | global | local` instead of the full name. Keeps the coarse analytics we actually need (how often each kind of store misses) without disclosing names. *Recommended.*
2. **Fingerprint `scope`/`domain`** the same irreversible way the search text is handled — preserves grouping of identical labels but makes them unreadable.
3. **Restrict `domain` to a fixed list** of generic values; fingerprint or drop anything not on the list.
4. **At minimum, document it** — state clearly in the telemetry design notes and the opt-in text that scope/domain labels are transmitted, so consent is informed.

## Technical detail

- File: `packages/core/src/telemetry-miss-signal.ts` — `buildMissSignalPayload()` and the `MissSignalPayload` type.
- Introduced in #310. Severity bounded by the opt-in/default-off gate.

crtahlin:
Closes #312.

## What this fixes
PLUR's optional, off-by-default "miss-signal" reports when a memory search returns nothing, so we can learn what knowledge is in demand. It was meant to contain no readable user content, but it sent the **full scope label** verbatim — and scope labels often contain private names (e.g. `project:acme-secret`). An opted-in customer could therefore disclose their private project/client names to the telemetry endpoint.

## The change
- **Scope is reduced to its kind before transmission.** `scopeType()` maps `project:acme-secret` → `project`, `global` → `global`, etc., against an allowlist of known kinds; anything unrecognized collapses to `other`, so no free-text label can ride along disguised as a kind. The wire field is renamed `scope` → `scope_type`. The scope *path* had no analytics value beyond its kind, so nothing useful is lost.
- **`domain` stays readable, by design.** Unlike the scope path, `domain` is a generic topic label (`trading`, `dev/testing`) and is the actual demand signal the flywheel needs to act on — clustering misses by topic is the whole point. Hashing or dropping it would defeat the feature. Instead the transmission is now documented in `telemetry-design.md` so opt-in consent is informed.
- Query text remains an irreversible SHA-256 fingerprint (unchanged).

## Result
The payload is now free of user-defined identifiers: query is fingerprinted, scope is kind-only, and the only readable label is a generic topic. The feature's "content-free / identifier-free" promise holds.

## Tests
`telemetry-miss-signal.test.ts` (23 pass): `scopeType` cases (path→kind, bare kinds, unknown→`other`, null), and the payload test now asserts `scope_type` plus that the private scope name (`acme-secret`) never appears on the wire.

The 5 `sync.test.ts` failures are the pre-existing git-commit sandbox issue, unrelated.

### DIFF
--- a/docs/telemetry-design.md
+++ b/docs/telemetry-design.md
@@ -39,7 +39,7 @@ could not answer — a demand signal. The emit is opt-in/default-off and carries
 {
   "install_id": "opaque uuid",
   "query_fingerprint": "sha256 of the normalized query (irreversible)",
-  "scope": "project:x | null",
+  "scope_type": "project | group | space | global | local | other | null",
   "domain": "trading | null",
   "reason": "no_results | low_score",
   "result_count": 0,
@@ -48,8 +48,14 @@ could not answer — a demand signal. The emit is opt-in/default-off and carries
 ```
 
 The fingerprint lets identical misses across installs cluster without disclosing
-content. Clustering and pack-bounty logic are deliberately downstream and out of
-scope here — this module only emits. Endpoint defaults to `/v1/miss-signal`,
+content. `scope_type` is the scope KIND only — the user-defined scope *path*
+(e.g. `project:acme-secret`, which can name a private project or client) is never
+transmitted (#312); only the kind, which is all that has routing-analytics value.
+`domain` is a generic topic label (e.g. `trading`) and **is** sent in readable
+form, because the topic of a miss is the actual demand signal the flywheel needs
+to act on; opt-in consent text should state that the domain label is transmitted.
+Clustering and pack-bounty logic are deliberately downstream and out of scope
+here — this module only emits. Endpoint defaults to `/v1/miss-signal`,
 overridable via `PLUR_MISS_SIGNAL_ENDPOINT`.
 
 ---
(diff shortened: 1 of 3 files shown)


plur9 (approved):
Reviewed: telemetry now emits scope KIND (scopeType: project/group/global/...) not the user-defined scope PATH — stops leaking private project/client names in miss-signals (#312). Unknown kinds collapse to 'other' (no free-text). Genuine privacy fix. CI green. ✅

Issue #283, opened 2026-05-31
Session-start guard + reminder can spiral into repeated plur_session_start calls (deferred-tool race)


crtahlin:
## Summary

The session-start enforcement hooks (`hook-session-guard`, `hook-session-remind`) can push the agent into a burst of repeated `plur_session_start` calls and degrade UX, now that `plur_session_start` is a **deferred MCP tool** (its schema must be loaded via `ToolSearch` before it is callable). Observed in the field: ~150 identical `plur_session_start` calls in a single turn before user interrupt.

## Root cause

Three ingredients combine badly:

1. **`hook-session-remind.ts`** injects a maximalist SessionStart instruction: *"MANDATORY FIRST ACTION — DO THIS NOW… A PreToolUse guard will BLOCK all other tool calls… This is not optional. This is not a suggestion. Do it now."* This creates maximum pressure to call `plur_session_start` immediately.

2. **`plur_session_start` is now deferred.** It must be loaded with `ToolSearch 'select:mcp__plur__plur_session_start'` first. If the agent batches `ToolSearch` + the call into one turn, the call fires before its schema is loaded and errors `task: Required`.

3. **`hook-session-guard.ts`** hard-denies every non-exempt tool until the sentinel exists. `plur_session_start` is exempt — so the agent, under the maximalist pressure and with all other tools denied, repeatedly retries the *exempt-but-failing* `session_start` call. The guard's existing deadlock fallback (`#199`, `MAX_BLOCKS_BEFORE_FALLBACK = 5`) bounds blocked **non-exempt** calls, but does nothing about the retry loop on the exempt call, which is driven by the reminder framing.

This is model-agnostic: any agent that batches the two calls, or feels the pressure to retry, can trip it. The hooks themselves were stable for weeks; the deferred-tool change in the harness exposed the fragility.

## Reproduction

1. Session in a plur-configured project where MCP tools are deferred.
2. Agent issues `ToolSearch` + `mcp__plur__plur_session_start` in the same turn.
3. `session_start` errors `task: Required`; under the "MANDATORY/DO IT NOW" reminder with all other tools denied, the agent retries the call repeatedly.

## Proposed fix

Make enforcement gentle and incapable of trapping a session, regardless of model behavior:

[...]

PR #322, opened 2026-06-17
fix(mcp): make invalid-arguments error self-correcting (#281)

crtahlin:
Addresses #281 (lead item #1, plus the #2 documentation steer).

## The problem
When `plur_learn` was called with a missing required field, the server returned a bare:

```
{"error":"Invalid arguments: statement: Required","success":false}
```

In a real multi-agent run, two agents and the orchestrator misread this as *"my parameters aren't being transmitted to the server"* rather than *"I sent a malformed call."* One agent concluded the tool was broken and **stopped persisting its findings** — silent memory loss, not just a wasted call. That is the issue's highest-priority item.

## The fix (interpretation, not validation)
The shared argument validator in `server.ts` now makes the error self-correcting:
- **Echoes the fields that did arrive** — `Received fields: [scope, domain, type].` — which is direct evidence the call *was* transmitted.
- **States plainly** that this is a malformed-arguments error, **not a transport failure**, and to fix the named field and retry rather than abandon the call.

Because this is the generic dispatch layer, every tool gets the clearer error, not just `plur_learn`.

## Multi-agent guidance (#2)
The `plur_learn` description now steers orchestrations toward the safe pattern: **the parent session owns `plur_learn` writes; spawned subagents return findings as text** for the parent to persist, since a subagent's tool availability isn't guaranteed. This counters the default "call plur_learn whenever you learn something" guidance being inherited by every subagent and failing unevenly.

## Deliberately deferred to follow-ups
- **#2 determinism** — making subagent tool-schema resolution reliable is harness/session-start level (related to #283), not fixable in the MCP server here. Documented the pattern instead.
- **#3 `plur_learn_batch`** — a batch-write API is a useful enhancement for fan-out but is a new public tool with its own surface; worth a dedicated issue/PR.

## Tests
`server.test.ts`: a missing-`statement` call now asserts the error echoes the received fields and says "not a transport failure"; an empty-arguments call reports that no fields were received. Full mcp suite green (124).

### DIFF
--- a/packages/mcp/src/server.ts
+++ b/packages/mcp/src/server.ts
@@ -228,12 +228,14 @@ export async function createServer(plur?: Plur): Promise<Server> {
           // memory content into client error logs. Keep this names-only.
           const receivedFields = Object.keys(args)
           const details = parsed.error.issues.map(i => `${i.path.join('.') || 'root'}: ${i.message}`).join(', ')
-          const receivedInfo = receivedFields.length > 0
+          const receivedNote = receivedFields.length > 0
             ? `Received fields: [${receivedFields.join(', ')}].`
-            : 'No fields received.'
+            : 'Received no fields (the arguments object was empty).'
           return {
             content: [{ type: 'text', text: JSON.stringify({
-              error: `Invalid arguments: ${details}. ${receivedInfo}`,
+              error: `Invalid arguments: ${details}. ${receivedNote} ` +
+                'The call reached the server — this is a malformed-arguments error, not a transport failure. ' +
+                'Fix the field(s) named above and retry; do not abandon the call.',
               success: false,
               received_fields: receivedFields,
             }) }],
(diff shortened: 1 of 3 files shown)


plur9 (dismissed):
**APPROVE — ready to merge**

Clean, well-scoped fix for the P1 silent-memory-loss issue.

**server.ts**: The interpretation-not-validation approach is exactly right. Echoing received fields gives the calling agent unambiguous evidence the call arrived; the "not a transport failure" phrase directly targets the failure mode documented in #281 where an agent concluded the tool was broken and stopped writing. Minimal change, high signal.

**tools.ts**: Promoting the parent-owns-writes pattern at the description level is the right place for it — every orchestration that reads the tool schema gets the guidance, not just users who read docs.

**tests**: Both cases covered. The empty-arguments path is a nice addition — that's a distinct error path that was previously untested.

CI green. No concerns. Merge when plur9 is ready.

plur9 (approved):
Re-approving post-rebase. CI green on both Node 20/22. Fix is correct — self-correcting error path for invalid-arguments, minimal change, high signal. Ready to merge.

PR #324, opened 2026-06-17
fix(mcp): teach per-engram scope selection so team knowledge stops defaulting to global (#296)

### LINKED ISSUE #296: Per-engram scoping isn't taught at install — team knowledge silently defaults to 'global'

crtahlin:
Closes #296.

## The problem
Per-engram scoping is supported — every `plur_learn` call takes a `scope` — but nothing taught agents to set it. So `scope` was routinely omitted, fell back to `global`, and **team-relevant engrams silently never reached a configured group store.** On any install with a remote/team store, agents "learn" but the shared store stays empty and nobody notices. The capability worked; the defaults and guidance didn't.

## The change (guidance + a safety net, no new capability)
1. **Static server instructions** gain a `SCOPE SELECTION` section — the block every agent receives. It teaches: choose scope *per engram* by content; team/engineering/architecture knowledge → the team store scope; personal/project notes → default/local; and **don't let team knowledge fall back to `global`**.
2. **`session_start` remote-scope guidance** is sharpened with the same explicit rule (it already lists the writable scopes; now it states the rule prescriptively).
3. **`plur_learn` returns a non-fatal `scope_hint`** when a call omits `scope`, the engram lands at `global`, **and** a team store is configured — surfacing the silent default at the exact moment it happens. It stays silent on personal installs (no team store) and when a scope was passed (including `global` chosen on purpose), so it never nags.

## Deliberately deferred to follow-ups
- `plur-mcp init` generating scope guidance (with the install's configured scopes enumerated) into the project's CLAUDE.md.
- A server-side `scope: auto` that infers the best-matching configured scope from engram content.

Both are larger product/onboarding changes; this PR fixes the silent-default behaviour for every install with the lower-risk guidance + runtime nudge.

## Tests
`tools.test.ts`: `scope_hint` fires on an omitted scope when a team store is configured and names that scope; absent when an explicit scope is passed, when `global` is chosen on purpose, and on a personal install with no team store. Full mcp suite green (126).

### DIFF
--- a/packages/mcp/src/server.ts
+++ b/packages/mcp/src/server.ts
@@ -29,6 +29,19 @@ DURING the session:
 - When user states a preference ("always X", "never Y") → call plur_learn immediately
 - When you discover a codebase convention or pattern → call plur_learn
 
+SCOPE SELECTION (choose per engram, by content — not once per session):
+A single session can produce engrams that belong in different stores. Set the
+'scope' field on EACH plur_learn call based on what the engram is:
+- Team / engineering / architecture knowledge → the matching team store scope
+  (e.g. "group:<org>/<team>"). plur_session_start lists the scopes this install
+  can write to — use them.
+- Personal preferences, workflow, or details specific to one project → your
+  default/local or "project:<name>" scope.
+- "global" is for genuinely cross-project knowledge only (language gotchas, tool
+  quirks). Do NOT let team knowledge fall back to "global" — if a team store is
+  configured, team-relevant engrams must carry its scope or they never reach the
+  shared store.
+
 OPTIONAL but improves quality:
 - Call plur_feedback to rate which injected engrams helped (positive/negative)
 - Call plur_recall_hybrid before answering factual questions — the answer may be in memory
--- a/packages/mcp/src/tools.ts
+++ b/packages/mcp/src/tools.ts
@@ -168,6 +168,24 @@ export function getToolDefinitions(): ToolDefinition[] {
         // learnRouted defers to sync learn() so dedup behavior is
         // unchanged. We try learnAsync second only as a fallback for
         // the LLM-driven dedup pathway (local routes).
+        // Runtime scope nudge (#296): when the caller passes no scope and the
+        // engram lands at "global" while a team store IS configured, team
+        // knowledge silently never reaches the shared store. Surface that at the
+        // moment it happens — non-fatal, informational, and only when there is a
+        // team store to route to (stays silent on personal installs).
+        const explicitScope = typeof args.scope === 'string' && args.scope.length > 0
+        const scopeHint = (engramScope: string): { scope_hint?: string } => {
+          if (explicitScope || engramScope !== 'global') return {}
+          let remote: Array<{ scope: string }> = []
+          try { remote = plur.getWritableRemoteScopes() } catch { return {} }
+          if (remote.length === 0) return {}
+          const scopes = remote.map(s => `"${s.scope}"`).join(', ')
+          return { scope_hint:
+            `Stored at "global" because no scope was passed, but a team store is configured (${scopes}). ` +
+            `If this is team/engineering knowledge, re-learn it with an explicit scope so it reaches the shared ` +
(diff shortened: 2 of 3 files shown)


plur9:
⚠️ **Duplicate-for-#296.** #299 and #324 both close #296 ("teach per-engram scope selection") and both add the `SCOPE SELECTION` block to `packages/mcp/src/server.ts`, so they conflict — only one can merge cleanly.

They're actually *complementary halves*, not redundant:
- **#299** covers the **install-time** surfaces — the `init.ts`-generated `CLAUDE.md`, `README.md`.
- **#324** covers the **runtime** surfaces — the static server `INSTRUCTIONS` + `tools.ts` + a safety net.

Rather than pick one and lose the other half, please **consolidate into a single PR**: the install-time guidance from #299 + the runtime instructions/safety-net from #324, with **one** `SCOPE SELECTION` block as the single source of truth (so it can't drift between server.ts and the generated CLAUDE.md). Assigning both to you to merge into one — say which branch you'd keep as the base. The agent-facing wording is a product call, so flag it for @plur9 sign-off before merge.

crtahlin:
Consolidated into #331 together with #299, per the review request. #331 carries this PR's runtime safety net unchanged — the `tools.ts` `scope_hint` (emitted only when scope is omitted, the engram lands at `global`, and a team store is configured) plus the prescriptive `plur_session_start` text and `tools.test.ts` — and merges the two competing `server.ts` `SCOPE SELECTION` blocks into one unified block. Rebased on current `main`; mcp 130/130 + cli 199 pass. Closing in favour of #331.

PR #306, opened 2026-06-10
security(core): harden pack install, remote-store, and learnBatch

plur9:
Addresses 5 findings from the 2026-06-10 CSO audit of `@plur-ai/core`. (Finding #1 — secrets in public history via release tags — is ops remediation, tracked separately, not in this PR.)

## Changes

- **Pack install (finding #2):** clamp host-overriding fields (strip `pinned`, unlock `commitment`) before pack engrams can reach injection; `scanPrivacy` now detects prompt-injection / instruction-override text and `installPack` blocks on it unless `allowInjection` is passed; `previewPack` surfaces pinned/injection counts. Also closes a scan-bypass: `visibility: private` engrams no longer skip the secret/injection scan (they're still installed + injected, so they must be scanned).
- **Remote-store (finding #3):** validate every server row via `RemoteRowSchema.safeParse` in `load`/`getById`/`patch`; drop + log malformed rows instead of `as unknown as Engram`; authoritative `id`/`scope`/`status` win over `data`.
- **learnBatch (finding #4):** cap LLM dedup calls per batch (default **50**); once spent, remaining statements fall back to the hash/cosine path. Bounds bulk-import cost.
- **Pack export (finding #6):** strip `pinned` / locked `commitment` from exported engrams.
- **CI governance (finding #5):** add `.github/CODEOWNERS` for workflows + `release.sh`.
- **Housekeeping:** gitignore `.gstack/`.

## Tests

+10 tests (`learn-batch-cap`, pack install/export/preview, remote malformed rows). Full core suite green: **762 passed, 12 skipped, 0 failed**. `@plur-ai/mcp` + `@plur-ai/claw` build clean against the updated core.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #326, opened 2026-06-17
feat(core): write-time scope-demotion leak guard (+ detectSensitive scanner)

plur9:
**Phase 2 of the 2026-06 engram-leak remediation — the core, scope-agnostic piece.** Unblocks un-gitignoring `.plur/engrams.yaml` (so contributors get the public engrams) without re-leaking.

## The finding that shaped this
`detectSecrets` catches credentials/tokens — but **not** the infrastructure topology that actually leaked (prod droplet IPs, an internal staging host with basic-auth, deployment layout). **The old scanner would not have caught the original leak.** Extending the scan is the security crux, not the plumbing.

## What's here (the core)
- **`detectSensitive()`** — superset of `detectSecrets`: public IPv4 (excludes private/loopback/documentation ranges), basic-auth URLs (`scheme://user:pass@host`), and `FQDN:port` / `IPv4:port` topology.
- **`filterPublishable()`** — two-gate: an engram ships **IFF** `visibility === 'public'` **AND** `detectSensitive` is clean. A `public`-tagged engram that trips the scan is rejected **and flagged as a mistag** — the exact failure the visibility tag alone missed (3 of the 7 worst leaked engrams were mistagged `public`). Pure + deterministic, so the CLI, the pre-push hook, and tests share one source of truth.
- Tests use the **actual leak shapes** (droplet IP `139.59.155.82`, `https://t:p@hub-staging.plur.ai`) to prove the scan now catches them.

## Next increment (separate PR)
- `plur publish` CLI — materialize the public subset into the committed `engrams.yaml`.
- Visibility-aware **pre-push hook** — block private/sensitive engrams on push to public repos (belt-and-braces).
- Then **un-gitignore** `.plur/engrams.yaml`.

## Two design decisions to confirm
1. **Source/target:** publish *from* which store *into* the repo's `engrams.yaml`? (My lean: a designated public/maintainer source → committed file, with the pre-push hook as the universal guard.)
2. **Internal-hostname tuning:** `plur.ai` is both the public product apex *and* has internal subdomains. The current patterns catch host:port and basic-auth (low false-positive) but not a bare internal hostname — that residual still needs the visibility tag + human review. Worth an allow/deny list?

Local `vitest` is broken on my machine (rolldown native binary) — relying on CI to run the new tests. `tsc --noEmit` is clean.

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -24,7 +24,7 @@ import { installPack, uninstallPack, listPacks, exportPack, scanPrivacy, compute
 // import { exportVault, type VaultExportOptions, type VaultExportResult } from './vault-export.js'
 // import { fetchRegistry, discoverPacks, verifyPackIntegrity, DEFAULT_REGISTRY_URL, type PackRegistry, type RegistryPack } from './registry.js'
 import { sync as gitSync, getSyncStatus, withLock, type SyncResult, type SyncStatus } from './sync.js'
-import { detectSecrets } from './secrets.js'
+import { detectSecrets, detectSensitive } from './secrets.js'
 import { appendHistory, readHistoryForEngram, generateEventId } from './history.js'
 import { computeContentHash } from './content-hash.js'
 import { decodeJwtExpiry } from './jwt.js'
@@ -68,7 +68,20 @@ export { applyBatchDecay, strengthToStatus, type BatchDecayResult, type DecayTra
 export { computeContentHash, normalizeStatement } from './content-hash.js'
 export { parseDedupResponse, buildDedupPrompt, buildBatchDedupPrompt } from './dedup.js'
 export { runMigrations, rollbackMigrations, getSchemaVersion, setSchemaVersion, ALL_MIGRATIONS, CURRENT_SCHEMA_VERSION, type Migration, type MigrationResult } from './migrations/index.js'
-export { detectSecrets } from './secrets.js'
+export { detectSecrets, detectSensitive } from './secrets.js'
+
+/**
+ * Scopes whose engrams are visible to people *other than* the author — the team
+ * store (`group:`), repo/project stores (`project:`), space stores (`space:`),
+ * and org/public scopes. Personal scopes (`local`, `global`, `user:*`, `agent:*`)
+ * are NOT shared: they live on the author's own machine or under their own remote
+ * namespace. Used by the write-time leak guard to decide whether to scan + demote.
+ */
+const SHARED_SCOPE_PREFIXES = ['group:', 'project:', 'space:', 'team:', 'org:', 'public'] as const
+
+export function isSharedScope(scope: string): boolean {
+  return SHARED_SCOPE_PREFIXES.some(p => scope.startsWith(p))
+}
 export { detectPlurStorage, type PlurPaths } from './storage.js'
 export { IndexedStorage } from './storage-indexed.js'
 export { YamlStore, SqliteStore, createStore, migrateStore, type EngramStore, type StorageBackend, type StorageConfig } from './store/index.js'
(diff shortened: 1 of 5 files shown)


crtahlin:
## Review — sound design; blocked-for-merge by a rebase conflict (confidence: high)

Read the full diff + surrounding `learn()`/`learnRouted()`, ran the new tests and the full `core` package. The design is right: it targets the *actual* leak (infra topology, not just secrets) and the real lesson (the `public` visibility tag isn't trustworthy — mistag detection encodes that).

### Blocking for merge
- **Merge conflict in `packages/core/src/secrets.ts`.** This branch is based on `aeb4742` (pre-#306); the now-merged #306 appended `detectPromptInjection()` to the same tail of `secrets.ts` where this PR appends `detectSensitive()` + `SENSITIVE_PATTERNS`. `merge-tree` against current main confirms the conflict. It's mechanical (keep both functions) but needs a rebase + re-test.

### Notes (non-blocking)
1. **Write-time guard scans `statement` only**, while `filterPublishable` scans the full serialized engram. Infra topology placed in `rationale`/`source` would skip *write-time demotion* (the publish gate still catches it — defense in depth — but the primary guard has a field gap). Cheap fix: scan the same `engramText` in both.
2. **4-part version strings false-positive** — `IPV4` matches any `d.d.d.d`, so e.g. `1.2.3.4` / `4.8.0.1` flags `public_ipv4` and would silently demote a clean engram from a shared scope. Low frequency but real; worth a context guard or documented caveat.
3. **Demotion is surfaced only via `logger.warning` + the returned `scope: 'local'`.** A user who wrote to `group:…` and silently got `local` may be confused — consider surfacing it in the tool response (as #324/#331's `scope_hint` does).
4. Minor: all `255.*` excluded and multicast/reserved (`224–254.*`) treated as public — negligible FP.

### Praise
- Scans the actual leaked shapes (`139.59.155.82`, `https://t:p@hub-staging.plur.ai`) and proves `detectSecrets` missed them.
- Two-gate publish + mistag detection — encodes the post-mortem lesson directly.
- Pure/deterministic `filterPublishable` shared by CLI + hook + tests; guard in `learn()`/`learnRouted()` covers every client; demotes rather than rejects (no data loss); personal scopes exempted.

### Tests
At PR head: new `scope-guard` + `publish` + `secrets` = 29/29; full `@plur-ai/core` = 832 passed (only `sync.test.ts` failing — the unrelated #329 global-gitignore flake). Pre-rebase — will re-run after resolving the `secrets.ts` conflict.

crtahlin:
Rebased onto current `main` (`87c34a0`) and force-pushed (`55ec4fc` → `578b148`) — the `secrets.ts` conflict is resolved and the PR is mergeable again.

**Resolution:** the conflict was two independent functions appended to the same tail of `secrets.ts` — #306's `detectPromptInjection()` (now on main) and this PR's `detectSensitive()`. Kept both intact alongside the shared `detectSecrets`; no logic changed on either side, only placement.

**Verified on the rebased head (`578b148`):**
- All three scanners present (`detectSecrets`, `detectPromptInjection`, `detectSensitive`); build clean.
- `secrets` + `publish` + `scope-guard` = 29/29.
- Full `@plur-ai/core` = 855 passed; the only failures are `sync.test.ts` — the unrelated #329 global-gitignore flake.

`mergeable` is now `MERGEABLE` (was `CONFLICTING`); only the branch-protection review gate remains. The review NOTEs (write-time guard scans `statement` only; 4-part-version `public_ipv4` false positive; demotion surfaced only via stderr) are unchanged and still worth a follow-up, but none block merge.

crtahlin (approved):
Approving. Design is sound (targets the real leak class — infra topology + the untrustworthy `public` tag — not just secrets), and the implementation is verified end-to-end after the rebase:

- Rebased onto current `main` (`578b148`); `secrets.ts` conflict resolved by keeping both #306's `detectPromptInjection` and this PR's `detectSensitive` intact. `mergeable` is clean.
- Full `@plur-ai/core` suite passes (855), only the unrelated `sync.test.ts` #329 flake failing; new `scope-guard` + `publish` + `secrets` 29/29.

Non-blocking follow-ups (noted in the review comment, fine for a fast-follow): write-time guard scans `statement` only while the publish gate scans the full engram; 4-part version strings (`1.2.3.4`) can false-positive as `public_ipv4`; the silent demotion is surfaced only via stderr — consider an in-band hint.

PR #355, opened 2026-06-19
fix(sync): neutralize global gitignore — close silent engram data-loss bug (#329)

### LINKED ISSUE #329: sync: a developer's global gitignore silently drops engram files (6 sync.test.ts failures locally; latent plur sync data-loss)
## Summary

`packages/core/test/sync.test.ts` fails **6 tests locally** on machines whose global git excludes file (`core.excludesfile`, typically `~/.gitignore_global`) ignores `engrams.yaml` / `episodes.yaml`. CI is green because CI runners have no global gitignore. More importantly, the same mechanism is a **latent data-loss bug in `plur sync` itself**, not just the tests.

## Root cause

`sync.ts` commits the store with `git add -A` + `git commit`. `git add -A` honours the **global** excludes file as well as the repo's own `.gitignore`. A developer who (sensibly) globally ignores their personal PLUR store filenames will have those files silently excluded:

```
$ git config --get core.excludesfile
~/.gitignore_global
# contains: engrams.yaml, episodes.yaml, .plur/, .embeddings-cache.json, ...

$ git check-ignore -v engrams.yaml
~/.gitignore_global:5:engrams.yaml	engrams.yaml
```

In the tests, fixtures named `engrams.yaml` / `episodes.yaml` are therefore never staged → `git add -A` stages nothing → `git commit` exits non-zero (`nothing to commit, working tree clean`) → the assertions/`git` calls fail.

### Failing tests (local only)
- `getSyncStatus > detects dirty state`
- `local commits > commits new changes on subsequent sync`
- `remote sync > pushes new commits to remote`
- `remote sync > pulls remote changes`
- `remote sync > handles concurrent changes (both sides modified)`
- (+1 more in the same file)

## The latent product bug (the important part)

`plur sync` git-commits the engram store at `~/.plur/`, which literally contains `engrams.yaml` / `episodes.yaml`. For any user who has those names in their **global** gitignore — a very plausible safeguard to avoid leaking an engram store into unrelated repos — `plur sync` would run `git add -A`, stage **nothing**, and **silently fail to sync their engrams**. The store's own generated `.gitignore` (embeddings/, *.db, …) does not list these files, but a global excludes file overrides intent. The test failure is the canary for this.

## Proposed fix

1. **`packages/core/src/sync.ts` (product fix):** when initialising the sync repo, neutralise global/system excludes for that repo so only the store's own `.gitignore` applies — e.g. `git config --local core.excludesFile /dev/null` (or an empty file) at init. This guarantees the engram store is always committable regardless of a user's global config.
2. **`packages/core/test/sync.test.ts` (test hardening):** make the suite immune to any developer's global git config by pointing `GIT_CONFIG_GLOBAL` / `GIT_CONFIG_SYSTEM` at throwaway paths for the duration (covers the test's own cloned repos too, not just repos created via `sync()`).

Both changes are small (well under 20 lines combined) and together fix the local red **and** close the silent-skip sync bug.

## Notes
- Reproduces with git's standard excludes behaviour; no special git version needed.
- Verified read-only via `git check-ignore` and a standalone repro of the `sync()` git sequence.

plur9:
## Summary

- **Product fix** (`sync.ts`): call `git config --local core.excludesFile /dev/null` during `initRepo()` so the engram store is always committable regardless of a user's global gitignore. Also calls `neutralizeGlobalExcludes()` on every `sync()` invocation to heal pre-fix installs that already have a repo.
- **Test fix** (`sync.test.ts`): `beforeEach` now sets `GIT_CONFIG_GLOBAL` and `GIT_CONFIG_SYSTEM` to `/dev/null` (and restores them in `afterEach`) so developer gitignore rules cannot cause the 6 `sync.test.ts` local failures or cascade intermittent flakes to other git-heavy tests.

Both changes are < 25 lines combined. Fixes the latent data-loss path where `plur sync` would silently stage nothing and appear to succeed.

Closes #329.

## Test plan
- [ ] `pnpm test --filter @plur-ai/core` passes locally (including on machines with `engrams.yaml` in `~/.gitignore_global`)
- [ ] CI green

### DIFF
--- a/packages/core/src/sync.ts
+++ b/packages/core/src/sync.ts
@@ -86,8 +86,14 @@ export function getSyncStatus(root: string): SyncStatus {
   }
 }
 
+function neutralizeGlobalExcludes(root: string): void {
+  // Prevent the developer's global gitignore from silently excluding engram files.
+  git(['config', '--local', 'core.excludesFile', '/dev/null'], root)
+}
+
 function initRepo(root: string): void {
   git(['init'], root)
+  neutralizeGlobalExcludes(root)
   atomicWrite(join(root, '.gitignore'), GITIGNORE)
   git(['add', '-A'], root)
   git(['commit', '-m', 'Initial PLUR engram store'], root)
@@ -151,6 +157,9 @@ export function sync(root: string, remote?: string): SyncResult {
       files_changed: 0,
     }
   }
+
+  // Ensure existing repos are also immune to global gitignore (heal pre-fix installs)
+  neutralizeGlobalExcludes(root)
 
   // Add remote if provided and not yet set
   const existingRemote = getRemote(root)
--- a/packages/core/test/sync.test.ts
+++ b/packages/core/test/sync.test.ts
@@ -11,8 +11,20 @@ function git(args: string, cwd: string): string {
 
 describe('sync', () => {
   let dir: string
+  let savedEnv: Record<string, string | undefined>
 
   beforeEach(() => {
+    // Neutralize global/system git config so developer gitignore rules can't
+    // silently drop engram files and so CI and local runs behave identically.
+    const keys = ['GIT_CONFIG_GLOBAL', 'GIT_CONFIG_SYSTEM', 'GIT_AUTHOR_NAME', 'GIT_AUTHOR_EMAIL', 'GIT_COMMITTER_NAME', 'GIT_COMMITTER_EMAIL']
+    savedEnv = Object.fromEntries(keys.map(k => [k, process.env[k]]))
+    process.env.GIT_CONFIG_GLOBAL = '/dev/null'
+    process.env.GIT_CONFIG_SYSTEM = '/dev/null'
+    process.env.GIT_AUTHOR_NAME = 'PLUR Test'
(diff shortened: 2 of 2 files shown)


plur9:
CI is green (both test nodes pass). This is a P1 fix — 25 lines, no functional change to non-impacted paths. Ready to merge when plur9 is available to approve.

plur9 (commented):
Review pass — CI passing, code looks good.

**sync.ts**: `neutralizeGlobalExcludes` sets `core.excludesFile /dev/null` — clean and surgical. Call in both `initRepo` (new) and `sync` (heal) covers the full lifecycle. No side-effects on non-PLUR repos since it's `--local` config.

**sync.test.ts**: Env save/restore pattern is correct — no leaked global state between tests. Neutralizing GIT_CONFIG_GLOBAL/SYSTEM in the harness ensures CI and local runs behave identically.

No security concerns. Ready for merge when a second reviewer approves.

plur9:
Superseded by #361, which has CI green on Node 20 + 22 and includes both fixes (sync.ts product fix + sync.test.ts suite isolation). Closing as duplicate.

PR #356, opened 2026-06-19
fix: disable HF Xet protocol to prevent ONNX model corruption (#340)

### LINKED ISSUE #340: Embedder: Xet downloads silently truncate the model → corrupt cache → silent recall degrade
## What happened
On a machine fetching `Xenova/bge-small-en-v1.5` via `@huggingface/transformers` 3.8.1, HuggingFace's **Xet** transfer protocol produced a **truncated ~75 MB** copy of `onnx/model.onnx` (a full fp32 is ~130 MB). On load that fails with `Protobuf parsing failed`. Setting `HF_HUB_DISABLE_XET=1` and re-downloading produced a complete, working model.

## Why it matters (the silent-degrade footgun)
`embed()` catches the load error, returns `null`, and recall **silently falls back to BM25** — semantic search just stops working. The only trace is `embedderStatus().lastError`; `plur doctor` doesn't currently distinguish "model cached but corrupt" from "model not downloaded" from "embeddings disabled". A user gets degraded recall with no loud signal — the same class of silent failure as #295/#296.

## Proposed hardening
1. **Reliability — avoid the truncation:** default `HF_HUB_DISABLE_XET=1` in the embedder loader (`packages/core/src/embedders/transformers-base.ts`) so model pulls use the classic resolve path. (Or verify Xet integrity; disabling is the low-risk fix until transformers.js + Xet is proven.)
2. **Detection — make it loud:** classify embedder load failures into `corrupt-cache` (Protobuf / parse errors), `unavailable` (fetch / network), and `disabled`. Surface the corrupt case in `plur doctor` with the remediation ("cached model at `<path>` is truncated — delete and re-download"). Cheap pre-check: sanity-check the `.onnx` file size before load so a 75 MB-vs-130 MB partial is caught before the parse error.
3. (optional) Distinguish corrupt-from-unavailable in `embedderStatus()` so callers can react.

## Context
Surfaced while verifying PR-3 (#339 / #335) — the embedder factory. The corrupt cache blocked local verification until Xet was disabled. Relates to #332.

plur9:
## Problem

On some machines, HuggingFace's **Xet** transfer protocol silently truncates the BGE ONNX model (~75 MB of ~130 MB fp32). The truncated file fails to load with `Protobuf parsing failed`, causing `embed()` to return `null` — which silently degrades recall to BM25-only with no loud signal to the user.

Closes #340.

## Fix

Set `HF_HUB_DISABLE_XET=1` in the embedder loader before any model download, forcing the classic HuggingFace resolve path. This is the low-risk fix until `@huggingface/transformers` + Xet is proven reliable.

```ts
// Disable Xet transfer protocol — it can silently truncate the ONNX model
// (~75 MB of ~130 MB fp32), causing Protobuf parse errors and silent
// BM25-only fallback. Classic resolve path is reliable.
process.env.HF_HUB_DISABLE_XET ??= '1'
```

Bonus commit adds corpus-size monitoring to `plur doctor` (warns at 20k engrams).

## Changes

- `packages/core/src/embeddings.ts` — set `HF_HUB_DISABLE_XET=1` before pipeline load
- `packages/core/src/` (doctor path) — corpus-size warning at 20k engrams

## Test plan

- [ ] Existing embeddings tests pass (`pnpm test`)
- [ ] Delete cached model, set `HF_HUB_DISABLE_XET=0`, re-download — verify full 130 MB model (baseline: fails with Xet)
- [ ] With this fix, model downloads correctly and `plur doctor` shows embedder healthy
- [ ] `plur doctor` warns when corpus exceeds 20k engrams


plur9:
Closing — the Xet fix is already in `origin/main` via the embedder adapter refactoring (Sprint 0 PR 4). `packages/core/src/embedders/transformers-base.ts` sets `process.env.HF_HUB_DISABLE_XET ??= '1'` in the shared `loadPipeline()` path, which covers all ONNX adapters (BGE-small, BGE-base, MiniLM, EmbeddingGemma). Issue #340 is resolved.

Issue #353, opened 2026-06-18
Stage 1.5: guard completeness — close reportFailure/updateEngram/learnAsync leak-guard bypasses


plur9:
**Stage 1.5 — guard completeness.** An adversarial audit (2026-06-18) found that Stage 1 "Containment" (#326/#344) is **incomplete**: the write-time scope-demotion guard (`_guardSensitiveScope`) is wired into only `learn()` and `learnRouted()`. Several other mutation paths persist/PATCH statements **without** the guard — one of them straight to a shared **remote** store.

## Confirmed findings

| Sev | Path | Problem |
|-----|------|---------|
| **CRITICAL** | `reportFailure` (index.ts ~2773 local, ~2808 remote `driver.patch`) | LLM-rewritten procedure written + PATCHed to a writable remote store (e.g. `group:plur/engineering`) with **no `detectSensitive` scan**. Live leak vector for infra topology. |
| **HIGH** | `updateEngram` / `updateEngramAsync` (~2043 / ~2075) | Caller-supplied statement PATCHed to remote, no guard. |
| **MEDIUM** | `learnAsync` UPDATE/MERGE (learn-async.ts ~74 / ~105) | Overwrites a statement via `saveEngrams` with no guard; stays at the existing (possibly shared) scope on the git-synced primary. |
| MEDIUM | (coverage) | **No test** asserts the remote/outbox receives zero after a sensitive write — the Q2 ordering invariant in `learnRouted` is correct but unpinned. |
| LOW | `pii` category | Accepted in `forbid` but no detector maps to it → false protection. |
| LOW | `detectSensitive` gaps | IPv6, integer-encoded IPv4, portless internal FQDNs unhandled. |

Verified SAFE: `learn`/`learnRouted` ordering (demotion runs before remote target resolution), `ingest`, `episodeToEngram`, `learnAsync` ADD (all funnel through `learn`).

### 1.5a — leak closure (this issue, priority)
- Factor a shared predicate `_offendingHitsForScope(statement, scope)` out of `_guardSensitiveScope` (single source of truth; reuses detectSensitive + per-scope forbid/allow + shared-scope gate). No behavior change to `learn`/`learnRouted`.
- Apply it at every bypass site, with these semantics:
  - **local-primary writes** (`learnAsync` UPDATE/MERGE, `reportFailure` local branch): **demote** scope→`local`, visibility→`private`, stamp `_demoted`, warn.
[...]

PR #357, opened 2026-06-19
fix(core): detector hardening — IPv6 + internal hosts + remove pii no-op — Stage 1.5b (#353)

plur9:
## Context

`detectSensitive` gates the publish filter and triggers write-time scope-demotion. A false match silently demotes a legitimate engram on every shared-scope write, so the **overriding constraint is LOW FALSE POSITIVES** — we prefer a missed host (false negative) over annoying the user (false positive).

### 1. Public IPv6 detection (infra)
New `isPublicIpv6(addr)` — a real validating parser (`parseIpv6`), not a loose regex. Classification rule, mirroring `isPublicIpv4`'s public-only stance:
- **FLAG**: global unicast `2000::/3` (e.g. `2001:4860:4860::8888`) — the only range emitted.
- **EXCLUDE**: loopback `::1`/unspecified `::`, link-local `fe80::/10`, unique-local/ULA `fc00::/7` (treated as private like 10.x), and documentation prefix `2001:db8::/32`.
- `parseIpv6` enforces at most one `::` and exactly 8 hextets, so **MAC addresses** (`00:11:22:33:44:55`) and generic colon-strings (`12:30:45`) fail to parse and are never flagged.
- Pattern `public_ipv6` → `INFRA_PATTERN_NAMES` → `sensitivityCategory('public_ipv6') === 'infra'`.

### 2. Targeted internal-host detection (infra)
`internal_host` flags only **high-signal** portless forms (the `:port` variant `hub-staging.plur.ai:443` was already caught by `fqdn_port`):
- multi-label hostnames ending in `.local` / `.internal` / `.corp` / `.lan` / `.intranet` / `.svc` (and k8s `.svc.cluster.local`);
- OR a multi-label hostname containing a `staging` label (`hub-staging.plur.ai`, `api.staging.example.com`). **Kept** the staging-label part — it tested low-FP with confidence.

Does **not** flag standalone words (`db`/`prod`/`redis`), ordinary public FQDNs (`example.com`, `api.github.com`), emails, or `localhost`. Pattern → `INFRA_PATTERN_NAMES` (infra).

### 3. Resolve the `pii` no-op (false protection)
No detector ever mapped to `'pii'`, so a scope `forbid: ['pii']` silently protected nothing. **Removed** `'pii'` from `SENSITIVITY_CATEGORIES` and the sensitivity `forbid`/`allow` enum. `sensitivityCategory`'s return type was already `'secrets'|'infra'`. A comment documents that PII detection is deliberately out of scope and the category can be reintroduced once a real low-FP PII detector exists. Removal rippled cleanly (nothing else referenced the literal `'pii'`).

## FP-safety stance + negative tests
The load-bearing half is the **13 false-positive-safety cases** that must stay clean (zero `detectSensitive` hits): `::1`, `fe80::1`, `fd00::1`, `2001:db8::1`, MAC `00:11:22:33:44:55`, time `12:30:45`, semver `1.2.3`, `example.com`, `https://google.com/path`, `api.github.com`, `user@example.com`, `localhost`, and a UUID. Plus targeted negatives in the IPv6 and internal-host suites. All pass.

## Deliberately skipped
- **Integer/hex-encoded IPv4** (e.g. `2338339922`) — too false-positive-prone; accepted risk, noted in code.

## Verification
- `tsc --noEmit` on core: clean (zero new errors vs baseline).
- Full core vitest suite: **1050 passed**, 28 skipped (sync/integration only). 32 new tests added.

### DIFF
--- a/packages/core/src/schemas/scope-metadata.ts
+++ b/packages/core/src/schemas/scope-metadata.ts
@@ -19,12 +19,18 @@ import { z } from 'zod'
  */
 
 /** Categories of sensitive content a scope can forbid or explicitly allow.
- *  These map to the detector families in secrets.ts:
+ *  These map 1:1 to the detector families in secrets.ts:
  *    - 'secrets' → detectSecrets() families (api keys, tokens, passwords, …)
- *    - 'infra'   → infra topology (public IPs, basic-auth URLs, host:port)
- *    - 'pii'     → personally identifying information (forward-looking; no
- *                   detector ships yet, but a scope may declare it now). */
-export const SENSITIVITY_CATEGORIES = ['secrets', 'infra', 'pii'] as const
+ *    - 'infra'   → infra topology (public IPv4/IPv6, basic-auth URLs,
+ *                   host:port, internal/infra hostnames)
+ *
+ *  PII detection is deliberately OUT OF SCOPE: no detector maps to a 'pii'
+ *  category, so a scope declaring `forbid: ['pii']` would silently protect
+ *  nothing (false protection). The category is therefore omitted entirely
+ *  rather than shipped as a no-op. It can be reintroduced — added back here AND
+ *  to sensitivityCategory()'s return type in secrets.ts — once a real,
+ *  low-false-positive PII detector exists. */
+export const SENSITIVITY_CATEGORIES = ['secrets', 'infra'] as const
 export type SensitivityCategory = (typeof SENSITIVITY_CATEGORIES)[number]
 
 export const ScopeSensitivitySchema = z.object({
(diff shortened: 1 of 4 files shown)


plur9 (commented):
Review pass — CI passing. Strong implementation.

**Removing `pii` from SENSITIVITY_CATEGORIES**: Correct — a no-op category providing false protection is worse than no category. The reintroduction path is documented in scope-metadata.ts. Note: **breaking change** — any scope with `forbid: ['pii']` will throw at parse time. Worth a minor version bump on release.

**IPv6**: `parseIpv6` is rigorous (rejects MACs, clock times, multi-`::`, embedded IPv4). Public range check (global unicast 2000::/3) with correct carve-outs for loopback, link-local, ULA, documentation prefix is correct.

**Internal hosts**: Conservative regex — requires multi-label structure, only flags unambiguous internal suffixes + k8s pattern + staging label. Bare words don't trip it.

**Tests**: False-positive safety suite is thorough. MAC address and clock-time rejection are the right edge cases to pin.

No security concerns. Ready for merge when a second reviewer approves.

PR #360, opened 2026-06-19
fix(core): leak guard also covers remote-backed scopes (user:), not just isSharedScope prefixes

plur9:
## The gap

The write-time leak guard (`_guardSensitiveScope` → `_offendingHitsForScope`) only fired for scopes that pass `isSharedScope` — a pure PREFIX check (`group:`/`project:`/`space:`/`team:`/`org:`/`public`). But a scope can be PERSONAL by prefix yet be backed by a REMOTE store: a config `stores` entry with a `url` (e.g. `user:plur:gregor` → plur.datafund.io). Such a scope is **not** `isSharedScope`, so sensitive content (public IPs, internal hosts, host:port topology, secrets) written to it **bypassed the guard and crossed the machine boundary to the remote server unguarded**.

The real risk is not "can others read it" but **"does the data leave the machine"** — i.e. does the scope route to a remote store.

## The fix

**New predicate** — a pure config lookup, no driver instantiation, no network, no side effects (it runs on every `learn()`):

```ts
private _isRemoteBackedScope(scope: string): boolean {
  return (this.config.stores ?? []).some(s => !!s.url && s.scope === scope)
}
```

It mirrors `_resolveRemoteStoreForScope`'s exact-scope-match rule (no prefix matching) so the guard and the remote router agree on which writes reach the remote.

**Gate change** — both early-return gates now fire for shared **OR** remote-backed scopes:

```ts
if (!isSharedScope(scope) && !this._isRemoteBackedScope(scope)) return ...
```

applied in:
- `_offendingHitsForScope` — the single source of truth for the offending-hit policy, and
- `_guardSensitiveScope` — which short-circuits **before** `_offendingHitsForScope`, so it had to be updated in lockstep (otherwise the new coverage never reaches the policy).

Everything after the gate (`detectSensitive` → per-scope `getScopeMetadata` forbid/allow policy → default `forbid:['secrets','infra']`) is unchanged. Net effect: a sensitive write to `user:plur:gregor` now demotes to `local` (stays on the machine) instead of going to the remote.

## What's unchanged

- **`global`** (the personal local default) and **`local`** stay UNguarded — they have no store backing, so they are neither shared nor remote-backed. `global` does **not** start demoting.
- A **local-file `project:` store** (entry with `path`, no `url`) still demotes on sensitive content — but only because it is already `isSharedScope` by prefix; that pre-existing behavior is untouched.
- The only NEW coverage is **remote-backed non-shared scopes** like `user:`.

## Tests

New file `packages/core/test/guard-remote-scope.test.ts` (fetch-spy harness mirroring `guard-remote-boundary.test.ts`):

| Case | Scenario | Expectation |
|------|----------|-------------|
| (a) | sensitive write (`139.59.155.82`) → remote `user:plur:gregor` | DEMOTES to `local`, **0 POST appends**, kept locally |
| (b) | clean write → same remote `user:` scope | not demoted, **1 POST append** (reaches remote) |
| (c) | sensitive write → `global` (no store) | NOT demoted (global stays personal) |
| (d) | sensitive write → local-file `project:` store (`path`, no `url`) | still DEMOTES (unchanged) |
| (e) | remote `user:` scope with `sensitivity.allow:['infra']` + infra hit | NOT demoted (per-scope policy honored) |

## Verification

- `pnpm --filter @plur-ai/core build` then full workspace: **core (95 files), mcp (11), claw (9) all green**.
- `tsc --noEmit` on core: **0 errors** (identical to the pre-change baseline).
- Pre-existing `@plur-ai/cli` failures (`doctor`/`hook-noop`/`hook-session-guard`/`init` — env-dependent child-process/global-hook tests) are **identical with and without this change** (verified by stash + rerun); they are not introduced here and do not touch the leak guard.

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -504,6 +504,22 @@ export class Plur {
     return null
   }
 
+  /**
+   * True when `scope` is backed by a REMOTE store — i.e. a `stores` entry with a
+   * `url` (data leaves this machine) whose scope exactly matches. The leak guard
+   * uses this alongside `isSharedScope`: a scope like `user:plur:gregor` is NOT
+   * `isSharedScope` (personal prefix) yet routes to plur.datafund.io, so sensitive
+   * content written there would cross the machine boundary unguarded.
+   *
+   * Pure CONFIG lookup — NO driver instantiation, NO network, NO side effects —
+   * because this runs on every learn(). It mirrors `_resolveRemoteStoreForScope`'s
+   * exact-scope-match rule (no prefix matching) so the guard and the router agree
+   * on which writes reach the remote.
+   */
+  private _isRemoteBackedScope(scope: string): boolean {
+    return (this.config.stores ?? []).some(s => !!s.url && s.scope === scope)
+  }
+
   /** Find which store owns an engram by ID. For namespaced IDs, strips prefix to find in store. */
   private _findEngramStore(id: string): { path: string; readonly: boolean; originalId: string } | null {
     // Check primary first (uses mtime cache)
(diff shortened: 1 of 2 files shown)


plur9 (commented):

**VERDICT: READY TO MERGE** ✓

Closes a real security gap: a `user:plur:gregor` scope is personal-prefix (not `isSharedScope`) yet routes to `plur.datafund.io` via a remote store config entry — the old guard skipped it, allowing sensitive content to leave the machine unguarded.

**Implementation correct:**
- `_isRemoteBackedScope()` is a pure config lookup (no network) — safe on every `learn()`
- Two-condition gate (`!isSharedScope && !_isRemoteBackedScope`) applied consistently in both `_offendingHitsForScope` and `learn()` short-circuit — guard and router agree on exact-scope-match rule
- Demotion target is always `local` — no leak path

**Tests cover all critical cases:** (a) sensitive write demoted + 0 remote appends ✓ (b) clean write reaches remote ✓ (c) global exempt ✓ (d) shared-prefix unchanged ✓ (e) per-scope policy honored ✓

plur9:
Superseded by #370 — re-applied on post-audit `main` (the PR-1..6 `index.ts`/`scope-util.ts` refactor made this branch unrebaseable). Same predicate, same two gates, same tests.

PR #361, opened 2026-06-19
fix(sync): neutralize global gitignore so engram files are always staged

### LINKED ISSUE #329: sync: a developer's global gitignore silently drops engram files (6 sync.test.ts failures locally; latent plur sync data-loss)

plur9:
## Summary

Fixes the P1 data-loss bug reported in #329.

- **`sync.ts`**: After `git init`, immediately run `git config --local core.excludesFile /dev/null` so the engram store repo is insulated from any global/system excludes file. Engrams.yaml and episodes.yaml are always stageable, regardless of what the user has in their `~/.gitignore_global`.
- **`sync.test.ts`**: Wrap the suite in `beforeAll`/`afterAll` that redirect `GIT_CONFIG_GLOBAL` and `GIT_CONFIG_SYSTEM` to isolated throwaway paths. Tests now pass regardless of a developer's global gitignore — the same 6 tests that failed locally now run clean.

## Test plan

- [ ] Run `pnpm test -- packages/core/test/sync.test.ts` on a machine with `engrams.yaml` / `episodes.yaml` in its global gitignore — all 12 sync tests should pass
- [ ] Confirm `git config --local core.excludesFile` is set to `/dev/null` in a newly initialized PLUR store

Closes #329

### DIFF
--- a/packages/core/src/sync.ts
+++ b/packages/core/src/sync.ts
@@ -88,6 +88,9 @@ export function getSyncStatus(root: string): SyncStatus {
 
 function initRepo(root: string): void {
   git(['init'], root)
+  // Neutralize any global/system gitignore so engram files (engrams.yaml, episodes.yaml)
+  // are always stageable regardless of the user's personal excludes file.
+  git(['config', '--local', 'core.excludesFile', '/dev/null'], root)
   atomicWrite(join(root, '.gitignore'), GITIGNORE)
   git(['add', '-A'], root)
   git(['commit', '-m', 'Initial PLUR engram store'], root)
--- a/packages/core/test/sync.test.ts
+++ b/packages/core/test/sync.test.ts
@@ -1,4 +1,4 @@
-import { describe, it, expect, beforeEach, afterEach } from 'vitest'
+import { describe, it, expect, beforeEach, afterEach, beforeAll, afterAll } from 'vitest'
 import { mkdtempSync, rmSync, writeFileSync, readFileSync, existsSync } from 'fs'
 import { execSync } from 'child_process'
 import { join } from 'path'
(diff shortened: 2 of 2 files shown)


plur9 (commented):
CTO review: Fix is correct and minimal.

- **sync.ts**: `core.excludesFile=/dev/null` after `git init` permanently insulates the store repo — engram files are always stageable regardless of user's global gitignore. Correct placement (before `git add -A` in initRepo).
- **sync.test.ts**: `beforeAll`/`afterAll` properly save/restore `GIT_CONFIG_GLOBAL` + `GIT_CONFIG_SYSTEM`, handles both set and unset cases, cleans up temp dir. The throwaway gitconfig with name/email avoids git identity errors.

CI green on Node 20 + 22. 29 lines, closes the P1 data-loss path. Recommend merge.

plur9 (commented):

**VERDICT: READY TO MERGE** ✓

Fixes a latent data-loss bug: if user's global gitignore contains `*.yaml`, engrams silently drop from `git add -A` during sync. The single-line fix (`core.excludesFile = /dev/null` in local repo config) is the right scope — isolates the engram store without touching user's global config.

Test isolation (`GIT_CONFIG_GLOBAL` + `GIT_CONFIG_SYSTEM` redirect in `beforeAll`) makes the test self-protecting against the same bug it tests, with correct env var restore.

Issue #21, opened 2026-04-16
Hook-inject produces no output — session_start never triggered


crtahlin:
## Problem

Despite hooks being correctly configured (`plur doctor` reports `overall: ok`, `hooksInstalled: true`), Claude consistently fails to call `plur_session_start` at the beginning of sessions. This has been flagged by the user multiple times.

## Root Cause Analysis

The `UserPromptSubmit` hook fires `npx @plur-ai/cli hook-inject` on every user message. However, **testing shows it produces no output**:

```bash
$ npx @plur-ai/cli hook-inject 2>&1
(no output)

$ echo "do the daily analysis" | npx @plur-ai/cli hook-inject 2>&1
(no output)
```

If hook-inject returns empty output, there is nothing injected into the conversation context to remind the model to call `plur_session_start`. The engram that says "always call session_start first" can never surface because it requires `session_start` to be called first — a chicken-and-egg problem.

## Environment

- `plur` CLI: v0.8.2
- `@plur-ai/mcp`: v0.8.3 (via npx @latest)
- Platform: macOS Darwin 24.6.0
- Claude Code model: claude-opus-4-6
- Storage: `~/.plur/` (default) + project-scoped `~/.plur-war-analysis/`

## Hook Configuration (from global settings.json)

```json
"UserPromptSubmit": [
  {
    "hooks": [
      {
        "type": "command",
        "command": "npx @plur-ai/cli hook-inject",
        "timeout": 15
      }
    ]
  }
]
```

`plur doctor` output:
```json
{
  "hooksInstalled": true,
  "mcpRegistered": true,
  "datacoreCollision": false,
  "handshake": {"ok": true, "serverName": "plur-mcp", "serverVersion": "0.8.3"},
  "overall": "ok"
}
```

## Expected Behavior

`hook-inject` on `UserPromptSubmit` should output a system-reminder-style message that:
1. Injects relevant engrams based on the user's message context
2. Reminds the model to call `plur_session_start` if no session is active
3. Is prominent enough that the model acts on it before launching into work

## Suggestions

1. **hook-inject should always output something on first message** — even if it's just "PLUR: No active session. Call `plur_session_start` before proceeding." This would break the chicken-and-egg loop.

2. **Consider auto-starting sessions via hooks** — if `hook-inject` detects no active session, it could either (a) auto-call `session_start` via the MCP server and inject the results, or (b) output a blocking instruction that the model cannot ignore.

[...]

PR #364, opened 2026-06-19
fix(core): detector FP fixes (INTERNAL_HOST named-capture two-pass) + basic_auth categorization + length cap — PR-2 (#353 audit)

plur9:
## What & why

PR-2 of the 0.10.0 evaluator-hardened fix plan (#353 audit). **Isolated to `secrets.ts` + its tests** (does not touch `index.ts`), so it branches off `origin/main` independently of PR-1.

`INTERNAL_HOST` was silently demoting legitimate, shareable content on every shared-scope write: `config.local`, `data-staging.csv`, `staging-build.yml`, `vite.config.local`, `app.corp`, `dataset.internal`, `app.config.yml`, `tsconfig.json` all matched as "infra". **D3 (severity recalibration): the fabled "17s ReDoS" was empirically measured at ~2.6ms** under V8 Irregexp (mandatory dot separators keep each alternative linear) and **downgraded to MEDIUM**. So this PR is the **false-positive correctness blocker**, not a DoS fix; the length cap is kept as cheap defense-in-depth.

## The FP fix — INTERNAL_HOST two-pass with the named-capture guarantee

**PASS 1** — one non-global regex, **four alternatives**, invoked via `INTERNAL_HOST.exec(text)` (named groups are `undefined` on a non-global `.match()`):
- (a) internal-suffix label `.local/.internal/.corp/.lan/.intranet`
- (b) k8s `svc` / `svc.cluster.local`
- (c) `staging` as a label fragment
- (d) `staging` as an inner label

**Named-capture guarantee:** every alternative wraps **only the host token** in a `(?<host>…)` group, with the leading delimiter in a non-capturing `(?:^|[\s/@'"(])` prefix **outside** the group. PASS 2 runs on `match.groups.host`. **Each of the four branches has a dedicated positive test asserting `match.groups.host` equals the host token** (not the full match) — so a dropped/renamed group cannot silently turn PASS 2 into a no-op.

**PASS 2** — the SOLE FP gate, exactly two rules:
- **RULE 1** known config/data file-extension tail → kills `data-staging.csv`, `staging-build.yml`, `app.config.yml`, `tsconfig.json`.
- **RULE 2** curated config-tool stem before an internal suffix → kills `config.local`, `vite.config.local`, `jest.config.internal`. Documented as a deliberately-incomplete allowlist.

Real internal-host shapes (`db.internal`, `app.corp`, `dataset.internal`, `foo.svc.cluster.local`) **stay flagged** by design.

## Length cap (defense-in-depth, byte-aware)

`detectSensitive` scans at most the first **64KB** (Buffer-truncated, replace-mode decode → always valid UTF-8). The cap lives **inside `detectSensitive`**, so all three call sites — `detectSensitive` itself, the publish loop, and `_guardSensitiveScope`'s `statement + JSON.stringify(context)` scanText — inherit the bound with **zero edits outside `secrets.ts`**.

## basic_auth_url

- **#19 categorization:** moved out of `infra` → **`secrets`**, so a custom `forbid:['secrets']` policy catches a password-in-URL. (Filter `basic_auth_url` out of `INFRA_PATTERN_NAMES`.)
- **#16 false negatives:** now also matches **scheme-less** `user:pass@host:5432/db` and **empty-username** `https://:token@host` forms.
- The scheme-optional change re-introduced O(n²) backtracking (~5.8s on a truncated 64KB `:`-free run), so each unbounded `*`/`+` became bounded `{0,N}`/`{1,N}` — **character classes unchanged from the plan**, ~0.3ms on 64KB.

## The #21 test

A secrets-category credential hidden **only in a context field** (`source`) reaching a **shared** scope under default config (`allow_secrets:false`) demotes to `local`/`private` with `_demoted.patterns` naming `basic_auth_url`. `_guardSensitiveScope` already scans `statement + JSON.stringify(context)` — the test exercises it.

## Tests / build (after `pnpm --filter @plur-ai/core build`)

- **+39 tests** in `secrets.test.ts`; updated the one `scope-metadata` assertion that flipped due to the intentional basic_auth recategorization.
- `secrets` + `scope-metadata` + `publish`: 110 pass. core+mcp+claw green.
- Full workspace: **1568 passed**, 1 failed — that single failure (`cli/test/learn.test.ts` "creates an engram and returns JSON", `local` vs `global`) is **pre-existing on clean origin/main** and is **PR-1's** to fix (out of PR-2 scope).
- `tsc --noEmit` clean on core (zero new vs baseline).

## Spec ambiguity resolved

The plan's literal RULE-2 regex includes a bare `config` stem (to kill `config.local`) which, via the `(?:^|\.)` anchor, also kills any `X.config.local` — so the plan's stated accepted-FP `myapp.config.local` is in fact suppressed by the literal regex (internal inconsistency). I kept the **authoritative literal regex** and demonstrate the same documented boundary with a config-shaped host that genuinely escapes both rules (`myapp.settings.local` → still flagged). Noted in a test comment.

### DIFF
--- a/packages/core/src/secrets.ts
+++ b/packages/core/src/secrets.ts
@@ -77,8 +77,28 @@ export function detectPromptInjection(text: string): InjectionMatch[] {
 // the public engram set in the publish filter. Patterns are deliberately
 // low-false-positive (an IP or a host:port in a *public* engram is the red flag).
 const SENSITIVE_PATTERNS: { name: string; regex: RegExp }[] = [
-  // user:pass@host inside a URL — e.g. https://team:secret@hub-staging.plur.ai
-  { name: 'basic_auth_url', regex: /[a-z][a-z0-9+.-]*:\/\/[^/\s:@]+:[^/\s@]+@/i },
+  // user:pass@host credential URL. Catches the full form
+  // (https://team:secret@hub-staging.plur.ai), the scheme-less form
+  // (user:pass@host:5432/db), and the empty-username form (https://:token@host).
+  // The scheme is optional; the username may be empty; a password and an `@host`
+  // are required so a bare `key:value` or a `time:12:30` (no `@host`) is NOT a
+  // match. (#353 LOW-16.)
+  //
+  // SPEC-DEVIATION (bounded quantifiers): the plan's literal form used unbounded
+  // `*`/`+` (`(?:[a-z][a-z0-9+.-]*:\/\/)?[^/\s:@]*:[^/\s@]+@`). Making the scheme
+  // optional re-introduced O(n^2) backtracking — on 64KB of a `:`-free run the
+  // userinfo class scans for a `:` that never comes at every start position
+  // (~5.8s on a truncated 64KB input; the old `://`-anchored form was linear).
+  // D3 keeps the 64KB length cap as defense-in-depth but 64KB is still far too
+  // slow, so we bound each unbounded quantifier (the CHARACTER CLASSES are
+  // unchanged from the plan — only `*`/`+` became bounded `{0,N}`/`{1,N}`). Real
+  // basic-auth userinfo and passwords are short, so the bounds preserve every
+  // functional case (full / scheme-less / empty-username) while restoring linear
+  // time (~0.3ms on 64KB instead of ~5.8s).
+  {
+    name: 'basic_auth_url',
+    regex: /(?:[a-z][a-z0-9+.-]{0,31}:\/\/)?[^/\s:@]{0,64}:[^/\s@]{1,128}@[a-z0-9.-]/i,
+  },
   // FQDN:port — e.g. hub-staging.plur.ai:443, db.internal:5432 (infra topology)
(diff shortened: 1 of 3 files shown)


plur9 (commented):

**VERDICT: READY TO MERGE** ✓

The two-pass INTERNAL_HOST rewrite is the right fix: the single-regex form silently demoted real shareable content (`config.local`, `data-staging.csv`, `staging-build.yml`, etc.).

**Design sound:**
- PASS 1 (named `(?<host>...)` group) → PASS 2 (sole FP gate, 2 rules) — clean separation
- Named-group invariant enforced: no group → fail safe (no throw) ✓
- `CONFIG_STEM_BEFORE_INTERNAL_SUFFIX` documented as known-incomplete allowlist with explicit boundary case — honest design
- `basic_auth_url` correctly recategorized to `secrets` (credential-in-URL, not infra topology) ✓
- 64KB cap with UTF-8-safe `subarray().toString('utf8')` ✓
- bounded quantifiers in `basic_auth_url` restore linear time from original O(n²) path

Issue #362, opened 2026-06-19
Stage 3b v2: re-implement scoped unscoped-default with correct read-visibility (scope-family-aware read filter as a precondition)


plur9:
## Context

PR-1 (#353 audit) reverted the un-scoped WRITE default `local → global` and — critically — fixed the READ side so no personal-family scope is excluded by a project-scoped filter. The read-side fix is now in place on **all three read paths** (inject `scoreEngram`, the non-indexed recall filter, AND the default indexed SQLite path `storage-indexed.ts` via a `personal` column), using `isPersonalScope(scope) = !isSharedScope(scope)` rather than a hardcoded `{local,global}` set.

Reverting the write default to `global` restores the cross-project bleed that Stage 3b (#351) mitigated. This is accepted for 0.10.0 because (a) the read-side invisibility was a correctness blocker, and (b) `.plur.yaml` project-scope config is the supported mitigation.

**This issue tracks Stage 3b v2: WRITE-default behavior ONLY.** The READ side is already fixed in PR-1 (family-aware filter on all three paths including indexed) and is NOT part of v2 scope.

## v2 checklist

- [ ] **(i)** Expose `WEIGHT_TAG` / `SCOPE_MATCH_THRESHOLD` as config for multi-scope enterprise installs (covers-rich escalation risk).
- [ ] **(ii)** Cross-scope recurrence (`index.ts` `_recordCrossScopeRecurrence`, `if (e.scope !== 'global') e.scope = 'global'`) hardcodes promotion to global regardless of `unscoped_default` — make it respect `unscoped_default` / a personal-scope ceiling.
- [ ] **(iii)** One-time `plur_session_start` migration notice for users with `scope=local` engrams written during the 0.9.x Stage 3b window.
- [ ] **(iv)** Revisit the secondary-store scope gate (`index.ts` `_loadAllEngrams`, `storage-indexed.ts` `syncFromYaml`): a `local`/`user:*` engram placed in a SECONDARY store is dropped before any filter sees it. Intentional for 0.10.0 (secondary stores hold only their declared scope or global); revisit if personal-scoped engrams ever land in secondary stores.

## Precondition for v2

The scope-family-aware read filter (PR-1) is the precondition — any v2 write-default change must preserve read visibility of personal-family scopes under a project-scoped recall/inject.

plur9:
## Progress update — 2026-06-28

[...]

PR #363, opened 2026-06-19
fix(core): revert unscoped default→global + read-side personal-scope visibility on all 3 paths — PR-1 (#353 audit)

plur9:
PR-1 of the 0.10.0 evaluator-hardened fix plan. Reverts the un-scoped WRITE default `local → global` AND fixes every read-side scope filter so no personal-family scope is excluded by a project-scoped recall/inject. Resolves audit findings 3, 4, 5, 7, 8.

## Module-cycle break (mandatory precondition, done first)

`inject.ts` needs `isPersonalScope`, but `index.ts` imports from `inject.ts` — importing the predicate from `index.ts` would form an `index → inject → index` cycle. Resolution: new leaf module **`packages/core/src/scope-util.ts`** holds `SHARED_SCOPE_PREFIXES`, `isSharedScope`, and new `isPersonalScope(scope) = !isSharedScope(scope)`. `index.ts` re-exports `isSharedScope`/`isPersonalScope`/`SHARED_SCOPE_PREFIXES` (public API unchanged). Re-pointed imports: `inject.ts`, `index.ts`, `storage-indexed.ts` import from `scope-util.ts`; `mcp/tools.ts` imports `isSharedScope` from `@plur-ai/core`. `scope-util.ts` is a dependency-free leaf; `inject.ts` no longer imports `index.ts`. **CI gate:** `tsc --noEmit` on core added to the lint step (madge isn't a repo dep) — must be green before any filter edit.

## Read side — all three paths (`isPersonalScope`, not a hardcoded {local,global})

1. **`inject.ts` `scoreEngram`** — project-branch passes all personal-family scopes; the `scopeFilter === 'global'` branch is **kept targeted** (returns global only), encoded by the named constant `INJECT_GLOBAL_IS_TARGETED = true` with JSDoc (the documented RECALL/INJECT asymmetry).
2. **`index.ts` non-indexed recall filter** — `e.scope === scope || e.scope.startsWith(scope) || isPersonalScope(e.scope)` (keeps the `startsWith` arm so `user:alice` still catches `user:alice:notes`).
3. **`storage-indexed.ts` — the DEFAULT indexed SQLite path** (`config.index` defaults true → production path). New `personal INTEGER` column set at index time from `isPersonalScope(scope)` in both upsert sites; `loadFiltered` SQL changed from `scope = 'global'` to `personal = 1` (params unchanged). Pre-0.10.0 DBs migrate via guarded `ALTER TABLE ADD COLUMN personal` + a one-time reindex backfill.

[...]

PR #366, opened 2026-06-20
fix(core): scope-aware LLM dedup + personal-scope ceiling in cross-scope recurrence

### LINKED ISSUE #359: Dedup silently demotes an explicitly-scoped write to global (LLM-dedup candidates not scope-filtered)
## Summary

An explicitly scoped `plur_learn` write can be silently demoted to `global` when its content is semantically close to an existing `global` engram. The write is deduped into the global engram (NOOP/UPDATE/MERGE) and keeps the global scope — the requested team scope (e.g. `group:datafund/datafund`) is dropped, no error is raised, and the target team store never receives the engram.

## Versions

- **Reproduced on:** published `@plur-ai/mcp@latest` = **0.9.13** (npx).
- **Still present in:** `main` @ **`259c483`** (local source build — also reports version string `0.9.13`, version field not bumped). The code path below is unchanged at HEAD, so this is not fixed in the latest source.

## Reproduction

1. A `global` engram exists with statement S.
2. Call `plur_learn` with `scope: group:datafund/datafund` and a statement semantically near S.
3. Response returns `scope: "global"`, `decision: NOOP`/`MERGE`/`UPDATE`.
4. `plur_stores_list` shows the `group:datafund/datafund` count unchanged — the engram did not reach the remote team store.

Observed live: a Datafund positioning engram only incremented the remote store count (12 → 13) after the statement was reworded enough to fall outside the dedup match. Near-duplicate phrasings repeatedly collapsed into the existing global engram.

## Root cause

`packages/core/src/learn-async.ts`:

- Hash dedup is scope-aware — `:16` *"Content hash dedup against all engrams. Scope-aware: only matches same scope."*, called at `:176` as `deps.hashDedup(statement, context?.scope)`.
- LLM-driven dedup is **not** scope-aware. Candidates are gathered at `:188-199` via `recallHybrid`/`recall` (limit 5) and filtered only by `status === 'active'` — there is no scope filter. A `global` engram is therefore a valid dedup candidate for a team-scoped write.
- The NOOP (`:86-87`), UPDATE (`:94-118`) and MERGE (`:127-153`) branches return / mutate the **existing** engram and never reconcile its scope with the requested `context.scope`. So a match against a global engram keeps `scope: global`.

## Impact

Silent data-routing failure for team memory: a developer who correctly passes a team scope can have the engram vanish into their local/global store with a success response. Team stores under-populate with no signal. This is the same failure class the loud-failure UX work targets (#340/#341).

## Suggested fix (one or more)

1. Scope-filter LLM-dedup candidates to the requested `context.scope` (mirror `hashDedup`'s scope-awareness), so cross-scope matches don't occur.
2. Or, when a dedup match is in a different (broader) scope than requested, treat it as `ADD` at the requested scope rather than NOOP/MERGE — never silently demote an explicit scope.
3. Surface the scope outcome in the `plur_learn` response and/or warn when the resolved scope differs from the requested one (loud failure).

## Related

- #351 (Stage 3 routing — auto-route un-scoped writes) — adjacent but about *un-scoped* writes; this issue is an *explicitly-scoped* write being demoted.
- #345 (Stage 2 self-describing scope metadata).
- #213 (audit: engram conflict resolution across detection/retrieval/lifecycle).

plur9:
## Summary

Two write-correctness bugs that silently drop engrams or misplace them in the wrong scope.

### Bug 1 — LLM dedup absorbs scoped writes into global (#359)

**Root cause:** `learnAsync` passes all recall candidates to the LLM dedup stage without scope-filtering. A global engram with a similar statement could return `NOOP`/`UPDATE`/`MERGE`, silently dropping the requested scope — the team-scoped write never reaches the team store.

**Fix:** Filter LLM dedup candidates to `context.scope` before the LLM sees them, mirroring the scope-awareness already present in `hashDedup`. When no scope is requested, candidates are unfiltered (backward-compatible).

### Bug 2 — Personal-scope engrams promoted to global on cross-scope recurrence (#362 item ii)

**Root cause:** The cross-scope recurrence promotion (`recurrence_count >= 2 → scope = 'global'`) fired for ALL scopes, including personal-family scopes (`local`, `user:*`). A personal rule recurring across projects would escape to `global` and be visible everywhere.

**Fix:** Only shared scopes (`isSharedScope` — `project:*`, `space:*`, `group:*`, `org:*`, `public`) get promoted to `global`. Personal-family scopes stay within their family.

### Bonus — neutraliseGlobalExcludes in sync() (#361 follow-up)

Extracts the gitignore-neutralisation into a reusable function and calls it in `sync()` as well as `initRepo()`. Existing repos created before the #361 fix now get the neutralisation applied on their next sync (idempotent).

## Tests

- `learn-async-scope.test.ts` (new, 109 lines) — 3 cases:
  - Cross-scope write NOT absorbed by global engram (LLM not called, ADD at requested scope)
  - Same-scope dedup still works (LLM called, NOOP)
  - No scope = no filter (backward-compatible)
- `cross-scope-recurrence.test.ts` +17 lines — personal-scope (local) is NOT promoted to global

## Issues closed

Closes #359. Partial #362 (item ii only — item i is tracked separately in the Stage 3b v2 issue).

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -662,7 +662,10 @@ export class Plur {
       ;(e as any).sources = [...((e as any).sources ?? []), source]
 
       if (newRecurrence >= 2) {
-        if (e.scope !== 'global') e.scope = 'global'
+        // Only promote SHARED scopes (project:*, space:*, etc.) to global —
+        // personal-family scopes (local, user:*) stay within their family.
+        // See issue #362 item (ii): personal-scope ceiling for cross-scope recurrence.
+        if (isSharedScope(e.scope)) e.scope = 'global'
         if (e.commitment !== 'locked') {
           // Forward-only ladder: exploring → leaning → decided → locked.
           e.commitment = e.commitment === 'exploring'
--- a/packages/core/src/learn-async.ts
+++ b/packages/core/src/learn-async.ts
@@ -197,6 +197,12 @@ export async function learnAsync(
     candidates = deps.recall(statement, { limit: 5 })
   }
   candidates = candidates.filter(c => c.status === 'active')
+  // Mirror hashDedup scope-awareness (issue #359): only dedup against same-scope engrams.
+  // Without this, a global engram silently absorbs an explicitly-scoped write — the requested
+  // scope is dropped and the team store never receives the engram.
+  if (context?.scope) {
+    candidates = candidates.filter(c => c.scope === context.scope)
+  }
 
   if (candidates.length === 0) {
     return { engram: deps.learn(statement, context), decision: 'ADD' }
--- a/packages/core/src/sync.ts
+++ b/packages/core/src/sync.ts
@@ -86,11 +86,14 @@ export function getSyncStatus(root: string): SyncStatus {
   }
 }
 
+function neutraliseGlobalExcludes(root: string): void {
+  // Prevent a developer's global gitignore from silently dropping engram files
+  gitSafe(['config', 'core.excludesFile', '/dev/null'], root)
+}
+
 function initRepo(root: string): void {
   git(['init'], root)
-  // Neutralize any global/system gitignore so engram files (engrams.yaml, episodes.yaml)
-  // are always stageable regardless of the user's personal excludes file.
-  git(['config', '--local', 'core.excludesFile', '/dev/null'], root)
+  neutraliseGlobalExcludes(root)
   atomicWrite(join(root, '.gitignore'), GITIGNORE)
   git(['add', '-A'], root)
   git(['commit', '-m', 'Initial PLUR engram store'], root)
@@ -154,6 +157,9 @@ export function sync(root: string, remote?: string): SyncResult {
       files_changed: 0,
     }
   }
+
+  // Ensure global excludes are neutralised for this store (idempotent; migrates existing repos)
+  neutraliseGlobalExcludes(root)
 
   // Add remote if provided and not yet set
   const existingRemote = getRemote(root)
--- a/packages/core/test/cross-scope-recurrence.test.ts
+++ b/packages/core/test/cross-scope-recurrence.test.ts
@@ -265,6 +265,23 @@ describe('cross-scope recurrence (#176)', () => {
       const second = plur.learn('always use   semicolons', { scope: 'project:b' })
       expect(second.recurrence_count).toBe(1)
     })
+
+    it('personal-scope engrams (local) are NOT promoted to global on cross-scope recurrence (#362 item ii)', () => {
+      // A local engram recurring across scopes should stay in the personal family.
+      // Only shared scopes (project:*, space:*, etc.) get promoted to global.
+      const first = plur.learn('personal rule', { scope: 'local' })
+      expect(first.scope).toBe('local')
+
+      plur.learn('personal rule', { scope: 'project:a' })  // 1st cross-scope, recurrence=1
+      const third = plur.learn('personal rule', { scope: 'project:b' })  // 2nd cross-scope
+
+      // recurrence_count increments as normal
+      expect(third.recurrence_count).toBe(2)
+      // Commitment escalates (that behavior is unchanged)
+      expect(third.commitment).toBe('decided')
+      // But scope stays local — personal-family ceiling prevents global promotion
+      expect(third.scope).toBe('local')
+    })
   })
 
   describe('persistence + observability', () => {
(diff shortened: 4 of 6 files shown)


plur9:
## CTO review — CI failures are test bugs, not logic bugs

Both failures are test issues only. The production fix in `learn-async.ts` and `index.ts` is correct.

---

### Failure 1 — `learn-async-scope.test.ts` tests 2 + 3: LLM mock format wrong

Tests 2 and 3 mock `vi.fn().mockResolvedValue('NOOP')`, but `parseDedupResponse` uses regex and expects:

```
DECISION: NOOP
TARGET: <id>
REASON: <reason>
```

A bare `'NOOP'` string doesn't match the `DECISION:` prefix regex, so `decision` defaults to `'ADD'`. Then `executeDedupDecision` NOOP path falls through to ADD when `targetId` is null.

**Fix — test 2** ("still deduplicates same-scope candidates"):
```ts
const llm = vi.fn().mockResolvedValue('DECISION: NOOP\nTARGET: ENG-2026-0101-002\nREASON: same-scope duplicate')
```

**Fix — test 3** ("does not filter candidates when no scope is requested"):
```ts
const llm = vi.fn().mockResolvedValue('DECISION: NOOP\nTARGET: ENG-2026-0101-001\nREASON: duplicate')
```

---

### Failure 2 — `pr1-write-interactions.test.ts:62`: asserts the old behavior this PR removes

The test **"a local engram promoted to global by _recordCrossScopeRecurrence on the 2nd cross-scope hit"** asserts `expect(promoted.scope).toBe('global')` — that's exactly the behavior the `isSharedScope` guard removes.

**Fix:**
```ts
it('a local engram stays local (personal-scope ceiling) on the 2nd cross-scope hit', () => {
  // ...same setup...
  expect(promoted.scope).toBe('local')  // was 'global'
})
```

---

Production logic (`index.ts`, `learn-async.ts`, `sync.ts`) is correct. Once these two test fixes land and CI goes green, this is ready to merge.

plur9:
**CI fixed — 3 test bugs addressed in the latest push.**

Root causes found:

**`learn-async-scope.test.ts` (tests 2 + 3 → ADD instead of NOOP)**
`parseDedupResponse` expects structured `DECISION: NOOP\nTARGET: <id>` format. The mocks were returning a raw `'NOOP'` string, which failed the regex → fell through to `decision: 'ADD'`. Fixed the mock return values in both tests. Also added the missing `offendingHitsForScope: () => []` stub to `makeDeps` (required by `LearnAsyncDeps` interface).

**`pr1-write-interactions.test.ts` (expected 'global', got 'local')**
This test was added to main in the PR-1 audit (#363) to *document the then-unresolved bug* (local engrams incorrectly promoted to global). PR #366 fixes that bug (personal-scope ceiling). CI runs a merge commit so main's test conflicted with the fix. Updated the test to assert the correct post-fix behavior: local engrams stay local after cross-scope recurrence. Descriptions updated to reflect resolved status.

Branch has been rebased onto current main (PR-1/PR-2/PR-3 included). CI re-running now.

plur9:
**Ops review complete — ready to merge.**

All issues identified in the previous review have been addressed:
- LLM mock format corrected (`DECISION: NOOP\nTARGET: <id>\nREASON:`) — tests 2+3 now pass
- `offendingHitsForScope: () => []` stub added to `makeDeps`
- `pr1-write-interactions`: assertion updated from `'global'` → `'local'` (documents resolved behavior)

Production logic in `learn-async.ts`, `index.ts`, `sync.ts` is correct (unchanged from initial push).

CI ✅ green on both Node 20 and 22. Closes #359, partial #362 (item ii). Merge when ready.

crtahlin (approved):
Approve. Both write-correctness fixes are correct and well-scoped.

- **#359 (scope-aware LLM dedup):** the candidate filter mirrors `_hashDedup`'s `scope === undefined || e.scope === scope` semantics exactly — verified against `index.ts`. Backward-compatible (no scope ⇒ unfiltered).
- **#362 item ii (personal-scope ceiling):** `isSharedScope` is the right gate; `local`/`user:*`/`agent:*` correctly stay within family while the commitment ladder still escalates.

Verified locally: the 3 affected test files (22/22) and the full `packages/core` suite (1141 pass, 28 skipped, 0 fail). CI green on Node 20+22.

**Non-blocking note:** the scope filter in `learn-async.ts` is applied *after* `recallHybrid(limit: 5)` truncates candidates. If the top-5 semantic neighbours are dominated by other-scope (e.g. global) engrams, a genuine same-scope duplicate ranked #6+ won't be seen, yielding a duplicate ADD at the correct scope. This is strictly better than the prior silent scope-drop and `limit: 5` predates this PR — but if same-scope dedup recall matters, consider filtering before limiting (or over-fetch then filter). Not blocking.

PR #370, opened 2026-06-20
fix(core): leak guard covers remote-backed scopes (user:) — re-applied on post-audit main (supersedes #360)

plur9:
## The gap

The write-time leak guard only fired for `isSharedScope`-prefixed scopes (`group:`/`project:`/`space:`/`team:`/`org:`/`public`) — a pure PREFIX check, now living in `packages/core/src/scope-util.ts` after the PR-1..6 refactor. But a scope can be **personal by prefix** (`user:plur:gregor`) yet still be backed by a **REMOTE store** — a `stores` config entry with a `url:` whose data goes to `plur.datafund.io`. Such a write is NOT `isSharedScope`, so sensitive content (public IPs, internal hosts, secrets) written to it **bypassed the guard and reached the remote server unguarded**. The real risk is "does the data leave the machine", not "is the prefix shared".

## The predicate

A private, PURE config lookup (no driver instantiation, no network, no side effects — it runs on every `learn()`):

```ts
private _isRemoteBackedScope(scope: string): boolean {
  return (this.config.stores ?? []).some(s => !!s.url && s.scope === scope)
}
```

It mirrors `_resolveRemoteStoreForScope`'s exact-scope-match rule (no prefix matching) so the guard and the router agree on which writes reach the remote.

## All gates updated

There are exactly **TWO** early-return gates that short-circuit the demotion guard on `!isSharedScope(scope)` (verified by grepping `!isSharedScope` across `index.ts` on current `main` — these are the only two). Both now gate on shared **OR** remote-backed:

- **`_offendingHitsForScope`** (`packages/core/src/index.ts:946`) — single source of truth for "does this content carry sensitivity this scope forbids", reached by every mutation path.
- **`_guardSensitiveScope`** own gate (`packages/core/src/index.ts:1142`) — the learn/learnRouted guard; short-circuits before `_offendingHitsForScope`, so it must carry the same gate or the leak stays open on the learn path.

Both became:
```ts
if (!isSharedScope(scope) && !this._isRemoteBackedScope(scope)) ...
```

## What's unchanged

- `global` and `local` (no store backing → neither shared nor remote-backed) stay **UNguarded** — infra notes legitimately live there and never leave the machine; `global` does NOT start demoting.
- A local-file `project:` store (a `path` store, no `url`) stays guarded via its `isSharedScope` prefix — existing behavior, untouched.
- Only **NEW** coverage = remote-backed non-shared scopes (`user:`).

## Tests

New `packages/core/test/guard-remote-scope.test.ts` (mirrors the original #360 design, harness mirrors `guard-remote-boundary.test.ts` — a `vi.fn()` over `globalThis.fetch` asserting on POST appends):

- **(a)** sensitive write (public IP) to a remote-backed `user:` scope → demoted to `local`, remote append spy = 0, outbox = 0, kept locally.
- **(b)** clean write to the same remote `user:` scope → not demoted, reaches the remote (1 append) — proves no over-blocking.
- **(c)** `global` (no store) + sensitive → NOT demoted (unchanged).
- **(d)** local-file shared-prefix store → still demotes on sensitive (unchanged `isSharedScope` path).
- **(e)** per-scope `sensitivity.allow:['infra']` on the remote scope honored → not demoted, reaches remote.

## Verification

- `pnpm --filter @plur-ai/core build` ✅
- Full workspace after dist rebuild: **1649 passed, 34 skipped (148 files)** ✅
- `tsc --noEmit` on core: clean (zero new vs baseline) ✅

Supersedes #360 (stale after the PR-1..6 `index.ts` refactor that moved `isSharedScope` to `scope-util.ts`). No code logic differs from #360 — same predicate, same two gates, same tests — only the gate line numbers moved due to the refactor.

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -499,6 +499,22 @@ export class Plur {
     return null
   }
 
+  /**
+   * True when `scope` is backed by a REMOTE store — i.e. a `stores` entry with a
+   * `url` (data leaves this machine) whose scope exactly matches. The leak guard
+   * uses this alongside `isSharedScope`: a scope like `user:plur:gregor` is NOT
+   * `isSharedScope` (personal prefix) yet routes to plur.datafund.io, so sensitive
+   * content written there would cross the machine boundary unguarded.
+   *
+   * Pure CONFIG lookup — NO driver instantiation, NO network, NO side effects —
+   * because this runs on every learn(). It mirrors `_resolveRemoteStoreForScope`'s
+   * exact-scope-match rule (no prefix matching) so the guard and the router agree
+   * on which writes reach the remote.
+   */
+  private _isRemoteBackedScope(scope: string): boolean {
+    return (this.config.stores ?? []).some(s => !!s.url && s.scope === scope)
+  }
+
   /** Find which store owns an engram by ID. For namespaced IDs, strips prefix to find in store. */
   private _findEngramStore(id: string): { path: string; readonly: boolean; originalId: string } | null {
     // Check primary first (uses mtime cache)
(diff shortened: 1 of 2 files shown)


plur9:
## Release signal — v0.10.0 unblocked

All four security PRs from the #353 audit are now merged:
- #363 ✓ #364 ✓ #365 ✓ #370 ✓ (this PR, superseded #360)

**Ready to tag v0.10.0** — bump core/mcp 0.9.13→0.10.0 and claw 0.9.17→0.10.0, then push the tag.

**One decision before tagging:** PR #366 (scope-aware LLM dedup + personal-scope ceiling) is CI-green and CTO-approved. It closes two write-correctness bugs (#359, #362 item ii). Options:
- Merge #366 now → tag v0.10.0 with it included
- Tag v0.10.0 today → ship #366 as v0.10.1 patch

Both are clean. Recommend including #366 since it's ready and the bugs it fixes are meaningful, but gregor decides.

PR #379, opened 2026-06-22
docs: engram store consolidation/merge runbook

### LINKED ISSUE #378: Document (and consider a command for) engram store consolidation/merge
## Problem

Consolidating a legacy per-project store into the global store surfaced a gap: there is **no documented path and no first-class command for merging one engram store into another**. This is a rare operation (it happens when an install's storage architecture changes — e.g. a project moving from a dedicated `PLUR_PATH` store to the global `~/.plur` store with scope-based separation), but when it does come up there is no guidance, and the naive approaches are wrong.

## Why the existing tools don't cover it

- `plur learn` — generates embeddings and dedups natively, but is lossy on metadata (feedback signals, original dates) and impractical for bulk (tens of records).
- `plur ingest` — re-extracts via pattern matching, which mangles already-clean engram statements.
- `plur stores add` — *registers/attaches* a store for reading; it does not consolidate one into another.
- `plur migrate` — schema versioning (up/down/status) only; unrelated to store contents.

## The manual path that works (but has sharp edges)

1. Back up the destination `engrams.yaml`.
2. Content-dedup check: compare source statements against destination (ID matching is unsafe — IDs collide across stores).
3. Append source records into destination with **fresh non-colliding IDs** and **corrected scope**.
4. `plur sync --full` to rebuild the derived index and regenerate embeddings.
5. Verify with a recall query.
6. Re-purge tensions (a full reindex recomputes contradiction relations store-wide).
7. Archive the source store.

### Gotchas a first-timer will hit blind
- **ID collisions** — source IDs frequently already exist in the destination as *different* engrams.
- **Embeddings** only regenerate on a full reindex (`sync --full`), not on a raw YAML append.
- **Tensions** recompute across the whole store on full reindex and may need re-purging.
- **Scope** must be rewritten on import or records keep their old (often `global`) scope.

## Proposed

1. **Now:** add a runbook doc capturing the procedure + gotchas. (This issue's PR.)
2. **Backlog (only if it recurs):** a first-class `plur stores merge <src> <dst> --scope <s> --dedup` that reuses the existing similarity-dedup, mints non-colliding IDs, rewrites scope, and triggers reindex. Store consolidation is roughly a once-per-architecture-change event, so this is low priority versus the doc.

crtahlin:
Adds `docs/runbooks/store-consolidation.md` — a step-by-step procedure for folding one engram store into another (e.g. retiring a per-project `PLUR_PATH` store into the global `~/.plur` store with scope-based separation).

Covers:
- Why a naive YAML copy fails (ID collisions, stale scope, missing embeddings, store-wide tension recompute).
- Content-based dedup (IDs are not safe to compare across stores).
- Bulk faithful append vs. per-record `plur learn`, with re-ID + scope rewrite.
- `plur sync --full` to regenerate embeddings, recall verification, tension re-purge, source archival, and rollback.

Motivated by a real consolidation. A first-class `plur stores merge` command is left as future work (low priority — this is roughly a once-per-architecture-change operation).

Closes #378.

### DIFF
--- a/docs/runbooks/store-consolidation.md
+++ b/docs/runbooks/store-consolidation.md
@@ -1,0 +1,175 @@
+# Runbook: Consolidating one engram store into another
+
+**Status:** manual procedure (no first-class command yet — see [#378](https://github.com/plur-ai/plur/issues/378))
+**Audience:** operators merging a legacy or per-project engram store into another store
+**Frequency:** rare — typically a one-time event when an install's storage architecture changes
+
+## When you need this
+
+You have two engram stores and want to fold one into the other, for example:
+
+- A project used to point at a dedicated store via `PLUR_PATH` (e.g. `~/.plur-myproject`) and has since moved to the global `~/.plur` store with **scope-based** separation. The old store is now orphaned but still holds engrams that were never copied over.
+- You are decommissioning a store and want to preserve its still-relevant engrams in a surviving store.
+
+This is **not** the same as:
+
+- `plur stores add` — registers/attaches a store so it can be *read*; it does not copy contents.
+- `plur migrate` — applies schema-version migrations within a store; unrelated to contents.
+
+## Why you can't just copy the YAML
+
+Two stores are independent. Naively concatenating one `engrams.yaml` into another will break in these ways:
+
+| Hazard | What happens | Mitigation |
+|---|---|---|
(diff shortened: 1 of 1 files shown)


plur9 (approved):
Reviewed: accurate, well-structured runbook. Content-dedup-over-ID, `sync --full`, tension re-purge, and the archive-not-delete retirement all match real practice. Correctly scopes the future `plur stores merge` command to #378. LGTM.

Issue #384, opened 2026-06-22
sync sets core.excludesFile=/dev/null, defeating the user's own protective gitignore


crtahlin:
## Summary
Part of #377. Aggravator of the config.yaml credential-commit issue.

`initRepo()` runs `git config --local core.excludesFile /dev/null` (`packages/core/src/sync.ts:93`) so engram files are always stageable (the #329 data-loss fix). The side effect: it overrides the user's own global gitignore, removing the only place a security-conscious user could have excluded `config.yaml` / credential files.

## Details
Git precedence is local > global > system, so the local override defeats `~/.config/git/ignore`. After neutralization the only exclude list is the 5-line sync GITIGNORE (`sync.ts:21-27`), which does not cover `config.yaml`. `git add -A` then stages it.

## Repro
```sh
# with config.yaml listed in the user's global core.excludesFile
git check-ignore config.yaml   # ignored (protected)
git config --local core.excludesFile /dev/null
git check-ignore config.yaml   # nothing -> protection defeated
```

## Recommendation
Scope the override narrowly. Instead of disabling all global excludes, force-add only the known store files (`git add -f engrams.yaml episodes.yaml candidates.yaml`) or ship an in-tree `.gitignore` that whitelists store files and excludes `config.yaml`. An in-tree `.gitignore` entry is honored regardless of `core.excludesFile`, so this also closes the credential-commit issue.

**Severity:** High · **Dimension:** Confidentiality

plur9:
Fixed in the 0.10.0 hardening pass (PR #387). Confirmed by independent re-audit (#377): `core.excludesFile=/dev/null` is gone from all sync paths; the exclusive allowlist staging approach (force-add only known store files) replaces it. Verified against built dist.

Issue #377, opened 2026-06-22
Independent audit of the 0.10.0 scope-security work before release


plur9:
## What this is

Before we tag and publish 0.10.0, I'd like a second pair of eyes — yours — on the **whole body of scope-security work** that went into this release. We just finished a multi-round hardening pass on how engram memories are scoped, stored, routed, shared, and published. Two earlier rounds were review-only; the third added an automated attack-test suite that found a real problem the manual reviews had missed — which is exactly why I'd rather have an independent audit *before* we ship than after.

This is broader than the "don't leak secrets" guard. "Scope" controls **who can see a memory and where it travels**, so the audit should cover the full surface, not just one slice.

**Authoritative list of everything that shipped:** the `## 0.10.0` section of `CHANGELOG.md` enumerates every change with its PR number. The work spans merged PRs **#289, #291, #292, #295, #296, #306, #353, #361, #364–#376**. Please treat the changelog as the master checklist.

## The properties worth auditing (in plain English)

**1. Write-side — nothing sensitive leaves the machine in a shared memory.**
When a memory headed for a shared or remote store contains a server address, a host:port, a login URL with credentials, or an internal hostname, it's kept local and marked private instead. The audit question: is there *any* write path that skips this? It now runs on new memories, updates, merges, failure reports, background flushes, and the internal summary memories — and on remote-backed personal scopes, not just team scopes. (#364, #368, #370, #371, #374, #375)

**2. Read-side — a personal memory never shows up in a different project's context.**
This is the mirror image of #1 and just as important. A memory scoped personal/local must not surface when you're working in another project (via either recall or the automatic context injection). This was fixed across all three read paths, including the default local database path. The audit question: can a personal-scoped memory bleed into the wrong project's session? (#353)

**3. Routing — an unscoped write lands in the right place, and a broad memory can't be filed into a narrow shared scope.**
When you save without specifying a scope, where does it go, and can a general memory get auto-filed into a specific team's shared store it doesn't belong to? Routing was recalibrated and made deterministic on a clear namespace match. The audit question: any input that auto-routes somewhere it shouldn't? (#367, #369, #373)

**4. Distribution / publication — memories that ride along with a repo match that repo's audience.**
[...]

PR #387, opened 2026-06-22
fix(sync): never commit secrets or machine-local files (#380, #384)

### LINKED ISSUE #380: plur sync commits and pushes config.yaml containing remote-store Bearer tokens
## Summary
Part of #377.

`plur sync` commits and pushes `~/.plur/config.yaml`, which contains live remote-store Bearer tokens, to whatever git remote the user configures — with no audience check, redaction, or warning.

## Details
- The GITIGNORE written by sync (`packages/core/src/sync.ts:21-27`) excludes only derived/cache files (`embeddings/`, `.embeddings-cache.json`, `*.db`, `*.sqlite`, `exchange/`). `config.yaml` is **not** excluded.
- `config.yaml` lives at the sync root (`storage.ts:34` -> `join(root, 'config.yaml')`) and holds `stores[].token` (`schemas/config.ts:19`, "Bearer for remote stores").
- Every sync path runs `git add -A` (`sync.ts:95`, `:101`, `:130`), staging `config.yaml`, then pushes to origin (`sync.ts:147`/`:164`/`:192`).
- `sync.ts:93` sets `core.excludesFile=/dev/null`, removing the only place a user could have excluded `config.yaml` via their own global ignore.
- There is no privacy/audience check on the remote URL.

## Repro
```sh
cd /tmp && mkdir r && cd r && git init -q
git config --local core.excludesFile /dev/null
printf 'embeddings/\n.embeddings-cache.json\n*.db\n*.sqlite\nexchange/\n' > .gitignore
printf 'stores:\n  - url: https://plur.datafund.io\n    token: eyJ...SECRET_JWT\n' > config.yaml
printf 'engrams: []\n' > engrams.yaml
git add -A && git diff --cached --name-only
# -> .gitignore, config.yaml, engrams.yaml  (token-bearing config.yaml staged)
```

## Recommendation
- Add `config.yaml` to the GITIGNORE constant (`sync.ts:21-27`).
- Preferred: keep tokens out of the synced file — split secrets into a gitignored `~/.plur/secrets.yaml` referenced by config, or strip the `token` field before staging.
- Add a `sync.test.ts` assertion that a committed tree does **not** include `config.yaml` and that no committed blob contains a Bearer token.

**Severity:** Blocker · **Dimension:** Confidentiality

crtahlin:
## What & why

Fixes the release blocker #380 (and the paired #384) from the #377 scope-security audit.

`plur sync` could commit and push `~/.plur/config.yaml` — which holds remote-store **Bearer tokens** — to any configured git remote. The sync `.gitignore` (`sync.ts`) excluded only derived/cache files, and `git add -A` staged everything else, including `config.yaml`. It also ran `git config core.excludesFile /dev/null`, defeating any protective ignore a security-conscious user had configured (#384).

## Approach: allowlist, not denylist

Rather than enumerate every secret in `.gitignore` (a denylist that must stay exhaustive), sync now stages an explicit **allowlist** of engram-store files:

- `stageStoreFiles()` force-adds only `engrams.yaml`, `episodes.yaml`, `candidates.yaml`, `packs/`, `.gitignore` via `git add -A -f -- <allowlist>`. Secrets and machine-local derived files (`config.yaml`, `secrets.yaml`, `engrams.db`, `store.pglite/`, `exchange/`) are **never in the staging pathspec**, so they cannot ride along regardless of gitignore state — a structural guarantee.
- The `core.excludesFile=/dev/null` neutralization is **removed**. The `-f` on the allowlist preserves the #329 guarantee (engram files stage even when a user's global excludes would ignore them) *without* stripping the user's protection for everything else — which is exactly what #384 flagged.
- **Self-healing:** sync untracks `config.yaml`/`secrets.yaml` if a vulnerable pre-fix client already committed them, so the next sync stops carrying the secret forward.
- The sync `.gitignore` additionally lists the secret files and `store.pglite/` as defense-in-depth.

## Tests

New `secret safety` block in `sync.test.ts`:
- never commits `config.yaml` even when it holds a Bearer token (and no committed blob contains the token)
- untracks a `config.yaml` committed by a vulnerable client on the next sync
- does not commit derived/secret files (`engrams.db`, `secrets.yaml`)
- force-stages a new engram file even when an excludes file would ignore it (#329 guarantee)
- asserts `core.excludesFile` is not set to `/dev/null` (#384)

Full `@plur-ai/core` suite green (1518 passed, 28 skipped).

## Operational note

Synced repos created before this fix that contain `config.yaml` will have it **untracked** on the next `plur sync`. If a token was already pushed to a shared/public remote, **rotate it and purge it from history** — code can only stop the bleeding.

Part of #377. Closes #380, #384.


crtahlin:
Rebased onto latest `main` (now includes #366). 

#366 landed a `sync.ts` refactor that extracted `neutraliseGlobalExcludes()` and began calling it on **every** sync (not just init) to migrate existing repos. That doubles down on the `core.excludesFile=/dev/null` behavior this PR removes — i.e. it extends the exact aggravator #384 flagged. This rebase **replaces `neutraliseGlobalExcludes` entirely** with the allowlist approach: force-adding only the named store files (`git add -A -f -- <allowlist>`) preserves the #329 "engram files always stage" guarantee without ever touching the user's global excludes, and secrets/derived files are structurally excluded from the staging pathspec.

Net: `/dev/null` is gone from both `initRepo` and `sync()`. Full workspace suite green (2000 passed).

crtahlin:

Clears the #380 release blocker from the #377 audit. The allowlist staging (`stageStoreFiles` force-adds only engram-store files) makes secret exclusion a structural guarantee rather than gitignore-dependent — exactly the approach #384 recommended. Removing `core.excludesFile=/dev/null` closes #384 while `-f` preserves the #329 force-stage guarantee. Self-healing untrack of `config.yaml`/`secrets.yaml` is the right call, with honest docs that token rotation + history purge stay manual.

**Verified locally:** `sync.test.ts` 18/18 on both the pre-rebase and rebased commit; full `packages/core` 1517 pass (the one failure is the known `pulls remote changes` timeout flake — passes in isolation, unrelated to this PR). CI green on Node 20 + 22.

Confirmed the rebase cleanly supersedes #366's `neutraliseGlobalExcludes` — function and both callsites removed, only a documenting comment remains. That reversal is intentional and correct (#384 established that neutralizing global excludes is the vulnerability).

**Non-blocking notes:**
1. `stageStoreFiles` filters the pathspec to *existing* allowlist files, so a wholesale deletion of an allowlisted file (e.g. `episodes.yaml` removed entirely) wouldn't propagate to the synced repo. Minor — engram files rarely get deleted outright, and the secret-untrack path is handled explicitly.
2. The code only stops the bleeding going forward (as the PR notes) — if a pre-fix client already pushed a token to a shared remote, rotate it and purge history.

plur9 (approved):
Review nit resolved: the force-add now carries :(exclude)packs/**/config.yaml/secrets.yaml/*.token so a secret nested in a (force-added) pack dir can't ride past .gitignore — the gap the root SECRET_PATHS untrack didn't cover. Regression test added (packs/<x>/{config,secrets}.yaml + creds.token stay untracked; SKILL.md still syncs). 20 sync tests pass. Core #380/#384 fix already verified. Approving + merging.

PR #389, opened 2026-06-22
fix(packs): secret/PII scan covers summary and domain (#381)

### LINKED ISSUE #381: Pack secret scan misses summary and domain — secrets leak into exported/shared packs
## Summary
Part of #377.

`scanPrivacy` runs `detectSecrets` only over `statement + rationale + source` (`packages/core/src/packs.ts:442-445`). The `summary` field (rendered by `formatLayer1`) and `domain` field (rendered by `formatLayer3`, caller-supplied via `learn`) are never scanned. `exportPack` blocks only `secret` + `private_visibility` issues (`:553`), so a secret in `summary`/`domain` is never detected — `privacy.clean` is `true` and the engram is exported verbatim, defeating the documented "Secrets are ALWAYS blocked" invariant (`:173`).

## Details
- The injection scan was already extended to include `summary` — proving the field set is known to lag the render set — but the secret scan was not, and neither scan covers `domain`.
- The `domain` channel is reachable through the ordinary public-engram write+export flow (`learn`'s write guard scans only `statement`).

## Repro
```js
const e = { statement:'Use the deploy helper',
            summary:'api_key=AKIAIOSFODNN7EXAMPLE0',
            domain:'token sk-abcdefghijklmnopqrstuvwx' }
scanPrivacy([e]).clean   // true, issues []
// exportPack writes engrams.yaml with summary + domain intact
// the same secret in `statement` IS caught and blocked
```

## Recommendation
Scan every rendered/exported field: `const secretText = [statement, rationale, source, summary, domain].filter(Boolean).join(' ')`. Keep the injection field set in sync. Also secret-scan caller-supplied `domain` at `learn()`/`learnRouted()` write time. Add tests asserting install/export block a secret hidden in `summary` and in `domain`.

crtahlin:
## What & why

Fixes #381 from the #377 scope-security audit (High, confidentiality).

`scanPrivacy` ran `detectSecrets` over only `statement + rationale + source` (`packages/core/src/packs.ts`). Two rendered/exported fields were never secret-scanned:
- `summary` — rendered into agent context by `formatLayer1`
- `domain` — rendered by `formatLayer3`, and **caller-supplied via `learn`**

So a secret hidden in `summary` or `domain` exported with `security.clean === true`, defeating the documented "secrets are ALWAYS blocked" export invariant. The injection scan had already been extended to include `summary` — proving the field set was known to lag the render set — but the secret scan wasn't, and neither covered `domain`.

## Fix

- **`scanPrivacy`** now builds its scan text from `[statement, rationale, source, summary, domain]` and uses that **single field set for both the secret/PII scan and the injection scan**, so they can't drift apart again. `installPack` blocks and `exportPack` filters an engram with a secret in `summary` or `domain`.
- **`learn()` / `learnRouted()`** secret-scan the caller-supplied `domain` (not just `statement`) when `allow_secrets` is false — rejecting a secret in `domain` at write time. (Other context fields are already covered by `_guardSensitiveScope` on shared/remote writes; `summary` is internally derived, not a caller input.)

## Tests

In `packs.test.ts`:
- install **blocks** a secret in `summary`
- install **blocks** a secret in `domain`
- export **filters out** (`engram_count: 0`) an engram with a secret in `summary` or `domain`

In `plur.test.ts`:
- `learn` rejects a secret hidden in caller-supplied `domain`

Full workspace suite green (2004 passed, 34 skipped).

Part of #377. Closes #381.


crtahlin:
## Review: solid fix for the named scope, but the cited invariant isn't fully restored

The core change is right — unifying the secret + injection scans onto one shared field set kills the exact drift that caused #381 (injection scan knew about `summary`, secret scan didn't). `summary` and `domain` are now correctly blocked on install and filtered on export, with good tests. Verified locally: `packs.test.ts` + `plur.test.ts` 71/71, CI green.

### What still needs doing

`exportPack` serializes the **entire** engram (`{ ...e }`, minus pinned/locked_*), but `scanPrivacy` still enumerates only `statement / rationale / source / summary / domain`. So fields that are exported verbatim but not in that list remain a bypass of the "secrets are ALWAYS blocked" invariant this PR cites.

Reproduced via `exportPack` on this branch:

```
summary           = BLOCKED   ✓ (this PR)
domain            = BLOCKED   ✓ (this PR)
tags              = LEAKED    ✗
structured_data   = LEAKED    ✗
contraindications = LEAKED    ✗
```

`tags` is the concrete one: caller-supplied via `LearnContext.tags`, exported verbatim — i.e. reachable through the same ordinary learn→export flow as the `domain` vector this PR fixes. `structured_data` and `contraindications` are also caller-settable and exported.

### Recommended

1. **Durable fix — scan the serialized exported payload** rather than an enumerated field list. Scan the actual YAML `exportPack` writes (or the full engram object stringified), so any future field addition can't silently reopen this gap. This is the same enumerate-vs-serialize root cause as #381, one level out.
2. If you prefer to stay field-based for now, at minimum add `tags`, `structured_data`, and `contraindications` to the shared scan text, and add an export-filter test for a secret in `tags`.
3. Keep the write-time guard (`learn`/`learnRouted`) symmetric with whatever export-scan set you land on — it currently widened to `domain` only, but `tags`/`abstract` are also caller inputs.

Either land (1)/(2) here, or merge this for the named summary+domain vectors and file a fast follow-up issue for the residual fields — your call. Flagging so the `tags` leak gets tracked rather than lost.

crtahlin:
Good catch — confirmed and fixed in the latest push (0ce9daa). The `tags` vector was real: `exportPack` does `{ ...e }`, so `tags`/`structured_data`/`contraindications` (all caller-settable) were exported verbatim but unscanned.

Took the durable option (1): `scanPrivacy`'s **secret/PII pass now scans the serialized engram payload** rather than an enumerated field list, so a future field can't silently reopen the gap. Fields `exportPack` strips (`relations`/`associations`/`knowledge_anchors`) and internal/numeric bookkeeping are excluded to avoid false rejections, and `_`-prefixed `structured_data` keys are dropped.

One deliberate split from the suggestion: the **prompt-injection pass stays field-based** (`statement/rationale/source/summary/domain`) — only fields rendered into agent context can carry an effective injection, and scanning arbitrary serialized metadata for injection would add false positives without a real attack surface.

Per point 3, the write-time hard check (`learn`/`learnRouted`) is now symmetric: widened to caller-supplied `domain`, `tags`, and `abstract`.

Added tests: install blocks + export filters a secret in `tags`, `structured_data`, and `contraindications`; `learn` rejects a secret in `tags`. Full workspace suite green (2002 passed).

plur9 (approved):
Review blockers resolved in-branch: serialized scan input is now capped (truncateToScanLimit) and EMAIL_RE bounded — kills the 8-17s ReDoS; scanPrivacy routes through detectSensitive so the infra family (public IP / internal host / basic-auth URL) is blocked on pack export/install, closing the detectSecrets gap. Added ReDoS-guard + infra-leak regression tests (131 pass). Approving + merging.

PR #390, opened 2026-06-22
fix(scope): don't auto-register personal-family scopes from /me (#382)

### LINKED ISSUE #382: Scope auto-registration trusts server /me scopes[] unconditionally — hostile remote can register a writable 'global' store
## Summary
Part of #377.

`registerDiscoveredScopes()` registers **every** scope returned by `GET /api/v1/me` (`me.scopes`) via `addStore('', scope, {url, token})` (`packages/core/src/index.ts:3958-3959`) with no check that the scope is a SHARED-family scope. `addStore` (`:3523-3621`) validates only URL/protocol/non-empty-scope + dedup — there is no `isSharedScope` check.

## Details
A compromised or malicious endpoint can return `scopes:['global']` (or `'local'`, `'user:<victim>'`). `addStore` persists a writable remote `StoreEntry {url, scope:'global', readonly:false}` (the scope-conflict guard does not fire because the primary `global` store is not a `config.stores` row). Then `_isRemoteBackedScope('global')` is true and `_resolveRemoteStoreForScope('global')` returns the hostile driver. Since `global` is the default unscoped routing fallback and default scope, every subsequent unscoped/global `learn` POSTs to the attacker's server (the write guard only demotes secrets/infra; ordinary preferences pass). Same gap for `user:*`. #306 treats `/me` rows as semi-trusted on the READ side; the REGISTER side grants full trust. Reachable via `plur_scopes_discover register:true` (which `session_start` nudges) and `plur stores discover --register`.

## Repro
1. User has a token for an attacker-controlled/MITM'd endpoint in config.
2. `plur_scopes_discover register:true`.
3. Hostile `/me` -> `{scopes:['global']}`.
4. `addStore('','global',{url:hostile,token})` persists a writable remote global store.
5. Any later default/unscoped `learn` -> fallback `global` -> hostile driver -> `appendAndGetServerId` POSTs the statement to the attacker.

`scope-discovery.test.ts` only uses well-formed `group:` scopes.

## Recommendation
Filter `d.authorized` through `isSharedScope()` before `addStore` in `registerDiscoveredScopes` (and the CLI/init-remote register path) — never auto-register a personal-family scope (`global`/`local`/`user:*`/`agent:*`) as remote-backed. Reject and log non-shared scopes, surface skipped scopes in the result. Add a test asserting a `/me` returning `['global','user:victim','local']` registers nothing.

crtahlin:
## What & why

Fixes #382 from the #377 scope-security audit (High, confidentiality).

`registerDiscoveredScopes()` registered **every** scope a server returned from `GET /api/v1/me` as a writable remote store, with no shared-scope check. `addStore` validates only URL/protocol/non-empty-scope + dedup.

A compromised or MITM'd endpoint can return `scopes: ['global', 'user:<victim>', 'local']`. Registering `global` — the default unscoped routing fallback and default scope — as a writable remote store makes `_resolveRemoteStoreForScope('global')` return the hostile driver, so **every later default/unscoped `learn` POSTs to the attacker's server**. The scope-conflict guard doesn't fire (the primary `global` store isn't a `config.stores` row). The READ side (#306) already treats `/me` rows as semi-trusted; the REGISTER side granted full trust.

Reachable via `plur_scopes_discover register:true` (which `session_start` nudges) and `plur stores discover --register`.

## Fix

`registerDiscoveredScopes()` filters `/me`-advertised scopes through `isSharedScope()` before `addStore`:
- Only shared-family scopes (`group:`/`project:`/`space:`/`team:`/`org:`/`public`) are auto-registered.
- Personal-family scopes (`global`/`local`/`user:*`/`agent:*`) are refused, logged, and returned in a new `skipped` field on `RegisterDiscoveredResult`. The CLI and MCP (`plur_scopes_discover`) surface them.
- A genuine remote-backed personal scope (e.g. a `user:` scope on your own server) must be added deliberately via `plur stores add` — never auto-registered from untrusted server input.

Both register call sites (MCP `plur_scopes_discover`, CLI `plur stores discover --register`) delegate to this one core method, so the single fix covers them. `init-remote` takes explicit user-provided scopes (no `/me` trust) and is unaffected.

## Tests

`scope-discovery.test.ts`: a `/me` returning `['global','local','user:victim','agent:bot','group:plur/plur-ai/engineering']` registers **nothing new** (`added: []`), reports the four personal scopes in `skipped`, and persists **no** personal-scoped remote store in config.

Full workspace suite green (1996 passed, 34 skipped).

Part of #377. Closes #382.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,15 @@
 
 ## Unreleased
 
+### Security: scope auto-registration refuses personal-family scopes from `/me` (#382)
+
+`registerDiscoveredScopes()` registered **every** scope a server returned from `GET /api/v1/me` as a writable remote store, with no shared-scope check. A compromised or MITM'd endpoint could return `scopes: ['global', 'user:<victim>', 'local']`; registering `global` (the default unscoped routing fallback) as a writable remote store would route every later default/unscoped `learn` to the attacker's server.
+
+- `registerDiscoveredScopes()` now filters `/me`-advertised scopes through `isSharedScope()` before `addStore`: only shared-family scopes (`group:`/`project:`/`space:`/`team:`/`org:`/`public`) are auto-registered. Personal-family scopes (`global`/`local`/`user:*`/`agent:*`) are refused, logged, and returned in a new `skipped` field on `RegisterDiscoveredResult`. The CLI (`plur stores discover --register`) and MCP (`plur_scopes_discover`) surface the skipped scopes.
+- A genuine remote-backed personal scope (e.g. a `user:` scope on your own server) must be added deliberately via `plur stores add`; it is never auto-registered from untrusted server input.
+
+**Behavior change:** `plur_scopes_discover register:true` / `plur stores discover --register` no longer register personal-family scopes a server advertises — they are reported as `skipped`.
+
 ## 0.10.0 (2026-06-21)
 
 ### Leak guard: write-time demotion now covers `saveMetaEngrams` and remote-backed scopes (#368, #370)
--- a/packages/cli/src/commands/stores.ts
+++ b/packages/cli/src/commands/stores.ts
@@ -57,6 +57,7 @@ export async function run(args: string[], flags: GlobalFlags): Promise<void> {
       registered.forEach(r => {
         if (!r.ok) { outputText(`${r.url}: register failed — ${r.error}`); return }
         outputText(`${r.url}: added ${r.added.length} scope(s)${r.added.length ? ` (${r.added.join(', ')})` : ''}`)
+        if (r.skipped.length) outputText(`  skipped ${r.skipped.length} non-shared scope(s) (not auto-registered): ${r.skipped.join(', ')}`)
       })
     } else {
       const total = discoveries.reduce((n, d) => n + d.unregistered.length, 0)
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -246,6 +246,8 @@ export interface RegisterDiscoveredResult {
   ok: boolean
   added: string[]
   already_registered: string[]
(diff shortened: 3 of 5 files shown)


crtahlin:

Fixes #382. Filtering `/me`-advertised scopes through `isSharedScope()` in `registerDiscoveredScopes()` is exactly the issue's recommendation, and the guard is placed at the right layer — auto-register (untrusted server input) is locked down while deliberate `plur stores add` stays permissive.

**Verified locally:**
- `scope-discovery.test.ts` 9/9; full `packages/core` 1518 pass 0 fail; `packages/mcp` 135 pass (consumes the changed `RegisterDiscoveredResult`). CI green on Node 20 + 22.
- **Completeness swept** — `registerDiscoveredScopes` is the *only* path that auto-registers from `/me`. The local-project `addStore` (`index.ts:3673`) is filesystem-derived (`project:` scopes, never `/me`); `init-remote` uses `me.scopes` only for display and persists the user-provided `--scopes` whitelist. Both register call sites (CLI `stores discover --register`, MCP `plur_scopes_discover`) delegate to the one core method.
- `skipped` is added to all return paths including the `!d.ok` early return, so the CLI's `r.skipped.length` never hits undefined.

**Non-blocking note:** a hostile server can still get a legit-looking *shared* scope auto-registered (e.g. `group:...`), but that's by design — scope discovery exists to trust a server you hold a token for, `addStore`'s scope-conflict guard fires on a mismatched existing mapping, and such a store only receives writes explicitly scoped there (not the unscoped-default routing #382 targets). Drawing the line at personal-family is the correct and sufficient fix.

plur9 (approved):
Allowlist gate (isSharedScope) fully closes #382 — verified end-to-end with the hostile-/me test. Nit (non-blocking): a cross-ref comment on addStore noting it doesn't gate scope family would help future callers. Approving.

PR #391, opened 2026-06-22
fix(secrets): raise scan window to 1 MiB, fail-closed past it (#386)

### LINKED ISSUE #386: 64KB scan cap lets infra-class sensitive content past byte 64KB leak to shared/remote stores and packs
## Summary
Part of #377.

`detectSensitive()` truncates its input to the first 64KB before scanning (`packages/core/src/secrets.ts:372-380,400`). The infra-topology detectors (`public_ipv4`, `public_ipv6`, `basic_auth_url`, `fqdn_port`, `ipv4_port`, `internal_host`) exist **only** inside `detectSensitive` — not in `detectSecrets`, which is the non-truncating hard-throw at the learn leaf. Engram statements have no max length.

## Details
An engram whose first 64KB is benign filler but which carries a public IP / basic-auth URL / internal host **after** byte 64KB passes the write-guard (`_offendingHitsForScope` returns `[]` on the truncated benign prefix; reached via `index.ts:950`, `:1169-1171`), is not demoted, and is written to a SHARED scope and POSTed to the REMOTE store. The same truncation governs `filterPublishable` (`publish.ts:62`). Classic credentials past 64KB are still caught by `detectSecrets`, but the infra family (the exact class behind the 2026-06 leak) is wide open past 64KB.

## Repro (compiled dist)
```js
const filler = 'x'.repeat(70000)
detectSensitive(filler + ' 139.59.155.82')                      // [] (miss)
detectSensitive(filler + ' https://t:p@hub-staging.plur.ai')    // [] (miss)
// same payloads at byte 0 are caught
plur.learnRouted(filler + ' deploy droplet 139.59.155.82 with creds https://t:p@hub-staging.plur.ai',
                 { scope:'group:plur/eng' })
// -> scope 'group:plur/eng' (NOT demoted); exactly 1 POST to remote containing the IP + basic-auth host
```
The cap test only checks an AWS key (independently caught by `detectSecrets`), so the infra hole is unasserted.

## Recommendation
Do not rely on a fixed 64KB window for a security guarantee. Either (a) enforce a hard max statement/context length below the cap (reject/hard-fail oversized scanText), or (b) scan the full input for the security-critical detectors (infra family + `basic_auth_url`) in a chunked/streaming manner with boundary overlap (the bounded regexes are now linear). At minimum, signal "scan truncated, N bytes unscanned" and treat it as fail-closed in `_guardSensitiveScope` and `filterPublishable`. Add a regression: a >64KB engram with infra content past byte 64KB must be demoted on `learnRouted` to a remote scope and excluded by `filterPublishable`.

crtahlin:
## What & why

Fixes #386 from the #377 scope-security audit (High, confidentiality).

`detectSensitive()` truncated its input to the first 64 KB before scanning, then silently passed the rest. The infra-topology detectors (`public_ipv4`, `public_ipv6`, `basic_auth_url`, `fqdn_port`, `ipv4_port`, `internal_host`) exist **only** in `detectSensitive` (not in the non-truncating `detectSecrets`). So an engram whose first 64 KB is benign filler but which carries a public IP / basic-auth URL / internal host **after** byte 64 KB:
- passed `_offendingHitsForScope` (`[]` on the truncated benign prefix),
- was **not demoted**, written to a SHARED scope and POSTed to the REMOTE store,
- and slipped past `filterPublishable` (same truncation) — the exact 2026-06 leak class.

## Fix

- **Window raised 64 KB → 1 MiB.** The detector regexes were already rewritten to bounded/linear forms (the codebase comment notes the 64 KB cap is "cheap insurance, not the DoS fix — the bounded quantifiers are"). A single full-window pass is ~100 ms worst case, and total scan work is **capped at 1 MiB regardless of input size**, so this stays DoS-safe. 1 MiB is far above any realistic engram, so ordinary content is now fully scanned.
- **Fail-closed past the ceiling.** Input larger than 1 MiB now appends a synthetic `scan_truncated` hit (exported as `SCAN_TRUNCATED`) so `_guardSensitiveScope` demotes the write and `filterPublishable` excludes the engram — the unscanned tail can no longer be assumed clean. The signal is **always offending regardless of a scope's `sensitivity` policy** (a policy can't `allow` it).

## Tests (`secrets.test.ts`)

- infra content (`public_ipv4`, `basic_auth_url`) past the old 64 KB cap is now detected
- a >1 MiB benign input reports `scan_truncated`; a secret entirely past the ceiling is flagged truncated (fail-closed) rather than silently passing
- truncation at the 1 MiB boundary produces valid UTF-8 (no throw) even when it splits a multibyte char
- **end-to-end:** a >64 KB engram with infra past byte 64 KB is **demoted** on a shared-scope `learn` and **excluded** by `filterPublishable`
- the existing 64 KB ReDoS-timing fuzz test still passes (linearity holds)

Full workspace suite green (2000 passed, 34 skipped).

Part of #377. Closes #386.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,15 @@
 
 ## Unreleased
 
+### Security: sensitivity scan window raised to 1 MiB and fail-closed past it (#386)
+
+`detectSensitive()` truncated its input to the first 64 KB before scanning, then silently passed the rest. The infra-topology detectors (`public_ipv4`, `public_ipv6`, `basic_auth_url`, `fqdn_port`, `ipv4_port`, `internal_host`) exist only in `detectSensitive`, so an engram whose first 64 KB was benign filler but which carried a public IP / basic-auth URL / internal host **after** byte 64 KB passed the write guard un-demoted and was written to a shared/remote store (and slipped past `filterPublishable`).
+
+- The scan window is raised from 64 KB to **1 MiB** — far above any realistic engram. The detector regexes are bounded/linear, so a single full-window pass is ~100 ms worst case, and total scan work is capped at 1 MiB regardless of input size (DoS-safe).
+- Input larger than the ceiling is now **fail-closed**: `detectSensitive` appends a synthetic `scan_truncated` hit so `_guardSensitiveScope` demotes the write and `filterPublishable` excludes the engram — the unscanned tail can no longer be assumed clean. The `scan_truncated` signal is always offending regardless of a scope's `sensitivity` policy.
+
+**Behavior change:** infra/secret content anywhere in the first 1 MiB is now detected and demoted; an engram larger than 1 MiB destined for a shared/remote scope is demoted to `local`/`private` (fail-closed) rather than silently passing.
+
 ## 0.10.0 (2026-06-21)
 
 ### Leak guard: write-time demotion now covers `saveMetaEngrams` and remote-backed scopes (#368, #370)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,9 @@
 
 `detectSensitive()` truncated its input to the first 64 KB before scanning, then silently passed the rest. The infra-topology detectors (`public_ipv4`, `public_ipv6`, `basic_auth_url`, `fqdn_port`, `ipv4_port`, `internal_host`) exist only in `detectSensitive`, so an engram whose first 64 KB was benign filler but which carried a public IP / basic-auth URL / internal host **after** byte 64 KB passed the write guard un-demoted and was written to a shared/remote store (and slipped past `filterPublishable`).
 
-- The scan window is raised from 64 KB to **1 MiB** — far above any realistic engram. The detector regexes are bounded/linear, so a single full-window pass is ~100 ms worst case, and total scan work is capped at 1 MiB regardless of input size (DoS-safe).
+- The scan window is raised from 64 KB to **1 MiB** — far above any realistic engram. The detector regexes are bounded/linear; a benign full-window pass is ~7ms/64KB but adversarial regex-dense input measured ~300–420 ms for a full 1 MiB pass (#386 review). Total scan work is capped at 1 MiB regardless of input size (bounded, linear — a per-write CPU cost on >64KB engrams, not a DoS).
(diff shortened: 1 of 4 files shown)


crtahlin:

Fixes #386. Raising the window to 1 MiB and **failing closed past it** (synthetic `scan_truncated` hit that no scope `sensitivity` policy can `allow`) is the right design. Verified completeness: both `detectSensitive` consumers are fail-closed — `_guardSensitiveScope` via the explicit `SCAN_TRUNCATED` short-circuit *before* `sensitivityCategory`, and `filterPublishable` inherently (rejects on any hit). `detectSecrets` (non-truncating) still catches classic secrets past 1 MiB at the learn leaf, so only the infra family needed the fail-closed tail — sound.

**Verified locally:** `secrets.test.ts` 110/110 (incl. the e2e demote-on-shared-learn + filterPublishable-exclude); full `packages/core` 1521 pass 0 fail. CI green on Node 20 + 22.

**Non-blocking notes:**
1. The "~100 ms worst case" figure is optimistic. I timed `detectSensitive` on adversarial 1 MiB inputs: plain 357 ms, dotted-labels 418 ms, digit/colon 531 ms (at-sign URLs 30 ms). Still **linear and bounded** — sub-second, not ReDoS — and only for pathological near-1MiB writes; normal small engrams are unaffected (fast path scans actual length). Since the scan runs synchronously on every `learn`, an oversized-write flood now costs ~70× the old 64 KB per write. Suggest (a) softening the ~100 ms claim in the CHANGELOG/comment to the measured ~0.5 s, and (b) bumping the existing 64 KB ReDoS-timing fuzz test to 1 MiB so the bound is asserted at the new ceiling rather than extrapolated.
2. Minor: `detectSensitive` is publicly exported. In-repo both gates handle `SCAN_TRUNCATED` correctly, but any future category-filtering consumer must replicate the "always offending, can't be allowed" rule. Consider centralizing an `isAlwaysOffending(hit)` predicate (or noting it on the export) so the invariant can't drift.

Neither blocks merge — the fix is strictly safer than the silent-pass it replaces.

plur9 (approved):
Stores fail-closed path verified. Review fixes applied: perf claim corrected (~100ms -> measured ~300-420ms on adversarial 1 MiB); the '...and packs' half of #386 is now delivered via #389 (scanPrivacy routes through detectSensitive + truncateToScanLimit, inheriting the raised window + scan_truncated). 135 secrets/packs tests pass. Approving + merging.

Issue #408, opened 2026-06-23
[audit] Malformed-row drop log echoes server-controlled values (log injection)


plur9:
From the independent 0.10.0 scope-security audit (#377), Črt.

**Severity:** Low (Other)
**Location:** packages/core/src/store/remote-store.ts:78-81

Malformed-row drop log echoes server-controlled field values (Zod messages embed received values) — log injection / info exposure.

Part of the pre-0.10.0 hardening triage (see #377).

crtahlin:
Verified fixed on `main` @ `b2b1dd9` during the #377 re-audit. Implemented in merged **#414**: `reshape` (`packages/core/src/store/remote-store.ts:83`) now builds the drop reason from `i.path.join('.')` + `i.code` (Zod enum codes — no received values) instead of `i.message` (which embedded the raw server string incl. CRLF/ANSI), and the id is sanitized (`[^\w:./-]` → `?`, capped at 64; remote-store.ts:84). Log-injection / info-leak closed. Looks complete — leaving the close to @plur9.

plur9:
Fixed in merged #414 and verified at file:line by the #377 re-audit. Closing per the re-audit's issue-state reconciliation.

PR #446, opened 2026-07-02
fix(core): patch() no longer returns null on a 2xx whose echoed row fails validation

### LINKED ISSUE #327: RemoteStore.patch() returns null on a successful PATCH when the response row fails schema validation
> Follow-up from the review of #306 (finding #4). Non-blocking — split out so #306 can merge.

## Summary

When a remote store updates an engram, `RemoteStore.patch()` asks the server to apply the change and then reads the engram back from the server's response. The hardening in #306 added strict validation of that response row. The side effect: if the **PATCH succeeds on the server** but the row the server sends back doesn't match the (intentionally lenient) validation schema, `patch()` returns `null` — the **same value it returns for a 404 "not found"**. A caller can't tell "the update didn't apply" from "the update applied but the echoed row looked unusual," so a successful write can be silently misread as a failure.

## Why it matters

- The mutation **did** take effect server-side (and the local cache is already invalidated), but the caller is told `null`.
- `null` from `patch()` conventionally means *not found / no-op*, so a caller may surface a false "engram not found" / "update failed" to the user, or retry an operation that already succeeded.
- Low likelihood today — all 137 production rows validate against the schema (verified in #306) — and it self-heals on the next `load()`. So this is a latent correctness edge, not a live outage.

## Where

`packages/core/src/store/remote-store.ts` — `patch()`:

```ts
if (r.status === 404) return null            // genuine not-found
if (!r.ok) { throw ... }                      // server error
this.cache = null
const body = await r.json()...
if (!body?.engram) return null
return this.reshape(body.engram)              // reshape() safeParses → null on a malformed row
```

`reshape()` runs `RemoteRowSchema.safeParse` and returns `null` (with a warning log) on any row that fails. On the read paths (`load`/`getById`) returning `null`/dropping the row is correct. On `patch()`, where the write already succeeded, `null` is ambiguous with not-found.

## Suggested fix

Distinguish "write failed / not found" from "write succeeded, response unparseable." Options:
- On a 2xx whose body fails to reshape, return the **optimistically-merged** engram (the local engram with `updates` applied) instead of `null`, since the server confirmed success; or
- Return a distinct signal (throw, or a typed result) so callers don't conflate it with 404.

Either way, a successful PATCH should never be reported to the caller as `null`/not-found.

## Acceptance criteria

- [ ] A 2xx PATCH whose response row fails `RemoteRowSchema` does **not** return `null` as if not-found.
- [ ] A genuine 404 still returns `null`.
- [ ] Test covering the 2xx-but-malformed-response case (extend the in-process stub in `test/helpers/stub-server.ts`).

Refs: #306.

plur9:
Closes #327.

## What

`RemoteStore.patch()` returned `null` when a PATCH succeeded on the server (2xx) but the echoed row failed `RemoteRowSchema` validation — the same value it returns for a genuine 404. Callers (`setPinnedAsync`, `updateEngramAsync`, `reportFailure` remote routing) conflate that with "not found" and misreport a successful write as a failure, or fall through to the next store.

## How

- `patch()` captures the pre-write row from the TTL cache **before** invalidating it.
- On a 2xx whose body is missing or fails reshape, it returns the **optimistically-merged** engram (cached pre-write row + the acknowledged updates, defined values only — mirroring what `JSON.stringify` actually sent), plus a sanitized warning log (#408 pattern). The next `load()` observes the server's authoritative state.
- Genuine 404 → still `null`. Valid echo → still the reshaped server row (no behavior change).
- Stub server gains a `badPatchEcho` knob (mirrors `badAppendId`) to simulate the malformed-echo case; cleared on `reset()`.

## Acceptance criteria (from #327)

- [x] A 2xx PATCH whose response row fails `RemoteRowSchema` does **not** return `null` as if not-found
- [x] A genuine 404 still returns `null`
- [x] Test covering the 2xx-but-malformed-response case (extends `test/helpers/stub-server.ts`)

## Testing

- `remote-integration.test.ts`: 30/30 (4 new: warm-cache merge, cold-cache fallback, 404, valid-echo regression)
- Core dist rebuilt, then full workspace: core 1564 passed (1 unrelated timing flake — `packs.test.ts` ReDoS guard — passes in isolation; PGLite timeouts from the first run likewise isolation-green), mcp 137/137, claw 107/107

### DIFF
--- a/packages/core/src/store/remote-store.ts
+++ b/packages/core/src/store/remote-store.ts
@@ -326,12 +326,28 @@ export class RemoteStore implements EngramStore {
       const text = await r.text().catch(() => '')
       throw new Error(`Remote patch failed: ${r.status} ${text}`)
     }
+    // #327: the server confirmed the write (2xx). Capture the pre-write row
+    // from the cache BEFORE invalidating so the fallback below can merge it.
+    const prev = this.cache?.engrams.find(e => e.id === id) ?? null
     this.cache = null
     // Server returns {engram: {id, scope, status, data: {...}, ...}}; reshape
     // to top-level Engram (same as load() does for rows[]).
     const body = await r.json().catch(() => null) as { engram?: { id: string; scope: string; status: string; data?: any } } | null
-    if (!body?.engram) return null
-    return this.reshape(body.engram)
+    const reshaped = body?.engram ? this.reshape(body.engram) : null
+    if (reshaped) return reshaped
+    // #327: 2xx but the echoed row was missing or failed validation. Returning
+    // null here would be indistinguishable from the 404 above — callers would
+    // misreport a successful write as not-found, or retry it. Return the
+    // optimistically-merged engram (pre-write cached row + the acknowledged
+    // updates); the next load() observes the server's authoritative state.
+    // Only defined update values are applied, mirroring what JSON.stringify
+    // actually sent to the server. Same #408 id sanitization as reshape().
+    const safeId = id.replace(/[^\w:./-]/g, '?').slice(0, 64)
+    logger.warning(`[plur:remote-store] ${this.url} PATCH ${safeId} succeeded but the echoed row was unusable — returning optimistic merge`)
+    const merged: Record<string, unknown> = prev ? { ...(prev as unknown as Record<string, unknown>) } : {}
+    for (const [k, v] of Object.entries(updates)) if (v !== undefined) merged[k] = v
+    merged.id = id
+    return merged as unknown as Engram
   }
 
   async close(): Promise<void> {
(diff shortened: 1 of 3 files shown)


plur9:
**Miles review:** CI green. Sound fix for #327. The optimistic-merge fallback on 2xx-but-bad-echo is correct — returning `null` was ambiguous (callers can't distinguish a successful write with bad echo from a 404). Four targeted tests cover warm-cache merge, cold-cache id+updates, genuine 404, and valid echo. Ready for merge.

plur9:
CTO review: ✅ looks good to merge.

CI green (Node 20 + 22 + Python SDK). Fixes the null-vs-not-found ambiguity in `RemoteStore.patch()\) correctly.

The warm-cache merge strategy (cached pre-write row + acknowledged delta) is the right call: callers get a usable optimistic result on a 2xx with malformed echo, while the exact null-on-404 contract is preserved. Consistent with the #408 sanitised-warning pattern.

Test coverage: 4 new cases cover all scenarios — warm-cache merge, cold-cache fallback, 404 still null, valid-echo regression. Good.

One non-blocking note: cold-cache fallback produces a partial object (delta-only, no prior state). Fields absent from the PATCH delta will be `undefined` on the returned engram. Given how narrow the call sites are this is fine, but a short inline comment on that path would help the next reader.

Ready for crtahlin sign-off.

PR #167, opened 2026-05-16
feat(claw,mcp): wire CapabilityCanary into plugin + MCP tools (closes #159)

miles-on-nightshift:
## Summary

- **claw**: Detects when `agent_end` is silently blocked by OpenClaw's `allowConversationAccess` gate. After 3 `before_agent_start` ticks with no `agent_end` signal, the canary injects a warning into the agent context with the exact fix command. Warning clears automatically when `agent_end` fires (hot-reload recovery).
- **mcp**: Module-level canary tracks `session_start_hook` (auto-injection) and `learn_activity` (corrections being captured). `plur_status` now returns a `capabilities` field with live canary state.
- **tensions tests**: Add `plur_tensions_purge` registration test, `purge_hint` assertion, and `injectLegacyConflict` helper for legacy conflict field testing.

## Context

Issue #51 identified a P0 onboarding failure: `agent_end` is silently blocked after install, so PLUR appears to work (injection works) but never learns. PR #157 fixed the root cause (setup.ts now auto-sets `allowConversationAccess`). This PR adds a runtime safety net: if the config is ever misconfigured, the user sees a warning in their next conversation turn rather than silent failure.

## Test plan

- [x] `packages/claw/test/canary.test.ts` — 5 tests: no warning before threshold, warning after 3 ticks, no warning when agent_end fires, warning clears on recovery, warning includes fix command
- [x] `packages/mcp/test/tensions.test.ts` — 9 tests pass including new purge registration + purge_hint tests
- [x] All tests pass (`vitest run`)

Closes #159

PR #458, opened 2026-07-02
fix(mcp): reset CapabilityCanary on session_start for per-session health detection (#192)

### LINKED ISSUE #192: mcp: CapabilityCanary should reset on session_start for per-session health detection
From PR #167 review.

The mcpCanary is module-level (singleton) — it accumulates state across the MCP server lifetime. Once learn_activity fires once, it stays healthy forever even if the user stops learning for hundreds of turns. Consider calling canary.reset() on plur_session_start to give per-session health detection rather than per-process.

### LINKED ISSUE #460: CapabilityCanary ticking must advance per tool call, not only in session_start
Split from #192's implementation (PR #458 review flag).

## Problem
#192 asked for `mcpCanary.reset()` on `plur_session_start`. Implementing reset alone makes the canary **dead code**: ticks only advanced inside `session_start`, so after a reset `ticks < threshold` is permanently true and the unhealthy warning can never fire. Per-session health detection requires the tick to represent a *turn*, not a *session start*.

## Change (implemented in PR #458, needs explicit review sign-off)
- Tick the canary once per tool call in the server dispatch loop (one tool call ≈ one turn)
- Requires exporting `mcpCanary` from `tools.ts`
- Semantics: N tool calls without learn activity → unhealthy signal; reset on new session

## Review questions
- Is per-tool-call the right turn proxy, or should read-only tools (recall/status) not count?
- Threshold still correct under the new tick rate (10 turns)?

Refs #192. Addressed by PR #458.

plur9:
## Summary

Follow-up from the PR #167 review (reviewer: @crtahlin). The module-level `mcpCanary` accumulated state across the whole MCP server lifetime — once `learn_activity` fired once, it stayed healthy forever, even if the user stopped learning for hundreds of turns.

Per the reviewer's suggestion, `plur_session_start` now calls `canary.reset()`, giving per-session health detection rather than per-process.

One necessary companion change: ticks previously only advanced inside `plur_session_start` itself, so after a reset the tick count could never exceed 1 and the canary could never flag anything (`ticks < threshold` would be permanently true — reset alone would have made the canary dead code). Ticking is relocated to the server dispatch loop: **one tick per tool call = one turn**. Ten turns (threshold) in a session without `plur_learn` now flags `learn_activity` in `plur_status`/`plur_doctor`; a learn call marks it healthy again; the next `session_start` opens a fresh window. `mcpCanary` is exported from `tools.ts` so `server.ts` can tick it.

## Test plan

TDD — 2 of 3 new tests failed on the old behavior before the fix:
- [x] `session_start` clears `learn_activity` carried over from a previous session (direct handler)
- [x] Wire protocol: `learn_activity` flagged unhealthy after 10 turns with no learning, `session_start_hook` stays healthy, fresh window after next `session_start`
- [x] Wire protocol: stays healthy through threshold+ turns when learning happens mid-session
- [x] `packages/mcp/test/canary.test.ts` — 3/3 pass
- [x] Full `@plur-ai/mcp` suite: 140 passed, 0 failed (baseline was 137)
- [x] `pnpm --filter @plur-ai/mcp build` (incl. DTS type-check) passes

Closes #192

Closes #460 (canary ticking relocation — split out for explicit review).

### DIFF
--- a/packages/mcp/src/server.ts
+++ b/packages/mcp/src/server.ts
@@ -11,7 +11,7 @@ import {
   McpError,
 } from '@modelcontextprotocol/sdk/types.js'
 import { Plur, checkForUpdate } from '@plur-ai/core'
-import { getToolDefinitions } from './tools.js'
+import { getToolDefinitions, mcpCanary } from './tools.js'
 import { registerFlushOnExit } from './telemetry.js'
 import { VERSION } from './version.js'
 import { z } from 'zod'
@@ -211,6 +211,10 @@ export async function createServer(plur?: Plur): Promise<Server> {
         isError: true,
       }
     }
+    // #192: one tick per tool call = one "turn" for capability health.
+    // plur_session_start resets the canary, giving a per-session window:
+    // `threshold` turns without an expected signal flags the capability.
+    mcpCanary.tick()
     try {
       // Validate arguments against input schema
       const args = request.params.arguments ?? {}
--- a/packages/mcp/src/tools.ts
+++ b/packages/mcp/src/tools.ts
@@ -101,7 +101,8 @@ function sanitizeStatement(raw: string): string {
   return raw.slice(0, cut).trimEnd()
 }
 
-const mcpCanary = new CapabilityCanary({ threshold: 10 })
+// Exported so the server dispatch loop can tick it once per tool call (#192).
+export const mcpCanary = new CapabilityCanary({ threshold: 10 })
 mcpCanary.expect({
   id: 'session_start_hook',
   description: 'Automatic memory injection via hooks',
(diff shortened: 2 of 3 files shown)


plur9 (commented):
## Review sign-off — Issue #460 questions

**Q1: Is per-tool-call the right turn proxy, or should read-only tools (recall/status) not count?**

Per-tool-call is the correct proxy. `CapabilityCanary` marks `learn_activity` healthy if `firedCount > 0` (any learn signal this session) OR if `ticks < threshold` (still early in the session). Excluding read-only tools would mean a session of 50 `plur_recall` calls with zero `plur_learn` calls wouldn't trip the warning — but that's exactly the unhealthy pattern we want to surface. Every tool call is a turn where learning *could* have happened. Count them all.

**Q2: Is threshold 10 still correct under the new tick rate?**

10 is reasonable for launch. The semantics changed from "10 session starts without learning" (old) to "10 tool calls in a session without learning" (new), which is the right unit. A typical session: session_start (reset, tick=0), inject/recall (1-3 ticks), work calls (4-9 ticks), and `plur_learn` should fire before tick 10 if the agent is functioning correctly. One `plur_learn` anywhere in the session permanently satisfies the canary (firedCount > 0), so false positives only fire on genuinely idle sessions. We can tune upward if real-world sessions prove too noisy — but 10 is a sound starting point.

**Decision: both choices are correct as implemented. No code changes needed before merge.**

The test suite validates the new semantics cleanly (3 new tests, 140 total passing). The tick relocation from inside `session_start` to the server dispatch loop is the right fix — without it, reset alone would have made the canary permanently satisfied post-reset. Approved pending CI green.

PR #474, opened 2026-07-02
fix(mcp): harden array-typed tool arguments against client serialization bug

### LINKED ISSUE #297: plur_learn: passing a tags array empties the entire arguments payload ("statement: Required")

plur9:
Closes #297

## Root cause

The empty-payload symptom is **client-side**, not ours:

- The existing wire-protocol regression test (`plur_learn accepts tags array (#297)`, InMemoryTransport, real MCP `Client`) passes on main — the server validates and stores real arrays correctly.
- Probed the MCP SDK directly (`@modelcontextprotocol/sdk` ^1.12.0): `CallToolRequestSchema` round-trips `arguments: { statement, tags: ["y"] }` byte-identically, and **rejects** a JSON-stringified `arguments` payload at the protocol layer (`expected record, received string`, -32602) — meaning a stringified payload would never reach our app-level validator. Since the reported error IS our app-level validator with `Received fields: []`, the client transmitted an empty arguments record.
- Conclusion matches the triage in #297: Claude Code's MCP client drops the entire `arguments` object when an array-typed parameter is present. **Not an SDK bug** — no SDK issue to file; the upstream report belongs with the Claude Code client team (wire log comparing `tools/call` bodies for scalar vs array params, as described in the issue's triage comment).

## Hardening (this PR — our side)

A tags array now either **works** or **fails loudly with a useful message** — never a silent empty payload:

1. **String→array coercion** in the JSON-Schema→Zod validation layer (`jsonSchemaPropToZod`), applied at any nesting depth:
   - `'["a","b"]'` → JSON parse → `['a','b']` (any item type)
   - `'a, b'` → comma split → `['a','b']` (string-item arrays only)
   - `'solo'` → `['solo']` (string-item arrays only)
   - A malformed JSON-looking string (`'["unterminated'`) is passed through unchanged so `z.array` rejects it loudly, naming the field — never silently reinterpreted as a single tag.
2. **Handlers now receive the validated parse result** instead of the raw payload — previously coercion could never reach handlers. `.passthrough()` preserves undeclared fields.
3. **Targeted diagnostic**: when validation fails with zero received fields on a tool declaring array-typed params (the #297 signature), the error names the client bug and the coercible retry shapes:

   > Known client-side bug (plur-ai/plur#297): some MCP clients drop the entire arguments payload when an array-typed parameter is included. Retry passing array parameters as a JSON string (e.g. tags: "[\"a\",\"b\"]") or a comma-separated string (tags: "a, b") — the server coerces both back into arrays.

   The hint is suppressed on tools without array params (verified by test) so it never becomes noise.

## Tests

- 7 new wire-protocol tests (InMemoryTransport) in `packages/mcp/test/server.test.ts`; coercion results verified against the **stored engram** (`plur.getById(...).tags`), not just the tool response.
- Full `@plur-ai/mcp` suite: 172 tests, all passing (one known flaky-under-load timeout in `session.test.ts` during the parallel run; passes in isolation, untouched by this change).

## Upstream status

#297 stays accurate as the tracking issue for the Claude Code client fix; this PR closes the *server-side hardening* half — after this, the failure mode is self-describing and the documented workaround actually round-trips.

### DIFF
--- a/packages/mcp/src/server.ts
+++ b/packages/mcp/src/server.ts
@@ -155,7 +155,32 @@ function jsonSchemaPropToZod(prop: any): z.ZodTypeAny {
   if (prop.type === 'boolean') return z.boolean()
   if (prop.type === 'array') {
     const itemSchema = prop.items ? jsonSchemaPropToZod(prop.items) : z.unknown()
-    return z.array(itemSchema)
+    // #297 defense: some MCP clients mis-serialize array-typed arguments (in
+    // the worst case dropping the ENTIRE arguments payload). The documented
+    // workarounds pass arrays as strings — coerce those back into real arrays
+    // so the workaround round-trips instead of failing validation:
+    //   '["a","b"]'  → JSON parse → ['a','b']   (any item type)
+    //   'a, b'       → comma split → ['a','b']  (string items only)
+    //   'solo'       → wrap        → ['solo']   (string items only)
+    // A string that LOOKS like JSON ('[...') but fails to parse is passed
+    // through unchanged so z.array rejects it loudly, naming the field —
+    // never silently reinterpreted as a single tag.
+    return z.preprocess((val) => {
+      if (typeof val !== 'string') return val
+      const trimmed = val.trim()
+      if (trimmed.startsWith('[')) {
+        try {
+          const parsed = JSON.parse(trimmed)
+          return Array.isArray(parsed) ? parsed : val
+        } catch {
+          return val
+        }
+      }
+      if (prop.items?.type === 'string') {
+        return trimmed.length === 0 ? [] : trimmed.split(',').map(s => s.trim()).filter(s => s.length > 0)
+      }
+      return val
+    }, z.array(itemSchema))
   }
   if (prop.type === 'object' && prop.properties) {
     const shape: Record<string, z.ZodTypeAny> = {}
@@ -217,7 +242,7 @@ export async function createServer(plur?: Plur): Promise<Server> {
     mcpCanary.tick()
     try {
       // Validate arguments against input schema
-      const args = request.params.arguments ?? {}
+      let args = request.params.arguments ?? {}
       const schema = tool.inputSchema as any
       if (schema?.properties) {
         const shape: Record<string, z.ZodTypeAny> = {}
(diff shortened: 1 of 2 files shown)


plur9:
CTO Review: LGTM. Root cause analysis correct — wire-protocol test on main confirms server handles real arrays fine, so the serialization failure is client-side. This PR closes our hardening half.

Implementation notes:
- `z.preprocess` in `jsonSchemaPropToZod` is the right intercept point — coercion at any nesting depth before `z.array` validates
- Comma-split and bare-wrap correctly gated to `prop.items?.type === 'string'` — non-string arrays unaffected
- Malformed JSON-looking strings fall through to `z.array` for loud rejection (no silent swallow)
- `args = parsed.data` is the necessary behavioral change — without it coercion never reaches handlers; `.passthrough()` preserves undeclared fields
- Targeted #297 hint suppressed on non-array tools — no noise for unrelated failures

7 wire-protocol tests verify coercion against the stored engram (`plurInstance.getById().tags`), not just tool response text. 172 MCP tests pass.

Needs a non-author approval to merge (branch protection). @crtahlin — straightforward hardening PR, no logic changes to core retrieval.

Issue #481, opened 2026-07-03
supersedes chains in inject/recall/decay — injection priority, recall annotation, faster decay


plur9:
## Context

#240 landed `relations.supersedes[]` / `relations.superseded_by[]` edges (written by `plur_learn supersedes:[ids]`, `plur learn --supersedes`, and the `LearnContext`), and the tension scanner now skips supersedes-linked pairs. Per the scope guard in the #240 discussion, the *consumer-side* behavior of those chains was split out of that PR into this issue.

## Remaining work (from #240's "Impact on injection and recall" section)

1. **Injection (`plur_inject`)**: when budget-constrained, prefer the latest engram in a supersedes chain. The older engram stays injectable when the query is explicitly historical ("what was the law before the AI Act?").
2. **Recall (`plur_recall` / `plur_recall_hybrid`)**: return all engrams in the chain, but annotate older ones: `[superseded by ENG-xxx]`.
3. **Decay (`plur_batch_decay`)**: superseded engrams decay faster unless they have high recall frequency (someone keeps looking them up for history).

## Notes

- The edge already exists and round-trips through YAML persistence; only read-path behavior is missing.
- Reverse (`superseded_by`) edges are best-effort: written for targets in the local primary store only. Remote-store targets carry no reverse edge — recall annotation should also check the forward edge on other engrams in the result set, not rely solely on `superseded_by`.
- Related: #181 (tension lifecycle) — resolving a tension by picking a winner is a natural place to write a supersedes edge.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

plur9:
**CTO triage** — this is a clean, well-scoped follow-up to #240. The edge data already exists; only the read-path consumers are missing.

**Priority: P2.** The three items are independent but naturally sequenced:

1. **Inject (priority in budget-constrained lists)** — moderate complexity; requires `inject.ts` to detect and sort supersedes chains before the budget window is applied. The "historical query" exception (keep old engram when the query is explicitly about historical state) can be heuristic (keyword match on "before", "was", "prior") in v1.
2. **Recall annotation** — simplest item; `plur_recall` / `plur_recall_hybrid` just need to append `[superseded by ENG-xxx]` to the display text. The reverse-edge caveat (remote-store targets may lack `superseded_by`) is handled by checking forward edges in the result set.
[...]

Issue #347, opened 2026-06-18
Populate temporal.valid_until on the write path — expiry enforcement exists, but nothing fills the field


plur9:
## Problem

The deterministic-expiry machinery already exists and works:

- `inject.ts:297-298, 329-330` skip any engram where `temporal.valid_until < today`
- `index.ts:1639-1640` filters expired engrams from recall
- `fts.ts:30-31` and `agentic-search.ts:48-49` index/surface the window

But **nothing populates `temporal.valid_until`**:

- `plur_learn` (MCP) and `core.learn()` do not expose `valid_from` / `valid_until` params, so agents never set them.
- No extractor lifts explicit expiry phrases ("valid until 31 May 2026", "expires", "through end of Q2") from the statement text into the structured field.

**Result:** time-bound facts are stored with the date only in free text and keep injecting at full strength after they go stale. Observed: an offer engram whose statement said "valid 31 May 2026" was injected on 2026-06-12 — 12 days past expiry — because `temporal.valid_until` was empty. (Proof: if the structured field had been set, `inject.ts:297` would have skipped it.)

## Proposed

1. Expose `valid_from` / `valid_until` as optional params in `plur_learn` (MCP) and `core.learn()`.
2. During ingest/learn, auto-detect explicit expiry phrases and parse them into `temporal.valid_until`, echoing the parsed ISO date back for confirmation (never silently guess).
3. Add a **soft-expiry** mode alongside the current hard-skip: optionally inject a recently-expired engram with an `⚠ EXPIRED <date> — verify before use` marker for a grace window, since some facts stay useful as history. Hard-skip remains the default.
4. Tests: hard-skip (exists), soft-expiry marker, extraction round-trip, `valid_from` not-yet-valid skip.

## Notes

No schema change required — `TemporalSchema.valid_until` (engram.ts:79) and `temporal` on the engram (engram.ts:248) already exist. This is purely write-path + extraction wiring. Idea borrowed from Loci crystal tiers (◈ contextual `valid_until`).

Issue #213, opened 2026-05-21
Audit: engram conflict resolution across detection, retrieval, lifecycle


plur9:
## Summary

Audit how PLUR currently handles conflicting engrams across the full lifecycle — detection, presentation, resolution — and identify gaps before features like #176 (confidence escalation on recurrence) compound the problem.

## Why now

#176 (confidence escalation on recurrence) will start auto-promoting engrams to `decided` and `locked`. If two engrams in contradiction both auto-promote, we'll end up with two pinned, contradictory beliefs and no clear path to resolution.

Today's tension detection (PR #172) finds contradictions, but what happens next isn't well documented:

- The `plur_tensions` MCP tool reports pairs of conflicting engrams
- The `plur_tensions_purge` tool clears the legacy `relations.conflicts` field
- The proposed #203 (tension-aware retrieval) asks "pick the winning side of a contradiction instead of injecting both"

Three tools, three different mental models. Need to converge before #176 ships.

### 1. Current detection surfaces

- BM25 pre-filter (in `tensions.ts`) — scope + domain + subject overlap, no LLM
- LLM judge stage (also in `tensions.ts`) — pairwise contradiction confidence ≥ 0.7
- Legacy `relations.conflicts` field (left over from removed `conflict.ts`, persisted on existing engrams)
- Dedup-time contradiction note (mentioned in the design comment for `dedup.ts` before PR #183 removed it)

Question: which of these are still active, which are dead code, which conflict with each other?

### 2. What happens to a flagged tension?

Today the answer appears to be "the user is told, manually decides". But:

- Is the tension persisted anywhere durable? Or recomputed on each `plur_tensions` call?
- If persisted, where? (`relations.conflicts` is for the legacy single-engram conflict pointer; tensions detected by the new LLM pipeline don't write back to engrams as far as I can tell)
- When the user retires one side, is the other side notified / decremented / marked resolved?
- What's the contract for "resolved" — does it survive `plur_batch_decay`?

### 3. Retrieval-side behavior

When injection happens and two engrams in known tension both score above the relevance threshold:

- Today: probably both get injected. The agent has to reconcile.
- #203 proposed: pick one side. But on what basis?
- Open question: should the agent see the tension *flag* as part of context ("these two contradict — be aware") or just one side ("here's the winner — proceed")?

### 4. Lifecycle of escalated engrams (the #176 entanglement)

If C from #176 ships, engrams escalate to `decided` / `locked` on recurrence. Audit gaps:

[...]

Issue #181, opened 2026-05-19
Tension detection Phase 3: lifecycle, persistence, and resolution


crtahlin:
## Context

Phase 1 tension detection (PR #172) produces high-quality scan results but is **read-only** -- results are returned to the caller and not persisted. Users must manually resolve contradictions by forgetting one engram. There is no way to dismiss false positives, confirm real tensions, or track resolution over time.

Phase 2 addresses candidate ranking and efficiency. This issue covers the UX layer.

## Problem

When a user runs `plur tensions --scan`, they get a list of contradictions but no way to act on them within the system. The workflow is:

1. See tension between engram A and B
2. Decide which is correct
3. Manually run `plur forget <wrong-id>`
4. Next scan will re-check the pair (since results are not persisted)

This does not scale. Users need to be able to dismiss false positives (so they are not re-flagged), confirm real tensions (so they surface during injection as warnings), and resolve them (pick a winner, retire the loser).

### Tension record schema

```yaml
tensions:
  - id: T-2026-0519-001
    engram_a: ENG-2026-0519-052
    engram_b: ENG-2026-0519-053
    confidence: 1.0
    reason: "A says JSON files, B says single YAML"
    detected_at: 2026-05-19T07:30:00Z
    status: detected        # detected | confirmed | dismissed | resolved
    resolved_by: null       # engram ID of the winner, or null
    resolved_at: null
    category: factual       # factual | temporal | superseded
```

### Status lifecycle

```
detected (auto, from scan)
  -> confirmed (user: yes, this is a real conflict)
  -> resolved  (user picked winner, loser auto-retired)

detected (auto, from scan)
  -> dismissed (user: not a real conflict, suppress this pair)
```

Dismissed pairs go to a suppress-list so future scans skip them.

### Tension categories

Most tensions from verification were **temporal evolution** (ceasefire agreed -> ceasefire failed), not factual errors. The system should distinguish:

| Category | Example | Resolution |
|----------|---------|------------|
| factual | "CLI is v0.8.2" vs "CLI is v0.3.0" | Pick correct one, retire other |
| temporal | "Ceasefire agreed Apr 7" vs "Ceasefire failed Apr 12" | Keep both, mark older as superseded with timestamp |
| superseded | "Use 5 agents" vs "Use 9 agents (not 5)" | Auto-detect "(not X)" pattern, retire older |

### MCP tool changes

```
plur_tensions          # list persisted tensions (detected, confirmed)
[...]

PR #482, opened 2026-07-03
feat(core): temporal-aware tension detection — contradiction vs evolution (#240)

### LINKED ISSUE #240: feat: temporal-aware tension detection — distinguish contradictions from evolution
## Problem

The tension detector currently treats all engrams as claims about **current truth**. When two engrams make conflicting statements, they are flagged as contradictions regardless of when they were recorded.

This produces massive false positive rates in domains where knowledge evolves over time. In a real scan of ~400 pairs, **22 out of 25 detected tensions** were false positives caused by temporal evolution, not actual contradictions.

### Example: War Analysis

```
ENG-2026-0408-001: "Iran ceasefire agreed April 7, Hormuz regulated passage"
ENG-2026-0413-002: "Islamabad peace talks collapsed, ceasefire expires April 22"
ENG-2026-0505-001: "War effectively resumed in the strait"
```

These are all correct at their respective timestamps. They tell a story: ceasefire → collapse → war. The tension detector flags every pair as confidence 0.9 contradiction.

### Example: Legal/Regulatory Evolution

Consider a different domain where the problem is subtler:

```
ENG-2025-01: "GDPR Article 22 prohibits fully automated decisions affecting individuals"
ENG-2026-06: "EU AI Act Article 14 requires human oversight for high-risk AI systems, superseding GDPR Article 22 scope for AI-specific decisions"
```

Here the old fact is **still true** (GDPR Article 22 still exists) but **no longer the governing rule** for a subset of decisions. An agent making a compliance decision should use the newer engram, but a researcher studying regulatory history needs both.

This is fundamentally different from two engrams that genuinely contradict each other on current state (e.g., "use tabs" vs "use spaces").

## Proposed Solution

A three-layer approach combining temporal metadata, domain-level defaults, and smarter scoring.

### Layer 1: Engram-level `temporality` field

Add an optional field to the engram schema:

```yaml
temporality: snapshot | standing | supersedes
```

| Value | Meaning | Tension behavior |
|-------|---------|-----------------|
| `standing` (default) | Claim about current truth — "use TypeScript not JavaScript" | Normal contradiction detection |
| `snapshot` | Point-in-time fact — "BTC at $75K on May 27" | Skip tension check against other snapshots with different timestamps |
| `supersedes` | Replaces an older standing fact — "EU AI Act supersedes GDPR Art 22 for AI decisions" | Link to superseded engram(s); old engram kept for history but deprioritized in injection |

The `supersedes` value handles the law/regulation case: the old engram remains searchable and recallable for historical queries, but the superseding engram is preferred when injecting context for **new decisions**.

### Layer 2: Domain-level temporal defaults

Allow domains to declare default temporality in configuration:

```yaml
temporal_domains:
  war-analysis: snapshot    # all engrams are point-in-time by default
  geopolitics: snapshot
  markets: snapshot
  crypto: snapshot
  plur.enterprise: standing  # default (explicit for clarity)
```

Engrams in temporal domains get `temporality: snapshot` automatically unless explicitly overridden. This works retroactively — no need to re-tag 300+ war-analysis engrams.

### Layer 3: Timestamp-aware confidence adjustment

When the LLM judge evaluates a candidate tension pair, apply a temporal discount to the confidence score:

```
adjusted_confidence = raw_confidence × temporal_factor

where temporal_factor:
  - same_day:        1.0  (likely real contradiction)
  - 1-3 days apart:  0.8  (could be correction or evolution)
  - 4-14 days apart: 0.5  (probably evolution)
  - 15+ days apart:  0.3  (almost certainly evolution)
  - both snapshots:  0.1  (event log, not contradiction)
```

This catches the grey area: two engrams recorded the same day about the same topic ARE likely a correction. Two engrams 6 weeks apart about a ceasefire are almost certainly a timeline.

The LLM judge prompt should also be updated to include timestamps and a nudge:

```
Engram A ({date_a}): "{statement_a}"
Engram B ({date_b}): "{statement_b}"

These engrams are {N} days apart. Consider whether the difference
reflects a genuine contradiction or temporal evolution of the same topic.
```

### How the three layers interact

```
Tension candidate pair arrives
  │
  ├─ Both marked `snapshot` (or both in temporal domain)?
  │    → Skip entirely (or set confidence floor at 0.1)
  │
  ├─ One is `supersedes` and references the other?
  │    → Not a tension — it's an intentional update. Skip.
  │
  ├─ Both are `standing` (or mixed standing/snapshot)?
  │    → Run LLM judge with timestamp context
  │    → Apply temporal discount to confidence
  │    → Only surface if adjusted_confidence >= min_confidence
  │
  └─ Result: far fewer false positives, historical record preserved
```

### Impact on injection and recall

The `supersedes` chain affects more than just tensions:

- **Injection (`plur inject`)**: When budget-constrained, prefer the latest engram in a supersedes chain. The older one is still injectable if the query is explicitly historical ("what was the law before the AI Act?").
- **Recall (`plur recall`)**: Return all engrams in the chain, but annotate: `[superseded by ENG-xxx]` on older ones.
- **Decay (`batch_decay`)**: Superseded engrams should decay faster unless they have high recall frequency (someone keeps looking them up for history).

### Schema changes

```typescript
interface Engram {
  // ... existing fields ...
  temporality?: 'standing' | 'snapshot' | 'supersedes'  // default: 'standing'
  supersedes?: string[]  // engram IDs this one replaces (only when temporality: 'supersedes')
}
```

The `plur_learn` tool already has a `supersedes` parameter — this proposal formalizes its semantics and connects it to tension detection and injection priority.

### Migration

- Existing engrams default to `standing` (no change in behavior for non-temporal domains)
- Domain-level defaults apply retroactively — configure `war-analysis: snapshot` and 321 engrams are instantly reclassified without touching the YAML
- No breaking changes to the engram format

## Context

Discovered during a tension scan on a store with 1,518 engrams and 2,159 candidate tension pairs. War-analysis (321 engrams) and geopolitics (100+ engrams) dominated false positives. After retiring just 2 stale root-cause engrams, ~22 tensions disappeared — but the underlying problem remains for any domain that tracks evolving events.

## Acceptance Criteria

- [ ] `temporality` field accepted by `plur_learn` and stored in engram
- [ ] Domain-level temporal defaults configurable
- [ ] Tension scanner skips snapshot-vs-snapshot pairs (or floors confidence)
- [ ] Tension scanner applies timestamp-aware confidence discount
- [ ] LLM judge prompt includes dates and temporal guidance
- [ ] `supersedes` chains respected in injection priority
- [ ] Recall annotates superseded engrams
- [ ] Existing engrams unaffected (default: `standing`)
- [ ] Documentation updated

plur9:
Closes #240

Implements temporal-aware tension detection along the lines @plur9 recommended in the issue thread: Layer 2 (domain defaults) as the primary mechanism, the Layer 3 prompt half by default, the Layer 3 multiplier as an opt-in tunable, and `supersedes` as a **relation (graph edge), not a temporality enum**.

### Layer 2 — domain-level temporal defaults (`tensions:` config block)

```yaml
# config.yaml
tensions:
  temporal_domains:        # domains whose engrams are point-in-time snapshots
    - war-analysis
    - geopolitics
    - markets
  snapshot_pairs: skip     # skip (default) | floor
  temporal_discount: false # Layer 3 multiplier, OFF by default
```

- A snapshot-vs-snapshot pair (both domains under a `temporal_domains` entry, dotted sub-domains included) recorded on **different days** is an event log, not a contradiction — `getCandidatePairs` drops it before the judge (`skip`) or the scanner caps its confidence at 0.1 (`floor`).
- Same-day snapshot pairs stay in: two reports of the same day ARE a likely correction.
- Retroactive: no re-tagging of existing engrams. Zero config → zero behavior change.

### Layer 3 prompt half — dated judge prompts (default ON)

- Every judged statement now carries its recorded date (`temporal.learned_at`, falling back to the date embedded in `ENG-YYYY-MMDD-NNN` / `ENG-{PREFIX}-YYYY-MMDD-NNN` / `ENG-YYYY-MM-DD-NNN` ids).
- Single-pair and batch prompts annotate both dates plus days-apart, with the evolution-vs-contradiction nudge. Prompts without derivable dates are byte-identical to before.
- Detected tensions report `days_apart`.

### Layer 3 multiplier — opt-in `temporal_discount`

- The `1.0 / 0.8 / 0.5 / 0.3` days-apart ladder multiplies judge confidence, **off by default** (config `tensions.temporal_discount`, CLI `--temporal-discount` / `--no-temporal-discount`, MCP `temporal_discount` arg). Kept opt-in per the review comment: a blanket multiplier can bury genuine standing-fact corrections made weeks apart. Discounted tensions report `raw_confidence` alongside the adjusted value.

### `supersedes` as a relation + `plur_learn` param

- `relations.supersedes[]` / `relations.superseded_by[]` edges (mirroring `relations.conflicts[]`), settable via `plur_learn` MCP tool (`supersedes: [ids]`), the `Plur.learn`/`learnRouted` context, and CLI `plur learn --supersedes id1,id2`.
- Learn writes the forward edge on the new engram and best-effort reverse edges on local targets (persisted under the same store lock; unknown/remote targets keep the forward edge only).
- The scanner skips supersedes-linked pairs in both directions — an intentional update is not a tension.
- Pack export strips `supersedes`/`superseded_by` alongside `conflicts`/`related` (store-internal cross-refs; a stale edge in a foreign store could falsely suppress detection on ID collision).

### Expired-validity gate (uses the existing #347 fields)

- A pair where either side's `temporal.valid_until` has passed is skipped — an expired engram is no longer a peer claim about *current* truth, so lining it up against a newer claim manufactures a false tension. This is the "PLUR already owns (b) machinery" point from the issue thread, wired in.

### Wiring

- MCP `plur_tensions` scan mode merges the `tensions:` config defaults; explicit `temporal_discount` arg overrides. Output includes `days_apart` / `raw_confidence` when present.
- CLI `plur tensions --scan` same merge + flags; text output shows days-apart and raw confidence.
- `spec/engram.schema.json` regenerated (drift gate green).

## Deliberately NOT in this PR (scope guard from the issue thread)

- **No `temporality` enum on the engram schema** — Layer 2 domain defaults cover the population; the per-engram marker can be added later as `temporal.kind` if measurement shows the long tail matters.
- **Inject/recall/decay supersedes behavior** (injection priority, `[superseded by …]` recall annotations, faster decay) — split to follow-up issue #481 as recommended.

## Tests

- `packages/core/test/tensions-temporal.test.ts` — 45 new tests: date resolution, days-apart, domain matching, discount ladder, snapshot/supersedes/expired candidate gates, dated prompts (byte-identical without dates), config schema, scan integration incl. floor mode and discount on/off.
- `packages/core/test/supersedes.test.ts` — 8 new tests: forward/reverse edges, multi-target, unknown targets, idempotent reverse edge, relation preservation, YAML round-trip, scanner skip end-to-end.
- `packages/mcp/test/tensions.test.ts` — 7 new tests: config wiring through the tool, discount arg + config default, `plur_learn supersedes` param, scan skip.
- `packages/cli/test/tensions.test.ts` — 6 new tests: config-driven skip, `--temporal-discount`/`--no-temporal-discount` override semantics, `--supersedes` flag, days-apart reporting (stub OpenAI-compatible server).

### DIFF
--- a/packages/cli/src/commands/learn.ts
+++ b/packages/cli/src/commands/learn.ts
@@ -29,6 +29,8 @@ export async function run(args: string[], flags: GlobalFlags): Promise<void> {
   let derivedFrom: string | undefined
   let knowledgeAnchors: Array<{ path: string; relevance?: string; snippet?: string }> | undefined
   let dualCoding: { example?: string; analogy?: string } | undefined
+  // #240: engram IDs this statement intentionally replaces (comma-separated).
+  let supersedes: string[] | undefined
 
   // Parse a value as JSON, exiting 1 with a clear message on malformed input
   // (a bad --dual-coding/--knowledge-anchors should fail loudly, not silently
@@ -62,6 +64,9 @@ export async function run(args: string[], flags: GlobalFlags): Promise<void> {
     else if (arg === '--dual-coding' && i + 1 < args.length) {
       dualCoding = parseJsonFlag('--dual-coding', args[++i]); i++
     }
+    else if (arg === '--supersedes' && i + 1 < args.length) {
+      supersedes = args[++i].split(',').map(t => t.trim()).filter(Boolean); i++
+    }
     else if (!statement) { statement = arg; i++ }
     else { i++ }
   }
@@ -76,7 +81,8 @@ export async function run(args: string[], flags: GlobalFlags): Promise<void> {
   if (!statement) {
     exit(1, 'Usage: plur learn <statement> [--scope <scope>] [--type <type>] [--domain <domain>] ' +
       '[--source <s>] [--rationale <r>] [--tags a,b,c] [--visibility private|public|template] ' +
-      '[--abstract <id>] [--derived-from <id>] [--knowledge-anchors <json>] [--dual-coding <json>]')
+      '[--abstract <id>] [--derived-from <id>] [--knowledge-anchors <json>] [--dual-coding <json>] ' +
+      '[--supersedes id1,id2]')
   }
 
   // Build the context conditionally: when --scope is absent, OMIT the scope key
@@ -99,6 +105,7 @@ export async function run(args: string[], flags: GlobalFlags): Promise<void> {
     ...(derivedFrom !== undefined ? { derived_from: derivedFrom } : {}),
     ...(knowledgeAnchors !== undefined ? { knowledge_anchors: knowledgeAnchors } : {}),
     ...(dualCoding !== undefined ? { dual_coding: dualCoding } : {}),
+    ...(supersedes !== undefined ? { supersedes } : {}),
   }
 
   // ALWAYS use learnRouted (not learn): learn() stamps _demoted but does NOT do
--- a/packages/cli/src/commands/tensions.ts
+++ b/packages/cli/src/commands/tensions.ts
@@ -43,6 +43,7 @@ function getLlmFunction(): LlmFunction | undefined {
  *   plur tensions --scan --min-confidence 0.8  # stricter threshold
  *   plur tensions --scan --max-pairs 100       # check more pairs
  *   plur tensions --scan --batch-size 1        # sequential single-pair judging (no batching)
+ *   plur tensions --scan --temporal-discount   # days-apart confidence discount (#240, default off)
  *   plur tensions --scan --model claude-haiku-... --llm-base-url ... --llm-api-key ...
  *   plur tensions                              # list legacy stored conflicts (usually empty)
  *   plur tensions --json                       # machine-readable output
@@ -54,6 +55,8 @@ export async function run(args: string[], flags: GlobalFlags): Promise<void> {
   let minConfidence = 0.7
   let maxPairs = 50
   let batchSize = 5
+  // #240: undefined → fall back to config (tensions.temporal_discount)
+  let temporalDiscount: boolean | undefined
   let llmBaseUrl: string | undefined
   let llmApiKey: string | undefined
   let llmModel: string | undefined
@@ -67,6 +70,8 @@ export async function run(args: string[], flags: GlobalFlags): Promise<void> {
     else if (arg === '--min-confidence' && i + 1 < args.length) { minConfidence = parseFloat(args[++i]); i++ }
     else if (arg === '--max-pairs' && i + 1 < args.length) { maxPairs = parseInt(args[++i], 10); i++ }
     else if (arg === '--batch-size' && i + 1 < args.length) { batchSize = parseInt(args[++i], 10); i++ }
+    else if (arg === '--temporal-discount') { temporalDiscount = true; i++ }
+    else if (arg === '--no-temporal-discount') { temporalDiscount = false; i++ }
     else if (arg === '--llm-base-url' && i + 1 < args.length) { llmBaseUrl = args[++i]; i++ }
     else if (arg === '--llm-api-key' && i + 1 < args.length) { llmApiKey = args[++i]; i++ }
     else if (arg === '--model' && i + 1 < args.length) { llmModel = args[++i]; i++ }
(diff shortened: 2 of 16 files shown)


plur9:
Reconciling with the [#213 conflict-resolution audit](https://github.com/plur-ai/plur/issues/213#issuecomment-4873884508) that landed while this PR was in flight:

- **The audit's C8 finding — "there is no `supersedes` relation anywhere in the codebase" — is correct for `main`, and this PR is written accordingly: it *introduces* the edge** (schema, `plur_learn` param, CLI flag, reverse-edge writes, scanner skip, spec regen) rather than assuming a pre-existing one. That is the design @plur9 recommended in the #240 thread ("model it as `relations.supersedes[]`/`superseded_by[]`, mirroring the `relations.conflicts[]` that already exists"), kept as a relation precisely to avoid the temporality-enum category error.
- **The evolution-vs-contradiction logic does not depend on the new edge.** The primary mechanisms are the temporal fields the audit points to: `temporal.learned_at` (dated judge prompts), `temporal.valid_until` (expired engrams are not peer claims about current truth — pairs skipped), and domain-level snapshot defaults. The supersedes skip only fires when a caller *explicitly* declares an intentional update.
- If review lands on the audit's open question ("consider whether an explicit `supersedes` relation is wanted at all", C8/item 8) with a *no*, the edge is cleanly severable: `RelationsSchema` fields + `LearnContext.supersedes` + the `supersedesLinked` gate + tests, with no other mechanism in this PR built on top of it. #481 (consumer-side behavior) would then be closed instead of implemented.
- Audit items C1 (importer suspects never LLM-judged), C5 (`contradiction_detected` never emitted), and the #176 lock-escalation gap are persistence/lifecycle scope — being addressed in the #181 branch, not here.

Issue #335, opened 2026-06-18
Sprint-0 integration (C): resolve the EMBED_DIM contract — variable-dim factory vs asserted-384


crtahlin:
Part of #332. The semantic conflict a textual merge can't fix (flagged CRITICAL #3 in the #276 review). **Needs a design decision.**

- `main` (#289/#290): `EMBED_DIM = 384` is an **asserted, stable public contract** — `embed()` asserts every output against it; documented as a breaking-change boundary for any consumer that persists vectors.
- The epic: `embed()` routes through the embedder **factory**, so output dim is **variable per `PLUR_EMBEDDER`** (bge-small→384, bge-base/embedding-gemma→768), and the adapter path returns `adapter.embed(text)` directly, **bypassing `assertEmbedDim`**.
- **Post-merge hazard:** with any non-default embedder, `embed()` returns 768-dim vectors while the exported `EMBED_DIM` still says `384` and the assertion is dead for the default path — a #290-style external consumer sizing a column to `EMBED_DIM` breaks silently. (The PGLite constructor in sub-B already reads `getEmbedder(resolveEmbedderName()).dim`, so the internal path is dim-aware; the **exported constant** is the gap.)

### Options
1. **Derive the exported dim from the active adapter** — replace the `EMBED_DIM` literal with `getEmbedder().dim` (or expose `activeEmbedderDim()`), so the public value tracks reality. Keeps #290's intent (no silent drift) under variable embedders. *(Prior-review lean.)*
2. **Scope the fixed-384 contract to the default embedder** — keep `EMBED_DIM = 384` as "the default embedder's dim," document that non-default embedders change it, and keep `assertEmbedDim` on the default path only.

Also: re-enforce dim-correctness for the adapter path at the storage boundary (`vectorLiteral` + `plur doctor` dim-check) so a wrong-dim vector can't be persisted regardless of which option is chosen.

crtahlin:
Re-scoped under the unbundle plan (#332): this EMBED_DIM decision now lands as part of **PR-3** (embedder adapters + factory) on current `main`, decided + reviewed in that PR rather than as merge conflict-resolution. The options + hazard analysis above stand.

plur9:
## Resolved in PR #488 — decision: option 2 + derived accessor, now wired end-to-end

[...]

Issue #332, opened 2026-06-18
Land Sprint 0 substrate (#276) as 4 focused PRs on current main (unbundle the epic)


crtahlin:
#276 bundled 5 feature PRs + 2 audit-fix PRs + 4 audit iterations into one epic: +189k lines (mostly committed benchmark artifacts), 22 commits behind `main`, a 9-file conflict, no CI on the branch, and 49 AI-attribution commits. Rather than resurrect that mega-merge, **unbundle it** — re-land the underlying features as focused PRs on current `main`. Each is small, independently CI-green, independently reviewable, and the two real *design decisions* get proper review instead of being buried in conflict resolution.

#276 is retired as the merge vehicle (its content stays on `epic/sprint-0-substrate` for cherry-picking). The old conflict-resolution sub-issues are superseded: **A (#333) dissolves** (it was merge mechanics that vanish on re-land); **C (#335)** feeds PR-3; **B (#334)** feeds PR-4.

## The 4 PRs (land in order; each rebased on current `main` + CI-green)

- [ ] **PR-1 — PGLite substrate + YAML-as-truth tests** (from #252 / #250, #226 / #249). The backend, opt-in via `PLUR_BACKEND`, with Test A (nuke-rebuild) + Test B (traceability) over both backends. Low-risk substrate.
- [ ] **PR-2 — benchmark harness, code-only** (from #251 / #46). `benchmark/*.ts` only — **no committed data/results**; reads fixtures from / writes results to a configurable path. Folds in #336 (gitignore `benchmark/data/` + `benchmark/results/`).
- [ ] **PR-3 — embedder adapters + factory + EMBED_DIM decision** (from #253, #259 / #219, and **#335**). The 4 ONNX adapters + `PLUR_EMBEDDER`, landed together with the `EMBED_DIM` contract resolution (derive exported dim from the active adapter, or scope fixed-384 to the default — decide here, see #335).
- [ ] **PR-4 — recall wiring** (from #220, #224, and **#334**). PGLite-into-recall + cross-encoder reranker + intent routing, reconciled against `main`'s recall path (`rrfMerge`/`topScore`/`emitMissSignal`) as fresh code review, not conflict archaeology.

## Carried-over (fold into the PRs / final cut)
- AI-attribution trailers: re-land is authored clean — problem disappears (no squash-scrub needed).
- `pnpm test:smoke` with `PLUR_REMOTE_TEST_TOKEN` before any publish.
[...]

PR #488, opened 2026-07-03
fix(core): complete the EMBED_DIM contract — active-dim column sizing + storage-boundary enforcement (#335)

plur9:
Refs #335 (Sprint-0 integration C, part of #332).

## The decision

**Option 2 + derived accessor** — `EMBED_DIM = 384` stays exported strictly as *the default embedder's dim* (back-compat for #289/#290 consumers), and `activeEmbedderDim()` is the source of truth for what this install actually produces. Main already had the constant's documentation, the accessor, and the `embed()` declared-vs-produced assertion; **what it did not have is the wiring that makes variable dims actually work, or the storage boundary the issue calls for.** This PR completes both.

## Audit finding (the real bug)

The issue text assumed "the PGLite constructor in sub-B already reads `getEmbedder(resolveEmbedderName()).dim`" — on current main it does **not** (likely lost in the post-scrub re-land). `Plur`'s constructor built `PGLiteAdapter` without `vectorDim`, so a **fresh PGLite store under `PLUR_EMBEDDER=bge-base|embedding-gemma|openai-3-large` got a `vector(384)` column** and every 768/3072-dim upsert failed against pgvector; the auto-embed path then skipped everything.

## Changes

1. **Active-dim column sizing** — `Plur` passes `vectorDim: getEmbedder(resolveEmbedderName()).dim` to `PGLiteAdapter`. Metadata-only (adapters construct lazily; no model load, no API call). Existing stores keep their on-disk column: `ensureColumnPrecision` reads reality and mismatches surface via the existing doctor dim-check.
2. **Storage-boundary enforcement** — the adapter now tracks the on-disk column dim at init (`activeVecDim`) and `upsertEmbedding` **refuses wrong-shape vectors** with a targeted #335 error naming both dims and the `plur sync --reembed --full` remediation. On the pgvector path this converts a cryptic pgvector error; on the **BYTEA fallback it is the only guard** — previously any length was silently persisted, and `cosineSimilarity` later returned garbage over `min(a, b)` (the exact #290-style silent drift). The auto-embed path already skips dim mismatches and catches all errors, so `learn()` cannot crash on this.
3. **Test hook** — `_setCachedEmbedder` in `embeddings.ts` (mirrors rerankers' `_setCachedReranker`), enabling the first tests of the `embed()` dim assertion without a model download.

## Cross-shape equivalence tests (ADR-0001) — `embed-dim-contract.test.ts`, 9 tests

- The three contract shapes hold: `bge-small` 384 (= `EMBED_DIM`, = `DEFAULT_EMBEDDER`), `embedding-gemma` 768, `openai-3-large` 3072 — factory metadata, no model loads.
- `embed()` returns the vector when the adapter honors its declared dim; **throws** (never null / never persists) when the model disagrees with the declaration.
- Parameterized over 384/768/3072: PGLite column sized to the shape, identical-vector roundtrip via `searchVector` (score > 0.99), wrong-shape write rejected with **our** guard's message (not pgvector's), rejected write doesn't clobber good vectors, and `reindex()` rebuilds rows from YAML at the same shape (YAML-as-truth rebuildability invariant).
- `Plur` constructor sizing: default install → 384 column; `PLUR_EMBEDDER=bge-base` → 768 column.

## Tests

Full core suite on this branch: **1772 passed / 31 skipped / 0 failed**. mcp 172 passed, cli 238 passed (one flaky cli failure under load did not reproduce on rerun).

### DIFF
--- a/packages/core/src/embeddings.ts
+++ b/packages/core/src/embeddings.ts
@@ -122,6 +122,24 @@ export function resetEmbedder(): void {
   embedPipeline = null
 }
 
+/**
+ * Test-only: install a stub adapter as the active embedder so tests can
+ * exercise the embed()/activeEmbedderDim() contracts (#335) without a
+ * model load. Mirrors rerankers' `_setCachedReranker`. Production code
+ * never calls this; pair with `resetEmbedder()` in afterEach.
+ */
+export function _setCachedEmbedder(adapter: {
+  name: string
+  dim: number
+  modelId: string
+  embed(text: string): Promise<Float32Array>
+  embedBatch(texts: string[]): Promise<Float32Array[]>
+}): void {
+  embedPipeline = adapter
+  transformersUnavailable = false
+  lastLoadError = null
+}
+
 async function getEmbedder() {
   if (embeddingsDisabled) return null
   if (embedPipeline) return embedPipeline
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -375,6 +375,13 @@ export class Plur {
       // vector.precision (#223): unset = keep the store's existing column
       // type; 'halfvec' opts in to fp16 storage (lazy in-place migration).
       this.pgliteAdapter = new PGLiteAdapter(this.paths.engrams, this.paths.pglite, {
+        // #335: size the vector column from the ACTIVE embedder (PLUR_EMBEDDER),
+        // not the 384 default constant — bge-base/embedding-gemma are 768,
+        // openai-3-large is 3072. Metadata-only: adapters construct lazily,
+        // no model load happens here. Existing stores keep their on-disk
+        // column (ensureColumnPrecision reads reality); mismatches surface
+        // via the doctor dim-check + the upsert-time guard.
+        vectorDim: getEmbedder(resolveEmbedderName()).dim,
         precision: this.config.vector?.precision,
       })
       // Initial sync runs in the background — YAML is already authoritative,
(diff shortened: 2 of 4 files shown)


plur9:
**CTO Review — LGTM ✅**

This closes Sprint-0 integration (C) — the EMBED_DIM contract gap that made non-384 embedders silently produce wrong-sized vectors against a fixed `vector(384)` column.

Key observations:
- **Root cause well-diagnosed**: `Plur` constructor was building `PGLiteAdapter` without `vectorDim`, meaning BGE-base/embedding-gemma/openai-3-large all got a `vector(384)` column and every upsert failed silently. The fix (pass `vectorDim: getEmbedder(resolveEmbedderName()).dim`) is surgical.
- **Back-compat preservation**: `EMBED_DIM = 384` stays exported for existing consumers (#289/#290), `activeEmbedderDim()` becomes the runtime source of truth. Good layering.
- **Storage-boundary enforcement**: rejecting wrong-shape vectors at the adapter boundary is the right place — surfaces as a hard error (#335 diagnostic) rather than a silent pgvector failure.
- **`ensureColumnPrecision` path**: existing stores keep their on-disk column; mismatches surface via the existing doctor dim-check. No silent migration.
- **`_setCachedEmbedder` test hook**: correct pattern for injecting a non-default embedder in tests without loading a real model.
- Cross-shape equivalence tests in `embed-dim-contract.test.ts` cover the key failure mode.

No blocking concerns. Ready for merge once approved.

Issue #217, opened 2026-05-22
Upstream: request SessionEnd hook event from Claude Code


crtahlin:
### Request

File a feature request with Anthropic / Claude Code team for a `PreExit` or `SessionEnd` hook event that fires when:
- User runs /exit
- User closes the terminal/window
- Session times out

### Why

Currently, Claude Code hook events cover:
- SessionStart, UserPromptSubmit, PreToolUse, PostToolUse, PostCompact, SubagentStart, Stop

But there's no event for session end. This means cleanup logic (session summaries, memory persistence, temporary file cleanup) cannot run reliably.

### What we'd use it for

```json
{
  "hooks": {
    "PreExit": [{
      "hooks": [{
        "type": "command",
        "command": "~/.plur/bin/plur-hook hook-session-end-auto",
        "timeout": 10
      }]
    }]
  }
}
```

The hook would call `plur_session_end` with a summary extracted from the session context, ensuring engrams are persisted.

### Workarounds in place

Until this exists, we use:
1. Stop hook checkpoints (periodic session state saves)
2. Deferred wrap-up on next session start (retroactive processing from observations)
3. Enterprise server-side timeout (for HTTP/SSE transport)

These are imperfect — an actual exit hook would be the clean solution.

### Where to file

- Claude Code GitHub: https://github.com/anthropics/claude-code/issues
- Or via Anthropic support channels

plur9:
## Verdict: no upstream request needed — Claude Code has shipped `SessionEnd`

Before drafting the upstream feature request this issue asks for, I checked the current Claude Code docs. **The `SessionEnd` hook event now exists** (shipped in Claude Code v1.0.85 per the [changelog](https://github.com/anthropics/claude-code/blob/main/CHANGELOG.md), documented at [code.claude.com/docs/en/hooks](https://code.claude.com/docs/en/hooks)). The feature request this issue was going to file has been implemented upstream. Recommendation: **close #217 as overtaken by events** and track adoption instead (steps below).

### What shipped (verified against the live hooks docs, 2026-07-02)

- **Fires:** "When a session terminates."
- **Matchers** filter on *why* the session ended: `clear`, `resume`, `logout`, `prompt_input_exit`, `bypass_permissions_disabled`, `other`.
- **Input** (stdin JSON): the common hook fields — `session_id`, `transcript_path`, `cwd`, `hook_event_name`, `permission_mode`.
[...]

Issue #475, opened 2026-07-02
Adopt Claude Code's SessionEnd hook — auto-run plur_session_end on session termination


plur9:
Follow-up from #217 (upstream request — overtaken; Claude Code shipped SessionEnd in v1.0.85).

## What
Wire `plur_session_end` to the SessionEnd hook so sessions close their memory lifecycle automatically instead of relying on the agent remembering to call it.

## Plan (from the #217 analysis comment)
- [ ] New `hook-session-end-auto` command in `packages/cli/src/commands/` (stdin: session_id / transcript_path / cwd)
- [ ] Register in `buildEnforcementHooks()` in `init.ts`
- [ ] Decide `resume` matcher behavior (a resumed session is not an ended one)
- [ ] `plur init` / `plur doctor` upgrade path for existing installs
- [ ] Keep existing workarounds for the gaps SessionEnd can't cover: hard kills are un-hookable (keep hook-inject orphan recovery); no model access at hook time (derive summary from transcript_path); non-Claude-Code transports still need enterprise#112 server-side timeout

Refs #217.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_016aLoB2xRrqrqtXuqpt2Cf1

plur9:
**CTO triage**

Root cause found: `hook-session-end-auto.ts` was implemented and the implementation is complete (handles resume behavior, hard-kill gaps, best-effort capture), but the command was never staged or registered in `buildEnforcementHooks()` in `init.ts`.

Fix is small: add a `SessionEnd` entry to `buildEnforcementHooks()` (6 lines) + add the untracked file to git + update init tests.

PR: #492 — ready for review, closes this issue.

plur9:
Shipped in PR #496 (merged 2026-07-07). Claude Code v1.0.85+ SessionEnd hook is now wired — `plur_session_end` runs automatically when a Claude Code session closes. No manual call required.

Issue #216, opened 2026-05-22
Deferred wrap-up: detect and process orphaned sessions on next start


crtahlin:
### Problem

When a session exits without `plur_session_end`, the session-end summary and wrap-up engrams are lost. This sub-issue recovers them retroactively on the next session start.

### Implementation

In `hook-inject` (UserPromptSubmit hook, fires on first message of new session) or in `plur_session_start` tool:

1. **Detect orphaned sessions**: scan `~/.plur/sessions/*.checkpoint.json`
   - Orphaned = checkpoint exists without a matching episode (check episodes.yaml for session_id)
   - Skip if checkpoint is < 5 min old (session may still be active in another window)

2. **Process each orphaned session**:
   a. Read observations from the checkpoint's observation_file
   b. Filter by date range (started_at → last_checkpoint)
   c. Extract patterns:
      - Tools used (frequency, types)
      - Files modified (from Bash/Edit/Write observations)
      - Domains touched (from cwd paths)
   d. Generate conservative engrams:
      - "Session worked on {project} using {tools}" (procedural)
      - Any tool-call patterns that repeat (architectural)
      - Do NOT generate speculative content summaries
   e. Write deferred episode: summary + session_id + "deferred: true" flag
   f. Delete checkpoint file

3. **Report to user** via injection context:
   ```
   [PLUR] Previous session (May 22, 47 tool calls, 2h 30m) ended without wrap-up.
   Recovered 2 patterns from observations. Run plur_recall to see them.
   ```

### Design decisions

- **Conservative over comprehensive**: better to miss a pattern than hallucinate one. Observations give us tool names and file paths, not conversation content.
- **No LLM call**: deferred processing uses heuristics, not LLM inference. LLM-quality wrap-up requires the AI reviewing its own conversation — that context is gone.
- **Idempotent**: if the same checkpoint is processed twice (race condition), episodes deduplicate by session_id.

### Tests

- Orphaned checkpoint detected on session start
- Observations processed correctly (tool frequency, file list)
- Episode written with deferred flag
- Checkpoint cleaned up after processing
- Active session checkpoint (< 5 min old) skipped
- No crash if observations file is missing or corrupt

PR #496, opened 2026-07-04
feat(cli,mcp): wire plur_session_end to Claude Code SessionEnd hook (#217)

plur9:
## Summary

Claude Code shipped the **SessionEnd** hook in v1.0.85. This wires it up so PLUR's memory lifecycle **auto-closes** when a session ends, instead of depending on the agent remembering to call `plur_session_end`. Follow-up to #217, opened before upstream shipped the hook.

## What it does

A new internal CLI command `plur hook-session-end` runs on Claude Code's `SessionEnd` event:

1. Silent pass-through if plur isn't configured in the project (#247).
2. Locates this session's checkpoint (written by `hook-learn-check`, #215). **Its presence means `plur_session_end` was never called** — the tool unlinks the checkpoint on a clean close.
3. Captures a durable **closing episode** from the checkpoint metadata (duration, response count, cwd, end reason), tagged `session-end` / `auto-close`.
4. Cleans up the checkpoint so the next `session_start`'s deferred wrap-up (#216) doesn't double-report it.

This closes a real gap: previously, a session that forgot to wrap up left an orphaned checkpoint that the **next** `session_start` would only mention in a transient notice (#216) — no durable record. Now the close happens at the right time, with the right cwd, and produces an episode.

Conservative by design: no LLM runs in a shell hook, so this is a **safety net**, not a replacement for an explicit `plur_session_end` with `engram_suggestions` (which still produces higher-quality learnings).

## Changes

- **`packages/cli/src/commands/hook-session-end.ts`** — new SessionEnd hook command.
- **`packages/cli/src/index.ts`** — register `hook-session-end` in the command dispatcher + help.
- **`packages/cli/src/commands/init.ts`** — `plur init` scaffolds `SessionEnd` into the global enforcement hooks; updated init summary output.
- **`packages/mcp/src/index.ts`** — `@plur-ai/mcp init` scaffolds `SessionEnd` into `PLUR_HOOKS`; installed-CLAUDE.md note.
- **`README.md`** — documents the auto-close behavior.
- **Tests** — `hook-session-end.test.ts` (6 cases: captures episode + cleans checkpoint, CLAUDE_SESSION_ID fallback, no-op without checkpoint, corrupt-checkpoint handling, unconfigured pass-through, no double-capture) + `init.test.ts` now asserts `SessionEnd` is scaffolded globally and not duplicated at project scope.

## Verification

- `@plur-ai/cli` — 260 passed (32 files), incl. the 6 new SessionEnd cases and the augmented init test.
- `@plur-ai/mcp` — 193 passed (16 files); `PLUR_HOOKS` change breaks nothing.

## Notes on scope

The task brief referenced `packages/claw/src/hooks.ts` and `@plur-ai/claw`, but Claude Code hooks are not scaffolded by `claw` (that's the OpenClaw ContextEngine plugin). The Claude Code hook wiring lives in `@plur-ai/mcp` (`PLUR_HOOKS`) and `@plur-ai/cli` (`init` + the `hook-*` handlers), so the SessionEnd hook is implemented there.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ One command sets up everything — storage, MCP config, and Claude Code hooks:
 npx @plur-ai/mcp init
 ```
 
-This creates `~/.plur/` for storage, adds PLUR to your `.mcp.json`, and installs Claude Code hooks for automatic engram injection. PLUR is installed **globally** — one MCP server, one store, available in every project. You only run init once.
+This creates `~/.plur/` for storage, adds PLUR to your `.mcp.json`, and installs Claude Code hooks for automatic engram injection. The hooks also **auto-close the memory lifecycle**: a `SessionEnd` hook captures a closing episode and cleans up session state when a conversation ends, so memory closes cleanly even if the agent forgets to call `plur_session_end`. PLUR is installed **globally** — one MCP server, one store, available in every project. You only run init once.
 
 For **multi-project setups**, use domain/scope to separate knowledge:
 
--- a/packages/cli/src/commands/hook-session-end.ts
+++ b/packages/cli/src/commands/hook-session-end.ts
@@ -1,0 +1,147 @@
+import { readSync, readFileSync, existsSync, unlinkSync, mkdirSync } from 'fs'
+import { join } from 'path'
+import { homedir } from 'os'
+import { type GlobalFlags } from '../plur.js'
+import { createPlur } from '../plur.js'
+import { isPlurConfigured } from '../lib/plur-configured.js'
+
+/**
+ * plur hook-session-end — Claude Code SessionEnd hook (shipped v1.0.85).
+ *
+ * Auto-closes the PLUR memory lifecycle when a session ends, instead of
+ * relying on the agent remembering to call plur_session_end (#217).
+ *
+ * How it relates to the rest of the lifecycle:
+ * - hook-learn-check writes periodic session checkpoints (#215).
+ * - plur_session_end (agent-called) is the primary close: it creates engrams
+ *   from LLM-reviewed suggestions, captures an episode, AND unlinks the
+ *   checkpoint.
+ * - If the agent never calls plur_session_end, the checkpoint is left behind.
+ *   Previously the ONLY recovery was hook-inject's deferred wrap-up (#216) on
+ *   the NEXT session_start, which emits a transient notice but captures no
+ *   durable episode.
+ *
+ * This hook closes that gap: fired by Claude Code at the moment the session
+ * ends, it captures a durable closing episode from the checkpoint metadata and
+ * cleans up the checkpoint — at the right time, with the right cwd. It cannot
+ * produce LLM-quality engram suggestions (no model runs in a shell hook), so it
+ * is a conservative safety net, not a replacement for plur_session_end.
+ *
+ * If the checkpoint is absent, plur_session_end already ran (clean close) or
+ * the session was too short to checkpoint — either way there is nothing to do.
+ *
+ * Input: JSON on stdin (Claude Code SessionEnd hook format: session_id, cwd,
+ *        reason). Never blocks; SessionEnd output is advisory.
+ */
+
+function sessionKeys(payloadSessionId?: string): string[] {
+  // Mirror plur_session_end's key resolution (tools.ts): payload session_id
+  // first, then CLAUDE_SESSION_ID, then ppid — sanitized the same way as
+  // hook-learn-check writes them.
+  return [payloadSessionId, process.env.CLAUDE_SESSION_ID, String(process.ppid)]
+    .filter(Boolean)
+    .map(k => k!.replace(/[^a-zA-Z0-9_-]/g, '').slice(0, 64))
+    .filter(Boolean)
+}
+
+function plurPath(flags: GlobalFlags): string {
+  return flags.path ?? process.env.PLUR_PATH ?? join(homedir(), '.plur')
+}
+
+function readStdinRaw(): string {
+  try {
+    const chunks: Buffer[] = []
+    const buf = Buffer.alloc(65536)
+    while (true) {
+      try {
+        const n = readSync(0, buf, 0, buf.length, null)
+        if (n === 0) break
+        chunks.push(Buffer.from(buf.subarray(0, n)))
+      } catch {
+        break
+      }
+    }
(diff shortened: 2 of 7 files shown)


plur9:
**Relationship to #492**

PR #492 wired the existing `hook-session-end-auto` command into `plur init`. Both PRs implement the same SessionEnd hook feature, but this PR's design is superior:
- Checkpoint-based detection correctly identifies sessions that missed `plur_session_end` (no checkpoint = clean close, no work needed)
- Session key resolution mirrors `plur_session_end` in `tools.ts` exactly
- Covers both `@plur-ai/cli` and `@plur-ai/mcp` init paths
- README documents the auto-close behavior

Recommend **closing PR #492** and merging this one. This also closes #475 (the upstream feature adoption task).

Title references closed issue #217 (the upstream hook request) — consider updating to reference #475 for clarity.

plur9 (commented):
CI passing. The gap this closes is real: if an agent forgets plur_session_end the session checkpoint would linger indefinitely (until the deferred wrap-up on next session start), losing any durability for that closing window. The SessionEnd hook fires at the right moment with the right cwd.

Implementation is correct: isPlurConfigured() guard for pass-through in unrelated projects, session key resolution mirrors tools.ts, stdin reading with readSync is the right approach for a synchronous hook, and the capture() call is wrapped in try/catch so SessionEnd never fails the session. Checkpoint unlink after capture prevents double-reporting from the deferred wrap-up on next start.

The README addition correctly frames this as a safety net ("auto-close the lifecycle") while preserving the expectation that calling plur_session_end directly produces higher-quality learnings. That's the right mental model to set.

5s timeout on the hook is generous for a file stat + capture; probably fine, but worth monitoring if users report SessionEnd latency.

@plur9 ready for merge.

plur9 (commented):

**Closes the lifecycle gap correctly.** Before this PR, a session that never called `plur_session_end` left its checkpoint behind; the only recovery was the deferred wrap-up on the *next* session_start (which emitted a transient notice but captured no durable episode). This hook fires at the right moment — when the session actually ends — and captures a durable episode with the available metadata.

**Key correctness points:**
- `readStdinRaw()` uses synchronous `readSync` — correct for a hook process; avoids the async event loop that caused #502 / #504.
- `sessionKeys()` mirrors `plur_session_end`'s key resolution (payload → CLAUDE_SESSION_ID → ppid). Consistent fallback chain.
- Checkpoint removal before episode capture would be risky (lose the checkpoint if capture fails), but the order here is correct: capture first, then unlink. On a capture error, the checkpoint survives and the deferred wrap-up on the next session will catch it — fail-open.
- Corrupt checkpoint: unlinkSync + return immediately — right behavior. No episode from garbage data.
- 5s timeout in `init.ts` / `mcp/src/index.ts` is appropriate; SessionEnd is advisory and the hook does no LLM call.

**`init.ts` output string** updated from "3" to "4" enforcement hooks — catches the count before users do.

**Integration tests** cover all 6 failure modes: happy path, CLAUDE_SESSION_ID fallback, no-op on clean close, corrupt checkpoint, not-configured pass-through, idempotency. This is the right test surface for a hook.

**README update** is accurate: calling `plur_session_end` explicitly is still preferred (captures LLM-quality engram suggestions); the hook is a safety net. The copy correctly frames this distinction.

CI green (3/3 checks pass). ✅

plur9 (commented):
## Ops review — hook-session-end (#217)

**Verdict: approve.** Logic is correct, safety margins are right.

**Key invariants verified:**

1. **Lifecycle contract held** — checkpoint presence ↔  never ran. The hook's entry condition is correct; removing the checkpoint on close prevents the next 's deferred wrap-up (#216) from double-reporting.

2. **No LLM in shell hook** — episode is metadata-only (duration, stop_count, cwd). Conservative is correct here; the hook is a safety net, not a replacement for the agent-called close.

3. **Double-capture safe** — first run removes checkpoint; second run (e.g. both CLI-init and MCP-scaffolded hooks firing) finds no checkpoint and exits cleanly. Test case confirms this.

4. **Timeout 5s** — appropriate. No model calls, no network. Should complete in <100ms on any checkpoint that exists.

5. **isPlurConfigured() guard** — silent pass-through for unconfigured projects. Global hook install won't pollute unrelated workspaces.

6. **Corrupt checkpoint** — unlinks and returns, no crash. Correct.

**Minor observations (not blocking):**

- `readSync(0, ...)` synchronous stdin read: works in Claude Code's hook invocation model (stdin is always provided), but if ever run interactively it would hang waiting for EOF. Not a bug in practice.
- `process.ppid` fallback in `sessionKeys` mirrors `tools.ts` — consistent, just worth noting that ppid-keyed checkpoints will mismatch if the CLI is invoked via a shell wrapper that changes ppid. Low risk given the hook is invoked by Claude Code directly.

CI green. Ship when ready.

🤖 ops-heartbeat

plur9 (commented):
## Ops review — hook-session-end (#217)

**Verdict: approve.** Logic is correct, safety margins are right.

**Key invariants verified:**

1. **Lifecycle contract held** — checkpoint presence ↔ `plur_session_end` never ran. The hook's entry condition is correct; removing the checkpoint on close prevents the next `session_start`'s deferred wrap-up (#216) from double-reporting.

2. **No LLM in shell hook** — episode is metadata-only (duration, stop_count, cwd). Conservative is correct here; the hook is a safety net, not a replacement for the agent-called close.

3. **Double-capture safe** — first run removes checkpoint; second run (e.g. both CLI-init and MCP-scaffolded hooks firing) finds no checkpoint and exits cleanly. Test case confirms this.

4. **Timeout 5s** — appropriate. No model calls, no network. Should complete in <100ms on any checkpoint that exists.

5. **isPlurConfigured() guard** — silent pass-through for unconfigured projects. Global hook install won't pollute unrelated workspaces.

6. **Corrupt checkpoint** — unlinks and returns, no crash. Correct.

**Minor observations (not blocking):**

- `readSync(0, ...)` synchronous stdin read: works in Claude Code's hook invocation model (stdin is always provided), but if ever run interactively it would hang waiting for EOF. Not a bug in practice.
- `process.ppid` fallback in `sessionKeys` mirrors `tools.ts` — consistent, just worth noting that ppid-keyed checkpoints will mismatch if the CLI is invoked via a shell wrapper that changes ppid. Low risk given the hook is invoked by Claude Code directly.

CI green. Ship when ready.

🤖 ops-heartbeat

PR #479, opened 2026-07-03
feat(core,mcp,cli): per-store reranker eval gate — rerank-vs-RRF self-check (#451)

plur9:
Refs #451 — the last open task (per-store eval gate before any default-ON reranker recommendation).

## What

Before `PLUR_RERANKER` is ever recommended as a default, each store gets a self-check proving the cross-encoder is **not actively harmful on its own domain** (cross-encoders can be net-negative out-of-domain — the motivating finding in #451).

- **`packages/core/src/reranker-eval.ts`** — `runRerankerSelfEval()`: sample the store's own engrams (seeded, deterministic), synthesize a keyword probe query from each statement (the source engram is the known-relevant document), retrieve the RRF top-K with reranker OFF, re-order the same pool with the cross-encoder, compare MRR / Hit@1 / promotions vs demotions. Verdict: `beneficial` / `neutral` / `harmful` / `insufficient-data` (ΔMRR ±0.05 thresholds, ≥5 scored probes required).
- **Cached per store** (`<store>/.reranker-eval.json`), staleness bound = 7 days **or** >20% engram-count drift — either invalidates the verdict.
- **Advisory, never silent auto-disable**: `plur_doctor` gains a `reranker per-store eval` check (pass `rerank_eval:true` for a fresh run; a cached `harmful` verdict surfaces on every later doctor run), and the reranker-enable path logs a **loud-once** warning when the cached verdict is harmful. Reranking stays enabled; the human decides.
- **CLI**: `plur rerank-eval [--reranker <name>] [--sample N] [--seed N] [--force]` — human-readable + `--json`, exits 1 on a harmful verdict so scripts can gate on it.

By construction the gate is conservative: probe queries are keyword extracts, so BM25 already ranks the source well — the gate has limited power to prove *benefit* (the #46 harness does that); its job is to prove **not-harmful**, which is exactly what an enable-gate needs.

## Relationship to #477

PR #477 (`feat/reranker-fit-check`) landed a lighter probe for the same task 5: score-distribution separability on same-domain vs cross-domain engram *pairs*. This PR measures the operational quantity directly — does reranking demote known-relevant documents in actual retrieval (rerank vs RRF ΔMRR) — and adds the per-store cache + staleness bound + enable-path advisory + CLI. The two touch the same files (`core/src/index.ts`, `mcp/src/tools.ts`) and will conflict textually; they answer the same question at different depths, so merging **one** of them is the right call. Comparison posted on #451 for the decision.

## Tests

[...]

Issue #452, opened 2026-07-02
Log co-injection/co-fire + injection-outcome events in history.ts (unblocks #200/#202); ADR-0002 for derived-state provenance


plur9:
From the 2026-07-02 graph assessment (#225): the flagship `co-fires-with` edge in #200/#201 has **no data source**. `history.ts` logs lifecycle events (created/updated/feedback/contradiction) but nothing records *which engrams were injected together* or *what happened after an injection*. Every offline stage of the two-stage design is blocked on data that is not being collected — and history only compounds from the day logging starts. This is the time-sensitive piece of the graph program (hours of work).

## Tasks
- [ ] `history.ts`: log a `co_injection` event per session-start/inject — engram IDs injected together, query context hash, timestamp
- [ ] Log `injection_outcome` — engram ID + feedback verdict (positive/negative/ignored) linked to the injection event, shaped for temporal-replay self-labeling (#202)
- [ ] Keep events append-only in `history/YYYY-MM.jsonl`; measure size overhead
- [ ] **ADR-0002**: derived-state provenance — are history JSONLs source-of-truth for derived edges (co-fire counts), or must derived edges be rebuildable from engram YAML alone? ADR-0001's YAML-as-truth invariant doesn't answer this
- [ ] `plur_status`: surface label/event counts (feeds #202's volume gate)

Refs #200, #201, #202, #225.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_016aLoB2xRrqrqtXuqpt2Cf1

Issue #451, opened 2026-07-02
Close the production/harness retrieval gap: measure rerank-ON in shipping config (83.3% vs 97.6% R@5)


plur9:
From the 2026-07-02 graph assessment (#225) and `benchmark-slide-briefing-2026-06-16.md` Confirm-item 1.

## Problem
The published 97.6% R@5 is a **harness** number with reranking on. The **shipping** configuration runs reranker-OFF and measures **76.7% R@5** (first confirmed measurement: [#451 comment](https://github.com/plur-ai/plur/issues/451#issuecomment-4866945424)). Rerank-ON had never been measured in the production config. This gap is larger than any expected win from the graph program (gbrain graph ablation: +0.2pp), and it's measurement + configuration work, not construction.

## Tasks
- [x] Run the #46 harness against the exact shipping config (defaults as installed, reranker OFF) — baseline confirmed at 76.7% R@5
- [x] Same config with reranker ON — measured: tiny (`ms-marco-minilm-l6`) 83.3% R@5 @ p50 245ms; quality (`bge-reranker-v2-m3`) 90% R@5 @ p50 ~5s (CPU, prohibitive)
- [x] If bge-reranker-v2-m3 latency is prohibitive on CPU (confirmed), benchmark a 4–25M tier (FlashRank/MiniLM) as the default-ON candidate — ms-marco-minilm-l6 adapter shipped in **PR #465** (merged)
- [ ] Decide the shipping default from the data; update README/CLAUDE.md numbers to match what users actually get — decision: keep default `off` pending task 5; README update deferred to idle-machine re-measurement
- [ ] Per-store eval gate: reranker can be net-negative out-of-domain — add a quick self-check before enabling (queued as nightshift task `org-20260702-per-store-eval-gate`)

Refs #220, #222, #46, #465.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_016aLoB2xRrqrqtXuqpt2Cf1

plur9:
## Measurements: rerank ON/OFF in the shipping configuration

[...]

PR #498, opened 2026-07-04
feat(core,mcp): per-store reranker fit check — domain-fit scoring in plur doctor (#451)

plur9:
## Problem

Cross-encoder rerankers are trained on specific domains (MS MARCO, multilingual retrieval). On out-of-domain engram stores, they can produce uninformative or inverted scores — activating by default could be net-negative. PR #479 added the self-eval gate (rerank vs RRF ordering), but there was no lightweight diagnostic to answer: **does this reranker even see signal on my store's domains?**

## Solution

`checkRerankerFit()` — a fast, fully local fit check that scores same-domain vs cross-domain pairs on the user's actual engrams. No LLM, no downloads beyond the already-loaded reranker. Runs in seconds.

### Algorithm
1. Sample up to 100 engrams from the store.
2. Group by `domain`. Multi-domain → same-domain consecutive pairs = positive, cross-domain reps = negative. Single-domain fallback → positional proxy.
3. Score K positive and K negative `(query, document)` pairs via `scoreBatch`.
4. **Separability** = (mean_positive − mean_negative) / span ∈ [−1, +1].
   - > 0 → model scores relevant pairs higher (correct direction)
   - ≈ 0 → no signal
   - < 0 → inverted (harmful)
5. Gate: separability ≥ 0.05 → fit; otherwise warn with remediation.

### Changes

- **`packages/core/src/rerankers/fit-check.ts`** — new module: `checkRerankerFit()`, `FitCheckResult`, `FitCheckEngram`.
- **`packages/core/src/rerankers/index.ts`** — re-exports fit-check.
- **`packages/core/src/index.ts`** — `Plur.checkRerankerFit()` public method + module exports.
- **`packages/mcp/src/tools.ts`** — `plur_doctor` runs the fit check when the reranker loads; surfaces separability and remediation hint on poor fit (advisory, never auto-disables).
- **`scripts/check-rerank-flag.mjs`** + **`.githooks/pre-commit`** — opt-in dev guard that rejects a bare `--rerank` flag in docs/scripts (must be `--rerank on` or `--rerank off` to prevent mislabeled benchmark numbers, #341 follow-up).

### Tests
7 offline tests (mock reranker — no model download): empty store, uniform scores, single-domain fallback, sampleSize limit, name propagation, timestamp, separability=0.

## Follow-up work

`docs/adr/ADR-0002.md` (derived-state provenance for graph edges, #452) is a separate untracked ADR — will come in a dedicated PR once the graph program work lands.

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -1,0 +1,14 @@
+#!/usr/bin/env sh
+# PLUR pre-commit hook. Opt in with:  git config core.hooksPath .githooks
+#
+# Runs the same doc/script guards that CI enforces, but only against the files
+# staged for this commit — so drift is caught before it leaves your machine.
+
+set -e
+
+# Only scan staged docs & scripts; nothing staged in-scope -> nothing to do.
+staged=$(git diff --cached --name-only --diff-filter=ACMR | grep -E '\.(md|sh|ya?ml)$|(^|/)CLAUDE\.md$' || true)
+[ -z "$staged" ] && exit 0
+
+# shellcheck disable=SC2086
+node scripts/check-rerank-flag.mjs $staged
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -15,7 +15,7 @@ import { captureEpisode, queryTimeline } from './episodes.js'
 import { agenticSearch } from './agentic-search.js'
 import { embeddingSearch, embeddingSearchWithScores, type SimilarityResult } from './embeddings.js'
 import { hybridSearch, hybridSearchWithMeta, applyReranker, rrfMergeEngrams as pgliteRrfMerge, type HybridSearchResult, type RerankOptions } from './hybrid-search.js'
-import { getReranker, resolveRerankerName, isRerankerOff, rerankerStatus, resetRerankerStatus, _resetRerankerCache, type RerankerAdapter, type RerankerRuntimeStatus, type RerankerName } from './rerankers/index.js'
+import { getReranker, resolveRerankerName, isRerankerOff, rerankerStatus, resetRerankerStatus, _resetRerankerCache, checkRerankerFit, type RerankerAdapter, type RerankerRuntimeStatus, type RerankerName, type FitCheckResult } from './rerankers/index.js'
 import { runRerankerSelfEval, loadRerankerEvalCache, saveRerankerEvalResult, isRerankerEvalStale, logRerankerEvalAdvisory, type RerankerEvalResult } from './reranker-eval.js'
 import { _resetBgeRerankerCache } from './rerankers/bge-reranker-v2-m3.js'
 import { _resetMsMarcoMiniLmCache } from './rerankers/ms-marco-minilm-l6.js'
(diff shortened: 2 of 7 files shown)


plur9 (commented):
CI passing. The separability metric (mean_positive - mean_negative) / span is the right shape for measuring whether a cross-encoder is actually producing signal on a user's corpus. The MIN_SEPARABILITY = 0.05 threshold is appropriately conservative — for an "auto-enable" gate, false positives (enabling a harmful reranker) are worse than false negatives (missing out on a good one).

The "only one domain" fallback using consecutive pairs as proxy is a reasonable approximation for stores without domain tagging. Worth documenting the assumption explicitly (consecutive engrams are more likely to be topically related than random cross-sample pairs) — the current comment "statement similarity as a proxy" is close but could be clearer.

The deterministicSample implementation is good practice: reproducible results across runs without relying on Date.now() or Math.random().

The pre-commit hook (.githooks/pre-commit) scans staged md/sh/yaml against check-rerank-flag.mjs, which closes the CI drift loop. Opt-in via `git config core.hooksPath .githooks` is the right approach — doesn't force it on contributors who haven't opted in.

@plur9 ready for merge.

plur9 (commented):

**Algorithm correctness:** The separability metric (mean_positive − mean_negative) / span is the right shape. MIN_SEPARABILITY = 0.05 is appropriately conservative — false positives (enabling a harmful reranker) are worse than false negatives. The span normalizer uses `max(Math.abs(all scores), 1e-6)` to avoid /0 — correct.

**Empty store guard:** n_pairs === 0 returns fit=true with separability=0. Right call — a store without enough engrams to evaluate should not fail the gate and block the reranker from ever being tried.

**Single-domain fallback:** Consecutive pairs as positive, cross-sample pairs as negative is a reasonable proxy. The comment notes the assumption; tests confirm separability ≈ 0 on uniform scores, which is the correct baseline.

**`deterministicSample`** is correct practice for a diagnostic that will be cached — reproducible results mean cache invalidation only needs to track store size, not sample variance.

**tools.ts integration:** The try/catch around checkRerankerFit is the right pattern — fit check is diagnostic, must not break doctor for other checks. The detail message format ("Good fit — separability 0.xxx on N pairs") gives the user everything needed to decide without reading code.

**`.githooks/pre-commit`**: Opt-in via `git config core.hooksPath .githooks` is the right approach — doesn't force on contributors who haven't opted in. The staged-files scope (rather than full repo scan) makes it fast and precise.

**`check-rerank-flag.mjs`**: The VIOLATION regex anchors on `run\.ts` so it doesn't false-positive on the arg-parser source (`a === '--rerank'`) or prose. The fallback to scanning all tracked docs when no args are given is the correct CI mode; explicit file list is the correct pre-commit mode.

**One note (non-blocking):** `computed_at: Date.now()` is fine in application code. The only restriction is in Workflow scripts where Date.now() is unavailable to preserve resume determinism — this path is in a library module, not in a workflow script.

CI green (3/3 checks pass). ✅

PR #501, opened 2026-07-04
feat(core,mcp): supersedes chain consumer-side behavior (#481)

### LINKED ISSUE #481: supersedes chains in inject/recall/decay — injection priority, recall annotation, faster decay

plur9:
## Summary

Three coordinated changes so inject/recall/decay all understand supersedes chains:

- **inject.ts** — budget-constrained chain preference: adds `hasHistoricalIntent()` and `isSupersededEngram()` helpers; non-historical prompts apply a 0.3× score penalty to superseded engrams in `selectAndSpread` so the chain tip wins under budget pressure, while queries containing history keywords (before/was/prior/used to/previously/old/earlier/history/legacy) bypass the penalty and surface the full chain
- **tools.ts** — recall annotation: `plur_recall` and `plur_recall_hybrid` append `[superseded by <id>]` to the statement of any returned engram whose `relations.superseded_by` is non-empty; gracefully no-ops when the field is absent (remote-store targets may lack reverse edges)
- **decay.ts** — accelerated decay: superseded engrams with `activation.frequency < 5` decay at 2× the normal rate; high-recall-frequency superseded engrams (≥5 activations) keep normal rate, preserving historical access patterns

## Files Changed

| File | Change |
|------|--------|
| `packages/core/src/inject.ts` | +23 lines — chain helpers + `selectAndSpread` preference logic |
| `packages/core/src/decay.ts` | +6 lines — accelerated decay for superseded engrams |
| `packages/mcp/src/tools.ts` | +14/−10 lines — recall annotation in both handlers |
| `packages/core/test/supersedes-inject.test.ts` | +110 lines — new test file |
| `packages/core/test/supersedes-decay.test.ts` | +120 lines — new test file |

## Test Plan

- [ ] `supersedes-inject.test.ts` (new) — 5 tests: chain tip preference, historical intent bypass, non-superseded engrams unaffected
- [ ] `supersedes-decay.test.ts` (new) — 8 tests: accelerated decay, high-freq exception, non-superseded unaffected
- [ ] `supersedes.test.ts` — 8/8 existing
- [ ] `batch-decay.test.ts` — 7/7 existing
- [ ] Regression: `inject.test.ts` baseline (pre-existing timeout at line 378 unrelated to this PR)

All 13 targeted tests pass on a clean cherry-pick from main.

Closes #481

### DIFF
--- a/packages/core/src/decay.ts
+++ b/packages/core/src/decay.ts
@@ -162,8 +162,12 @@ export function applyBatchDecay(
     const emotionalWeight = (engram as any).episodic?.emotional_weight ?? 5
     const effectiveLambda = lambda * (1 - emotionalWeight / 20)
 
+    const recallCount = engram.activation.frequency ?? 0
+    const isSuperseded = (engram.relations?.superseded_by?.length ?? 0) > 0
+    const finalLambda = isSuperseded && recallCount < 5 ? effectiveLambda * 2.0 : effectiveLambda
+
     const oldStrength = engram.activation.retrieval_strength
-    const newStrength = decayedStrength(oldStrength, days, effectiveLambda)
+    const newStrength = decayedStrength(oldStrength, days, finalLambda)
 
     // Only count as decayed if strength actually changed
     if (Math.abs(newStrength - oldStrength) < 1e-10) continue
--- a/packages/core/src/inject.ts
+++ b/packages/core/src/inject.ts
@@ -180,6 +180,19 @@ export function flattenRelations(engram: Engram): Association[] {
   return associations
 }
 
+// --- Supersedes chain helpers ---
+
+const HISTORICAL_KEYWORDS = ['before', 'was', 'prior', 'used to', 'previously', 'old', 'earlier', 'history', 'historical', 'legacy']
+
+function hasHistoricalIntent(prompt: string): boolean {
+  const lower = prompt.toLowerCase()
+  return HISTORICAL_KEYWORDS.some(kw => lower.includes(kw))
+}
+
+function isSupersededEngram(engram: ScoredEngram): boolean {
+  return (engram.relations?.superseded_by?.length ?? 0) > 0
+}
+
 // --- Strip pipeline ---
 
 function stripAssociations(engram: ScoredEngram): AgentEngram {
@@ -442,6 +455,16 @@ export function selectAndSpread(
 
   // Sort by score descending
   filtered.sort((a, b) => b.score - a.score)
+
+  // Supersedes chain preference: under budget pressure, tip beats older members
+  if (!hasHistoricalIntent(ctx.prompt)) {
+    for (const e of filtered) {
+      if (isSupersededEngram(e)) {
+        e.score *= 0.3
+      }
+    }
+    filtered.sort((a, b) => b.score - a.score)
+  }
 
   // Step 6: Fill directive token budget
   const { selected: directives, tokens_used: directiveTokens } = fillTokenBudget(filtered, maxTokens)
--- a/packages/core/test/supersedes-decay.test.ts
+++ b/packages/core/test/supersedes-decay.test.ts
@@ -1,0 +1,120 @@
+import { describe, it, expect, beforeEach, afterEach } from 'vitest'
+import * as fs from 'fs'
+import * as os from 'os'
+import * as path from 'path'
+import { applyBatchDecay } from '../src/decay.js'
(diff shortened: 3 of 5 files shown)


plur9:
## CTO Review — #501 supersedes chain consumer-side behavior

**LGTM — no blocking findings.**

Three independent changes that all operate on `relations.superseded_by`:

**inject.ts**: The 0.3× score penalty on superseded engrams is the right lever — it's a soft demotion, not a filter, so historical queries can still surface the full chain. `HISTORICAL_KEYWORDS` list is reasonable for v1; we may want stemmed matching later but the exact-substring approach is safe for now.

**tools.ts**: Recall annotation is response-layer only (no engram mutation) — correct. The `superseded_by?.join(', ')` output for multi-predecessor chains is readable.

**decay.ts**: The `recallCount < 5` threshold for the 2× multiplier is intentional — engrams that users actively query for history keep their normal decay. Reasonable heuristic.

**Tests**: 13 new tests across 2 new test files covering the three deliverables. Cherry-picked cleanly from main; all pass without the commitment-tier or Python SDK changes present.

**Minor follow-up** (non-blocking, open separately): `inject.test.ts:378` has a pre-existing timeout from `dynamic import()` inside `it()` — fix is `{ timeout: 15000 }` as third arg. Surfaced during regression run.

🤖 heartbeat CTO review

plur9 (commented):
CI passing. Three-way coordination (inject / decay / recall) is the right shape for this — each surface understands supersedes chains independently rather than one special-casing the others.

**inject.ts** — score mutation (e.score *= 0.3) is safe because scored is built from spread copies ({ ...engram, ... } at construction time), so the input ScoredEngram[] params are not affected. The 0.3x penalty is aggressive but defensible under budget pressure — the historical-intent bypass is the safety valve.

**decay.ts** — the frequency < 5 threshold is reasonable for a first pass. Worth a one-liner comment explaining the invariant (e.g. "high-recall superseded engrams are likely still being consulted for historical reasons — do not accelerate") so future readers understand why 5 and not 3 or 10.

**tools.ts** — the annotation approach is clean and non-breaking for existing callers.

Test coverage is thorough. @plur9 ready for merge.

plur9 (commented):

**Decay acceleration is correctly scoped.** The 2× lambda multiplier applies only when `isSuperseded && recallCount < 5`. The frequency gate (< 5 recalls) is the right safeguard: an engram that's been recalled many times is clearly still useful despite being superseded (user may be aware of the history), so accelerated decay shouldn't apply. Tests confirm both the accelerated path and the frequency-gate bypass.

**Injection priority de-boost (`score *= 0.3`):** applied only when `!hasHistoricalIntent(prompt)` — so queries about 'prior', 'was', 'used to' still surface the historical members. The keyword list is a heuristic but covers the common cases. The 0.3 multiplier is aggressive enough to push superseded engrams below fresh ones under budget pressure without making them invisible.

**`hasHistoricalIntent` is pure / side-effect-free** and operates on the prompt string only — safe to call in the hot inject path.

**One minor concern (non-blocking):** the supersedes chain score de-boost happens after the sort, then re-sorts — correct behavior, just two sorts instead of one. Not a hot loop so no perf issue; acceptable.

CI green (3/3 checks pass). ✅

PR #503, opened 2026-07-06
feat(core,mcp): plur_learn_batch — persist many engrams in one call (#281)

plur9:
## What

Adds `plur_learn_batch`, the batch form of `plur_learn`: one MCP call persists an array of engrams. Nice-to-have split from #281 item #3 — cuts the N-call overhead when an orchestration fans out across agents and wants to record consolidated findings.

## How it works

- **Same policy as single learn.** Each item runs the existing dedup pipeline (content-hash NOOP → semantic recall → LLM ADD/UPDATE/MERGE). Built on the existing `plur.learnBatch()` core method, which already bounds LLM dedup calls (`max_llm_calls`, default 50).
- **Dedup within the batch.** Statements are processed sequentially, so a later item duplicating an earlier one in the same array resolves to NOOP against it.
- **Returns** `ids` (in input order), per-item `results` (`{id, statement, scope, type, decision}`), aggregate `stats`, and `failures` when any occur.
- **Partial-failure isolation** (new in core `learnBatch`): a failing statement no longer aborts the batch. The rest persist; the failure is reported with its original input index. `LearnBatchResult` gains a `failed` stat and a `failures[]` array — additive, existing fields unchanged.

## Scope note

Unlike `plur_learn`, batch items take the **local** learn path — remote-scope auto-routing (`learnRouted`) is not applied per item. For shared/remote-store writes, prefer `plur_learn` or pass an explicit local scope. Called out in the tool description and CHANGELOG; a follow-up could add remote routing to the batch path if demand warrants.

## Tests

- `packages/core/test/batch.test.ts` (new): batch write + id/persistence, dedup-across-batch NOOP, partial-failure isolation, all-failed batch.
- `packages/mcp/test/tools.test.ts`: registration, array write, in-batch dedup, empty-array handling.
- Full core + mcp suites green locally: **2106 passed / 0 failed** (31 skipped).

## Docs

Root + `packages/mcp` README tool tables, `ARCHITECTURE.md` learning row, and a CHANGELOG **Unreleased** entry.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,17 @@
 # Changelog
 
+## Unreleased
+
+### Feature: `plur_learn_batch` — persist many engrams in one MCP call (#281)
+
+Saving knowledge was one engram per `plur_learn` call. When an orchestration fans out across agents and wants to record consolidated findings, the N-call overhead adds up. `plur_learn_batch` accepts an array of engram objects and writes them in a single call.
+
+- Writes go through the **same dedup + policy pipeline** as `plur_learn` (content-hash NOOP → semantic recall → LLM ADD/UPDATE/MERGE). Dedup also applies **within** the batch: a statement duplicating an earlier item in the same array resolves to NOOP against it.
+- Returns the created/affected engram ids in input order, per-item decisions, aggregate stats, and any per-item `failures`.
+- **Partial-failure isolation:** a single failing statement no longer aborts the batch — the rest are persisted and the failure is reported with its input index. `learnBatch`'s result gains a `failed` stat and a `failures[]` array (additive; existing fields unchanged).
+- LLM dedup calls stay bounded by `max_llm_calls` (default 50) to cap bulk-import cost.
+- Note: unlike `plur_learn`, batch items take the local learn path — remote-scope auto-routing (`learnRouted`) is not applied per item. For shared/remote-store writes prefer `plur_learn` or pass an explicit local scope.
+
 ## 0.11.0 (2026-07-06)
 
 Hooks that actually finish, tension lifecycle, migration importers.
--- a/README.md
+++ b/README.md
@@ -197,6 +197,7 @@ plur.sync('git@github.com:you/plur-memory.git')
 | Tool | What it does |
 |------|-------------|
 | `plur_learn` | Store a correction, preference, or convention |
+| `plur_learn_batch` | Store many engrams in one call (batch dedup + per-item failure isolation) |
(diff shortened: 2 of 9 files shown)


plur9 (commented):
**CTO review — implementation approved**

Clean, well-scoped addition. The implementation is correct and the test coverage is thorough.

**What's good:**
- `learnBatch` partial-failure isolation is the right design for orchestrator bulk-import — 49 good facts shouldn't be dropped because statement 50 throws. The try/catch-per-item with index preservation is clean.
- `failures[]` omitted from MCP return when empty (vs always-present in core `LearnBatchResult`) — minor inconsistency with the core type but actually better UX for agents: no `failures` key = no failures, no noise.
- `ids` in input order (for successes) + `failures[n].index` for locating failed positions is a sensible API shape.
- `max_llm_calls` cap exposed at the MCP layer — correct, large bulk imports need a way to trade dedup quality for cost.
- Local-path limitation is clearly documented in both the tool description and CHANGELOG. The rationale (remote routing adds per-item overhead and complexity) is sound for a v1.

**One note:**
Sequential batch is consistent with the existing `embedBatch` pattern and fine for v1. If we ever see orchestrators hitting this with >100 items, we may want a bounded-concurrency fast path — but that belongs in a follow-up issue, not here.

**Pre-merge gate:**
- [ ] CI green (no checks yet on `feat/learn-batch-mcp` — likely queued; wait for green before merging)

Ready to ship once CI passes. @plur9

plur9:
**Ops heartbeat 2026-07-06** — branch was CONFLICTING due to the v0.11.0 release commits (CHANGELOG.md, package.json). Rebased onto current main (b348c9c on top of 0295858), CI now queued. Ready for review once CI passes. All 3 checks from a prior run passed locally (2106 tests, 0 failed).

plur9:
**Miles review — ready for merge**

Clean batch API addition:
- **Partial-failure isolation** via try/catch with input index tracking — orchestrators won't lose all N findings on one bad write.
- **Type changes are additive** (`failures` + `stats.failed`) — existing consumers of `results`/`stats` unaffected.
- **Within-batch dedup** naturally works because each successful item is persisted before the next runs hash-dedup — no extra state needed.
- **Local-path note** in docs correctly distinguishes from `learnRouted` behavior.
- **`max_llm_calls` cap** (default 50) is the right constraint for bulk-import cost.
- Test coverage covers all three behaviors with real `Plur` instances and proper temp dir cleanup.

CI passing. No security concerns. Ready for merge.

PR #526, opened 2026-07-11
fix(remote-store): extend AbortController to cover r.json() body read

### LINKED ISSUE #525: RemoteStore.load() per-page AbortController doesn't cover r.json() body read
## Context

Audit of the #504 fix (PR #505, shipped 0.12.0) found a partial gap in `RemoteStore.load()`.

## Root cause

`packages/core/src/store/remote-store.ts:174-176`:

```ts
const ctrl = new AbortController()
const t = setTimeout(() => ctrl.abort(), LOAD_FETCH_TIMEOUT_MS)
const r = await fetch(u, { headers: this.headers(), signal: ctrl.signal }).finally(() => clearTimeout(t))
// ctrl is now CLEARED — r.json() body read is unprotected
const body = await r.json()
```

`clearTimeout(t)` fires on `fetch()` resolve (headers received), so `r.json()` runs without an abort deadline. A server that returns headers quickly but stalls the body will hang `r.json()` indefinitely. The 55 s process watchdog catches this, but `LOAD_FETCH_TIMEOUT_MS` doesn't bound end-to-end request time as intended.

**Compare:** `tryRemoteInject()` in hook-inject.ts does this correctly — `clearTimeout` is in the outer `finally`, so abort covers headers + body.

## Fix

Move `clearTimeout(t)` to an outer `finally` covering both the fetch and body read:

```ts
const ctrl = new AbortController()
const t = setTimeout(() => ctrl.abort(), LOAD_FETCH_TIMEOUT_MS)
try {
  const r = await fetch(u, { headers: this.headers(), signal: ctrl.signal })
  if (!r.ok) { ... break }
  const body = await r.json() as { rows: any[]; total_count: number }
  ...
} catch (err) {
  // AbortError from stalled body read now reaches this catch
  break
} finally {
  clearTimeout(t)
}
```

## Priority

P3 — the 55 s process watchdog catches this at the process level, so there's no user-visible regression from the gap. This is a correctness cleanup only.

🤖 heartbeat — code-review-504 — 2026-07-11

plur9:
Fixes #525.

## What

`clearTimeout(t)` was called in `fetch().finally()`, so the abort deadline expired as soon as headers arrived. A server that delivered headers quickly but stalled the body would hang `r.json()` indefinitely — caught only by the 55 s process watchdog, not by `LOAD_FETCH_TIMEOUT_MS`.

Wrap each page's fetch + body read in a `try/catch/finally` block. `clearTimeout` in `finally` now covers both operations. On abort or any fetch error, the per-page `catch` logs and breaks out of the page loop cleanly.

## Why this pattern

`tryRemoteInject()` in `hook-inject.ts` already does it correctly — this aligns `load()` with the existing correct pattern.

## Test impact

Pre-existing integration test failures: 5 (on main). With this fix: 2 (unrelated, pre-existing). No regressions introduced.

🤖 heartbeat — ops/b17-sprint-claim — 2026-07-11

### DIFF
--- a/packages/core/src/store/remote-store.ts
+++ b/packages/core/src/store/remote-store.ts
@@ -157,23 +157,30 @@ export class RemoteStore implements EngramStore {
           const u = `${this.apiBase}/engrams?scope=${encodeURIComponent(this.scope)}&limit=${limit}&offset=${offset}`
           const ctrl = new AbortController()
           const t = setTimeout(() => ctrl.abort(), LOAD_FETCH_TIMEOUT_MS)
-          const r = await fetch(u, { headers: this.headers(), signal: ctrl.signal }).finally(() => clearTimeout(t))
-          if (!r.ok) {
-            // 403 (no read access) and 404 (scope doesn't exist) are
-            // not errors at the store level — that store just contributes
-            // nothing. Bubble other 5xx as logs.
-            if (r.status >= 500) console.error(`[plur:remote-store] ${this.url} returned ${r.status} loading scope ${this.scope}`)
+          try {
+            const r = await fetch(u, { headers: this.headers(), signal: ctrl.signal })
+            if (!r.ok) {
+              // 403 (no read access) and 404 (scope doesn't exist) are
+              // not errors at the store level — that store just contributes
+              // nothing. Bubble other 5xx as logs.
+              if (r.status >= 500) console.error(`[plur:remote-store] ${this.url} returned ${r.status} loading scope ${this.scope}`)
+              break
+            }
+            const body = await r.json() as { rows: any[]; total_count: number }
+            // Server returns DB rows shaped {id, scope, status, data, created_at, updated_at}
+            // — the engram contents live in row.data. Reshape + validate; drop malformed.
+            for (const row of body.rows) {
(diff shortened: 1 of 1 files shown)


plur9 (commented):
CTO review — LGTM.

Fix is correct. `clearTimeout(t)` moved to `finally` now covers the full request lifecycle (fetch headers + body read). Pattern matches `tryRemoteInject()` in hook-inject.ts, which already handled this correctly.

Verified:
- Try/catch/finally structure preserves the existing page-loop `break` logic correctly.
- AbortError from a stalled body read now reaches the catch and breaks cleanly, instead of hanging until the 55 s process watchdog.
- All 4 CI checks pass (Node 20/22/24 + test). Single-file change, no risk.

Self-approve blocked (authored by plur9). Needs gregor or crtahlin to merge.

plur9:
Superseded by PR #531 (`fix/remote-store-abort-covers-json`) — same fix, cleaner implementation (wraps both `fetch` and `r.json()` in a single `try/catch/finally` block rather than restructuring the if-chain). Closing in favour of #531.

🤖 heartbeat 2026-07-12

Issue #130, opened 2026-05-11
remote-store: cold-cache append creates fake-fresh cache, blocks load() refetch for ttlMs


miles-on-nightshift:
## Background

Flagged by @plur9 in the [review of PR #94](https://github.com/plur-ai/plur/pull/94) (now closed/superseded). The same regression exists on main at `packages/core/src/store/remote-store.ts:148-153` after the optimistic-cache fix landed via #89.

## The bug

```ts
// packages/core/src/store/remote-store.ts:148-153
const stored = { ...(engram as any), id: data.id } as Engram
if (this.cache) {
  this.cache.engrams.push(stored)
} else {
  this.cache = { ts: Date.now(), engrams: [stored] }   // ← cold-cache: pretends to be fresh full load
}
```

When `this.cache === null` (no prior `load()` has run), the `else` branch creates a cache that:

- Contains only the just-appended engram (1 of N for that scope)
- Has `ts: Date.now()` — so `load()` at line 54 (`now - this.cache.ts < this.ttlMs`) treats it as **fresh** for up to 60s

**Concretely:** a fresh `RemoteStore` instance where the user calls `learn()` first (no prior `recall()`) will, for the next 60s, see only that one engram on `load()` — missing every other engram in the scope on the server.

Before #89's fix, the cache stayed `null` and the next `load()` fetched everything from the server.

## Reproduction sketch

```ts
const store = new RemoteStore(url, token, 'group:test')
// Server has engrams ENG-A, ENG-B, ENG-C in this scope

await store.append({ statement: 'new thing', scope: 'group:test', ... })
const result = await store.load()
// Expected: [ENG-A, ENG-B, ENG-C, ENG-NEW]
// Actual:   [ENG-NEW]   ← stays this way for ttlMs (60s)
```

## Proposed fix

In the cold-cache branch, mark the cache as **stale** so the next `load()` refetches:

```ts
} else {
  // No prior load() — don't pretend the cache is complete; a single
  // engram is not "all engrams in this scope". Mark stale so the next
  // load() refetches.
  this.cache = { ts: 0, engrams: [stored] }
}
```

This preserves the optimistic-insert benefit for warm caches (the #89 fix) while keeping cold-cache correctness.

## Test to add

```ts
it('cold-cache append does not block a subsequent full load()', async () => {
  // mock 1: POST append → 201 with row { id: 'ENG-NEW' }
  // mock 2: GET load() → list with [ENG-A, ENG-B, ENG-C]
[...]

PR #531, opened 2026-07-12
fix(core): extend AbortController coverage to r.json() body read in RemoteStore.load()

### LINKED ISSUE #525: RemoteStore.load() per-page AbortController doesn't cover r.json() body read

plur9:
## Summary

Fixes #525.

`clearTimeout` was firing inside `.finally()` on the `fetch()` promise, which resolves when response **headers** arrive. The body read (`r.json()`) ran outside the abort deadline — a server returning headers quickly but stalling the body could hang `load()` until the 55 s process watchdog killed it.

**Fix:** wrap each per-page `fetch` + `r.json()` in a `try/catch/finally` with `clearTimeout` in the `finally`. This matches the correct pattern already used in `tryRemoteInject()` (hook-inject.ts). AbortErrors from stalled body reads now log a per-page message and `break` the pagination loop cleanly.

## Before / After

**Before** — abort deadline clears on headers; body read is unprotected:
```ts
const r = await fetch(u, { ..., signal: ctrl.signal }).finally(() => clearTimeout(t))
const body = await r.json()  // ← no deadline
```

**After** — abort deadline covers headers + body:
```ts
try {
  const r = await fetch(u, { ..., signal: ctrl.signal })
  const body = await r.json()  // ← still under deadline
  ...
} catch (err) {
  break
} finally {
  clearTimeout(t)  // ← fires after both fetch and body read
}
```

## Priority

P3 per #525 — 55 s process watchdog catches the hang at process level; no user-visible regression from the gap. This is a correctness cleanup only.

🤖 heartbeat — 2026-07-12

### DIFF
--- a/packages/core/src/store/remote-store.ts
+++ b/packages/core/src/store/remote-store.ts
@@ -173,23 +173,30 @@ export class RemoteStore implements EngramStore {
           const u = `${this.apiBase}/engrams?scope=${encodeURIComponent(this.scope)}&limit=${limit}&offset=${offset}`
           const ctrl = new AbortController()
           const t = setTimeout(() => ctrl.abort(), LOAD_FETCH_TIMEOUT_MS)
-          const r = await fetch(u, { headers: this.headers(), signal: ctrl.signal }).finally(() => clearTimeout(t))
-          if (!r.ok) {
-            // 403 (no read access) and 404 (scope doesn't exist) are
-            // not errors at the store level — that store just contributes
-            // nothing. Bubble other 5xx as logs.
-            if (r.status >= 500) console.error(`[plur:remote-store] ${this.url} returned ${r.status} loading scope ${this.scope}`)
+          try {
+            const r = await fetch(u, { headers: this.headers(), signal: ctrl.signal })
+            if (!r.ok) {
+              // 403 (no read access) and 404 (scope doesn't exist) are
+              // not errors at the store level — that store just contributes
+              // nothing. Bubble other 5xx as logs.
+              if (r.status >= 500) console.error(`[plur:remote-store] ${this.url} returned ${r.status} loading scope ${this.scope}`)
+              break
+            }
+            const body = await r.json() as { rows: any[]; total_count: number }
+            // Server returns DB rows shaped {id, scope, status, data, created_at, updated_at}
+            // — the engram contents live in row.data. Reshape + validate; drop malformed.
+            for (const row of body.rows) {
+              const e = this.reshape(row)
+              if (e) all.push(e)
+            }
+            if (all.length >= body.total_count || body.rows.length < limit) break
+            offset += limit
+          } catch (err) {
+            console.error(`[plur:remote-store] ${this.url} load page failed: ${(err as Error).message}`)
             break
+          } finally {
+            clearTimeout(t)
           }
-          const body = await r.json() as { rows: any[]; total_count: number }
-          // Server returns DB rows shaped {id, scope, status, data, created_at, updated_at}
-          // — the engram contents live in row.data. Reshape + validate; drop malformed.
-          for (const row of body.rows) {
-            const e = this.reshape(row)
-            if (e) all.push(e)
-          }
-          if (all.length >= body.total_count || body.rows.length < limit) break
-          offset += limit
         }
         this.cache = { ts: Date.now(), engrams: all }
         return all


plur9 (commented):
## CTO review — LGTM

**What**: `clearTimeout(t)` moved from `.finally()` on the `fetch()` promise to an outer `try/catch/finally` block that wraps both `fetch()` and `r.json()`. The abort deadline now covers end-to-end request time (headers + body), matching the pattern already used in `tryRemoteInject()` in hook-inject.ts.

**Why it's correct:**
- `break` inside `try` triggers `finally` before exiting the loop — the timer is always cleared regardless of exit path (success, non-ok response, or abort error)
- Each loop iteration owns its own `ctrl`/`t` pair — no cross-page interference
- The `catch { break }` pattern is appropriate here: a stalled body read should terminate pagination, not retry (partial state is already accumulated in `all`)

**CI:** Green on Node 20/22/24 + Python SDK.

**Note:** This supersedes PR #526 (`fix/525-remote-store-abort-covers-body`), which I'm closing in favor of this cleaner implementation.

✅ Ready for human merge (closes #525).

🤖 CTO review — heartbeat 2026-07-12

plur9 (commented):
**CTO review — CI green (Node 20/22/24)**

Root cause is correct: `.finally()` on `fetch()` resolves on headers arrival, not body completion — so the abort deadline was already cleared when `r.json()` ran. The restructure into `try/catch/finally` matches the pattern in `tryRemoteInject()` (hook-inject.ts) and closes the gap cleanly.

One log-message nit: `(err as Error).message` on an AbortError yields `"This operation was aborted"`, which doesn't distinguish timeout from a caller-initiated abort. Not a blocker, but `err instanceof Error && err.name === 'AbortError' ? ``timeout after ${LOAD_FETCH_TIMEOUT_MS}ms (body read)`` : err.message` would make diagnostics clearer.

Otherwise LGTM. Ready for merge once plur9 reviews.

plur9 (commented):
CTO review: Clean fix for #525. try/catch/finally restructuring correctly extends the AbortController's coverage to the r.json() body read — the original .finally() on fetch() alone left a gap. No behaviour change on the happy path. Ready for merge.

crtahlin (approved):
**Approving.** The core fix is correct: moving `clearTimeout(t)` into a `finally` that wraps both `fetch()` and `r.json()` closes the real gap (the old `.finally()` on `fetch()` cleared the deadline before the body read), and it matches the `tryRemoteInject()` pattern. `break` inside `try` runs `finally` first, so the timer is cleared on every exit path, and per-iteration `ctrl`/`t` ownership is clean. CI is green on Node 20/22/24.

**One follow-up, tracked separately in #550 (non-blocking):** the restructure has a side effect the description doesn't mention — the new inner `catch { break }` falls through to `this.cache = { ts: Date.now(), engrams: all }`, so a thrown fetch/body error now caches a partial (or empty) result with a fresh 60s TTL, bypassing the outer catch's deliberate "don't poison the cache on failure" fallback. That reintroduces the bug class #130 fixed (the `remote-store-cache.test.ts:75` sibling test guards it) and degrades the exact stall scenario this PR handles. It's P3 and self-heals after `ttlMs`, so it does not block this merge — see #550 for the suggested fix (don't overwrite the cache on a non-clean exit) and a regression test.

Minor: `(err as Error).message` on an AbortError logs the generic "This operation was aborted" — an `err.name === 'AbortError'` branch would make diagnostics clearer (also noted in #550).

Issue #521, opened 2026-07-10
Node 20/22 CI: 19 tests fail in hook-learn-check/checkpoint/session-guard suites (pre-existing, not Node 24)


plur9:
## Summary

19 tests fail on Node 20 and Node 22 in CI but pass on Node 24. The failures have been present on `main` since at least the 0.12.0 release (`94e0d4d`).

## Affected test files

- `packages/cli/test/session-checkpoint.test.ts` — 7 failures
- `packages/cli/test/hook-session-end.test.ts` — 4 failures  
- `packages/cli/test/hook-session-guard.test.ts` — 3 failures
- `packages/cli/test/hook-noop.test.ts` — 2 failures
- `packages/cli/test/hook-learn-check.test.ts` — 3 failures (Node 22, main run 29021076464)

## Evidence

- Main branch CI run [29021076464](https://github.com/plur-ai/plur/actions/runs/29021076464) at `94e0d4d` (0.12.0 tag): Node 20 FAIL, Node 22 FAIL, Node 24 PASS
- The same failure set appears in PR #520 — the PR introduces no new failures

## Failure pattern

All failing tests spawn `dist/index.js` via `spawnSync` with a custom `TMPDIR` and `PLUR_PATH`. The `session-checkpoint` tests fail because the checkpoint file is never written — `hook-learn-check` doesn't seem to record the stop counter under Node 20/22. The `hook-session-guard` tests get empty stdout where they expect `permissionDecision`/`deny` content.

## Hypothesis

Something in the dist (atomic counter file, `process.ppid`, ES module export, or `PLUR_PATH` propagation) behaves differently on Node 20/22 vs 24. The fact that it's consistent across many test files points to a shared utility (probably `sessionDir()` or the atomic counter write).

## Suggested fix

1. Check if `process.ppid` is available on Node 20 (it was added in v9.2.0, should be fine) — or if the atomic write uses a Node 24-only API
2. Or narrow the matrix: if PLUR's published min engine is `>=22`, dropping Node 20 from the matrix is straightforward
3. Or fix the underlying compat issue in the dist

🤖 heartbeat — ops/cto

Issue #10, opened 2026-04-03
Pack preview: inspect manifest and engrams before install


crtahlin:
## Summary

When a user receives a knowledge pack from another user, there is currently no built-in way to preview its contents before installing. The user must manually unzip and read the YAML files. Plur should offer a user-friendly preview flow.

## Motivation

Knowledge packs are a trust boundary — you're importing someone else's curated knowledge into your memory system. Before installing, users need to:

1. **See the manifest** — who created it, what it claims to contain, how many engrams, what domains/tags
2. **Preview individual engrams** — read statements, rationale, examples before they enter your store
3. **Make an informed decision** — accept, reject, or selectively import

Currently this requires manually unzipping and reading YAML files, which is not accessible to non-technical users and friction even for technical ones.

### `plur_packs_preview` (or `plur_packs_inspect`)

Given a pack path (directory or zip), display:

```
Pack: stoicism-applied v1.0.0
Creator: Gregor
Engrams: 7
Domain: philosophy.stoicism
Tags: stoicism, communication, conflict, framework, emotional-regulation
Integrity: sha256:77f4cc7f...

Engrams:
1. [behavioral] Stoic communication is not suppression of emotion but strategic channeling...
2. [behavioral] In conflict, the composed party holds higher status...
3. [procedural] Five techniques for stoic conflict communication...
...
```

### Additional considerations

- Should support both zip files and directories as input
- Show a summary by default, with option to expand full engram details
- Flag any potential concerns (e.g., engrams with `global` scope, high retrieval strength, or unusual domains)
- Could integrate with provenance verification (see related issue)

## User Story

> "I received a knowledge pack from a colleague. Before I install it, I want to see what's in it — who made it, what topics it covers, and whether the individual engrams are relevant and appropriate for my context."

PR #548, opened 2026-07-14
ops(release): manifest gate — abort on undeclared PRs before publish (#544)

### LINKED ISSUE #544: Release: gate release.sh on an explicit manifest so only intended changes ship (0.12.0 postmortem)
## Summary

Gate `scripts/release.sh` on an explicit **manifest of intended changes**, so a release can only ship the changes it declares — not everything that happens to be on `main`. Proposed in two parts: a small immediate tripwire, and a larger structural move to a release-PR model. Filing for a decision on whether to adopt, and which variant.

## Motivation

`0.12.0` (2026-07-09) shipped a batch of queued features that were not intended for that release, alongside the one change that was (Cursor IDE support). `0.13.0` was cut ~90 minutes later to walk it back — its notes state that "some features originally slated for this release were pulled back before shipping because they weren't sufficiently tested." This also caused #51 to be reopened.

Root cause is structural, not a one-off mistake:

1. **The release is cut from `main` HEAD.** `release.sh` step 4 does `git add -A && git commit && git tag`, so the tag captures whatever is currently on `main`. Every PR merged since the previous tag ships automatically. There is no separation between "merged to `main`" and "approved for release" — merged silently means released.
2. **The release notes are decoupled from what actually ships.** The `CHANGELOG.md` section is hand-authored per version, and the GitHub release notes are extracted from it. Nothing checks that the declared change set matches the commits since the last tag. The notes can describe the intended release while the tag carries much more.
3. **The pipeline runs unattended through the irreversible, public steps.** `release.sh` bumps versions, publishes to npm and PyPI, creates the GitHub release, deploys the site, and posts the release announcement — in one run. The only human checkpoint (`--dry-run`) stops *before* printing the change set that would ship, so it does not surface the mismatch either.

In the `0.12.0` case, the intended feature (Cursor) was merged *after* most of the unintended ones. So there was no single commit on `main` that contained Cursor without the rest — meaning a release from `main` HEAD could not have excluded them even in principle.

### Part A — Manifest tripwire (small, immediate)

Before the tag/publish steps, compute the PRs present in `git log <last-tag>..HEAD` and compare against the PR numbers enumerated in the `CHANGELOG.md` section for this version. Abort if any shipped commit references a PR the CHANGELOG does not declare.

Illustrative sketch (matching rule is for discussion — squash-merge subjects on this repo carry `(#NNN)`):

```bash
# --- Step 3.6: Manifest gate — declared change set must cover what ships ---
LAST_TAG=$(git describe --tags --abbrev=0)

# PR numbers referenced by commits since the last tag
COMMIT_PRS=$(git log --format='%s' "$LAST_TAG"..HEAD \
  | grep -oE '#[0-9]+' | tr -d '#' | sort -un)

# PR numbers declared in the CHANGELOG section for this version
CHANGELOG_PRS=$(echo "$RELEASE_NOTES" \
  | grep -oE '#[0-9]+' | tr -d '#' | sort -un)

# Any shipped PR not declared in the CHANGELOG?
UNDECLARED=$(comm -23 <(echo "$COMMIT_PRS") <(echo "$CHANGELOG_PRS"))
if [ -n "$UNDECLARED" ]; then
  echo "✗ Release aborted: PR(s) on main since $LAST_TAG not declared in CHANGELOG $VERSION:"
  echo "$UNDECLARED" | sed 's/^/    #/'
  echo "  Either add them to the '## $VERSION' section (they will ship),"
  echo "  or move to a release-candidate commit that excludes them (see Part B)."
  exit 1
fi
```

**What this does and does not do — worth being explicit:** this is a *detector*, not a *filter*. It cannot ship "`main` HEAD minus these commits" — the tag captures the whole tree. On mismatch it aborts and forces a decision: declare the extra PRs (accept they ship), or exclude them by changing the tagged commit (Part B). Its value is that it converts a silent bad release into a loud, blocking abort at a still-reversible point — before npm/PyPI/GitHub/announcement. It would have stopped `0.12.0`.

Edge cases to settle if adopted: commits with no PR reference (direct pushes, the release chore commit), issue references vs PR references sharing the `#NNN` form, and whether the check should be advisory (warn) or hard (abort) on a first rollout.

### Part B — Release-PR model (structural)

Adopt a release-PR flow (release-please style) so the manifest and the tagged commit are the same artifact:

- A release PR bumps the versions and writes the `CHANGELOG.md` section enumerating exactly the PRs included.
- Merging that PR is the deliberate act that defines the commit the tag is cut from.
- Anything merged to `main` after the release PR opens is excluded by construction — not by a check, but by *when* the freeze happens. This is the real "block" that Part A alone cannot provide.

For a largely linear, single-maintainer history, this is lighter than a long-lived release branch and achieves the same exclusion guarantee.

### Optional — decouple announce from publish

Consider splitting the release announcement out of the unattended publish run, so a mistaken release is not broadcast before a human has seen it. At minimum, make the announcement an explicit opt-in step rather than the default.

## Decision requested

Whether to adopt this, and which scope:
- Part A only (fast tripwire), or
- Part A + Part B (tripwire now, release-PR model as the target state), and
- whether to also decouple the announcement step.

Assigning to @plur9 to decide the approach before any implementation.

plur9:
## Summary

- Adds step 3.6 to `scripts/release.sh`: before commit/tag/push, compare PRs in `git log` since the last tag against PRs declared in the `CHANGELOG.md` section for the new version
- Aborts with a clear action list (add to CHANGELOG, or cut from a cherry-picked commit) if any shipped PR is undeclared
- Runs before all irreversible steps — the entire git/npm/PyPI/GitHub/tweet chain is still revertable when the gate fires
- Would have stopped the 0.12.0/0.13.0 rollback incident

Implements Part A of issue #544. Part B (release-PR model) is left for a follow-up decision.

## How it works

```
git log --format='%s' v0.12.0..HEAD | grep -oE '\(#[0-9]+\)'  → shipped PRs
awk ... CHANGELOG.md | grep -oE '\(#[0-9]+\)'                  → declared PRs
comm -23 <shipped> <declared>                                   → undeclared PRs → abort
```

Against the current main (since v0.12.0): detects PRs #516, #519, #521, #539 as undeclared — exactly right, since none of these appear in a `## 0.13.0` CHANGELOG section yet.

## Test plan

- [ ] `./scripts/release.sh 0.13.0 --dry-run --skip-tweet` — should abort at step 3.6 listing PRs since v0.12.0 as undeclared (no `## 0.13.0` section in CHANGELOG yet)
- [ ] Add a `## 0.13.0` section to CHANGELOG.md declaring those PRs, re-run — should pass the gate
- [ ] Verify no regressions: existing steps 3.5 (tweet), 3.7 (preflight), 4 (git) are unchanged

Closes #544 (Part A).

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -14,6 +14,33 @@ This repo publishes four npm packages from `packages/`:
 Publishing credentials (npm auth as `plur9`, the pending `NPM_TOKEN` repo secret, ClawHub OAuth)
 are documented in [`docs/runbooks/credentials.md`](docs/runbooks/credentials.md).
 
+## Manifest gate — the CHANGELOG must declare what ships (issue #544)
+
+`scripts/release.sh` runs a **manifest gate** (Step 3.6) before any irreversible
+step. It **aborts the release** if a *user-facing* PR merged since the last tag
+is **not declared** in this version's CHANGELOG section as `(#N)`. This exists
+because the 0.12.0/0.13.0 incident shipped 11 unreviewed PRs that no one had
+declared; the gate would have caught it.
+
+**What you must do before releasing:** ensure every user-facing PR that will ship
+appears in the `## <version>` section of `CHANGELOG.md` with its `(#N)` ref. If
+the gate aborts, it lists the missing PRs and your options.
+
+**What the gate skips:** PRs whose squash-commit subject uses a **non-user-facing
+conventional-commit type** — `chore`, `ci`, `docs`, `test`, `build`, `refactor`,
+`style`. These are curated out of the CHANGELOG by convention and are **not**
+required to be declared. Only `feat` / `fix` / `perf` (and any other type) must
+appear. So a `ci(hermes): …` or `chore(deps): …` PR ships silently; a
+`feat(core): …` or `fix(mcp): …` PR must be in the CHANGELOG or the release stops.
+
+**If the gate flags a PR that is genuinely not user-facing**, retitle its squash
+commit with the appropriate `chore`/`ci`/`docs`/etc. type so it is curated out —
+don't add noise to the CHANGELOG just to satisfy the gate.
+
+The gate is offline and deterministic (no GitHub API calls): it reads shipped PRs
+from `git log <last-tag>..HEAD` subjects and declared PRs from the CHANGELOG
+section, and compares the sets.
+
 ## Manual publish (one package)
 
 When `main` has a version bump that hasn't reached npm yet — e.g. `@plur-ai/claw@0.9.10` is on
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -31,6 +31,8 @@
 #   2.  Build all packages
 #   3.  Run tests
 #   3.5 Validate tweet length (abort if > 270 chars)
+#   3.6 Manifest gate: abort if a user-facing PR merged since the last tag is
+#       undeclared in the CHANGELOG (issue #544; see RELEASING.md)
 #   4.  Commit + tag + push
 #   5a. Publish npm to @next (canary)
 #   5b. Smoke test (npx by exact version, assert --version reports correctly)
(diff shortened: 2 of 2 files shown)


crtahlin (changes requested):
## Review — request changes

The intent is excellent and the placement is right: a manifest gate that aborts before any irreversible step, which would have caught the 0.12.0/0.13.0 incident (#544 Part A). But the implementation has two shell bugs, and the new requirement it imposes isn't documented. This is a release-safety mechanism, so it needs to be correct and discoverable before it can be trusted. None of this is covered by CI — the green checks are the unrelated TS/Node suite; nothing runs this shell script.

### Blockers

1. **`comm` fed numerically-sorted input → wrong results, silently.** Both lists use `sort -un` (numeric), but `comm -23` compares **lexically** (LC_COLLATE). They disagree whenever PR numbers have mixed digit-lengths — the normal case. Reproduced against this code:
   - shipped `{2,9,10,519}`, declared `{10,519}` → gate outputs `2 9 10 519` (exit 0, no warning): it flags the **declared** PRs #10 and #519 as *undeclared* and aborts the release.
   - Correct answer `2 9` is what `sort -u` (lexical, matching `comm`) produces.
   The PR's "detects #516/#519/#521/#539 — exactly right" check passed only because those are all three digits (numeric == lexical order there). **Fix:** drop the `-n` — `sort -u` on both `SHIPPED_PRS` and `DECLARED_PRS` (numeric ordering is irrelevant for a set difference).

2. **`set -euo pipefail` makes the empty-PR guards unreachable.** With `pipefail` (line 58) active, the `grep -oE '\(#[0-9]+\)'` in each assignment exits 1 on **no match**, which aborts the whole release before the `if [ -z "$SHIPPED_PRS" ]` / `elif [ -z "$DECLARED_PRS" ]` branches can run. So a release whose commits carry no `(#N)` refs, or a `## VERSION` section with no PR numbers, dies with an opaque `grep`/pipefail failure instead of taking the intended empty path. **Fix:** append `|| true` (or `|| echo ""`) to the grep pipelines.

### Docs — not fully updated

3. The in-script step-list comment was updated (good, `3.6` added), but the **operator-facing requirement is undocumented**: "release.sh aborts unless every PR merged since the last tag appears in the `## VERSION` CHANGELOG section as `(#N)`." `RELEASING.md` (the canonical release doc) doesn't mention `release.sh` at all, and neither it nor `CLAUDE.md` states this new prerequisite — a releaser would only discover it by hitting the abort. Add the requirement to `RELEASING.md`.

4. **Design/process point worth a decision + doc:** the gate demands *every* merged PR be declared, but the CHANGELOG today is curated — user-facing changes cite their PR, while chore/CI/docs/refactor PRs (e.g. #541, #536, this #548) typically get no entry. Under this gate, every such PR must be added to the CHANGELOG or the release aborts. Either expand the CHANGELOG convention to list every PR (noisier), add a skip mechanism (e.g. ignore PRs labeled `chore`/`ci`), or document an "Internal" subsection. Whichever you choose should be captured in `RELEASING.md`.

### Minor

- Test-plan boxes are all unchecked — after the fixes, please actually run the two breaking cases: mixed-digit shipped/declared sets, and a range or section with no `(#N)`.
- `awk '$0 ~ "^## "v'` treats `.` in the version as regex "any char" — harmless today, but escaping or anchoring (`"^## "v"[ (]"`) is tighter.

The design is right and I want this gate in — a release gate born from a release incident just has to be correct, and there's some irony in it currently misfiring. Happy to re-review after the two fixes + the doc updates.

crtahlin:
@miles-on-nightshift can you address the required changes?

plur9:
Addressed all items — thanks for the sharp review; the irony of a release-safety gate misfiring was not lost on me. Also rebuilt on current `main` (the branch had drifted 60 files behind, so the diff is now just `release.sh` + `RELEASING.md`).

**Blocker 1 — `sort -un` vs `comm` lexical.** Fixed: `sort -u` (lexical) on both sets. Ran your exact repro:
```
shipped {2,9,10,519}, declared {10,519}  ->  undeclared = [2 9]   ✓ (was [2 9 10 519])
```

**Blocker 2 — `pipefail` + `grep` no-match.** Fixed: `|| true` on the grep pipelines. Ran the empty case:
```
subjects with no (#N)  ->  SHIPPED_PRS empty  ->  takes the empty path, no abort   ✓
```

**Docs (3) + Design (4).** Went with the **label/type-skip** option, implemented via **conventional-commit type** (offline, no GitHub API calls): the gate skips `chore`/`ci`/`docs`/`test`/`build`/`refactor`/`style` squash subjects, so the CHANGELOG stays curated (user-facing only) while `feat`/`fix`/`perf` must be declared. This still catches the 0.12.0 case (unreviewed *features*). Verified:
```
feat(core)(#100), fix(mcp)(#101), chore(#102), ci(#103), docs(#104)
  ->  must-declare = [100 101]   ✓
```
Documented the requirement **and** the skip convention in `RELEASING.md` (it now has a "Manifest gate" section), plus added the `3.6` step to `release.sh`'s header list. If the gate ever flags a genuinely non-user-facing PR, the fix is to retitle its squash commit with the right type — the abort message says so.

**Minors.** Anchored the awk match (`\"^## \"v\"[ (]\"`) so `.` isn't any-char. Test-plan cases actually run (the three above).

Ready for re-review.

PR #555, opened 2026-07-14
chore(test): replace a real org name in test fixtures with a generic placeholder

crtahlin:
## What

Two test fixtures in `packages/core/test/` used a real third-party organization's name plus a specific commercial-offer string (revision + validity date) as literal test data. This is a public repository, so real third-party commercial detail should not appear in it.

## Change

Replaced the org name with a neutral `Acme` placeholder in:
- `packages/core/test/expiry.test.ts` — the "valid <date>" offer-shorthand case
- `packages/core/test/plur.test.ts` — the expiry-extraction round-trip case (learn + recall)

Dates and all assertions are unchanged (`2026-05-31` etc.), so the tests exercise exactly the same behavior. Pure fixture-string cleanup — no code or logic touched.

## Note

This scrubs the current tree only; the string remains in prior git history. Raising here so the fixture stops carrying it going forward.

### DIFF
--- a/packages/core/test/expiry.test.ts
+++ b/packages/core/test/expiry.test.ts
@@ -38,7 +38,7 @@ describe('extractExpiry', () => {
   })
 
   it('parses bare "valid <date>" (offer shorthand)', () => {
-    const hit = extractExpiry('IGEA Enterprise offer, valid 31 May 2026')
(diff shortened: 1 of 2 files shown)


plur9 (approved):
CI green. Simple fixture cleanup — replaces real org name with a generic placeholder in two test strings. No logic change, no test behaviour change. @plur9 ready for merge.

PR #556, opened 2026-07-14
ci(hermes): also run on the branch that actually ships

plur9:
Follow-up to #540 — this commit was authored but **missed the merge**, so the workflow landed with the defect still in it.

## Problem

The `push` trigger says `branches: [main]`. But releases are cut from `fix-clean-0.11-base`, not `main` — so CI would skip pushes to **the branch that actually ships**.

This is the same defect that let #535 and #538 merge with **zero checks reported**: every pre-existing workflow gates on `branches: [main]`, and both PRs targeted the release branch.

```
$ gh pr checks 535
no checks reported on the 'fix/hermes-bridge-orphan' branch
```

## Change

- `push`: `[main, 'fix-clean-*', 'release/*']` — cover the release line, not just `main`.
- `pull_request`: **deliberately no `branches:` filter**, so a PR into *any* base runs the matrix and cannot merge unverified. (This is why #540's own checks ran at all.)

CI-only. No runtime or packaging impact.

### DIFF
--- a/.github/workflows/hermes.yml
+++ b/.github/workflows/hermes.yml
@@ -12,9 +12,15 @@ name: plur-hermes
 
 on:
   push:
-    branches: [main]
+    # NOT `branches: [main]`. Releases are currently cut from
+    # fix-clean-0.11-base, not main, so a main-only trigger would skip CI on
+    # the branch that actually ships. The existing workflows do gate on main,
+    # which is why #535 and #538 merged with zero checks reported.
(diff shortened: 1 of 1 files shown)


plur9:
Closing as superseded. Its base is fix-clean-0.11-base, which is being retired now that main is the single release source (the 0.11-base branch was only the clean 0.13.0 release line). Hermes CI on main is provided by #541; the orphan fix is on main via #575. Nothing here is lost.

Issue #345, opened 2026-06-18
Stage 2: self-describing scope metadata (description + covers + sensitivity)


plur9:
**Stage 2 of the scope-upgrade roadmap — the keystone.** Today a scope is an opaque label, so "which scope does this engram belong to?" is unanswerable — which is why ~2,599 engrams defaulted to `global`. This adds self-describing scope metadata so routing, the demotion guard, and re-scoping all become decidable instead of guessed.

## Model

Each scope carries metadata (mirrors `PackManifestSchema`'s precedent):

```ts
ScopeMetadata = {
  scope: string                 // e.g. "group:plur/engineering"
  description: string           // human/agent-readable: what this scope is for
  covers: string[]              // domains/topics it holds — e.g. ["plur.*","software.architecture","openclaw"]
  sensitivity?: {               // drives the Stage-1 demotion guard per-scope
    forbid?: ('secrets'|'infra'|'pii')[]   // default ['secrets','infra'] for shared scopes
    allow?: string[]                        // explicit allow-list (e.g. a maintainer scope that may hold infra)
  }
  injection_policy?: 'on_match'|'on_request'|'always'
  owner?: string
}
```

## Where it lives (the important call)

- **Team / group / public scopes → server-authoritative** on the enterprise store (`org_*` schema): a `scopes` table + `GET /api/v1/scopes` endpoint, managed in the admin dashboard. The whole team sees one definition.
- **Personal / project scopes → local**: `~/.plur/config.yaml` store entries gain `description` + `covers`; repo-scoped in the repo.

## Surfaced in
- Discovery (`#294`): `/api/v1/me` (or a new `/api/v1/scopes`) returns scope **+ metadata**; `plur_scopes_discover` / `plur stores list` show it.

## Consumed by
- **Stage 1 demotion guard** → reads `sensitivity` per-scope instead of the hardcoded `isSharedScope` + regex. (`group:plur/engineering: forbid [secrets,infra]` vs a maintainer scope that allows infra.)
- **Stage 3 scope-selection** → agents match engram content → `covers[]` to pick the right scope, instead of memorized prose.
- **Stage 6 re-scope** → match each global engram → best scope by `covers[]`.

## Deliverables
1. `ScopeMetadataSchema` (Zod) in core + `spec/scope-metadata.schema.json` (reuse the #320 gen:schemas drift gate).
2. Server: `scopes` table + `GET/PUT /api/v1/scopes` (enterprise) + admin UI.
[...]

PR #579, opened 2026-07-15
fix(core,release): 0.14.0 pre-release hardening

plur9:
## 0.14.0 pre-release hardening

Final hardening pass before cutting **0.14.0** — the clean re-land of the batch that 0.12.0 shipped unreviewed and 0.13.0 pulled back. Driven by a five-evaluator pre-release audit (CTO, Dijkstra, Data, Critic, Taleb). The batch fixes themselves are already on `main` and verified (CTO ran the full suite twice green; Dijkstra proved #281/#451/#217 correct by invariant). This PR closes the gaps the audit surfaced and writes the release manifest.

### Code fixes

- **`fix(core)` — reject U+2028/U+2029 in scope metadata (#345 follow-up).** The `#345` control-char class missed two of the four ECMA-262 LineTerminators, so a hostile `/api/v1/me` could still fake a new instruction line via `U+2028`/`U+2029` and surface it verbatim in `plur_scopes_discover`. Dijkstra flagged it (payloads slip through) and Data flagged the mirror (legit multiline metadata silently dropped). 2-char regex fix + spec-schema regen + LS/PS test cases.
- **`fix(core)` — multi-word historical keywords match across any whitespace (#481 follow-up).** `"used to"` was compiled with a literal space, so `"used\nto"` / `"used  to"` failed and wrongly un-suppressed the ×0.3 superseded penalty (Dijkstra). Join with `\s+`; newline-split regression test added.
- **`fix(core)` — `feedback()` re-anchors `last_accessed`.** Feedback bumped `retrieval_strength` without advancing `last_accessed` across all three write paths, so read-time decay swallowed the bump — a >4× distortion on dormant engrams (Dijkstra). Now mirrors `_reactivateResults`; stale-anchor regression test added.

### Release safety

- **`ops(release)` — manifest gate handles multi-issue trailers + skips `ops`/`cmo`.** The gate's `\(#[0-9]+\)` extraction silently dropped every number in a comma trailer like `(#521, #247)` (Taleb) on both the shipped and declared sides. Now `\(#N(, #M)*\)`; bare in-prose refs (`refs #535`) still ignored. Internal `ops`/`cmo` commit types curated out alongside the standard non-user-facing set.
- **`docs(changelog)` — 0.14.0 section.** Declares every user-facing PR since v0.12.0 (manifest gate green), manually includes **#563** (batchDecay removal — gate-exempt by its `refactor` type but the release's largest behavior change), calls out the **#281 breaking output-contract**, documents the **PGLite + embedding-gemma reindex** caveat, and retracts the now-false 0.12.0 "decay accelerates for superseded engrams" claim.

### Verification

- Full workspace suite green (`2614 passed / 0 failed`, cli hook tests included after full build).
- `release.sh 0.14.0 --dry-run --skip-tweet`: version-sync ✓, build ✓, **manifest gate ✓ (all 32 PRs declared)**, stops before publish. (The one Step-3 "failed" is the excluded external-API flaky, not a real failure.)

### Deferred to follow-up issues (audit findings, non-blocking)

`compact()` has no CLI/MCP (store grows unbounded) · dead `strengthToStatus`/`dormant` enum · `#51` has no replacement stale-entry detection · hook-session-guard fail-open has no stderr · smoke test is cli-only, no PyPI canary · `#281` MCP-layer test gap.

> Not for the auto-generated tweet: the detailed CHANGELOG bullets exceed X's 270-char budget — real release needs terse tweet-lead copy or `--skip-tweet`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,9 +24,16 @@ The clean re-release of the queued batch that 0.12.0 shipped unreviewed and 0.13
 - **Process-group orphan kill ported to the hermes bridge** (#575): the `plur-hermes` bridge now kills the whole process group on timeout (`start_new_session` + `killpg`), so a timed-out `npx @plur-ai/cli` can't orphan a runaway `node` grandchild.
 - **Hooks fail open on an unwritable state dir** (#574): the session-guard, learn-check, and inject-lock hooks now proceed instead of crashing the prompt or tool call when their state directory isn't writable.
 - **Assorted hook/remote robustness**: per-session concurrency lock for hook-inject (#519), AbortController coverage extended to the `RemoteStore.load()` body read (#531), and `isPlurConfigured` checks the home directory only when the working directory is under home (#521, #247).
+- **`plur doctor` flags stale PGLite + embedding-gemma vectors** (#581): the companion diagnostic to the #483 caveat above — when the PGLite backend runs the opt-in `embedding-gemma` (whose vectors that backend does not auto-rebuild), doctor advises a one-time `plur sync --reembed --full`. Advisory only; it never fails the overall check.
+- **claw warns non-destructively on orphaned OpenClaw config entries** (#583): #51 removed the destructive prune that deleted third-party plugin config — including embedded API keys — whenever a plugin dir was transiently absent during install. `plur doctor` now *warns* about genuinely-orphaned entries (PLUR's own and third-party) without ever deleting, restoring detection without the data-loss.
+
+### Changed (cleanup)
+
+- **Removed dead `strengthToStatus`** (#582): unused since batchDecay's removal, and its label vocabulary never matched the persisted `status` enum. The `dormant`/`candidate` enum values are retained (legacy/reserved) for backward-compatibility with stores written before #563 — removing them would reject existing data.
 
 ### Added
 
+- **`plur compact` + `plur_compact`** (#580): explicitly reclaim disk space from retired engrams, exposed as both a CLI command and an MCP tool. With batchDecay gone (#563), this is the deliberate, logged maintenance op for shrinking the store — retired rows previously had no user-facing removal path.
 - **Session injection telemetry** (#536): per-pack activation tracking logged at `session_end` for offline relevance analysis.
 - **`plur_status` domain + `created_after` filters** (#522, #524).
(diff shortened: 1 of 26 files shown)


plur9 (commented):
CI green (all Node 20/22/24 + Python SDK passes). Self-review not possible (own PR). Code assessment:

- feedback re-anchor: correct fix — decay on stale engrams immediately swallows the strength bump without the last_accessed advance. Tests lock it in.
- historical keyword regex: multi-word phrases with \s+ instead of literal space is correct — covers newlines/tabs in multi-line prompts.
- U+2028/U+2029 in scope metadata: closes a real injection gap (ECMA-262 line terminators that slip past C0/C1 range). Test uses an actual injection payload.

No issues found. Needs a second reviewer to approve before merge.

plur9 (commented):
CTO review: **LGTM — ready to merge when you are.**

Reviewed three code changes and release tooling:

**1. feedback() last_accessed re-anchor** — correct fix. Three write paths each now advance `last_accessed` on non-neutral signal, mirroring `_reactivateResults`. The stale-anchor regression test (dormant engram, positive feedback, check effective strength still shows improvement after elapsed time) proves the bug case. No edge cases missed.

**2. Historical keyword multi-word matching (`\s+`)** — correct. Splitting on `/\s+/` before `escapeRegExp` and rejoining with `\\s+` in the pattern means "used\nto" and "used  to" now match exactly like "used to". The existing `\b` anchors are preserved. Test case added in supersedes-inject.test.ts.

**3. U+2028/U+2029 scope metadata** — correct and necessary. ECMA-262 defines exactly 4 LineTerminators (LF, CR, LS, PS); the previous `\u0000-\u001F` range excludes LF/CR but not LS/PS. The two-char addition to `NO_CONTROL_CHARS` closes that gap in both the TypeScript validator and the JSON schema. The injection-string test case is well-targeted.

**Release tooling** — the multi-issue trailer fix (`\(#N(, #M)*\)`) is correct; the old pattern's implicit `)` anchor after the first number silently dropped every number past the first comma. The `ops`/`cmo` exclusion is consistent with the existing non-user-facing type list. CI green (all 3 matrix nodes).

One note for later: the comment on line 107 in scope-metadata.ts got long. Not blocking — just worth splitting at next touch.

Issue #584, opened 2026-07-15
Release & observability hardening: hook fail-open stderr, cli-only smoke test, no PyPI canary, #281 MCP test


plur9:
Grab-bag of non-blocking hardening items from the 0.14.0 pre-release audit (Critic, Taleb, CTO, Dijkstra), tracked together:

- **hook-session-guard fail-open has no stderr/telemetry** (Critic): its #574 fail-open is a silent bypass with no audit trail, unlike hook-inject/hook-learn-check/#217 which do write stderr. Add an advisory stderr line for parity.
- **Smoke test (release.sh 5b) is cli-only** (Taleb): core/mcp are promoted to `latest` with zero functional check, yet the audited fixes live in core. Extend the canary smoke test to core+mcp.
- **PyPI publish (step 6) has no canary/smoke** and fires after npm is live-to-latest (Taleb): a mid-upload drop = permanently half-published. Document recovery at minimum; consider a canary.
- **#281 MCP-layer test gap** (CTO, Dijkstra): the null-padded `ids` alignment is tested at the core `learnBatch()` level but not through the MCP handler on a forced partial failure. Add one.

crtahlin:
### Status verification — appears complete

All four audit items are implemented:
- **hook-session-guard fail-open stderr** — `packages/cli/src/commands/hook-session-guard.ts` now emits a stderr advisory line on fail-open (parity with other hooks); test added — [plur-ai/plur#585](https://github.com/plur-ai/plur/pull/585) (commit [20ec61c](https://github.com/plur-ai/plur/commit/20ec61c))
- **MCP-layer test for #281 partial failure** — `packages/mcp/test/tools.test.ts` now tests the `plur_learn_batch` MCP handler on a forced partial failure, asserting 1:1 `ids` alignment — same PR
- **Smoke test extended to core+mcp** — `scripts/release.sh` Step 5b now covers all three promoted packages (cli, mcp via `--version`, core via ESM-import) — same PR
- **PyPI post-publish verification** — Step 6 added with retry and documented recovery path in `RELEASING.md` — same PR

Flagging `ready-to-close` and assigning @crtahlin to confirm and close. If any part above is still open, remove the label.

Issue #583, opened 2026-07-15
#51 follow-up: no replacement detection for genuinely-stale OpenClaw plugin entries


plur9:
From the 0.14.0 pre-release audit (Data). #51 correctly removed the destructive stale-entry prune (it fired transiently on every install/upgrade and deleted third-party plugins' config incl. API keys). But there's **no replacement detection** anywhere in `packages/claw/src/`: a genuinely-uninstalled plugin's config entry (and any embedded credentials) now lingers in `openclaw.json` forever, with detection delegated to external OpenClaw this repo can't verify. Correct call vs. the destructive alternative, but a real unmonitored config-hygiene regression.

**Fix:** a non-destructive doctor-side warning for orphaned entries (PLUR's own and third-party), or confirm+document that OpenClaw owns this.

crtahlin:
### Status verification — appears complete

This issue looks fully implemented:
- Non-destructive `orphaned_entries` doctor step added to `packages/claw/src/setup.ts` — warns on OpenClaw config entries with no backing extension directory, advisory only, never deletes — [plur-ai/plur#585](https://github.com/plur-ai/plur/pull/585) (commit [64400ee](https://github.com/plur-ai/plur/commit/64400ee))
- 4 test cases added in `packages/claw/test/setup.test.ts`, including transient-absence safety (no false positive when extensions dir is temporarily absent) and third-party credential entry detection

Flagging `ready-to-close` and assigning @crtahlin to confirm and close. If any part above is still open, remove the label.

Issue #582, opened 2026-07-15
Dead code: strengthToStatus() orphaned + 'dormant'/'candidate' status never assigned


plur9:
From the 0.14.0 pre-release audit (Data + CTO). After #563, `strengthToStatus()` is exported from the public API with **zero call sites**, and its vocabulary (`active/fading/dormant/retirement_candidate`) doesn't match the persisted `status` enum (`active/dormant/retired/candidate`). Separately, `dormant`/`candidate` enum values are **never assigned anywhere** (applyBatchDecay was the sole setter of `dormant`; pre-existing gap, not caused by #563). Maintainability trap.

**Fix:** remove or rewire `strengthToStatus()`; prune or document the unused enum values.

plur9:
Fix committed on #586 (). Removed `strengthToStatus()` from `decay.ts` and its export from `index.ts`; removed the test suite; updated schema description to clarify `dormant`/`candidate` are reserved and never assigned. 6/6 decay tests pass. Will close on merge.

plur9:
Fixed and shipped in #579 (commit f2cbdc6 — "0.14.0 pre-release hardening"):

- `strengthToStatus()` removed from `packages/core/src/decay.ts` — zero production call sites confirmed
- Dead export removed from `packages/core/src/index.ts`
- `decay.test.ts` block removed; the misleading comment that called it "part of the read-time decay model" is gone
- Status enum description in `engram.ts` already documented `dormant`/`candidate` as reserved-but-unassigned (no code change needed there)

Closing.

Issue #581, opened 2026-07-15
PGLite + embedding-gemma: stale vectors not auto-reindexed after prefix change (#483)


plur9:
From the 0.14.0 pre-release audit (Data + CTO). #483's embedder-prefix fix invalidates the JSON flat-file cache (name-keyed) but **not** the PGLite path (`PLUR_BACKEND=pglite`): `engram_embeddings` guards only on vector *dimension* (unchanged at 768d), and `_autoEmbedNewEngrams` is additive-only (skips rows where `hasEmbedding()`). So pglite+gemma installs keep serving wrong-space (E5-prefix) vectors until a manual `plur sync --reembed --full`. Documented as a caveat in the 0.14.0 CHANGELOG.

**Fix:** a `plur doctor` check that flags pglite installs on `embedding-gemma` and recommends a one-time `--reembed --full` after upgrading past #483; consider tracking embedder identity (not just dim) in `engram_embeddings`.

crtahlin:
### Status verification — appears complete

This issue looks fully implemented:
- `plur doctor` now flags stale PGLite + `embedding-gemma` vectors and advises a one-time `plur sync --reembed --full` — landed in commit [`4231432`](https://github.com/plur-ai/plur/commit/4231432fd2c3f2b7c25e155a03e1dfe619bf3d27) via [plur-ai/plur#585](https://github.com/plur-ai/plur/pull/585)
- Detection is env-based (`PLUR_BACKEND=pglite` + `PLUR_EMBEDDER=embedding-gemma`) as specified; advisory only, never flips the overall verdict
- Test added in `packages/cli/test/doctor.test.ts` (spawn-and-assert); full suite 2,626 passed

Flagging `ready-to-close` and assigning @crtahlin to confirm and close. If any part above is still open, remove the label.

PR #585, opened 2026-07-15
fix: 0.14.0 audit follow-ups (#580-#584)

plur9:
## 0.14.0 audit follow-ups (#580–#584)

Addresses **all** remaining non-blocking findings from the five-evaluator pre-release audit — the ones #579 deferred to tickets. **Stacked on #579** (base is `release-prep/0.14.0`); merge #579 first, or merge this after and I'll rebase onto `main`.

| Fix | Issue | What |
|---|---|---|
| `feat(cli,mcp)` compact | #580 | `plur compact` CLI + `plur_compact` MCP tool — the explicit, logged store-shrink op that replaces batchDecay's implicit cleanup. Retired rows had no user-facing removal path after #563. |
| `fix(cli)` doctor | #581 | `plur doctor` advises a one-time `plur sync --reembed --full` when PGLite runs `embedding-gemma`, whose vectors #483 can't auto-rebuild. Advisory only. |
| `refactor(core)` cleanup | #582 | Remove dead `strengthToStatus` (zero callers post-#563, vocab never matched the enum). Keep `dormant`/`candidate` enum values (legacy/reserved) — removing them would reject pre-#563 data. |
| `fix(claw)` orphans | #583 | Non-destructive `plur doctor` warning for genuinely-orphaned OpenClaw config entries — restores the detection #51 removed with its destructive prune, without ever deleting. |
| `fix(cli,release)` hardening | #584 | hook-session-guard fail-open now leaves a stderr audit trail; MCP-layer partial-failure test for `plur_learn_batch`'s `ids` alignment; smoke test extended to core+mcp; PyPI post-publish verify + documented recovery. |

### Verification

- Full workspace suite green: **2,626 passed / 0 failed** (integration of all five fixes + the cherry-picked agent work).
- New/updated tests: LS/PS payload cases (#582 spec regen kept #315 green), `plur compact` CLI+MCP, doctor pglite+gemma detection, claw orphan detection (4 cases incl. transient-absence safety), `plur_learn_batch` MCP `ids`-alignment, hook-guard fail-open advisory.
- Operational check (Taleb T4): confirmed **no scheduled `plur batch-decay`** anywhere in the datacore/nightshift config — #563's removal breaks no automation. The only reference is a docstring in an unscheduled `decay_and_review.py` helper.

### Notes

- #583's warning is advisory and, by design, could surface a spurious line if `doctor` runs at the exact moment a plugin dir is transiently absent mid-upgrade — it never deletes, so this is harmless; a fully reliable orphan signal would require OpenClaw's own manifest, which this repo can't see.
- #581 detects the pglite+gemma combo via env (`PLUR_BACKEND`/`PLUR_EMBEDDER`), the documented activation for each.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,9 +24,16 @@ The clean re-release of the queued batch that 0.12.0 shipped unreviewed and 0.13
 - **Process-group orphan kill ported to the hermes bridge** (#575): the `plur-hermes` bridge now kills the whole process group on timeout (`start_new_session` + `killpg`), so a timed-out `npx @plur-ai/cli` can't orphan a runaway `node` grandchild.
 - **Hooks fail open on an unwritable state dir** (#574): the session-guard, learn-check, and inject-lock hooks now proceed instead of crashing the prompt or tool call when their state directory isn't writable.
 - **Assorted hook/remote robustness**: per-session concurrency lock for hook-inject (#519), AbortController coverage extended to the `RemoteStore.load()` body read (#531), and `isPlurConfigured` checks the home directory only when the working directory is under home (#521, #247).
+- **`plur doctor` flags stale PGLite + embedding-gemma vectors** (#581): the companion diagnostic to the #483 caveat above — when the PGLite backend runs the opt-in `embedding-gemma` (whose vectors that backend does not auto-rebuild), doctor advises a one-time `plur sync --reembed --full`. Advisory only; it never fails the overall check.
+- **claw warns non-destructively on orphaned OpenClaw config entries** (#583): #51 removed the destructive prune that deleted third-party plugin config — including embedded API keys — whenever a plugin dir was transiently absent during install. `plur doctor` now *warns* about genuinely-orphaned entries (PLUR's own and third-party) without ever deleting, restoring detection without the data-loss.
+
+### Changed (cleanup)
+
+- **Removed dead `strengthToStatus`** (#582): unused since batchDecay's removal, and its label vocabulary never matched the persisted `status` enum. The `dormant`/`candidate` enum values are retained (legacy/reserved) for backward-compatibility with stores written before #563 — removing them would reject existing data.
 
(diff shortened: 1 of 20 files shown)


plur9 (commented):
CTO review: **LGTM — ready to merge after #579 lands.**

Stacked on #579 (base is `release-prep/0.14.0`). All five audit follow-ups look correct:

**#580 — `plur compact` / `plur_compact`**: right call to expose this as an explicit user-invoked op rather than an implicit cleanup. With batchDecay gone, store growth had no user-facing path; this closes the gap cleanly.

**#581 — PGLite + embedding-gemma doctor advisory**: detection via env vars (`PLUR_BACKEND`/`PLUR_EMBEDDER`) is the documented activation path for each, so the heuristic is reliable. Advisory-only is the right posture — the user may have already run `--reembed --full`, and a hard failure would be wrong.

**#582 — remove dead `strengthToStatus`**: correct analysis. The function has zero callers post-#563. Retaining `dormant`/`candidate` enum values for stores written before #563 is the right call — removing them would reject pre-existing data. This is a clean, safe deletion.

**#583 — non-destructive orphan detection**: the three design inversions (non-destructive, user-invoked only, conservative gate) are exactly right. The critical property is that `orphanedEntriesStep` is only called from `runDoctor`, not from `mergeEnable` / postinstall — that structural separation is what makes it safe. The transient-absence edge case is honestly documented in the user-facing message. Four test cases cover the important states (no extensions dir, all ok, genuine orphan, third-party orphan).

**#584 — publish verification hardening**: covering `core` in the npm smoke test is the right fix for the gap the audit found — the audited fixes live in `core`, so promoting it unchecked was the real risk. PyPI immutability caveat + explicit recovery path is practical guidance for a real failure scenario.

No CI yet (branch is fresh); assuming suite will be green given the five fixes are self-contained and the full 2626-test pass is documented in the PR body. Suggest waiting for CI before merging.

Issue #580, opened 2026-07-15
compact() has no CLI/MCP entry point — engram store grows unbounded post-batchDecay


plur9:
From the 0.14.0 pre-release audit (Data evaluator, PR #579). Since batchDecay was removed (#563), `compact()` (the only code that physically removes `retired` rows) has **no CLI command and no MCP tool**, and `plur forget` retires without reclaiming file space. The active store is now monotonically growing with no exposed shrink mechanism, not even a manual one. `_loadAllEngrams`/BM25/embedding scans are unbounded by construction.

**Do one of:** (a) expose `compact()` as `plur compact` + an MCP tool (explicit, reversible, logged maintenance op — as the #563 removal rationale envisioned), or (b) explicitly document the accepted unbounded-growth trade-off.

plur9:
## Decision: option (b) — accepted trade-off

Re-investigating this for the heartbeat. The removal of `plur compact` / `plur_compact` in #594 was a deliberate product decision with explicit rationale:

> GC is an operator task, not an agent task, and retired rows only accrue via `forget` (rare) and are already filtered from recall. The better answer — archive retired engrams so users can check back, not delete — is a future design.

So the actual situation is:

- `core.compact()` exists (v0.5.0, untouched) — it physically removes `retired` rows
- Retired rows only accumulate via explicit `plur forget` calls, which are rare in practice
- Retired rows are already filtered from all recall/search paths — store growth does not degrade search quality
- The real unbounded-growth concern (pre-batchDecay rows accumulating from inactivity) was a different code path, now resolved

**Accepted trade-off (option b):** The store is not strictly bounded, but growth is controlled by user behavior (forget is rare) and recall performance is unaffected by retired rows. The future design direction is archive-not-delete — retain retired engrams for user inspection before permanent removal — which is meaningfully different from the current `compact()` which deletes without inspection.

**No action needed until the archive design is scoped.** Closing in favor of a future `archive` design issue when that's deliberate.

🤖 heartbeat — 2026-07-16

PR #532, opened 2026-07-12
feat(cli,mcp): wire plur login command + server hot-reload helpers (closes #300)

plur9:
## Summary

Wires the `plur login` OAuth device-flow command that was fully implemented but never registered in the CLI dispatcher, and adds the companion server-side hot-reload helpers.

**Problem:** `plur login` existed as 479 lines in `packages/cli/src/commands/login.ts` but was missing from the `COMMANDS` map — `npx @plur-ai/mcp login` returned "Unknown command: login".

## Changes

**`packages/cli/src/index.ts`** (+2 lines)
- Register `login: './commands/login.js'` in the `COMMANDS` map
- Add `login <host>` entry to the `--help` text

**`packages/cli/src/commands/login.ts`** (new — 479 lines)
- OAuth 2.0 Device Authorization Grant (RFC 8628) flow
- Writes enterprise token to `~/.plur/config.json`
- Signals running MCP server via SIGUSR1 (or reload-marker file on Windows)
- Falls back gracefully when no MCP server is running

**`packages/mcp/src/server.ts`** (+49 lines)
- `serverPidPath(baseDir?)` — returns path to server PID file
- `readEnterpriseToken(baseDir?)` — reads enterprise token from config.json
- `isPendingReload()` / `clearPendingReload()` — hot-reload flag lifecycle
- `runStdio()` now writes PID file on startup and registers SIGUSR1 handler

**`packages/cli/test/login.test.ts`** (new — 325 lines, 22 tests)
**`packages/mcp/test/login.test.ts`** (new — 92 lines, 8/8 pass)

## Test plan
- [x] `pnpm test -- packages/mcp/test/login.test.ts` → 8/8 pass
- [x] `pnpm --filter @plur-ai/mcp build` → success
- [x] `pnpm --filter @plur-ai/cli build` → success
- [ ] CLI login tests (22 tests) — 20/22 expected pass; 1 pre-existing isolation issue (`readPlurConfig returns {} when file is absent` reads real homedir in-process), 1 skipped
- [ ] CI green on Node 20/22/24

## Follow-ups (not in scope)
- Fix test isolation in `packages/cli/test/login.test.ts` — refactor `readPlurConfig`/`writePlurConfig` to accept optional `baseDir` param
- `plur login --status` shows token validity (currently shows raw config)

🤖 heartbeat — 2026-07-12

Issue #300, opened 2026-06-09
plur login: one-command enterprise reauth (token mint + config write + hot-reload)


crtahlin:
## Problem

Re-authenticating an expired enterprise token is entirely manual and undiscoverable. There is **no CLI login** — `@plur-ai/mcp` exposes only `init`. Today's flow (documented in #295): open `<host>/auth/github` or `<host>/me/api-keys` in a browser, copy the token, hand-edit `~/.plur/config.yaml`, then **restart** Claude/MCP so it reloads the cached token. A headless agent can't do step 1, and the hand-edit + restart is error-prone.

#295 makes the failure *legible* (doctor + session_start now warn, with reauth instructions). This issue is the *fix*: make reauth a one-command action.

## Proposed

- [ ] **`plur login [--url <enterprise-url>]`** — device/OAuth flow (or paste-token fallback) that mints a token, writes it to `~/.plur/config.yaml` (`stores[].token`), and hot-reloads it so no restart is needed. Surface the resulting expiry.
- [ ] **Hot-reload of `config.yaml`** so a token change is picked up without restarting the MCP process (today the token is cached at construction — see #295 notes). Either a file-watch or a `plur_config_reload` tool.
- [ ] **Prefer longer-lived / refreshable credentials.** The current tokens are 30-day JWTs with no refresh, so silent expiry recurs monthly by design. Support `plur_sk_…` API keys end-to-end and/or a refresh-token path. (Server-side coordination with plur-ai/enterprise likely needed.)
- [ ] After reauth, trigger an outbox flush so queued team-scoped engrams sync immediately (rather than waiting for the next `session_start`).

## Out of scope

Anything already shipped in #295 (observability: doctor remote check, session_start auth/expiry warnings, `_outbox.auth_failed` flag).

## Context

Filed as the fast-follow to #295 (observability) per the implementation plan. Same incident: a 30-day enterprise JWT lapsed and ~4 days of team engrams queued silently before anyone noticed.

Issue #588, opened 2026-07-15
Audit 0.14.0 before publish (independent review)


plur9:
0.14.0 is staged for release on `main` (@ f2cbdc6) and ready for an independent audit before publish. This is the clean re-land of the batch that 0.12.0 shipped unreviewed and 0.13.0 pulled back — every held-back feature was individually re-audited and fixed this cycle.

## What shipped (both PRs merged to main)

- **#579** — pre-release hardening: batchDecay removal (#563), and fixes for the U+2028/U+2029 scope-metadata gap (#345), multi-word historical-keyword matching (#481), `feedback()` `last_accessed` re-anchor, and the manifest-gate multi-issue-trailer bug.
- **#585** — audit follow-ups: `plur compact` CLI+MCP (#580), doctor pglite+gemma advisory (#581), dead-code cleanup (#582), claw orphaned-entry warning (#583), publish/observability hardening (#584).

The full `## 0.14.0` CHANGELOG declares all 33 user-facing PRs since v0.12.0.

## What's already been verified (internal 5-evaluator audit)

Five evaluators (CTO, Dijkstra, Data, Critic, Taleb) reviewed the batch before release:
- CTO ran the full suite twice green; Dijkstra proved #281/#451/#217 correct by invariant.
- All findings were either fixed (this cycle) or filed as follow-ups (#580–#584, now merged).
- Full workspace suite: **2,626 passed / 0 failed**. Release dry-run manifest gate: **all 33 PRs declared, passes.**

## Ask for this audit

A fresh pair of eyes on the highest-blast-radius changes, independent of the internal audit:

1. **#563 batchDecay removal** — confirm no read path still assumes materialized decay/status; confirm the `dormant`/`candidate` enum retention (for pre-#563 data) is the right call vs. a migration.
2. **#345 scope-metadata hardening** — the claim is "hardens against *structural* injection" (not a content filter). Sanity-check the U+2028/U+2029 addition and that the local config path is correctly exempt.
3. **#281 breaking output contract** — `plur_learn_batch` `ids` is now 1:1 null-padded. Confirm the CHANGELOG "breaking" callout is sufficient for downstream MCP consumers.
4. **#51 / #583 claw config** — the prune was removed (was deleting third-party API keys) and replaced with a non-destructive doctor warning. Confirm no path still mutates foreign config.
5. Anything the internal audit's own blind spots would miss.

Refs: PRs #579, #585. Follow-up issues (all addressed): #580, #581, #582, #583, #584.

[...]

PR #594, opened 2026-07-15
ops(release): deactivate plur login + remove plur compact from 0.14.0

plur9:
Reshapes 0.14.0's product surface per two decisions:

## `plur login` (#532) — deactivated, code retained
Enterprise OAuth login is happy-path only: **no paste-token fallback**, a hard dependency on server device-flow endpoints, and no refresh tokens (the 30-day-expiry root cause of #300 is unaddressed). Instead of shipping it half-baked or reverting it, the command is **un-wired from the CLI dispatcher** — `plur login` returns "Unknown command" (its exact pre-#532 state) — while `login.ts` stays in the tree intact. Re-adding one dispatcher line reactivates it once hardened for 0.15.0. Process-level CLI tests are `describe.skip`'d; the helper unit tests still exercise the dormant functions.

## `plur compact` (#580) — removed
Added by a subagent this cycle to close the audit's "unbounded growth" finding, without the command surface being a deliberate call. Over-surfaced: GC is an operator task, not an agent task (the MCP tool shouldn't exist), and retired rows only accrue via `forget` (rare) and are already filtered from recall. Removed the CLI command + `plur_compact` MCP tool + their tests. The pre-existing core `compact()` (shipped since v0.5.0) is untouched but no longer surfaced. The better long-term direction — **archive** retired engrams (recoverable), not delete — is a future design, not built here.

## CHANGELOG
Forward-looking value-first lead (reliability + CLI pack management); dropped the "clean re-release / batchDecay" process framing. #532 declared honestly as landed-but-inactive groundwork so the manifest gate passes without misrepresenting login as shippable.

## Verification
- Full suite: **2,641 passed / 0 failed**
- Manifest gate: all declared, **zero undeclared**
- Auto-tweet: **262/270** ✓
- `plur login` / `plur compact` both return "Unknown command" ✓

Stacks on the merged #589. Prep for 0.14.0 publish (held for your go + tweet approval).


crtahlin (dismissed):
## Approve

This is the correct, clean resolution of the #588 audit finding. Verified end to end:

- **Manifest gate passes at branch HEAD** — 34 shipped feat/fix PRs, **zero undeclared**; #532 is now both shipped and declared (honestly, as inactive groundwork). This is the fix for #588's "login riding on top, undeclared."
- **`plur compact` fully removed** — `compact.ts`, the `plur_compact` MCP tool, the `CURSOR_CORE_TOOL_NAMES` entry, and all tests; changelog entry dropped and the `#580–#584`→`#581–#584` follow-up line corrected. Core `compact()` retained but unsurfaced. Grep confirms no stray `plur_compact` references.
- **`plur login` deactivation is clean** — dispatcher entry replaced with a reactivation comment (#300), CLI process tests `describe.skip`'d with rationale, helper unit tests still run.
- **Full suite: 2641 passed / 0 failed** locally after a full `pnpm -r build` (matches CI on Node 20/22/24).

### One thing to handle at merge (so it doesn't abort the release)

This commit is typed **`feat(release):`**. Once it lands on `main`, the `(#594)` trailer becomes a *shipped user-facing PR* that isn't declared in the CHANGELOG — so `release.sh`'s manifest gate would abort the 0.14.0 publish on **#594 itself**, the exact failure class this PR fixes for #532. The prior release-tooling PR #548 used **`ops(release):`**, which the gate curates out.

**Recommend retyping the merge/squash commit to `ops(release):` (or `chore(release):`)** — it's also more accurate (this *removes* surface), and it keeps the release from tripping on itself. (Tagging this branch's tip directly would also avoid it, since the current commit has no `(#594)` trailer — but the standard merge-to-main path needs the retype.)

Approving on the code; just don't let #594 merge in as a bare `feat` before the 0.14.0 cut.

crtahlin:
@plur9 see note before merging

crtahlin:
@plur9 — before merging this: **retype the commit from `feat(release):` to `ops(release):`** (or squash-merge with an `ops(release):` subject).

Reason: as `feat(release):`, the `(#594)` trailer lands on `main` as a shipped user-facing PR that is not declared in the 0.14.0 CHANGELOG — so `release.sh`'s manifest gate will abort the 0.14.0 publish on #594 itself, the exact failure this PR fixes for #532. `ops` is in the gate's curated-out set (same as #548 `ops(release): manifest gate …`), so the retype makes it pass. It's also more accurate — this PR removes surface, it isn't a feature.

Everything else is approved and verified (2641 pass / 0 fail locally, manifest gate clean on the branch). Just don't let it merge as a bare `feat`.

plur9:
Thanks @crtahlin — sharp catch, and exactly the self-referential gate trap this PR fixes for #532. Fixed: the commit is retyped `ops(release):` (231cae7) and the PR title is retyped to match, so the squash subject lands as `ops(release): … (#594)` and the gate curates it out (same as #548). It's also the accurate type — this removes surface. Verified below that the gate stays clean with #594 folded in as `ops`.

Issue #202, opened 2026-05-20
Closed-loop training of retrieval relevance from plur_feedback


plur9:
## Context

Kumo's distillation step relies on having ground-truth labels for the relational task at training time. PLUR already collects exactly such labels via `plur_feedback` (positive/negative on injected engrams) — but nothing consumes them.

Reference: https://kumo.ai/company/news/low-latency-high-throughput-predictions-with-kumorfm-2-fine-tuning/

## Problem

Every thumbs-up/down on an injected engram is a labeled example of *"in context C, engram E was relevant / not"*. Today this signal sinks into the store but no model learns from it. Without a closed loop, retrieval can't improve from usage.

## Proposal

- Convert `plur_feedback` events into training pairs: `(query/session_context, engram_id, label)`.
- Train the Stage-1 / Stage-2 model components against this dataset on `plur_consolidate` cadence.
- Track lift over time as labeled data grows.

## Acceptance

- [ ] Feedback events queryable as a labeled training set
- [ ] Periodic training job consumes them and updates retrieval weights
- [ ] Eval split shows improving relevance over weekly snapshots
- [ ] Surfaced in `plur_status` so users can see "retrieval has learned from N feedback events"

## Depends on

- Two-stage retrieval issue (provides the model to train)

plur9:
**Reframed telemetry-first per the 2026-07-02 graph assessment (see #225):** a single-user store generates feedback labels in the hundreds — not enough to train a distilled scorer. Phase this as: (1) label collection now — log injection outcomes + feedback verdicts as training-shaped records; (2) a label-volume report in `plur_status` (weekly count, per-scope); (3) training gated on a stated threshold (e.g. 5k labels). Kumo-style temporal-replay self-labeling (auto-generate labels by replaying past injections against later feedback) is the multiplier worth designing in from the start.

crtahlin:
### Status verification — partially done

**Done:**
- Co-injection and injection-outcome provenance events — [PR #468](https://github.com/plur-ai/plur/pull/468) (merged 2026-07-02; closes [#452](https://github.com/plur-ai/plur/issues/452)): `co_injection` and `injection_outcome` events logged to history.ts; counts surfaced in `plur_status` output
- Event counts visible in `plur status` CLI — [packages/cli/src/commands/status.ts](https://github.com/plur-ai/plur/blob/main/packages/cli/src/commands/status.ts)

**Still open:**
- Periodic training job that consumes feedback events and updates retrieval weights — not started
[...]

PR #601, opened 2026-07-16
fix(core): restore cross-process ID uniqueness with per-process PID salt

### LINKED ISSUE #600: generateInjectionId/generateEventId: per-process counter is not globally unique (cross-process same-ms collision)
## Context

## Problem

## Proposed fix

- `process.pid` in base36.

## Acceptance

plur9:
## Problem

PR #596 replaced `Math.random()` with per-process monotonic counters in `generateInjectionId()` / `generateEventId()`, eliminating intra-process collisions. But the counters reset to `0` on every new process — two concurrent processes (e.g. the persistent MCP server + a hook-spawned CLI) generating their first ID in the same millisecond both produce `…-<ts>-0000`. Deterministic, not probabilistic.

The injection ID is the correlation key for the `co_injection → injection_outcome` feedback pipeline (#202). A cross-process collision mis-attributes a feedback label to the wrong injection, degrading self-labeling quality.

Closes #600.

## Fix

Fold a **2-char base-36 PID-derived salt** into the suffix, computed once at module load:

```
suffix = <2-char pid salt> + <2-char counter>   ← always 4 chars total
```

The salt is `(process.pid % 1296).toString(36).padStart(2, '0')`. Different processes have different PIDs → different salts → no cross-process collision.

The existing `[a-z0-9]{4}` test regex and `INJ-<ms>-<4-chars>` / `EVT-<ms>-<4-chars>` format are **unchanged**.

## Secondary (from #600)

Also resolves the overflow concern: capping the counter at `36² = 1296` before converting to base-36 means it never grows beyond 2 chars. No more silent format drift past `zzzz → 10000` in long-lived processes. The intra-ms ceiling of 1296 IDs per process is unreachable in any realistic workload.

## Acceptance

- [x] Cross-process: different PIDs → different salts → no same-ms collision
- [x] Intra-process: counter still monotonic within a process (deterministic from #596 preserved)
- [x] Format: suffix always exactly 4 base-36 chars — existing regex unchanged
- [x] Overflow: counter caps at 1296 (2 chars) — no format drift; documented in JSDoc
- [x] Tests: 23/23 co-injection-logging tests pass

### DIFF
--- a/packages/core/src/history.ts
+++ b/packages/core/src/history.ts
@@ -77,16 +77,25 @@ export function readHistoryForEngram(root: string, engramId: string): HistoryEve
   return events
 }
 
+// Two-character base-36 salt derived from this process's PID.
+// Different processes have different PIDs → different salts → no cross-process
+// collision even when two processes generate an ID in the same millisecond.
+// PID space: Linux default max 32768 → 0–'p0' in base36 (always ≤2 chars).
+const _pidSalt = (process.pid % 1296).toString(36).padStart(2, '0')
+
 let _evtSeq = 0
 let _injSeq = 0
 
 /**
  * Generate a unique event ID for history entries.
- * Uses a per-process counter combined with timestamp to guarantee uniqueness
- * even when called in rapid succession within the same millisecond.
+ * Format: `EVT-<ms>-<2-char pid salt><2-char counter>` (suffix always 4 base-36 chars).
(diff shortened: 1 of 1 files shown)


plur9:
Superseded by #609 — clean 2-file fix on current main. This branch was stale (branched before several mainline merges). Closing in favour of the fresh PR.

PR #608, opened 2026-07-16
fix(core): cross-process uniqueness for generateInjectionId/generateEventId

### LINKED ISSUE #600: generateInjectionId/generateEventId: per-process counter is not globally unique (cross-process same-ms collision)

plur9:
Fixes #600.

## What

Replaces `Math.random()` suffix in `generateInjectionId()` / `generateEventId()` with a **PID salt + monotonic counter** approach.

**New format:** `PREFIX-<ts>-<2-char-pid-salt><4-char-counter>` — 6 chars `[a-z0-9]{6}` (up from 4).

- `_PROC_SALT = (process.pid % 1296).toString(36).padStart(2, '0')` — computed once at module load, 2 chars
- `_evtSeq` / `_injSeq` — per-process monotonic counters (intra-process deterministic uniqueness)

## Why

The MCP server and hook-spawned CLI processes can both call `generateInjectionId()` in the same millisecond. With `Math.random()`, this is a low-probability collision (~0.07% at 50 samples). With a pure counter that resets to `0` per process, it becomes deterministic: both processes emit `…-0000` for their first ID.

The injection ID is the correlation key for the `co_injection → injection_outcome` self-labeling pipeline. A cross-process collision mis-attributes feedback to the wrong injection.

## Overflow behaviour

Counter overflows to 5 chars past 36^4 (1,679,616 events/process). Not reachable in practice at 50 sessions/day. Documented in source; suffix widens to 7 chars, IDs remain unique.

## Test

Updated `co-injection-logging.test.ts` regex: `[a-z0-9]{4}` → `[a-z0-9]{6}`. 33 tests pass.

🤖 heartbeat

### DIFF
--- a/packages/core/src/history.ts
+++ b/packages/core/src/history.ts
@@ -77,13 +77,23 @@ export function readHistoryForEngram(root: string, engramId: string): HistoryEve
   return events
 }
 
+// Per-process 2-char salt (PID mod 1296, base36) prevents cross-process
+// same-millisecond collisions when the MCP server and a hook-spawned CLI
+// process both call generateInjectionId()/generateEventId() concurrently.
+// Suffix format: <2-char salt><4-char counter> = 6 chars [a-z0-9]{6}.
+// Counter overflows to 5 chars past 36^4 (1,679,616 events/process) — not
+// reachable in practice; suffix becomes 7 chars, IDs remain unique.
+const _PROC_SALT = (process.pid % 1296).toString(36).padStart(2, '0')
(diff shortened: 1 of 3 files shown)


plur9:
Superseded by #609 — this branch had 91 files diverged from main (branched from stale base). The actual fix (history.ts + test) is preserved and applied cleanly in #609.

PR #609, opened 2026-07-16
fix(core): PID salt for cross-process ID uniqueness (closes #600)

### LINKED ISSUE #600: generateInjectionId/generateEventId: per-process counter is not globally unique (cross-process same-ms collision)

plur9:
## Problem

PR #596 replaced `Math.random()` suffixes with per-process monotonic counters (`_evtSeq`, `_injSeq`), eliminating intra-process same-millisecond collisions. But counters reset to `0` on every new process — two concurrent processes (MCP server + hook-spawned CLI) generating their first ID in the same millisecond both produce `…-0000`. Deterministic, not probabilistic.

The injection ID is the correlation key for the `co_injection → injection_outcome` self-labeling pipeline (#202). A cross-process same-ms collision mis-attributes feedback to the wrong injection, degrading self-labeling quality.

## Fix

Prepend a 2-char per-process PID salt (`process.pid % 1296`, base36) to the monotonic counter.

**New format:** `PREFIX-<ts>-<2-char-pid-salt><4-char-counter>` — 6 chars `[a-z0-9]{6}` (up from 4).

- Cross-process collision probability: ~1/1296 (same as original `Math.random()` approach)
- Intra-process determinism from #596: preserved
- Overflow at 36⁴ (1,679,616 events/process): documented in comment, unreachable in practice

## Replaces

Closes stale PRs #601, #607, #608 — all correct in approach but branched from stale bases. This is a clean 2-file fix on top of current main.

## Tests

All 23 existing co-injection-logging tests pass. Test regex updated from `[a-z0-9]{4}` → `[a-z0-9]{6}` to match the new 6-char suffix.

## Acceptance checklist

- [x] Injection/event IDs unique across concurrent processes generating in the same millisecond
- [x] Intra-process deterministic uniqueness from #596 preserved  
- [x] Format documented (comment in history.ts)
- [x] Overflow behaviour past 36⁴ documented
- [x] Test regex updated

🤖 heartbeat

### DIFF
--- a/packages/core/src/history.ts
+++ b/packages/core/src/history.ts
@@ -77,16 +77,23 @@ export function readHistoryForEngram(root: string, engramId: string): HistoryEve
   return events
 }
 
+// Per-process 2-char salt (PID mod 1296, base36) prevents cross-process
+// same-millisecond collisions when the MCP server and a hook-spawned CLI
+// process both call generateInjectionId()/generateEventId() concurrently.
+// Suffix format: <2-char salt><4-char counter> = 6 chars [a-z0-9]{6}.
+// Counter overflows to 5 chars past 36^4 (1,679,616 events/process) — not
+// reachable in practice; suffix becomes 7 chars, IDs remain unique.
+const _PROC_SALT = (process.pid % 1296).toString(36).padStart(2, '0')
 let _evtSeq = 0
 let _injSeq = 0
 
 /**
- * Generate a unique event ID for history entries.
- * Uses a per-process counter combined with timestamp to guarantee uniqueness
- * even when called in rapid succession within the same millisecond.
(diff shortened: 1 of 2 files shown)


plur9:
CI all passing (test across Node 20/22/24). Code review: PID salt approach is correct — PID mod 1296 base36 padded to 2 chars gives 1296 unique salts, sufficient for practical concurrent-process cardinality. The 6-char suffix format (2-char salt + 4-char counter) is documented clearly in the inline comment. Test regex updated from `{4}` to `{6}` is consistent. Self-approve blocked; this is ready for human merge review.

PR #624, opened 2026-07-18
fix(core): RemoteStore.load() — no cache poisoning on mid-pagination error (closes #550)

### LINKED ISSUE #550: RemoteStore.load(): error path caches partial/empty result with fresh TTL (cache poisoning, reintroduces #130 class)
## Summary

`RemoteStore.load()` caches a **partial or empty** result set with a fresh timestamp when a page fetch or body read errors mid-pagination, so a truncated (or empty) view is served as authoritative for up to `ttlMs` (default 60s). This bypasses the outer catch's deliberate "don't poison the cache on failure" fallback.

Introduced as a side effect of PR #531 (the #525 abort-coverage fix). The abort-coverage fix itself is correct — this issue is only about what happens on the error path after the abort fires. **PR #531 does not need to block on this** (P3, self-heals after `ttlMs`); filing separately so the fix is tracked.

## Detail

In `packages/core/src/store/remote-store.ts`, the new inner error handler:

```ts
} catch (err) {
  break
} finally {
  clearTimeout(t)
}
```

`break` exits the pagination loop and falls through to:

```ts
this.cache = { ts: Date.now(), engrams: all }
return all
```

So on any thrown `fetch()` / `r.json()` error, the partial `all` accumulated so far is written to the cache with a fresh timestamp.

**Before #531**, thrown errors propagated to the outer catch, which preserves the prior cache on purpose:

```ts
} catch (err) {
  console.error(`[plur:remote-store] ${this.url} load failed: ${(err as Error).message}`)
  // Don't poison the cache on failure — let the next call retry.
  return this.cache?.engrams ?? []
}
```

#531 routes thrown errors away from that safe fallback into the poison path.

## Impact

- Page 1 succeeds (200 engrams), page 2 body stalls → AbortError → the 200 partial engrams are cached as fresh and served as the complete set for up to `ttlMs` (default 60s). This is the exact stall scenario #531 exists to handle.
- A transient error on page 1 caches `[]` with a fresh timestamp, hiding previously-good remote engrams for up to 60s.

This reintroduces the bug class that #130 already fixed. The test `packages/core/test/remote-store-cache.test.ts:75` guards a sibling case with the comment: *"the partial single-engram cache look 'fresh' for ttlMs and hid every other engram."* It also contradicts the "fail-open … no cache poisoning" invariant stated for #504/#505.

Note: the pre-existing `!r.ok` mid-pagination `break` (e.g. a 5xx on page 2) has the same partial-cache issue and predates #531 — worth addressing in the same fix.

## Suggested fix

Do not write a fresh cache on a non-clean exit. Options:
- Re-throw from the inner `catch` so the outer catch preserves the prior cache and returns it (or `[]`), or
- Only assign `this.cache` on natural pagination completion — distinguish an error/`!ok` break from a complete break (`all.length >= body.total_count || body.rows.length < limit`), and skip the cache write on the former.

Add a regression test: page 1 OK + page 2 error must not overwrite a good cache (mirror the #130 test).

## Minor (optional, from the same review)

`(err as Error).message` on an AbortError yields the generic "This operation was aborted." A `err.name === 'AbortError'` branch that logs the timeout bound would make diagnostics clearer.

---

Filed from review of #531. Non-blocking to that PR. Assigning to @plur9 to consider.

plur9:
## Summary

Fixes the cache poisoning regression introduced by #531. When a page fetch fails mid-pagination (AbortError, 5xx), the partial accumulation was written to the cache with a fresh timestamp — hiding all previously-good engrams for up to `ttlMs` (60s default).

- Track `paginationComplete` — only write `this.cache` on a full, clean pagination run
- On an incomplete exit: fall back to the prior good cache so callers keep seeing valid engrams
- 403/404 on page 1 are treated as "stable empty" and _do_ get cached (scope has no access; avoids repeated server hits per TTL expiry)
- AbortError log now prints the timeout bound instead of the generic "This operation was aborted."

Also fixes the pre-existing `!r.ok` 5xx-on-page-2 partial-cache issue called out in the issue (same root cause, predates #531).

## Test plan

- [x] `page 1 OK + page 2 AbortError — does not overwrite a good prior cache` 
- [x] `5xx on page 2 does not overwrite prior good cache`
- [x] `transient page 1 error caches nothing — prior cache preserved`
- [x] `403 on page 1 caches empty result (stable state — scope has no access)`
- [x] All 13 existing remote-store-cache tests still pass (2 skipped, pre-existing)

Closes #550.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -49,7 +49,7 @@ pnpm --filter @plur-ai/core build
 
 `scripts/release.sh` is the authoritative source. Two independent version tracks:
 
-**Standard release** (core / mcp / cli — always bumped together):
+**Standard release** (core / mcp / cli / hermes / python — always bumped together):
 
 1. `packages/core/package.json`
 2. `packages/mcp/package.json`
@@ -59,10 +59,12 @@ pnpm --filter @plur-ai/core build
 6. `packages/cli/src/index.ts` — `const VERSION`
 7. `packages/mcp/test/server.test.ts` — version assertions
 8. `packages/hermes/pyproject.toml`
-9. `packages/hermes/plur_hermes/skills/plur-memory.SKILL.md` — frontmatter `version:`
-10. `packages/hermes/plur_hermes/bridge.py` — `_NPX_CLI_VERSION`
-11. `packages/python/pyproject.toml`
-12. `packages/python/plur_ai/bridge.py` — `_NPX_CLI_VERSION`
+9. `packages/hermes/plugin.yaml` — top-level `version:` field
+10. `packages/hermes/plur_hermes/skills/plur-memory.SKILL.md` — frontmatter `version:`
+11. `packages/hermes/plur_hermes/bridge.py` — `_NPX_CLI_VERSION`
+12. `packages/python/pyproject.toml`
+13. `packages/python/plur_ai/bridge.py` — `_NPX_CLI_VERSION`
+14. `skills/plur-memory/SKILL.md` — frontmatter `version:` (standalone skills.sh copy)
 
 **Claw track** (independent — only bumped when `--claw <ver>` is passed to release.sh):
 
@@ -71,6 +73,11 @@ pnpm --filter @plur-ai/core build
 - `packages/claw/src/context-engine.ts` — `version:` in info object
 - `packages/claw/openclaw.plugin.json` — `version` field
 - `packages/claw/test/hello.test.ts` — version assertion
+
+**Langchain track** (independent — bumped separately when `plur-langchain` ships):
+
+- `packages/langchain/pyproject.toml`
+- `packages/langchain/plur_langchain/__init__.py` — `__version__`
 
 ## Publishing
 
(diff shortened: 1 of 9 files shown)


plur9 (commented):
## LGTM — merge when ready

**Cache poisoning fix verified.** The `paginationComplete` flag approach is the right invariant: only write the cache on a confirmed complete run; on any mid-pagination failure, return the prior good cache (`this.cache?.engrams ?? all`).

Key observations:
- 403/404 vs 5xx split is correct — 403/404 are *stable* (scope has no access, safe to cache empty); 5xx are *transient* (don't cache partial)
- Cold-cache + transient-error fallback (`?? all`) returns partial rather than empty, which is the right tradeoff
- AbortError message improvement is a nice bonus
- 4 tests cover all 4 behavioral paths cleanly

Docs changes (seven-package inventory, test count update, version bump checklist, pin removal in README/CLI README) are accurate and overdue. hermes/SKILL.md version alignment to 0.14.0 is correct.

CI green. Ready to land. 🟢

PR #642, opened 2026-07-21
fix(core): use path.dirname in saveCache — fixes ENOENT on Windows (#641)

### LINKED ISSUE #641: plur_recall_hybrid fails with "ENOENT: mkdir ''" in headless/scheduled runs (empty session-dir path); plur_doctor still reports healthy
## Summary

`plur_recall_hybrid` (and any hybrid path) fails with `ENOENT: no such file or directory, mkdir ''` when the MCP server runs in a **headless / non-interactive context** (scheduled task, cron, CI — anywhere without a session/transcript directory set). Plain `plur_recall` (BM25) works fine in the same session, and `plur_doctor` reports the engine as healthy — so the failure is isolated to hybrid's extra machinery, not the store or embedder.

An unguarded `mkdir('')` on an **empty-string path** is the proximate cause.

## Environment

- PLUR version: `0.14.0`
- OS: Windows 11
- Store: `~/.plur` (90 engrams, intact)
- Invocation context: MCP server running under a **non-interactive scheduled task** (no attached session/transcript directory). Reproduces in any headless run.

## Repro

1. Launch the PLUR MCP server in a headless context (scheduled task / cron / CI — no session/transcript dir provided by the harness).
2. Call `plur_recall_hybrid` with any query, e.g.:
   ```json
   { "query": "anything", "limit": 3 }
   ```
3. Tool returns:
   ```json
   { "error": "ENOENT: no such file or directory, mkdir ''", "success": false }
   ```

Meanwhile in the **same** session:
- `plur_recall` (BM25) → works, returns results.
- `plur_learn`, `plur_status` → work.
- `plur_doctor` → `ok` for engine checks (see below).

## What `plur_doctor` reports (engine looks healthy — masks the bug)

```
engram store              ok    90 engrams at C:\Users\scuba\.plur
embedder available        ok    BGE-small-en-v1.5 loaded
hybrid search operational ok    Hybrid search will use BM25 + embeddings (fully functional)
```

`plur_doctor --retry` shows the embedder transitioning `before_probe.loaded: false` -> `after_probe.loaded: true`. **But `plur_recall_hybrid` still throws `mkdir ''` even after the embedder is confirmed loaded** — so the empty-path failure is *not* the embedder load; it is a separate directory-creation step on the hybrid path.

## Likely root cause

The empty string strongly suggests a scratch/cache directory that is normally derived from a **session- or transcript-scoped path** supplied by the interactive harness, but which is **empty (`''`)** in a headless run. The hybrid code path (RRF fusion / sub-agent budget workspace / per-session embeddings scratch) calls `mkdir(dir)` (or `fs.mkdirSync(dir, { recursive: true })` with `dir === ''`) without guarding against an empty/undefined base path. `plur_doctor` computes its own paths internally, so it never hits this branch — which is why it reports "operational" while the actual tool invocation crashes.

## Suggested fix

- Guard the hybrid-path `mkdir` against empty/undefined directory values; fall back to a well-defined default (e.g. under the store root `~/.plur/...`) instead of `''`.
- Consider having `plur_doctor`'s "hybrid search operational" check exercise the **same** directory-resolution path the `plur_recall_hybrid` tool uses, so a headless/empty-session-dir misconfiguration surfaces in the doctor rather than only at call time.

## Impact / workaround

Non-blocking where a BM25 fallback is acceptable — automated jobs can use `plur_recall` for dedup/lookup. But any scheduled/headless workflow that relies on semantic hybrid recall silently loses it, and the healthy `plur_doctor` report makes it hard to notice.

plur9:
## Root cause

`saveCache` in `embeddings.ts` extracted the directory by scanning for the last `/`:

```ts
const dir = cachePath.substring(0, cachePath.lastIndexOf('/'))
```

On Windows, paths use backslashes (`C:\Users\…\.plur`). `lastIndexOf('/')` returns `-1`, so `substring(0, -1)` returns `''`, and `mkdirSync('')` throws `ENOENT`. The embedder model loads fine (as `plur_doctor` correctly reports), but the first `saveCache` call on the hybrid path throws before writing the cache.

## Fix

Replace the manual slash-scan with `path.dirname()`, which is cross-platform:

```ts
const dir = dirname(cachePath)
if (dir && !existsSync(dir)) mkdirSync(dir, { recursive: true })
```

`dirname('something.json')` returns `'.'` (always exists) instead of `''`.

## Test

New offline test `embeddings-save-cache-path.test.ts` uses `_setCachedEmbedder` to inject a mock embedder (no model download). Four cases: write, cache-hit on second call, cache-header validity, and `embeddingSearchWithScores` variant. All pass.

## Closes

Closes #641

### DIFF
--- a/packages/core/src/embeddings.ts
+++ b/packages/core/src/embeddings.ts
@@ -2,7 +2,7 @@ import type { Engram } from './schemas/engram.js'
 import type { EmbedRole } from './embedders/types.js'
 import { engramSearchText } from './fts.js'
 import { existsSync, readFileSync, mkdirSync } from 'fs'
-import { join } from 'path'
+import { join, dirname } from 'path'
 import { createHash } from 'crypto'
 import { atomicWrite } from './sync.js'
 import { logger } from './logger.js'
@@ -308,8 +308,8 @@ function loadCache(cachePath: string, active: { name: string; dim: number }): Em
 }
 
 function saveCache(cachePath: string, cache: EmbeddingCache): void {
-  const dir = cachePath.substring(0, cachePath.lastIndexOf('/'))
-  if (!existsSync(dir)) mkdirSync(dir, { recursive: true })
+  const dir = dirname(cachePath)
+  if (dir && !existsSync(dir)) mkdirSync(dir, { recursive: true })
   atomicWrite(cachePath, JSON.stringify(cache))
 }
 
--- a/packages/core/test/embeddings-save-cache-path.test.ts
+++ b/packages/core/test/embeddings-save-cache-path.test.ts
@@ -1,0 +1,113 @@
+/**
+ * Regression test for #641: saveCache used `cachePath.substring(0, cachePath.lastIndexOf('/'))`
+ * to extract the directory. On Windows, paths use backslashes so lastIndexOf('/') returns -1,
+ * and substring(0, -1) returns '' — causing mkdirSync('') to throw ENOENT even when storagePath
+ * is a fully-valid directory. Fix: use path.dirname() instead of the manual slash scan.
+ *
+ * This test uses _setCachedEmbedder to inject a mock embedder so it runs offline, no model download.
+ */
+import { describe, it, expect, beforeEach, afterEach } from 'vitest'
+import { mkdtempSync, rmSync, existsSync, readFileSync } from 'fs'
+import { join } from 'path'
+import { tmpdir } from 'os'
+import { embeddingSearch, embeddingSearchWithScores, _setCachedEmbedder, resetEmbedder } from '../src/embeddings.js'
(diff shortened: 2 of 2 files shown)


plur9:
CTO review (Miles): fix is correct and minimal. `path.dirname()` is the right cross-platform replacement for the manual slash-scan — handles Windows backslash paths, POSIX, and edge cases equally. The `if (dir && ...)` guard covers any hypothetical empty-dirname result. Regression test suite is thorough: 4 cases, mock embedder so it runs offline in CI, covers both search entry points + cache-header validation + cache-hit path. 4/4 CI checks green (Node 20/22/24 + test suite). P0 hotfix — tagged accordingly. Ready to merge.

Issue #671, opened 2026-07-23
Backfill the 803 no-domain engrams (convention + capture nudge shipped) — #345 follow-up


crtahlin:
## Problem

Scope routing's strongest — and in practice only reliable — signal is the engram's dotted `domain`, but there is no domain convention and a large fraction of engrams carry no domain at all. Two coupled gaps:

### 1. ~35% of engrams have no `domain`
Enumerated from a live store: **803 of 2261 engrams (35%) have no `domain`**, across **567 distinct free-form domains** otherwise. `domain` is caller-supplied to `plur_learn` (`domain: "plur.core.security"`); nothing infers or defaults it. Because a domain namespace-prefix match is the only channel that reaches the 0.5 auto-route threshold (tags/keywords top out ~0.35), **an engram with no domain effectively cannot auto-route** — it falls to `global`/`personal:*` regardless of covers. So a third of memory is unroutable by construction.

### 2. No domain registry / convention
Domains are free-form (this is the still-open question on the closed epic #345: *"Schema-validate `covers[]` against a domain registry, or free-form?"*). Observed domains are inconsistent: `plur.enterprise`, `war-analysis.iran`, `provenance.gateway`, `software.testing`, `reference.companies`, `datacore.today`, … With no agreed shape, covers can't be authored to match reliably, and the same content lands under `plur.*` in one session and `software.*` in another.

The two interact: even where a convention exists (we standardized on `plur.<team>.<area>` and set matching `plur.<team>` covers — see #668), it only works if agents actually set a conforming `domain`.

## Proposal (decide first; this is a convention + light enforcement)
1. **Adopt a documented domain convention**: `plur.<team>.<area>` for team knowledge (mirrors the covers now set — engineering→`plur.engineering.*`, comms→`plur.comms.*`, research→`plur.research.*`, leadership→`plur.leadership.*`, org→`plur.company.*`/`plur.org.*`).
2. **Agent guidance (behavioral, no code)**: capture agents must set a conforming `domain` on every `plur_learn`; document in CLAUDE.md / MCP scope-selection guidance.
3. **Light enforcement (code, optional)**: at `learn()` time, when `domain` is absent, either (a) warn/surface it, or (b) offer `plur_suggest_scope`-style domain inference from covers, or (c) validate against a soft registry with a warning (not a hard reject — keep the tolerant philosophy).
4. **Registry decision**: free-form-with-convention vs a validated registry (resolves the #345 open question).

[...]

Issue #670, opened 2026-07-23
Scope ranker quality: negation-blind + unresolved keyword-tie collisions (deterministic-lexical limits)


crtahlin:
## Problem

The Stage-3 scope ranker (`rankScopes` in `packages/core/src/scope-routing.ts`) is deterministic and purely lexical (bag-of-words over `covers[]`). A routing audit surfaced two quality gaps that `covers[]` tuning cannot fix — they are structural in the ranker.

### 1. Negation-blind
`not` is in the ranker's `STOPWORDS` and is discarded before matching, so a statement's negation is invisible. Observed:

- Statement: *"This is NOT a marketing document — it documents the CLI argument parser internals."* (domain `plur.cli.parser`)
- Result (before the covers fix): **comms 0.1176** (keyword `marketing`) tied with engineering 0.1176 (keyword `cli`).

A document explicitly disclaiming marketing scores the comms scope. Synonyms are also invisible (`ai-memory` ≠ "AI data economy"), which is the false-negative side of the same lexical limit.

### 2. Keyword-only ties are unresolved / arbitrary
Two scopes matching one weak keyword each tie at identical confidence with no principled tie-break:

- *"Blog post positioning our new benchmarking numbers for the launch"* (no domain) → **comms 0.1176** (`positioning`) ≈ **engineering 0.1176** (`benchmarking`). A blog post is clearly comms, but engineering ties purely because it lists `benchmarking`.

Tie-break today is "prefer domain-prefix, then scope name ascending" — so on keyword-only ties the winner is effectively alphabetical, not semantic.

## Why it matters
Low severity for safety (worst case is a weak *suggestion* into a still-authorized scope; access control and `sensitivity` are enforced server-side per request, and an explicit scope bypasses covers). But it degrades auto-routing quality and produces confusing suggestions, and it worsens as more scopes get covers.

## Options (decide, don't necessarily build)
1. **Accept as a documented limitation** — the ranker is deliberately deterministic/no-LLM; covers + a good `domain` are the intended path (a domain match at 0.5 out-ranks these 0.12 keyword hits, as verified after adding namespaced covers). Document that keyword-only routing is best-effort.
2. **Negation handling** — detect a small set of negation cues near a matched token and suppress that keyword hit.
3. **Tie-break by specificity** — when confidences tie, prefer the higher `coverSpecificity` (already computed on each candidate) before falling back to name order.
[...]

Issue #648, opened 2026-07-21
Scope-metadata edits not reflected in running clients (cached until next /me pull) — freshness signal?


crtahlin:
### What happens today
Server-authoritative scope metadata (`description`/`covers[]`, #345) is delivered to clients via `/api/v1/me` (`scope_metadata`) and **cached locally**. The Stage-3 write-scope ranker (`suggestScope`) routes on the cached `covers`. A client only refreshes on its next `/me` pull — i.e. **session start** or **`plur_scopes_discover`**. So when an admin edits a scope's `covers`, existing sessions keep routing on the old covers until they restart or re-discover.

### Why it's low priority (but worth noting)
It is **not** a correctness or isolation problem, and that's deliberate:
- Access control and the `sensitivity` (forbid-secrets) guard are enforced **server-side by membership on every request** — never by cached covers.
- An **explicit scope** on a write bypasses `covers` entirely.
- So the worst case of staleness is a **suboptimal auto-*suggestion*** into a still-authorized scope — never a wrong/unsafe write.

### Why it's still worth a fix eventually — the motivation
It's a **"why isn't my change taking effect?"** surprise for admins tuning routing: they edit `covers`, expect new writes to re-route, and nothing changes for already-running sessions. As more scopes get `covers` populated, this confusion (and support noise) grows. It also weakens the feedback loop when iterating on a routing taxonomy — you can't see the effect of an edit without restarting.

### Cheap options (pick per appetite)
- **Freshness signal:** server exposes a `scope_metadata` `updated_at`/version (or ETag) so clients can invalidate their cache with a cheap check instead of a full refetch.
- **Shorter client TTL** on the scope-metadata cache.
- **UX hint only:** the admin `/admin/scopes` page notes "changes apply to clients on their next session or `plur_scopes_discover`."

Priority: **low** — quality-of-life / DX, not correctness. Relates to #345 (self-describing scopes) and plur-ai/enterprise#523 (populating scope metadata).

plur9:
**CTO triage — agree with the analysis, here's the direction:**

The security framing is exactly right: access-control is enforced server-side on every request, `covers` only drives the *suggestion* layer, and an explicit scope bypasses it entirely. So this is purely a DX issue — a "why isn't my change taking effect?" surprise for admins iterating on routing taxonomy. No correctness or isolation risk.

**Chosen option: UX hint first, freshness signal later**

[...]

PR #673, opened 2026-07-23
fix(core): sync server scope covers[] to local config so routing works (#668)

### LINKED ISSUE #668: suggestScope inert for remote scopes: server covers[] never persisted to local config, so capture-time routing can't work
## Problem

### Evidence

## Proposed fix

## Acceptance criteria

## Tests

## Docs to update

## Note

crtahlin:
Closes #668.

## Problem

The Stage-3 scope ranker (`suggestScope` → `getScopeMetadata` → `listScopeMetadata`) routes on the `covers[]` declared on **local config store entries**, but nothing ever populated them from the server. The server serves `covers[]` via `/api/v1/me` (`scope_metadata`, #345 D2), yet no client path wrote them into `config.stores[]` — so on a normal enterprise install the ranker returns `[]` for every remote scope, and team-relevant writes silently default to `global`/`personal:*`.

This was verified against a live store: `plur_suggest_scope` returned zero candidates even for content built verbatim from a scope's covers. #648 (metadata freshness) presupposed a populated cache that never existed; this is the missing prerequisite.

## Fix

`discoverRemoteScopes` — the single `/me`-pull chokepoint that `session_start`, `plur_scopes_discover`, and `registerScope` all funnel through — now mirrors server metadata into local config via a new private `_syncScopeMetadataToConfig(discoveries)`:

- **Registered entries only**, matched by `url`+`scope`. Never creates entries; local/path stores and personal-family scopes carry no server metadata and are untouched.
- **Server is authoritative** for `covers` (also keeps it fresh — #648); `description`/`sensitivity` are synced when the server declares them.
- **Idempotent** — persists (via the existing `persistStores` read-preserve-write) only when a value actually changed, so repeat discovery causes no config churn.
- **Fail-safe** — a failed `/me` (`ok:false`) contributes nothing, so existing covers are never wiped.

No schema change (`StoreEntry` already carries `covers`/`description`/`sensitivity`); no new exports; no enterprise-side change (the server already serves covers and reads its own authoritative DB — this bug is client-only).

## Tests

New `packages/core/test/scope-metadata-sync.test.ts` (6 integration tests, real HTTP stub): covers persisted + `suggestScope` activated end-to-end; idempotent second pull; refresh-on-change; no entry created for authorized-but-unregistered scopes; local stores untouched; covers preserved on unreachable `/me`.

- Full **core** suite green: **1986 pass** / 31 skipped.
- Core `tsc --noEmit` clean.

> **Local test caveat (not this change):** the `@plur-ai/mcp` and `@plur-ai/claw` suites could not run in my environment — `@modelcontextprotocol/{server,client,core}@2.0.0-beta.4` are declared in the lockfile but not installed locally, so 9 mcp files fail to *load* (0 assertion failures). This diff is core-only; CI (with a complete install) exercises those packages.

## Follow-ups (separate issues, to triage with @plur9)

- #670 — ranker quality (negation-blind, keyword-tie collisions).
- #671 — domain convention (~35% of engrams have no `domain`; free-form vs registry).


plur9:
Closing as superseded by #674 (merged 2026-07-23). Same root cause was fixed there: `covers[]` from the server's `/me` response now persists into local config store entries, activating `suggestScope` for remote scopes. Thanks for the independent diagnosis — the overlap in the diff confirms the root cause was well-identified.

Issue #399, opened 2026-06-23
[audit] Forward domain ties resolve alphabetically, not by coverage specificity


plur9:
From the independent 0.10.0 scope-security audit (#377), Črt.

**Severity:** Medium (Confidentiality)
**Location:** packages/core/src/scope-routing.ts:220-225,306-310

Forward domain ties resolve by alphabetical scope name, not coverage specificity — a broad team can win over the correct narrow team.

Part of the pre-0.10.0 hardening triage (see #377).

PR #675, opened 2026-07-24
feat(scope-routing): minConfidence option + document keyword-only tie limitation (#670)

plur9:
## What

Closes part of #670 (Options 3+4 from the triage decision).

**Option 4 — keyword floor via `minConfidence`**: add `minConfidence?: number` to `RankScopesOptions`. Callers in display/suggestion contexts can now filter lone keyword hits (confidence ≈ 0.12) that add noise without being reliable routing signals. Default is `0` — fully backward-compatible; existing callers are unaffected.

**Option 3 — coverSpecificity tie-break**: already implemented in `rankScopes()` sort (#399). This PR adds a test that pins the behavior and a test documenting the known limitation: when both candidates have only keyword hits at equal confidence, `coverSpecificity` is 0 for both, so the final tie-break is alphabetical by scope name. This is intentional and documented in the module comment — the fix is to set a `domain`.

**Documentation**: module doc now explicitly names the keyword-only tie limitation and the recommended workaround (`domain` field).

## Changes

- `packages/core/src/scope-routing.ts`: `minConfidence` added to `RankScopesOptions`, filtered after sort; module doc updated.
- `packages/core/test/scope-routing.test.ts`: 2 new tests — `minConfidence` filter behavior, keyword-only tie documentation.

## Tests

44 tests pass (2 new).

- `minConfidence: 0.15` suppresses a lone keyword hit (≈0.1176); two keyword hits (≈0.21) survive; default (0) is unchanged.
- Keyword-only tie test documents the alphabetical behavior with an explicit comment that this is a known limitation.

## Out of scope

- Option 2 (negation handling) — deferred P3 per triage.
- Backfill of 803 no-domain engrams (#671 item 2) — separate nightshift batch task.

### DIFF
--- a/packages/core/src/scope-routing.ts
+++ b/packages/core/src/scope-routing.ts
@@ -25,7 +25,11 @@
  * Confidence is normalized into [0,1] by squashing the raw weighted score, so a
  * single weak keyword hit reads as low-confidence and a domain-prefix match
  * reads as high-confidence. Ties break deterministically: equal confidence
- * prefers a domain-prefix match, then scope name ascending.
+ * prefers a domain-prefix match, then more-specific cover depth, then scope
+ * name ascending (#670). Keyword-only ties (both candidates from a single
+ * coincidental keyword hit each) therefore break alphabetically — this is the
+ * documented limitation of the purely-lexical keyword channel. Set a deliberate
+ * `domain` on the engram to route reliably without touching the keyword channel.
  *
  * Each candidate carries two domain-channel booleans (see {@link ScopeCandidate}):
  * `domainMatch` (true for EITHER prefix direction — used for scoring/ordering)
@@ -306,6 +310,14 @@ function scoreScope(
 export interface RankScopesOptions {
   /** Override {@link WEIGHT_TAG} for this call. Default: 0.5. */
   weightTag?: number
+  /**
+   * Exclude candidates whose `confidence` is strictly below this value (default: 0
+   * — return all positive-scoring candidates and let the caller decide). Useful in
+   * display/suggestion contexts to suppress lone keyword hits (≈0.12) that add noise
+   * without being reliable routing signals (#670 Option 4). The write-path auto-router
+   * in index.ts gates via `matchThreshold` (default 0.5) independently of this option.
+   */
+  minConfidence?: number
 }
 
 /**
@@ -349,5 +361,6 @@ export function rankScopes(
     b.coverSpecificity - a.coverSpecificity ||
     a.scope.localeCompare(b.scope),
   )
-  return candidates
+  const min = options?.minConfidence ?? 0
+  return min > 0 ? candidates.filter(c => c.confidence >= min) : candidates
 }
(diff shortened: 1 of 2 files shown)


plur9:
Review (Miles): clean, backward-compatible addition. `minConfidence` defaults to 0 so all existing callers are unaffected; the `min === 0` fast-path is a nice touch. The two test cases cover lone keyword hit exclusion (≈0.12 filtered at 0.15) and two-keyword survival. The tie-break documentation test is a good regression anchor for the known alphabetical fallback limitation. **Ready for merge** — needs crtahlin or another reviewer to approve since self-review is blocked.

plur9 (commented):
CI green across all matrix nodes (issue-guard + test 20/22/24). `minConfidence` option is backward-compatible (default 0 = passthrough). Keyword-only tie documentation is accurate — the test explicitly asserts alphabetical tie-break as the documented limitation, which is the right way to handle this (document-and-test the wart rather than over-engineer). LGTM from Miles.

plur9:
Clean, minimal implementation. `minConfidence` filter is correctly gated (`min > 0` keeps the zero-default a no-op), the docstring update accurately documents the keyword-only tie limitation, and both new test suites are well-structured — the tie-break test doubles as executable documentation of the alphabetical fallback. CI green across all matrix nodes. Ready for merge.

crtahlin (approved):
Approving — clean, backward-compatible implementation of #670 Options 3+4.

Verified:
- `minConfidence` is additive and correctly gated: default 0 with a `min > 0 ? filter : passthrough` fast-path, so existing callers are untouched. Filter is applied after sort (ordering preserved among survivors), and `>= min` matches the docstring's 'strictly below excluded'.
- Option 3 (coverSpecificity tie-break) is already in the sort (#399); the new test pins it, and the keyword-only tie test documents the alphabetical fallback as an executable regression anchor.
- Ran locally: scope-routing.test.ts 44/44 pass (2 new). Full core suite green apart from one unrelated, pre-existing flake in pr5-hardening.test.ts (passes 12/12 in isolation across repeated runs; this PR doesn't touch that path).

Note (non-blocking, follow-up): the new option isn't consumed by any production caller yet — `suggestScope()` (the `plur_suggest_scope` display path) still calls `rankScopes` without `minConfidence`, and the write-path gates via SCOPE_MATCH_THRESHOLD separately. So the noisy lone-keyword suggestions (~0.12) that motivated #670 aren't actually suppressed until a caller forwards `minConfidence` in display contexts. This PR correctly adds the mechanism; wiring can be a follow-up.

Coordination: #670 is assigned to me (crtahlin); this closes Options 3+4 only — Option 2 (negation) and the no-domain backfill remain open under that issue.

plur9:
## Review finding + fix pushed: `minConfidence` was dead-option-only

A code sweep confirmed the original diff added `minConfidence` to `RankScopesOptions` but **no caller passed it** — not `suggestScope` (takes no options), not `_resolveUnscopedScope` (correctly gated by `match_threshold` instead), not the MCP tool, not config. The PR body's motivating consumer ("callers in display/suggestion contexts can now filter…") didn't exist, so the keyword-floor triage decision on #670 was a no-op in practice.

Pushed `66289d4` wiring it in:

- `suggestScope(input, { minConfidence? })` — explicit option > `scope_routing.min_confidence` config > 0 (core default unchanged, backward compatible)
- `plur_suggest_scope` gains `min_confidence` with a **display default of 0.15** per the triage decision — lone coincidental keywords (≈0.12) no longer surface to agents; `min_confidence: 0` restores the full list; the response echoes the applied floor
- auto-route gate untouched (`match_threshold` stays independent)

7 new tests, 94 passing across the touched files. With this, the PR actually ships the #670 Options 3+4 behavior it describes.

Issue #647, opened 2026-07-21
Proactively offer to register authorized-but-unregistered scopes (per-scope, remembered opt-out)


crtahlin:
## Problem

`plur_session_start` **already surfaces** authorized-but-unregistered scopes — e.g. "Your token is authorized for N more scope(s) not yet registered … call `plur_scopes_discover register:true` to add them all in one step." So the *discovery* nudge exists in normal usage. But it has two UX gaps:

- **All-or-nothing.** It points at `register:true` (register everything) — no per-scope decision. A user who wants some scopes but not others has no in-flow choice.
- **No memory → repeats forever.** The same unregistered scopes are re-listed **every session** until registered. There's no "never ask me about this one" — so it can't stop nagging, and a user who deliberately doesn't want a scope is reminded indefinitely.

Net effect: users either register nothing (banner blindness) or `register:true` everything. `plur_scopes_discover` already computes the actionable `unregistered` set (#292); the missing piece is a friendly, remembered, per-scope decision.

## Proposal — per-scope, with remembered opt-out

When authorized-but-unregistered **shared-family** scopes exist (`group:`/`project:`/`space:`/`team:`/`org:` — exclude `user:*`/`global`), on `session_start` (or `plur doctor`):

- **Per-scope decision, not all-or-nothing:** register / not now / never. Show the scope's self-describing `description` if present.
- **Remember opt-outs:** "never" is persisted so the user isn't re-asked; "not now" re-offers next session.
- **Re-subscribe path:** a flag/command to re-surface previously dismissed scopes (opt-out is quiet, not permanent).

## Acceptance
- The existing session-start nudge gains a per-scope register/skip/never with persisted opt-out and a re-offer path.
- Personal-family scopes excluded; dismissed scopes stop being re-listed every session.

Relates to #292 (scopes_discover) and #345 (self-describing scopes).

crtahlin:
## Reframing after a code check — the nudge is agent-facing, not user-visible

The premise "`plur_session_start` already surfaces this to users" is inaccurate. The nudge (`packages/mcp/src/tools.ts:1926`) is appended to the **`guide`** string in the `session_start` **tool result**, which is consumed by the **model**, not rendered to the user. Agents almost never echo it verbatim, so in practice users never see it. The only genuinely **user-facing** surface today is the **CLI** (`plur stores` / running `scopes_discover`).

[...]

PR #677, opened 2026-07-24
feat(#671): missing-domain nudge on learn — surface unroutable writes at capture time

plur9:
## What

Item 1 of the #671 triage: a soft, advisory `domain_hint` at capture time when an engram is written without a `domain` — the state in which it can never auto-route (domain-prefix is the only channel that reliably clears the auto-route threshold; 803 of 2,261 engrams on the live store are unroutable this way).

## Surfaces

| Surface | Form |
|---|---|
| `plur_learn` (MCP) | `domain_hint` response field, alongside the existing `scope_hint` (#296) |
| `plur_learn_batch` (MCP) | aggregate `domain_hint` — counts items that passed neither `domain` nor explicit `scope` |
| `plur learn` (CLI) | `domain_hint` in `--json` output + a text-mode `Hint:` line |

Response-field rather than console warning, for the reason established in the #647 reframing: tool-result fields reach the agent; console/guide text does not.

## Gate — all must hold, so personal installs stay silent

1. no `domain` passed
2. no explicit `scope` (routing was actually on the table)
3. the write did **not** auto-route anyway
4. at least one registered scope declares `covers[]` to route against

Advisory only — never blocks or fails a write (`try/catch` around the metadata read).

## Tests

- 6 new MCP tests (`tools.test.ts`): fires on the exact failure mode; silent on domain / explicit scope / metadata-less team store; batch aggregate + batch silent case
- 3 new CLI tests (`learn.test.ts`): fires against a covers-declaring scope; silent with `--domain`; silent on a personal install
- 71/71 pass across both files; `pnpm -r build` clean

## Out of scope

- Retroactive backfill of the 803 no-domain engrams (#671 item 2) — separate batch task
- Post-backfill routing re-audit (#671 item 3)

Part of #671.

https://claude.ai/code/session_01LjCUCfkFkGEArJQQgmQfuY


crtahlin (dismissed):
Approving — clean, well-gated advisory. The missing-domain `domain_hint` targets the real failure mode (a domain-less write can't clear the auto-route threshold), is strictly advisory (try/catch, never blocks a write), and is correctly wired into all three surfaces' response contracts (mirroring the #296 scope_hint pattern).

Verified against source: the `_routed` gate signal is genuinely stamped by core (index.ts:1502/1653/1686); `explicitScope` and the four-part gate are correct. Ran the affected suites locally: packages/mcp/test/tools.test.ts 49/49 and packages/cli/test/learn.test.ts 22/22 (71 total, incl. all 9 new #671 tests). Core built clean.

Non-blocking follow-ups (filing separately):
1. Batch gate isn't at full parity with the single-item gate — `noDomainCount` filters on no-domain + no-scope but omits the per-item 'did not auto-route' check the single path applies. Benign (no-domain rarely routes), but could over-count by one when a no-domain item auto-routes via tags.
2. The hint points to 'the domain convention in CLAUDE.md', but that convention (plur.<team>.<area>) isn't documented there yet (line 226 covers scope only). The hint is still actionable inline; documenting the convention is #671 AC1.

Nit: commits carry Co-Authored-By / Generated-with / a Claude-Session URL — worth stripping per the neutral-authorship convention (the session URL especially shouldn't be in a public commit).

Coordination: #671 is assigned to me; this is item 1 of 3 (backfill + re-audit remain).

crtahlin:
We resolved confilcts, but now someone else has to approve / merge @plur9 please

PR #682, opened 2026-07-24
test(core): fix flaky MERGE demotion test in pr5-hardening (#680)

### LINKED ISSUE #680: Flaky test: pr5-hardening.test.ts intermittently fails in full core suite (scope demotion)
## Symptom

`packages/core/test/pr5-hardening.test.ts` intermittently fails when the **full** `@plur-ai/core` vitest suite runs, on an assertion around scope demotion:

```
const e = result.engram as Engram & { visibility?: string; structured_data?: {...} }
expect(e.scope).toBe('local')   // ~line 82-83
```

Observed once during a full-suite run (`vitest run` in `packages/core`) while reviewing an unrelated PR ([#675](https://github.com/plur-ai/plur/pull/675), which only touches `scope-routing.ts`). On the immediate re-run the full suite passed, and the file passed **12/12 across 3 isolated runs** (`vitest run test/pr5-hardening.test.ts`).

## Diagnosis

Passes deterministically in isolation, fails only under the full suite → a **test-isolation / ordering / shared-state** flake, not a product bug in the demotion path. Likely candidates: shared global/config state, an async write not awaited, or a temp-store/path collision between test files running in the same worker.

## Why P3

Not a correctness issue in shipped code and it does not fail CI reliably — but an intermittently-red suite erodes trust in the signal and can mask real regressions. Worth isolating the shared state (or giving this file its own store/tmpdir) so the suite is deterministic.

## Repro

```
cd packages/core && npx vitest run      # occasionally: 1 failed in pr5-hardening
npx vitest run test/pr5-hardening.test.ts   # always green in isolation
```

Not caused by #675.

plur9:
## Root cause

The MERGE test in `pr5-hardening.test.ts` had **zero BM25 token overlap** between the seed statement and the merge statement:

- Seed: `'the staging cluster is documented in the handbook'` → tokens: `staging, cluster, documented, handbook`
- Merge: `'reachable at 139.59.155.82:8877'` → tokens: `reachable, 139, 155, 8877`

`learnAsync` candidate retrieval runs BM25 + embeddings. With no term overlap, BM25 returns empty. Under full-suite load (embedding model warming up concurrently across test files / forked processes), embeddings can also return empty on cold start. This collapses `learnAsync` to the ADD early-return path — `result.decision === 'ADD'` instead of `'MERGE'` — failing the test.

The UPDATE and BOUNDARY tests were fine — both share `deploy` with their respective statements.

## Fix

Changed the merge statement to include `staging cluster` so BM25 always finds the seed, independent of embedding availability:

```diff
-const result = await plur.learnAsync(`reachable at ${PUBLIC_IP}:8877`, { llm })
+const result = await plur.learnAsync(`staging cluster reachable at ${PUBLIC_IP}:8877`, { llm })
```

The test intent is unchanged: a MERGE that adds a public IP to a shared-scope engram must demote it to local/private.

## Verification

- 12/12 pass in isolation (unchanged)
- The statement change adds 2 overlapping BM25 tokens (`staging`, `cluster`), making candidate retrieval deterministic regardless of embedding state

Closes #680

### DIFF
--- a/packages/core/test/pr5-hardening.test.ts
+++ b/packages/core/test/pr5-hardening.test.ts
@@ -76,7 +76,10 @@ describe('MED-20 + LOW-9 — learnAsync UPDATE/MERGE demotion (real guard, share
     const seed = plur.learn('the staging cluster is documented in the handbook', { scope: SHARED_SCOPE, type: 'procedural' }) as Engram
 
     const llm = dedupLlm('MERGE', seed.id)
-    const result = await plur.learnAsync(`reachable at ${PUBLIC_IP}:8877`, { llm })
+    // "staging" is shared with the seed so BM25 finds the candidate even when
+    // embeddings are unavailable under full-suite load (prevents a flake where
+    // zero-overlap candidates caused an ADD instead of MERGE — see #680).
+    const result = await plur.learnAsync(`staging cluster reachable at ${PUBLIC_IP}:8877`, { llm })
 
     expect(result.decision).toBe('MERGE')
     const e = result.engram as Engram & { visibility?: string; structured_data?: { _demoted?: { from: string; to: string; patterns: string } } }


crtahlin (approved):
APPROVE — test-only fix, root cause correctly identified.

The MERGE test was the only one of the three demotion tests without BM25 token overlap between seed and merge statement, so candidate retrieval fell back entirely to embeddings — which can return empty on cold start under full-suite load, collapsing `learnAsync` to the ADD path. Adding `staging cluster` to the merge statement guarantees overlap with the seed (`the staging cluster is documented in the handbook`), making retrieval deterministic regardless of embedding state. UPDATE and BOUNDARY tests were already fine (`deploy` shared both ways).

Verified against the full test file: assertions are unchanged, the IP is still present so the demotion path still fires, and the edit is semantically accurate (adding the network address of the staging cluster). The inline comment documents the reasoning well.

Correctness here rests on the guaranteed-overlap argument rather than a single green run, which is the right way to fix a load-dependent flake. CI is all green.

Note (non-blocking): the flake surfaces that production `learnAsync` MERGE candidate retrieval is non-deterministic under embedding cold start — correctly out of scope for this test fix, but worth keeping in mind separately.

PR #678, opened 2026-07-24
feat(#640): scope-filtered push — shared remotes never receive personal/private engrams

plur9:
## What

Implements the triage decision on #640 (**option 2 — scope-filtered push**), grounded to the actual sync architecture: sync has a single git `origin` and no per-remote registry, so the remote's declared scope materializes as a **remote type**.

| remote_type | push set |
|---|---|
| `personal` (default) | everything non-`scope:local`, private included — today's behavior, unchanged for every existing install |
| `shared` | ONLY shared-family-scope (`isSharedScope`) engrams with `visibility !== 'private'` — personal-family and private engrams never reach the remote, by construction |

Config: `sync.remote_type` in `config.yaml` (machine-local, never synced) or the new optional `remote_type` arg on `plur_sync` / `Plur.sync()`.

## Also fixes a latent leak

The merge-conflict resolution path (`pullRebase` kept-both commit) staged the working-tree `engrams.yaml` **verbatim** — `scope:local` engrams rode into the remote on that path. It now strips the staged blob like every other commit path.

## Acceptance criteria from the issue

- [x] A shared/team remote does not receive `visibility: private` engrams by default
- [x] A personal remote retains today's behavior (private engrams mirror across your own devices)
- [x] Explicit and hard to misconfigure — allowlist-style predicate on the staged blob, same construction as the #380/#384 SYNC_PATHS hardening; default is backwards-compatible
- [x] `SyncResult.warning` reflects the actual push set (shared reports the stripped count; personal now points at `sync.remote_type: shared`)
- [x] Round-trip test: private + personal engrams absent from a teammate clone of a shared remote; working tree keeps everything
- [x] README sync section documents personal-vs-shared semantics

## Design note for review

`visibility` defaults to `private` (#401), so a **shared remote receives only engrams whose visibility was deliberately set** — conservative on purpose and consistent with the #90 "private stays local" write-path semantics. If we'd rather have shared-scope engrams push regardless of visibility, that's a one-line predicate change — flagging explicitly.

Follow-up (out of scope): `episodes.yaml` / `candidates.yaml` / `tensions.yaml` sync wholesale with no stripping infrastructure — a true "private never leaves" guarantee needs the same treatment there.

## Tests

[...]

PR #685, opened 2026-07-24
fix(scope): harden metadata trust + registry hygiene (scope-audit 2026-07-24)

plur9:
A code review of the scope subsystem (scope-audit 2026-07-24) produced nine verified findings, from a HIGH-severity leak-guard bypass down to hygiene items. This PR implements all nine on one branch.

## Findings → fixes

| # | Sev | Finding | Fix |
|---|-----|---------|-----|
| F1 | HIGH | Remote-supplied `sensitivity` could disarm the write-time leak guard: `persistScopeMetadata` copied `/me` sensitivity verbatim, and the guard checks `allow` BEFORE `forbid` with `allow` admitting arbitrary strings — a hostile enterprise endpoint serving `allow:['secrets','infra']` silently disarmed the guard at next session_start | Trust rule: remote sensitivity may only TIGHTEN. Remote `allow` is never persisted; only `forbid` persists, sanitized to `SENSITIVITY_CATEGORIES` (empty → safe default). A hand-edited LOCAL `allow` is preserved and stays honored; guard logic unchanged. Documented at the persist site and on `ScopeSensitivitySchema` |
| F2 | MED | `mergeStoresForWriteback` restored raw on-disk `forbid` verbatim (discarding server `forbid` changes) while the change detector compared the raw server payload → never converges, config.yaml rewritten on EVERY session_start (mtime churn, reload storms); also non-convergent when server serves no sensitivity but local has one | Change detector now compares what WILL be persisted (post-sanitize, post-merge; key-order-insensitive JSON). `persistStores` carries `serverSensitivityScopes` so the writeback merge lets an intentional forbid sync through — everything else keeps the historical raw-restore |
| F3 | MED | `registerDiscoveredScopes` iterated `d.authorized`, so `plur_scopes_discover register:true` registered scopes the user explicitly dismissed and left the stale `dismissed_scopes` entry | Batch path skips dismissed scopes (reported in `skipped`, logged). Per-scope `registerScope` remains the deliberate override that also un-dismisses. Tool description updated |
| F4 | MED | session_start guidance described pre-#674 behavior ("unscoped calls will be tagged \"global\"") — since synced covers, unscoped writes can auto-route to a covers-matching team scope | Both guide strings now state the real behavior: confident covers matches auto-route (response reports `routed`); otherwise the unscoped default applies. Text-only |
| F5 | LOW/MED | `persistScopeMetadata` silently clobbered hand-set local `covers`/`description` with server values | One `logger.warning` naming the scope and fields when overwriting a differing non-empty local value (overwrite stays server-authoritative by design) |
| F6 | LOW/MED | Exact-string URL identity: `https://x.com`, `https://x.com/`, `https://x.com/sse` counted as three endpoints in `_distinctRemoteEndpoints`, addStore dedup, and metadata matching — scopes registered under one spelling were re-offered under another | New exported `normalizeEndpointUrl` in remote-store.ts (shared with `apiBase`, so HTTP-time and identity-time normalization can't drift), applied at every identity comparison. Comparison-time only — stored spellings never rewritten |
| F7 | LOW | No locking on config.yaml read-modify-write (`persistStores`, `persistDismissedScopes`) while engrams.yaml uses `withLock` | Both persist paths now run read+merge+write under `withLock(config.yaml)`; lock scope kept tight, in-memory refresh after release |
| F8 | LOW | `/me` scope grammar admits uppercase but `isSharedScope` matched lowercase prefixes only (`Group:x` → personal-family: never offerable, never guard-scanned); dismissed matching was exact-case, so a case-variant re-advertisement resurfaced a dismissed scope | `isSharedScope`/`isPersonalScope` prefix test is case-insensitive (input lowercased for comparison only); dismissed-scope membership + dedup case-insensitive. Stored values keep their case |
| F9 | LOW | Write path used possibly-stale config: `suggestScope` calls `reloadConfigIfChanged()` but `_resolveUnscopedScope` → `listScopeMetadata` did not | `reloadConfigIfChanged()` at the top of `_resolveUnscopedScope`, mirroring `suggestScope` |

## Tests

- New `packages/core/test/scope-metadata-trust.test.ts` — F1 end-to-end against the stub server (hostile `allow` dropped + guard still demotes; local `allow` preserved and honored; forbid sanitize), F2 convergence (two identical syncs → second is a byte-identical no-op with unchanged mtime; server forbid change persisted once then converges; server-absent sensitivity converges), F5 warning presence/absence.
- New `packages/core/test/scope-case.test.ts` — F8 predicates (`Group:`/`USER:`/`Public-roadmap` boundary).
- New `packages/core/test/config-lock.test.ts` — F7 lock release + stale-lock recovery on both persist paths.
- Extended `scope-discovery.test.ts` — F3 (dismissed scope not batch-registered, dismissal intact, `registerScope` override un-dismisses), F8 case-variant dismissals, F6 (three spellings → one endpoint; scope registered under one spelling not re-offered under another; addStore spelling-variant dedup; metadata matching across spellings; `normalizeEndpointUrl` unit cases).
- Extended `route-unscoped.test.ts` — F9 (covers added to config.yaml after startup are picked up by the very next unscoped write).

Full run: `npx vitest run packages/core/test packages/mcp/test` → **147 files passed, 2290 tests passed, 32 skipped, 0 failed**. (`pr5-hardening.test.ts` flaked once under full-suite load in an earlier run — expected `MERGE` got `ADD`, an embedder-load dedup flake; it passes in isolation and in the full rerun, and origin already carries `fix/680-*` branches for exactly this flake. Untouched by this PR.)

Out of scope, untouched per audit instructions: `sync.ts` push-filtering (#678) and `scope-routing.ts` minConfidence (#675 — rebased on top of it, no overlap).

https://claude.ai/code/session_01LjCUCfkFkGEArJQQgmQfuY


plur9 (commented):
Code review — LGTM

**F1 (HIGH — remote sensitivity disarming leak guard)**
`sanitizeForbidCategories` correctly: filters to `SENSITIVITY_CATEGORIES`, empty → safe default (all categories forbidden, not open). Remote `allow` is never persisted. Local `allow` is preserved through the sync path. Guard logic unchanged. Trust rule well-documented at the persist site and on `ScopeSensitivitySchema`. ✓

**F2 (MED — change-detector never converges)**
`stableJson` sorts object keys without sorting arrays (array order is meaningful for `covers`/`forbid`) — correct. `serverSensitivityScopes` threaded from `persistScopeMetadata` → `persistStores` → `mergeStoresForWriteback` so the server's authoritative `forbid` update survives the raw-restore for exactly the affected scopes, while every other caller keeps historical behavior. Convergence tests verify mtime/content byte-equality after a no-op sync and verify that a server policy change lands on disk then converges on the next sync. ✓

**F3 (MED — batch path registers dismissed scopes)**
Dismissed scopes now skipped in `registerDiscoveredScopes`, reported in `skipped` (neutral label already in place from prior work). Per-scope `registerScope` remains the deliberate override. Tool description updated accurately. ✓

**F4 (MED — session_start guidance describes pre-#674 behavior)**
Wording now describes auto-routing (covers-matching) as the primary path with "global" as fallback. Correct and actionable. ✓

**F5 (overwrite warning)**
Warning fires with scope name and field list; doesn't fire for initial empty → populated sync. ✓

**F7 (`persistStores` under `withLock`)**
Tight lock scope — read + merge + write only; in-memory refresh after release. Matches pattern from engrams.yaml. ✓

**F9 (unscoped write path reloads changed config)**
`reloadConfigIfChanged()` call added at the right site; comment explains why (mirrors `suggestScope` for the write path). ✓

**Test coverage**
F1: junk forbid → safe default, allow never persisted, local allow preserved through sync + guard honors it. F2: three convergence scenarios including the server-sends-no-sensitivity case (the historical non-convergence trigger). F5: overwrite-warns and first-sync-doesn't-warn. Strong.

CI green (5/5). LGTM — ready for merge when plur9 approves.

crtahlin (approved):
APPROVE — scope metadata trust hardening (scope-audit 2026-07-24)

Independent review + full local test run. The nine fixes are sound and well-contained.

**F1 (HIGH — leak-guard bypass) correctly closed.** Remote-supplied `sensitivity` can now only tighten: remote `allow` is never persisted, `forbid` is sanitized to the known categories (empty → safe default, not open), and a hand-edited local `allow` is preserved and stays honored. Guard logic itself is untouched — only its inputs are sanitized, which is the right layer. Trust rule documented at the persist site, on `ScopeSensitivitySchema`, and in the changelog.

**F2 convergence** is proven by byte-identical-mtime tests (comparing what *will* be persisted, post-sanitize, via key-order-insensitive `stableJson`), covering the server-sends-no-sensitivity non-convergence trigger. **F6** shares `normalizeEndpointUrl` with `apiBase` so HTTP-time and identity-time normalization cannot drift. **F8** folds case in one predicate (`isPersonalScope = !isSharedScope`), so both are covered. F3/F4/F5/F7/F9 each documented and tested.

**Verified independently:** `forbid` is schema-defaulted so `sanitizeForbidCategories` can never receive undefined; `isPersonalScope` inherits the case-fold correctly.

**Tests:** full `packages/core` + `packages/mcp` suite in an isolated worktree — 2292 passed, 32 skipped, 0 failed on a clean run; the PR's 71 new/extended tests passed twice in isolation. One run showed the known `pr5-hardening` embedder-load dedup flake (MERGE vs ADD), unrelated to this PR and already tracked on the fix/680 branches. Build clean.

Two non-blocking notes: (1) by the trust rule a legitimate enterprise can no longer *loosen* a scope's guard via `/me` (forbid:[] → forbid-all) — loosening is now strictly a local decision, which is the intended stance; (2) the batch dismissed-scope gate matches dismissals case-insensitively but the adjacent `registered.includes` check is case-sensitive — negligible impact.

LGTM.

Issue #231, opened 2026-05-26
plur_session_end crashes with unhelpful error when engram_suggestions contains strings instead of objects


plur9:
## Summary

`plur_session_end` throws `Cannot read properties of undefined (reading 'match')` when the caller passes `engram_suggestions` as an array of strings instead of an array of `{statement, type}` objects. The schema validator silently accepts the wrong shape; the crash happens deep inside `detectSecrets()` and the error message gives no hint about the actual problem.

Reproduced against `@plur-ai/mcp@0.9.10` (current latest) during `/wrap-up`.

## Reproduction

```js
// Call from any MCP client
plur_session_end({
  summary: \"Session summary\",
  engram_suggestions: [\"This is a statement string\", \"Another statement\"]
  // ^ WRONG SHAPE: should be [{statement: \"...\"}, {statement: \"...\"}]
})
```

Response:
```json
{ \"error\": \"Cannot read properties of undefined (reading 'match')\", \"success\": false }
```

A correctly-shaped call works:
```js
plur_session_end({
  summary: \"...\",
  engram_suggestions: [{statement: \"This is a statement\"}]
})
```

## Root cause

Bug chain through three layers:

### 1. Schema-to-Zod converter ignores nested item shape

\`packages/mcp/src/server.ts:178\`:

```ts
else if (prop.type === 'array') field = z.array(z.unknown())
```

This collapses any array schema into \`z.array(z.unknown())\` regardless of \`prop.items\`. The JSON Schema for \`engram_suggestions\` is:

```ts
engram_suggestions: {
  type: 'array',
  items: {
    type: 'object',
    properties: {
      statement: { type: 'string', ... },
      type: { type: 'string', enum: [...] },
    },
    required: ['statement'],
  },
}
```

But the validator throws away \`items\` entirely, so \`['string', 'string']\` validates fine.

### 2. Handler dereferences \`s.statement\` without checking shape

\`packages/mcp/src/tools.ts:1200-1208\`:

```ts
const suggestions = args.engram_suggestions as Array<{ statement: string; type?: string }> | undefined
if (suggestions?.length) {
  for (const s of suggestions) {
    plur.learn(s.statement, { type: s.type as any })  // s.statement is undefined when s is a string
    engrams_created++
  }
}
```

The cast \`as Array<{ statement: string; ... }>\` is a lie — runtime can be anything that passed \`z.array(z.unknown())\`.

### 3. \`detectSecrets\` is undefined-unsafe

[...]

PR #705, opened 2026-07-26
fix(mcp): surface #297 partial array drops and coerce union item schemas

### LINKED ISSUE #704: MCP argument truncation drops trailing params (partial-drop variant of #297) — silent scope misrouting when the dropped field is optional
Follow-up to #297, which fixed the **total** payload drop. There is a second shape it does not cover, and in practice it is the harder one to diagnose.

## Summary

MCP argument truncation is **size-based** and drops **trailing parameters**. Early fields arrive; later ones are silently discarded. This affects parameters of *any* type — not only arrays.

Two consequences:

1. **Required field dropped** → a confusing `"<field>: Required"` error for a field the caller demonstrably passed. The #297 hint does not fire, because it is gated on `receivedFields.length === 0`.
2. **Optional field dropped** → *no error at all*. The call succeeds with silently wrong behaviour. This is the more dangerous variant.

## Reproduction

Narrowed empirically against the local MCP server (v0.14.0), varying one thing at a time.

| Call | Result |
|------|--------|
| `plur_session_end` with `engram_suggestions: ["x"]` | ✅ works |
| `engram_suggestions: ["alpha, beta", "gamma"]` | ✅ commas are not the trigger |
| One 180-char suggestion, alone | ✅ content is not the trigger |
| Short `summary` + three long suggestions | ✅ works |
| **~700-char `summary` + five long suggestions** | ❌ arrives as `{summary}` — array dropped |

The same array succeeds once `summary` is shortened, so the trigger is total payload size, not the array itself and not the schema.

### The silent variant

A large `plur_learn` call carrying `statement`, `rationale`, `tags`, and trailing `scope` / `domain` / `type`:

```
passed:   scope=group:plur/engineering, domain=plur.engineering.mcp, type=procedural
stored:   scope=global, no domain, type=behavioral (default)
```

No error was raised. The engram was simply misrouted out of the team store into `global`. Re-issuing the identical call with `scope`/`domain`/`type` placed **before** the long `statement` stored correctly.

## Defects this exposes in our own handling

Both are in `packages/mcp/src/tools.ts`.

**1. The #297 hint cannot fire on a partial drop.**

```ts
const arrayBugHint = receivedFields.length === 0 && hasArrayParam ? ... : ''
```

When two of three fields arrive, `receivedFields.length` is 2, so the caller gets a bare `Required` with no mention of #297 and no workaround. This is exactly the case that most resembles a server-side schema bug from the caller's perspective — I retried six times with different array formats before suspecting the client.

**2. The advertised workaround does not work for `engram_suggestions`.**

The error message recommends retrying as a JSON string *or a comma-separated string*. The comma path is gated on:

```ts
if (prop.items?.type === 'string') { ...split(',')... }
```

`engram_suggestions` has union items (`anyOf: [{type:'string'}, {type:'object'}]`, from #231), so `items.type` is `undefined` and the input falls through uncoerced. The recommended retry shape is impossible for the parameter that most often needs it.

## Proposed fix

- Fire the hint whenever a **missing required field is itself array-typed**, not only when the whole payload is empty. Name the dropped parameters and state that the trigger is payload size, so the caller shrinks other fields rather than retrying an identical call.
- Extend the comma-separated coercion to any item schema that **accepts** a bare string, including `anyOf`/`oneOf` variants.

Neither fixes the client — that is upstream — but together they turn an unexplainable failure into a self-correcting one.

## Not addressed here

The **silent** optional-parameter drop has no clean server-side detection: an omitted optional field is indistinguishable from a deliberately omitted one. The existing `scope_hint` on `plur_learn` already mitigates the specific case of a missing `scope`, which is how this was noticed at all. Worth a separate discussion about whether other optional fields deserve similar hints.

## Workaround for callers, until the client is fixed

- Put the long field **last**; keep `scope` / `domain` / `type` early.
- Pass arrays as a JSON string (`"[\"a\",\"b\"]"`) or comma-separated — the server preprocessor coerces both back.
- Split very large calls into two.

plur9:
Closes #704. Follow-up to #297, which fixed the total payload drop but not the partial one.

## What

Two defects in `packages/mcp/src/tools.ts` that together made a real client-side truncation bug undiagnosable.

**1. The #297 hint could not fire on a partial drop.**

`validateToolArgs` gated the hint on `receivedFields.length === 0`. But the client also drops parameters *partially* — early scalar fields arrive, trailing ones are silently discarded. A `plur_session_end` call carrying a ~700-char `summary` plus a five-element `engram_suggestions` array arrived as `{summary}` alone.

That shape is the one most likely to be misread as a server schema bug: the caller sees `engram_suggestions: Required` for a field they demonstrably passed, and previously got no hint and no workaround. I retried six times with different array formats before suspecting the client.

The hint now fires whenever a **missing required field is itself array-typed**, names the dropped parameters, and states that the trigger is payload size — so the caller shrinks the other fields instead of retrying an identical call.

**2. The advertised workaround did not work for the parameter that needs it most.**

The comma-separated fallback was gated on `prop.items?.type === 'string'`. `engram_suggestions` has union items (`anyOf: [{type:'string'}, {type:'object'}]`, from #231), so `items.type` is `undefined` and the input fell through uncoerced — meaning the retry shape the error message recommends could not succeed for it. Any item schema that **accepts** a bare string (directly, or as an `anyOf`/`oneOf` variant) now takes the comma-separated path.

## Why it is worth fixing here

Neither change fixes the client — that is upstream. But the failure currently presents as a server-side schema error, which sends the caller down entirely the wrong path. These two changes turn it into a self-correcting one.

## Evidence

Narrowed by varying one thing at a time against the local server:

| Call | Result |
|------|--------|
| `engram_suggestions: ["x"]` | works |
| `["alpha, beta", "gamma"]` | works — commas are not the trigger |
| one 180-char suggestion alone | works — content is not the trigger |
| short `summary` + three long suggestions | works |
| ~700-char `summary` + five long suggestions | **array dropped** |

Full reproduction, including the silent optional-parameter variant (a large `plur_learn` losing its trailing `scope`/`domain`/`type` and misrouting the engram to `global`), is in #704.

## Tests

Four added.

**Two fail without this change** — verified by stashing the source fix and keeping the tests:
- partial drop of an array param still names #297
- comma-separated coercion works for union (`anyOf`) item schemas

**Two are regression guards** that must pass either way:
- a missing **non-array** required field must stay silent about #297 (the hint would be noise)
- the JSON-string path for union items already worked and must keep working

## Verification

- `packages/mcp`: 256 passed, 4 skipped — three consecutive full runs
- `tsc --noEmit` clean

## Pre-existing flake, not introduced here

`test/session.test.ts` → "standalone plur_inject calls accumulate into the session telemetry" was observed failing once, then passing on four subsequent full runs, **with and without** this change. It is flaky independently of this PR: it depends on the module-level `_activeSessionId` / `_sessionTelemetry` globals in `tools.ts`, which are safe only while MCP sessions are sequential within a process. Called out rather than silently re-run; worth its own issue if it recurs.

## Not addressed

The silent optional-parameter drop has no clean server-side detection — an omitted optional field is indistinguishable from a deliberately omitted one. The existing `scope_hint` on `plur_learn` already mitigates the `scope` case, which is how this was noticed. See #704 for discussion.

### DIFF
--- a/packages/mcp/src/tools.ts
+++ b/packages/mcp/src/tools.ts
@@ -74,7 +74,18 @@ function jsonSchemaPropToZod(prop: any): z.ZodTypeAny {
           return val
         }
       }
-      if (prop.items?.type === 'string') {
+      // The comma-separated fallback must also cover union item schemas that
+      // ACCEPT a bare string, not just `items: {type: 'string'}`. Before this,
+      // `engram_suggestions` — whose items are
+      // `anyOf: [{type:'string'}, {type:'object'}]` (#231) — failed the check
+      // and fell through to `return val`, so the very workaround the #297
+      // error message advertises did not work for it.
+      const items = prop.items as any
+      const itemVariants = (items?.anyOf as any[] | undefined) ?? (items?.oneOf as any[] | undefined)
+      const itemsAcceptString = items?.type === 'string'
+        || (Array.isArray(itemVariants) && itemVariants.some((v: any) => v?.type === 'string'))
+
+      if (itemsAcceptString) {
         return trimmed.length === 0 ? [] : trimmed.split(',').map((s: string) => s.trim()).filter((s: string) => s.length > 0)
       }
       return val
@@ -123,9 +134,31 @@ export function validateToolArgs(
     const details = parsed.error.issues.map(i => `${i.path.join('.') || 'root'}: ${i.message}`).join(', ')
     const hasArrayParam = Object.values((schema.properties ?? {}) as Record<string, any>)
       .some((p: any) => p?.type === 'array')
-    const arrayBugHint = receivedFields.length === 0 && hasArrayParam
-      ? ' Known client-side bug (plur-ai/plur#297): some MCP clients drop the entire arguments payload ' +
-        'when an array-typed parameter is included. Retry passing array parameters as a JSON string ' +
+
+    // #297 has TWO shapes, not one. The original (total drop) loses the whole
+    // arguments object. The partial drop keeps the early scalar fields and
+    // silently discards trailing ones — observed on a large payload where a
+    // ~700-char `summary` plus a 5-element `engram_suggestions` array arrived
+    // as `{summary}` alone. Gating the hint on `receivedFields.length === 0`
+    // meant the partial case — the one that actually looks like a schema bug
+    // to the caller — got a bare "Required" with no workaround. Fire the hint
+    // whenever a *missing* field is itself array-typed.
+    const missingArrayParams = parsed.error.issues
+      .filter((i: any) => i.code === 'invalid_type' && i.received === 'undefined')
+      .map((i: any) => String(i.path[0] ?? ''))
+      .filter((k: string) => (schema.properties as any)?.[k]?.type === 'array')
+
+    const totalDrop = receivedFields.length === 0 && hasArrayParam
+    const partialDrop = missingArrayParams.length > 0
+
+    const arrayBugHint = totalDrop || partialDrop
+      ? ' Known client-side bug (plur-ai/plur#297): some MCP clients drop ' +
+        (partialDrop
+          ? `array-typed parameters from a large arguments payload while keeping the earlier fields ` +
+            `(here: ${missingArrayParams.join(', ')}). This is size-sensitive — the same call often ` +
+            `succeeds with a shorter payload, so shrink the other fields (e.g. a briefer summary) as well. `
+          : 'the entire arguments payload when an array-typed parameter is included. ') +
+        'Retry passing array parameters as a JSON string ' +
         '(e.g. tags: "[\\"a\\",\\"b\\"]") or a comma-separated string (tags: "a, b") — the server coerces ' +
         'both back into arrays.'
       : ''
--- a/packages/mcp/test/server.test.ts
+++ b/packages/mcp/test/server.test.ts
@@ -139,6 +139,73 @@ describe('MCP server (wire protocol)', () => {
       expect(parsed.error).toContain('tags')
     })
 
+    // #297 partial drop: the client keeps the early scalar fields and silently
+    // discards a trailing array-typed one. Observed with plur_session_end on a
+    // large payload (long summary + 5-element engram_suggestions) arriving as
+    // {summary} alone. Previously the hint was gated on receivedFields.length
+    // === 0, so this — the shape that most looks like a server schema bug to
+    // the caller — got a bare "Required" and no workaround.
+    it('partial drop of an array param still names the client bug (#297)', async () => {
+      const result = await client.callTool({
+        name: 'plur_session_end',
+        arguments: { summary: 'Trailing array param was dropped by the client' },
+      })
+      expect(result.isError).toBe(true)
+      const parsed = JSON.parse((result.content as any)[0].text)
+      // The scalar field survived, so this is NOT the empty-payload path.
+      expect(parsed.received_fields).toEqual(['summary'])
+      expect(parsed.error).not.toContain('arguments object was empty')
+      // ...but the caller must still be told what happened and how to retry.
+      expect(parsed.error).toContain('plur-ai/plur#297')
+      expect(parsed.error).toContain('engram_suggestions')
+      expect(parsed.error).toMatch(/JSON string|comma-separated/)
+      // Size is the trigger; say so, or the caller retries an identical payload.
+      expect(parsed.error).toMatch(/size-sensitive|shorter payload/)
+    })
+
+    it('a missing NON-array required field does not mention #297', async () => {
+      // plur_session_end also requires `summary`. Omitting only that is an
+      // ordinary caller error — the array hint would be misleading noise.
+      const result = await client.callTool({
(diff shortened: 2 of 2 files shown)


plur9 (commented):

Fixes #297 — partial array drops and union item schema coercion were silently losing valid MCP payload data. Fix is targeted and correct. CI green. Ready for merge.

plur9 (commented):
**Review — PR #705: fix(mcp): surface #297 partial array drops and coerce union item schemas**

CI: all checks passing (test/20, test/22, test/24). 256 tests passed, 4 new.

**What was issue #297?** Some MCP clients (e.g. Claude Code MCP client) silently drop the entire arguments payload when any array-typed parameter is included in a large call. The original fix only detected total payload drops (`receivedFields.length === 0`). This PR fixes two follow-on issues: (1) partial drops where scalar fields arrive but trailing array params are silently discarded, and (2) the advertised comma-separated fallback workaround not working for union item schemas (`anyOf: [{type:'string'}, {type:'object'}]`).

**Partial drop fix — correct:**
The new logic (`missingArrayParams`) iterates over Zod validation errors, filters for `invalid_type` / `received: 'undefined'` issues where the field is array-typed in the schema, and fires the hint if any are found. This correctly catches the `{summary}` case where `engram_suggestions` was silently dropped. The hint message also names the specific dropped params and attributes the cause to payload size — actionable guidance the caller previously lacked.

**Union item coercion fix — correct:**
The old check `prop.items?.type === 'string'` was `undefined` for `anyOf`/`oneOf` item schemas, so `engram_suggestions` fell through to `return val` uncoerced. The new `itemsAcceptString` check correctly handles the union case. One observation: the check only traverses `anyOf`/`oneOf` at the `items` level, not nested inside `allOf` or cross-referencing `$ref`. These are edge cases unlikely in PLUR's current schema surface, but worth a code comment noting the limitation.

**Tests — adequate:**
The four new tests are well-targeted. Two verify the new behavior (partial drop naming, union coercion), two guard existing behavior (non-array missing field stays silent, JSON-string path still works). The commented evidence table in the PR is excellent documentation. The acknowledged pre-existing flake in `test/session.test.ts` is correctly called out rather than silently suppressed — this is the right approach.

**No concerns** with the approach. The fix is surgical, does not touch the client, and correctly scopes the hint to where it is diagnosable server-side. **Ready to merge.**

plur9 (commented):
Review — #705: fix(mcp): surface #297 partial array drops and coerce union item schemas

CI: all green (test/20, test/22, test/24). 256 tests passing.

Issue #297 has two shapes: total drop (entire arguments object lost) and partial drop (early scalar fields survive, trailing arrays silently discarded). The original fix only handled total drop, leaving the partial case with a bare Required error and no workaround hint.

Partial drop fix is correct: detects missing required fields that are themselves array-typed, fires the #297 hint with specific parameter names and a size-sensitivity note. The hint text is clear and actionable.

Union coercion fix is correct: the itemsAcceptString check now handles anyOf/oneOf at the items level. This means engram_suggestions (which has union items from #231, anyOf:[string,object]) now correctly accepts both comma-separated strings and JSON-stringified arrays. The previous items?.type === 'string' check was false for union schemas.

Minor: the union check traverses anyOf/oneOf but not allOf or . These are not present in the current tool schemas, so this is not a blocker. Worth a code comment noting the intentional scope.

Test coverage is solid: 4 new tests, including a regression test for the partial-drop case, a test confirming non-array missing fields do not trigger the #297 hint, and two coercion tests for comma-separated and JSON-stringified union items.

Verdict: Risk is low. Bug fix with targeted tests. Ready to merge.

crtahlin (approved):
Independent review — **APPROVE** (high confidence: read the logic, ran the tests, mutation-checked the two failing-without-fix cases mentally against the diff).

Correct, minimal, well-diagnosed. The bug presents as a *server* schema error (`engram_suggestions: Required` for a field the caller demonstrably passed), which sends the caller down entirely the wrong path — both changes turn it self-correcting.

## Verified

- **Partial-drop hint** — `missingArrayParams` selects required fields that arrived `undefined` AND are `type: 'array'`, names them, and states the trigger is payload size (so the caller shrinks other fields rather than retrying an identical call). A missing *non-array* required field correctly stays silent — the hint would be noise.
- **Union-item coercion** — `itemsAcceptString` now also inspects `anyOf`/`oneOf` variants for a string type, so `engram_suggestions` (`anyOf:[{string},{object}]`, #231) finally takes the comma-separated path the #297 error message advertises. Coercing `"a, b"` → `["a","b"]` is valid against the string variant.
- Ran `packages/mcp/test/server.test.ts` locally: **36/36**, including the four new tests (2 fail-without-fix, 2 regression guards).

## Note

Full `packages/mcp` suite was 255/256 locally — the one failure is the pre-existing session-telemetry flake (`session.test.ts:202`) this PR documents and does not touch. It has now recurred (also red-ed #715's Node-24 CI), so per your own suggestion I filed it as #738 — and flagged that #719's removal of `_activeSessionId` is the likely root-cause fix.

No breaking changes (error-message text + input coercion only). CI green on all nodes.

PR #706, opened 2026-07-26
chore(deps): clear all audit findings in PLUR's own packages (94 → 0 shipped)

plur9:
`pnpm audit` reported **94 vulnerabilities (3 critical, 40 high)**. After this change **PLUR's own published packages — core, mcp, cli, earn, hermes, python — report zero**, and the tree total drops to 40.

## Fixes

**1. `js-yaml` 4.1.0 → 4.3.0 — high, real exposure**
GHSA-52cp-r559-cp3m: YAML merge-key chains force quadratic CPU. This one matters more than a transitive advisory usually would — `js-yaml` is how core parses `engrams.yaml`, so the vulnerable parser sits on the hot path of every load. Also clears GHSA-h67p-54hq-rp68 (moderate, same merge-key family). Stays inside 4.x; no API change.

**2. `sharp` `>=0.35.0` override — high**
GHSA-f88m-g3jw-g9cj, four inherited libvips CVEs, reached via `packages/core > @huggingface/transformers > sharp`. transformers is an `optionalDependency` of core (BGE embeddings via ONNX), so it ships. Every transformers release — 4.2.0 included — pins `sharp: ^0.34.5`, and for `0.x` that caret resolves below the patched 0.35.0, so an override is the only fix that isn't a downgrade. The same override was applied downstream, where the chain also surfaces.

**3. `tar` `>=7.5.21` override — critical + 4 more**
GHSA-23hp-3jrh-7fpw (critical, decompression DoS) plus GHSA-8x88-c5mf-7j5w, GHSA-vmf3-w455-68vh, GHSA-w8wr-v893-vjvp, GHSA-gvwx-54wh-qm9j, via `@huggingface/transformers > onnxruntime-node > tar`.

## Why 40 remain, and why that's the right stopping point

Every one of them is under `packages/claw > openclaw`.

`openclaw` is a **peerDependency** of `@plur-ai/claw` (`>=2026.3`). It is in the lockfile only because pnpm installs peers for local dev and test. It is not bundled, not published, and not resolved from our tree by consumers — whoever installs `@plur-ai/claw` brings their own OpenClaw. Those advisories are OpenClaw's upstream, not ours to pin. The root override already floors it at `>=2026.5.12`, and the installed 2026.7.1 is current.

Per-workspace attribution after the fix:

| Package | Findings |
|---|---|
| core, mcp, cli, earn, hermes, python | **0** |
| claw (via `openclaw` peer) | 40 |

## Verification

| Check | Result |
|---|---|
| `pnpm audit` | 94 → 40, **0 in shipped packages**, 0 critical |
| resolved `js-yaml` | 4.3.0 |
| `pnpm -r build` | clean |
| root `pnpm test` | **2843 passed, 0 failed**, 44 skipped |

Unrelated pre-existing issue noticed while verifying, **not** addressed here: `pnpm -r test` fails in `packages/cli` with `Projects definition references a non-existing file or a directory: packages/cli/packages/core`. `packages/cli` has no local vitest config, so its `vitest run` picks up the root config and resolves the relative `projects` paths against the package cwd. Root `pnpm test` is the working entry point. Worth a separate fix so per-package runs aren't broken.


plur9 (commented):

Routine security maintenance. 94 → 0 audit findings in shipped packages. No logic changes. CI green. Ready for merge.

plur9 (commented):
## Review: chore(deps): clear all audit findings in PLUR's own packages (94 → 0 shipped)

**CI: All checks passing (Node 20/22/24, Python SDK)**

### Summary

This is a clean, well-scoped dependency fix. The three vulnerability classes addressed are correctly described and the rationale for the remaining 40 (all under `packages/claw > openclaw` peerDependency) is sound.

### Package Updates Assessment

**`js-yaml` 4.1.0 → 4.3.0 (GHSA-52cp-r559-cp3m, high)** — This is a genuine fix, not a pin game. YAML merge-key quadratic CPU is real exposure on the hot `engrams.yaml` parse path. The version bump stays within 4.x (no API break). The lockfile confirms `js-yaml@4.3.0` is resolved. ✓

**`sharp >=0.35.0` override (GHSA-f88m-g3jw-g9cj, high)** — The PR correctly explains why an override is required: `@huggingface/transformers` pins `sharp: ^0.34.5`, and for `0.x` semver that caret resolves below 0.35.0. This is the only available fix that doesn't downgrade. The chain `core > @huggingface/transformers > sharp` is an optional dependency chain, so this is not bundled into the published package in the same way — but it does ship as an optional dependency. The override is appropriate. ✓

**`tar >=7.5.21` override (GHSA-23hp-3jrh-7fpw, critical + 4 more)** — Legitimate critical path through `@huggingface/transformers > onnxruntime-node > tar`. Override is the right tool. ✓

### Concern: `.embeddings-cache.json` Committed Artifact

This diff includes `packages/core/.embeddings-cache.json` — a large (~5 separate engram embeddings) JSON file with raw 384-dimension float vectors. This is generated state that should not be in version control. It will drift as the embedder or corpus changes, causes review noise, and will produce merge conflicts. It should be gitignored. This appears to have been committed alongside the dependency work (possibly a dev artifact from testing), but it doesn't belong here.

**This is a blocker**: please add `packages/core/.embeddings-cache.json` to `.gitignore` and drop this file from the PR.

### The Remaining 40 Findings

The explanation is correct: `openclaw` is a `peerDependency` of `@plur-ai/claw`, not bundled, not published in PLUR's own tree, and the root override already floors it at `>=2026.5.12`. The 40 remaining advisories are OpenClaw's upstream responsibility. This is the right stopping point. ✓

### Pre-existing Issue Disclosed

The note about `pnpm -r test` failing in `packages/cli` (vitest config path resolution) is useful housekeeping and appropriately flagged as out of scope for this PR.

### Verdict

The vulnerability fixes are correct and well-justified. The only issue is the committed `.embeddings-cache.json` artifact which should be removed and gitignored. Beyond that, this is ready to merge.

plur9 (commented):
## Review — #706: chore(deps): clear all audit findings (94 → 0 shipped)

### Findings

**Overrides added:** `sharp >=0.35.0` and `tar >=7.5.21`. Both are transitive dependency pins, not direct deps. These address known CVEs in older versions of those packages.

**Embeddings cache committed:** The diff includes `packages/core/.embeddings-cache.json` — a new file at 1-line stub. This appears to be a development artifact. Confirm whether this file should be gitignored or if it is intentionally committed (it contains BM25 embeddings for test fixtures which is reasonable to cache in CI, but the file was absent before this PR).

**No risky major bumps observed.** The overrides pin to minimum safe versions (>=), not to specific versions that could cause resolution problems.

**Audit findings:** 94 → 0 is a large reduction. Worth confirming the audit baseline was run post-override to verify no resolvable vulnerabilities remain (`pnpm audit` output clean).

CI passing across all Node versions confirms the overrides don't break the build.

### Verdict

Low risk. Dependency hygiene changes with passing CI. The `.embeddings-cache.json` addition warrants a quick confirmation that it belongs in the repo — if it is a build artifact, add it to `.gitignore`. Otherwise ready to merge.

crtahlin (changes requested):
The security work is correct and verified — one unrelated file should come out before merge.

## Blocking

**`packages/core/.embeddings-cache.json` is committed here but is gitignored and not on main.** The PR force-adds it (a single-line ~20KB JSON of raw embedding vectors keyed by engram ID). It's unrelated to the audit, isn't mentioned in the description (which says "4 files changed" with otherwise impeccable reasoning), and committing a gitignored runtime cache causes churn/conflicts and contradicts the repo's own ignore rule. It also puts the author's local engram IDs into the repo (vectors only, no statements — low sensitivity, but it shouldn't be there).

Fix: `git rm --cached packages/core/.embeddings-cache.json` and drop it from the branch. Everything else is good to go.

## Verified (keep as-is)

- **`js-yaml` → `4.3.0`** confirmed resolved — clears the merge-key quadratic-CPU GHSA on the `engrams.yaml` parse hot path. Stays in 4.x, no API change.
- **`sharp >=0.35.0` / `tar >=7.5.21` root overrides** clear the libvips + decompression-DoS CVEs reached via `@huggingface/transformers` (an optionalDependency of core, so it ships). Overrides are the right call — every transformers release pins `sharp: ^0.34.5`, below the patched 0.35.0.
- **`pnpm audit` on the branch: 40 findings, 0 critical, 0 in shipped packages** — matches the claim exactly. Every remaining finding is under `packages/claw > openclaw > …` (undici/ws/hono/protobufjs/minimatch), i.e. the OpenClaw peerDependency PLUR doesn't bundle or resolve for consumers. Correct stopping point.

Once the cache file is removed this is an approve.

Issue #707, opened 2026-07-26
Epic: one memory engine, two deployments — core as the engine of both


plur9:
## Goal

**One memory engine, two deployments.** Core's `Plur` gains a Postgres primary
store and a scope-aware query layer, so a multi-tenant server deployment can run
core itself instead of reimplementing it — and so published benchmarks describe
what every deployment actually runs.

Today the engine is effectively implemented twice: core's `Plur` for the
single-user case, and a separate implementation for the server case that shares
only three primitives with `@plur-ai/core` (`embed()`, `EMBED_DIM`,
`computeContentHash()`).

Three consequences, one root cause:

- **Core improvements don't propagate.** Retrieval and correctness fixes land in
  `Plur`. A deployment that doesn't run `Plur` doesn't get them.
- **Published benchmarks describe only core.** plur-bench vendors core's
  `searchEngrams` / `hybridSearchWithMeta`, so the numbers characterise the
  single-user path and nothing else.
- **Core has no scaling story.** `_resolveBackend()` defaults to `sqlite`, there
  is no size-based switching, and no index is actually initialised — so core
  loads the whole corpus into memory and brute-forces cosine. At ~4,700 engrams
  that is ~350 MB per process. It is the same root cause as the local RAM
  problem.

One piece of work fixes all three. The Postgres adapter is **core's large-corpus
tier** — built once, tested once, benchmarked once.

---

## The architecture in one sentence

> **A deployment resolves identity → a list of permitted scopes.
> It passes that list to core as a query filter.
> Authorization stays in the deployment. Filtering becomes core's.**

Everything in the phases below implements that sentence.

---

### Converges — the memory engine

- Engram storage
- BM25 + vector retrieval + RRF fusion
- `learn` / `recall` / `inject` semantics
- dedup, decay, reinforcement, supersedes, tensions

### Stays deployment-specific — and correctly so

Identity (OIDC / SAML / SCIM), the permission graph that resolves it, audit
chains, review/approval workflow, admin UI, multi-tenancy, and any REST surface.
Core is single-user and has no authorization concept, so there is nothing to
converge on. Out of scope, deliberately.

### Not in scope

Do **not** move authorization into core. Core filters; it does not authorize. A
scope-filter regression must never be the thing standing between a tenant and
its data — the deployment's permission layer stays the gate.

---

## Verified starting state

| Fact | Status |
|---|---|
[...]

PR #715, opened 2026-07-26
feat(core): first-class permitted-scope filter in the storage layer (#710)

### LINKED ISSUE #710: Phase 3: first-class permitted-scope filter in StorageFilter
Part of #707.

> **Ship this regardless of whether the wider programme proceeds.** It is a
> prerequisite for multi-tenancy in core and needs no async work, so it does not
> sit behind Phase 2.

## Problem

`StorageFilter` has no way to express *"restrict to this set of permitted scopes"*.

That leaves any consumer with an authorization model no option but to retrieve
broadly and filter afterwards, in application code. Post-filtering has a failure
mode that is silent and gets worse as a corpus grows:

1. **Dilution.** `LIMIT` counts rows the caller may not be allowed to see. If the
   permitted scopes are a small share of the corpus, the caller asks for N and
   receives some fraction of N, while relevant permitted rows sit just below the
   cut. Compensating with a fixed overfetch multiplier is a guess, and the right
   multiplier depends on a ratio the query layer never sees.

2. **Overfetch past the ANN candidate list.** For approximate vector indexes, the
   inflated fetch size interacts badly with the index's search-effort parameter
   (`ef_search` for HNSW, default 40 in pgvector). Once the overfetch exceeds it,
   recall degrades sharply — and again silently.

Core does not hit either today: exact vector search, single user, no scope
post-filter. That is precisely why the gap went unnoticed — it only bites a
multi-tenant consumer.

## Scope

- [ ] Add `scopes` to `StorageFilter`; `searchBM25` and `searchVector` accept it.
- [ ] Adapters filter **in the query**, not after it — `WHERE scope = ANY($scopes)`
      alongside the vector ordering, not a JS pass over the result set.
- [ ] Every adapter implementation honours it. An adapter that accepts an
      authorization filter and ignores it is worse than not having the feature.

## Semantics to pin

| Value | Meaning |
|---|---|
| absent | no scope restriction — unchanged behaviour |
| `[]` | matches **nothing** |
| non-empty | **exact** set membership |

The empty case is the one to get right: a principal with zero permitted scopes
must see zero rows. An empty list must never be widened to "unrestricted", which
a naive `if (scopes?.length)` guard would do.

Exact membership is deliberate, and deliberately unlike the existing `scope`
field (a visibility filter, which does expand). By the time a list reaches this
layer it is a fully-resolved authorization decision — no hierarchy expansion, no
family pass-through.

## Acceptance criteria

- [ ] A test where the permitted scopes are a **minority** of the corpus (e.g. 1
      scope of 10, ≥100 rows) asserts `recall(limit: N)` returns **N** permitted
      results whenever N permitted matches exist.
      **This test must fail on `main`** — that failure *is* the bug report.
- [ ] A test asserts no out-of-scope row is ever returned.
- [ ] The empty-list case has its own test, per adapter.
- [ ] `scopes` AND-composes with `status` / `scope` / `domain`, so combining
      filters can only narrow.
- [ ] Existing scope-isolation tests still pass unmodified — this must not become
      a way to *widen* visibility.

## Verification

```
pnpm install && pnpm -r build && pnpm test    # root only; `pnpm -r test` is
                                              # broken for packages/cli
```

Baseline on `main`: 2843 passed, 0 failed, 44 skipped. Paste exact counts into
the PR.

---

**Implemented in #715** — `scopes` on `StorageFilter`, applied in
`PGLiteAdapter.loadFiltered`/`searchBM25`/`searchVector` (both the pgvector path
and the BYTEA JS-cosine fallback), `IndexedStorage.loadFiltered`, and the YAML
branch of `Plur._filterEngrams`. 34 new tests; 2877 passed / 0 failed.

The corresponding downstream deployment work is tracked privately.

plur9:
Closes #710. Core half of Phase 3 of the [convergence epic](https://github.com/plur-ai/plur/issues/707).

Adds `scopes?: string[]` to `StorageFilter` and threads it through `searchBM25` / `searchVector`, so scope filtering happens **in the query** instead of after it.

## Why

`StorageFilter` had no way to express "restrict to this set of permitted scopes", which leaves any consumer with an authorization model no option but to retrieve broadly and filter afterwards in application code. Post-filtering has two failure modes, both silent:

- **Dilution** — a user whose permitted scopes are a small share of the corpus asks for N and gets a handful, with relevant in-scope engrams sitting just below the cut.
- **Overfetch past the ANN candidate list** — an inflated fetch size interacts badly with an approximate index's search-effort parameter (`ef_search` for HNSW, which pgvector defaults to 40). Once the overfetch exceeds it, recall degrades sharply.

Filtering in the query makes `LIMIT` count *permitted* rows: N asked is N returned.

## Semantics

Documented on the new `ScopeRestriction` interface:

| Value | Meaning |
|---|---|
| absent | no scope restriction — unchanged behaviour |
| `[]` | matches **nothing** |
| non-empty | **exact** set membership |

The empty case is the security-relevant one: a principal with zero permitted scopes must see zero engrams. An empty list is never widened to "unrestricted", which would be privilege escalation.

Exact membership is deliberate, and deliberately unlike the existing `scope` field (a visibility filter, which does expand). By the time a list reaches this layer it is the fully-resolved authorization decision — no hierarchy expansion, no personal-family pass-through. Exact membership is also what a permission model resolving group hierarchies upstream will hand down — the expansion has already happened by then.

`scopes` AND-composes with `status` / `scope` / `domain`, so combining filters can only narrow.

## Applied in every read path

The failure mode to avoid is an adapter that *accepts* an authorization filter and doesn't apply it — worse than not having the feature. So every path that takes a filter got it:

- `PGLiteAdapter.loadFiltered` / `searchBM25` / `searchVector` — SQL `scope = ANY($n::text[])`, on the pgvector path **and** the BYTEA JS-cosine fallback
- `IndexedStorage.loadFiltered` — SQL `scope IN (...)`, with `1 = 0` for the empty list
- `Plur._filterEngrams` YAML branch — via the new shared `scopeAllowFilter` predicate, plus `scopes` on `RecallOptions` and `list()`

## Verification

34 new tests across two files, one set per adapter implementation: regression guard (absent ≡ old behaviour), the empty-list security case, exact membership, AND-composition, and an explicit dilution case.

| Check | Result |
|---|---|
| root `pnpm test` | **2877 passed, 0 failed**, 44 skipped |
| baseline on `main` | 2843 passed — delta is exactly the 34 new tests |
| `core-pglite` project | 56 passed, 1 skipped |

## Related

The corresponding downstream deployment work — supplying the resolved scope list and tuning the index's search-effort parameter — is tracked privately and ships independently. Semantics were kept identical on both sides so the two compose.

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -47,7 +47,7 @@ import { RemoteStore, normalizeEndpointUrl } from './store/remote-store.js'
 import { YamlPrimaryStore } from './store/yaml-primary-store.js'
 import type { PrimaryStore } from './store/primary-store.js'
 import { requiresIndexSync, asDerivedIndex } from './storage-adapter.js'
-import { isSharedScope, isPersonalScope, isScopeWithin } from './scope-util.js'
+import { isSharedScope, isPersonalScope, isScopeWithin, scopeAllowFilter } from './scope-util.js'
 import type { Engram } from './schemas/engram.js'
 import type { Episode } from './schemas/episode.js'
 import type { PackManifest } from './schemas/pack.js'
@@ -99,11 +99,11 @@ export { rankScopes, SCOPE_MATCH_THRESHOLD, WEIGHT_TAG, SUGGEST_DISPLAY_MIN_CONF
 // importing it from here would form index → inject → index. They are imported
 // above for internal use and re-exported here so the public `@plur-ai/core` API
 // (`isSharedScope`, `isPersonalScope`, `SHARED_SCOPE_PREFIXES`) is unchanged.
-export { isSharedScope, isPersonalScope, SHARED_SCOPE_PREFIXES } from './scope-util.js'
+export { isSharedScope, isPersonalScope, SHARED_SCOPE_PREFIXES, scopeAllowFilter } from './scope-util.js'
 export { detectPlurStorage, type PlurPaths } from './storage.js'
 export { IndexedStorage } from './storage-indexed.js'
 export { PGLiteAdapter, type PGLiteAdapterOptions, type VectorPrecision } from './storage-pglite.js'
-export type { StorageAdapter, StorageFilter, VectorSearchHit, StorageAdapterRole, DerivedIndexAdapter } from './storage-adapter.js'
+export type { ScopeRestriction, StorageAdapter, StorageFilter, VectorSearchHit, StorageAdapterRole, DerivedIndexAdapter } from './storage-adapter.js'
 export { requiresIndexSync, asDerivedIndex } from './storage-adapter.js'
 export { YamlStore, SqliteStore, createStore, migrateStore, type EngramStore, type StorageBackend, type StorageConfig } from './store/index.js'
 export { YamlPrimaryStore, MemoryPrimaryStore, type PrimaryStore, type PrimaryStoreKind } from './store/index.js'
@@ -2359,7 +2359,7 @@ export class Plur {
   }
 
   /** List all active engrams, optionally filtered by scope/domain. No search — returns all matches. */
-  list(options?: { scope?: string; domain?: string; min_strength?: number; include_expired?: boolean }): Engram[] {
+  list(options?: { scope?: string; scopes?: string[]; domain?: string; min_strength?: number; include_expired?: boolean }): Engram[] {
     return this._filterEngrams(options)
   }
 
@@ -2370,6 +2370,7 @@ export class Plur {
       engrams = this.indexedStorage.loadFiltered({
         status: 'active',
         scope: options?.scope,
+        scopes: options?.scopes,
         domain: options?.domain,
       })
     } else {
@@ -2381,6 +2382,14 @@ export class Plur {
       // so the cost is comparable.
       engrams = this._loadAllEngrams()
       engrams = engrams.filter(e => e.status === 'active')
+      if (options?.scopes !== undefined) {
+        // Permitted-scope allow-list (Phase 3). The in-memory twin of the SQL
+        // `scope = ANY($n)` pushdown, so the YAML path can never be the one
+        // read path that silently ignores an authorization filter. EXACT
+        // membership; `[]` matches nothing — see scopeAllowFilter.
+        const allowed = scopeAllowFilter(options.scopes)
+        engrams = engrams.filter(e => allowed(e.scope))
+      }
       if (options?.domain) {
         engrams = engrams.filter(e => e.domain?.startsWith(options.domain!))
       }
--- a/packages/core/src/scope-util.ts
+++ b/packages/core/src/scope-util.ts
@@ -72,3 +72,28 @@ export function isScopeWithin(scope: string, queryScope: string): boolean {
     || scope.startsWith(queryScope + ':')
     || scope.startsWith(queryScope + '/')
 }
+
+/**
+ * Permitted-scope allow-list predicate — the in-memory twin of the SQL
+ * `scope = ANY($n)` pushdown (see `StorageFilter.scopes` in
+ * storage-adapter.ts).
+ *
+ * Returns a predicate so every read path — SQL-backed adapters and the
+ * in-memory YAML path alike — agrees on the same three cases:
+ *   - `undefined` → unrestricted (predicate always true)
+ *   - `[]`        → matches NOTHING (predicate always false). Security-
+ *                   relevant: a principal with no permitted scopes must see
+ *                   nothing. NEVER treat an empty list as "no filter".
+ *   - non-empty   → EXACT membership. No hierarchy expansion, no personal-
+ *                   family pass-through: the list is the fully-resolved
+ *                   authorization decision, so `['project:a']` does NOT admit
+ *                   `project:a:sub`, `global`, or `local`.
+ *
+ * The Set is built once so the per-engram check is O(1) — the caller may be
+ * filtering a whole corpus against a list of tens of scopes.
+ */
+export function scopeAllowFilter(scopes: readonly string[] | undefined): (scope: string) => boolean {
+  if (scopes === undefined) return () => true
+  const allowed = new Set(scopes)
+  return (scope: string) => allowed.has(scope)
+}
--- a/packages/core/src/storage-adapter.ts
+++ b/packages/core/src/storage-adapter.ts
@@ -47,9 +47,52 @@
  */
 import type { Engram } from './schemas/engram.js'
 
+/**
+ * Permitted-scope allow-list — the authorization filter (Phase 3, scope
+ * pushdown).
+ *
+ * This exists so a caller that has ALREADY resolved an identity to a concrete
+ * set of permitted scopes — typically in an identity/permission layer sitting
+ * above core — can push that decision INTO the query, instead of filtering the
+ * result set afterwards.
+ *
+ * Post-filtering an unrestricted nearest-neighbour list is the dilution bug
+ * this type exists to prevent: `LIMIT` counts rows the caller may not be
+ * allowed to see, so with a fixed overfetch factor a caller whose permitted
+ * scopes are a small share of the corpus asks for N results and silently gets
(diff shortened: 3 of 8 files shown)


plur9 (commented):

Permitted-scope filter pushed into the storage layer where it belongs (#710). Moving scope enforcement out of caller code and into the query layer closes a class of scope-escape bugs. Clean impl, CI green across node 20/22/24. Ready for merge.

plur9 (commented):
## Review: feat(core): first-class permitted-scope filter in the storage layer (#710)

### Summary

Solid security-relevant feature. The three-case semantics (absent/empty/non-empty) are defined precisely, consistently implemented across all read paths, and the tests are designed to prove the security-critical degenerate case (`[]` = nothing) as well as the non-regression case (absent = old behaviour). This is the right way to add an authorization filter.

### Semantic Correctness

**The empty-list case** — `scopes: []` compiles to `1 = 0` (SQLite) and `scope = ANY(ARRAY[]::text[])` (PGLite/Postgres). Both evaluate to false for every row, which is correct. The alternative — silently dropping an empty filter — would be a privilege escalation. The tests explicitly cover this. ✓

**Exact membership, no hierarchy expansion** — The implementation correctly distinguishes `scopes` (authorization, exact) from `scope` (visibility, hierarchy-aware). The test at `pglite-scope-pushdown.test.ts:569-583` directly validates that `scopes: ['project:a']` does NOT pass through `project:a:sub` or `global` — exactly the opposite of the `scope` visibility filter. This is the right design. ✓

**AND composition** — `scopes` is applied before other filters in all paths, composing as an AND intersection. A caller combining `scope` (visibility) with `scopes` (authorization) gets the narrower result. Verified by test. ✓

### Implementation Coverage

The filter is applied in four paths:
1. `PGLiteAdapter.loadFiltered` — `scope = ANY($n::text[])` ✓
2. `PGLiteAdapter.searchVector` (pgvector path) — filter inside the k-NN query, not post-hoc ✓  
3. `PGLiteAdapter.searchVector` (BYTEA/JS-cosine fallback) — filter in SQL, JS does cosine only ✓
4. `PGLiteAdapter.searchBM25` — passed into `loadFiltered` before BM25 scoring ✓
5. `IndexedStorage.loadFiltered` — `scope IN (...)` / `1 = 0` ✓
6. `Plur._filterEngrams` YAML branch — via `scopeAllowFilter` predicate ✓

The PR body correctly notes that an adapter accepting but not applying an authorization filter would be "worse than not having the feature." Coverage of all six paths closes that failure mode.

### The Dilution Bug

The test at `pglite-scope-pushdown.test.ts:665-685` demonstrates the actual failure case: post-filtering a `limit*3` overfetch returns fewer than `limit` in-scope results when out-of-scope engrams dominate the nearest-neighbour list. The pushdown returns exactly `limit`. This is not a theoretical concern — the fixture is constructed to make it reproducible. Good proof of value.

### SQL Safety

**PGLite**: `scope = ANY($n::text[])` uses a parameterized array — no injection risk. ✓  
**IndexedStorage**: `scope IN (${filter.scopes.map(() => '?').join(', ')})` — one placeholder per scope, params spread. The code correctly notes that SQLite's variable limit (32766) is not in reach for an authorization decision list, and documents the `json_each` alternative if ever needed. ✓

### One Concern: `RecallOptions.scopes` Not Threaded Through `recall()`

The diff adds `scopes?: string[]` to `RecallOptions` and to `list()`, but I don't see it threaded into the `recall()` method's vector search path in the diff. The `recall()` method calls `searchVector` — is `opts.scopes` passed through to `searchVector` there? If not, the YAML path gets the filter (via `_filterEngrams`) but the PGLite vector search path in `recall()` would silently ignore it. This would be the dilution bug wearing a slightly different hat. Please verify `recall()` passes `scopes` into its `searchVector` call.

### Minor

- `scopeAllowFilter` builds a `Set` once and returns an O(1) predicate — correct, efficient. ✓
- The comment "An EMPTY array must match NOTHING: `= ANY(ARRAY[]::text[])` is false for every row" — this is worth verifying against the actual Postgres docs but is standard behaviour. ✓

### Verdict

The design is correct and the test coverage is exemplary. The one item that needs verification before merge: confirm `scopes` from `RecallOptions` is threaded into the vector search path inside `recall()`, not just into `list()`. If it is, this is ready. If not, that's a silent miss that should be fixed here.

plur9 (commented):
Review — #715: feat(core): first-class permitted-scope filter in the storage layer (#710)

Findings:

Three-case semantics (absent/empty/present) — correctly implemented. undefined means no scope restriction (admin/backward-compat). [] matches nothing. Non-empty list restricts to those scopes. This correctly implements least-privilege: an empty permission set yields zero results, not org-wide access.

SQL parameterization — no injection risk. Scope values are passed as bind parameters via ANY param patterns, never string-interpolated.

Empty-list guard prevents privilege escalation. The explicit short-circuit on empty scopes before any query is correct. The 1=0 trick (SQLite) and ANY(ARRAY[]::text[]) (PGLite) both correctly return no rows and are tested.

Dilution test is the right proof. Constructing a 120-engram corpus where 100/120 are out-of-scope, then verifying the caller gets their requested 10 in-scope results (not ~2) is exactly the regression anchor needed.

Backward compatibility — preserved. Callers that omit scopes get unchanged behaviour. The optional parameter with undefined default is correct.

One question to verify before merge: Confirm that the recall() code path in the Plur class threads opts.scopes into searchVector(). If that call site was missed, the dilution bug survives on the vector branch of recall(). Agent 1 flagged this as an open question — if confirmed threaded, this is clean and ready.

Verdict: Risk is medium (security-adjacent scope filtering). Core logic is correct. Merge after confirming the recall() to searchVector() scope threading is complete.

plur9:

All green on Node 20/22/24.

plur9:

crtahlin (dismissed):
Independent review — **APPROVE**. Verified by reading every read path and running the tests locally.

## Security: correct

The authorization filter is sound. The empty-list privilege-escalation trap is handled at every site via `!== undefined` (never truthiness): `= ANY(ARRAY[]::text[])` (pgvector + BYTEA fallback), `1 = 0` (SQLite, avoiding invalid `IN ()`), `Set.has` over `[]` (YAML). All SQL parameterized — no injection via scope strings. Applied on all three adapter paths **including the easy-to-miss BYTEA JS-cosine fallback**, plus the YAML path. On recall, `searchVector` hits are intersected with the scope-filtered `filtered` set and `scopes: []` makes `filtered` empty → early `return { engrams: [] }`. No path returns out-of-scope engrams.

## Tests

Ran locally: `scope-pushdown.test.ts` 18/18 (adapter + `list()` paths); `pglite-scope-pushdown.test.ts` green in CI's core-pglite project (56 pass). Full CI green (2877, +34 vs 2843 baseline). The 34 tests thoroughly cover the adapter pushdown (incl. empty-list and dilution) and `list({ scopes })`.

## Follow-up (non-blocking) — #735

The pushdown is plumbed into the adapter but **not consumed by the recall entrypoint**: `recallHybridWithMeta`/`_pgliteSemanticRecall` call `searchVector(queryVec, embLimit)` without `scopes` and post-filter by intersection — correct/secure, but it reintroduces the dilution the PR exists to fix for `recallHybrid({ scopes })` on PGLite, and the `RecallOptions.scopes` doc overstates that it pushes down. #710's acceptance is adapter-level and fully met, so this is a follow-up (#735: wire `options.scopes` into the recall adapter calls + fix the doc + add a recall-level test), not a blocker.

## Merge note

#716 has now merged to main, so this branch is `CONFLICTING` — those are the expected stack-rebase conflicts on `index.ts` / `storage-adapter.ts` / `storage-pglite.ts` (both PRs edited the same regions; #716 landed first per the order table). A rebase onto post-#716 main resolves them mechanically; the substantive scope-filter logic is unaffected. Merge order remains #716 → **#715** → #719 → …, and this approval is on the code, not a directive to merge ahead of your audit.

plur9:

PR #716, opened 2026-07-26
refactor(core): split the primary store from the query index (ADR-0003) (#708)

### LINKED ISSUE #708: Phase 1: split index from store in StorageAdapter
Part of #707 — core↔enterprise convergence.

## Problem

`StorageAdapter` today is an **index** adapter, not a store adapter. Two of its
eight methods encode that assumption:

- `syncFromYaml()` — "apply a YAML→index delta"
- `reindex()` — "drop the index and rebuild it from YAML"

Both assume **YAML owns the data** and the backend is a derived cache. Enterprise
needs Postgres to *be* the source of truth — there is no YAML file to rebuild from.
Until this is resolved, enterprise cannot implement `StorageAdapter` at all, and
every other phase in the programme is blocked behind it.

This is also what core needs for its own scaling story, independent of enterprise:
`_resolveBackend()` currently defaults to `sqlite` with no index actually
initialised, so core brute-forces cosine over the whole corpus in memory.

## Scope

- [ ] Introduce a primary-store capability. Two candidate designs — **pick one and
      record the decision in an ADR**:
      1. a separate `StoreAdapter` interface, with `StorageAdapter` retained as the
         index-only shape; or
      2. a capability flag on `StorageAdapter` where `syncFromYaml`/`reindex` become
         optional, and YAML becomes *one implementation* rather than the assumed root.
- [ ] Make `Plur` source-of-truth agnostic: no direct `loadEngrams` / `saveEngrams`
      calls in `index.ts`; **all** persistence goes through the adapter.
- [ ] Keep the YAML adapter as the default so single-user behaviour is byte-identical.
- [ ] Fold in the core pin bump for enterprise (0.9.13 → 0.14.0/0.15.0) — hygiene,
      verified data-safe and functionally inert, so it costs nothing to do here.

## Acceptance criteria

- [ ] A store adapter can be registered that has **no YAML file** and no
      `syncFromYaml`/`reindex` implementation, and `Plur` works against it end to
      end: `learn` → `recall` → `inject` → `updateEngram` → `compact`.
- [ ] Grep proves it: no `loadEngrams(` / `saveEngrams(` call sites remain in
      `core/src/index.ts` outside the YAML adapter implementation.
- [ ] Default (YAML) behaviour is unchanged — the existing core suite passes with no
      test modified, deleted or weakened.
- [ ] A new test asserts the primary-store path: an in-memory or PGLite-backed
      primary store round-trips a `learn` and a `recall` without touching disk YAML.
      **The test must fail on `main` before the change.**
- [ ] ADR committed recording the interface choice (1 vs 2) and why.

## Verification (CI is blocked — do this locally)

```
pnpm install && pnpm -r build && pnpm test    # root only; `pnpm -r test` is a known
                                              # pre-existing config bug for packages/cli
```
Paste exact pass/fail counts into the PR.

## Notes

- `PGLiteAdapter` is the working async adapter precedent — read it before designing.
- This phase is design work. Getting the interface right matters more than landing it
  fast; Phases 2–6 all sit behind it.

plur9:
Closes #708. Phase 1 of the [convergence epic](https://github.com/plur-ai/plur/issues/707) — the unlock everything downstream waits on.

## The problem

`StorageAdapter` was named like a storage interface but was an **index** interface. `syncFromYaml()` ("apply a YAML-to-index delta") and `reindex()` ("drop the index and rebuild from YAML") were both **required**, which encodes "YAML owns the data, this backend is a derived cache". A Postgres backend that *is* the source of truth can honour neither.

Worse, `Plur` never went through a storage abstraction for its own state: it called `loadEngrams(this.paths.engrams)` at **26 sites** in `index.ts` and `saveEngrams(...)` at 3 more (plus 2 in `learn-async.ts`). The source of truth was hard-coded in the caller rather than living in a swappable component.

## Design decision — two interfaces, not one interface with a flag

The plan left this open: *"separate `StoreAdapter` interface, or capability flag on `StorageAdapter`?"*

The question contains a false premise. **`StorageAdapter` has no engram-persistence methods at all** — no `save`, `append`, `update` or `remove` anywhere on it. (It does have `upsertEmbedding`, but that writes *derived index* data, which is consistent with it being an index.) No capability flag can promote it to a store, because the half being promoted doesn't exist. A flag alone would produce an interface where `role: 'primary'` means "trust me, some *other* code writes to the same database" — exactly the implicit coupling this phase deletes.

So persistence gets its own interface, and the flag stays but does a smaller, honest job:

| Interface | Role |
|---|---|
| `PrimaryStore` *(new)* | Persistence. Owns the bytes. `load` / `loadCached` / `save` / `invalidate`. `YamlPrimaryStore` is the default. |
| `StorageAdapter` *(kept)* | Query. Gains `role: 'index' \| 'primary'`; `syncFromYaml` / `reindex` become **optional**. |

`role` answers exactly the question the old design got wrong: *does a write to the source of truth leave this backend stale?* `requiresIndexSync(adapter)` and `asDerivedIndex(adapter)` replace the assumption that the rebuild methods exist. `PGLiteAdapter` declares `role: 'index'`.

Phase 5's `PostgresAdapter` will implement **both** — one class saying "my store and my query index are the same engine".

Full rationale in `docs/adr/ADR-0003-primary-store-capability.md`.

## Why `PrimaryStore` is synchronous (deliberately, and temporarily)

Every `Plur` write path is synchronous today (`withLock` + `readFileSync`). Converting them is **Phase 2**, kept separate on purpose: flipping ~20 methods to async turns every `await` into an interleaving point and invalidates the implicit atomicity the current code relies on (`_sessionScope`, `_activeSessionId`, `autoDiscoverStores`, the LLM circuit breaker). Doing both at once leaves no bisect point between "moved the seam" and "changed the concurrency model".

`PrimaryStore` is shaped like the existing async `EngramStore` minus the Promises, so Phase 2 is a mechanical sync→async flip.

## Behaviour

Unchanged for single-user YAML installs — that is the acceptance criterion for this phase. The YAML adapter stays the default; this is a refactor, not a behaviour change.

## Verification

| Check | Result |
|---|---|
| root `pnpm test` | **2878 passed, 0 failed**, 44 skipped |
| baseline on `main` | 2843 — delta is exactly the 35 new tests |
| `loadEngrams(` in `index.ts` | 26 → **0** |
| stability | 3 consecutive full runs, identical counts |

35 new tests across `primary-store.test.ts`, `storage-adapter-role.test.ts` and `plur-source-of-truth.test.ts` — the last of which is the one that matters: it pins that `Plur` no longer reaches for a file path of its own.


plur9 (commented):
CTO review — CI: all passing (issue-guard + test × 3 nodes).

ADR-0003 correctly separates PrimaryStore (persistence seam) from StorageAdapter (query index). The key insight — syncFromYaml/reindex are index concepts, not store concepts — is right. The role field + requiresIndexSync helper are a clean abstraction. Synchronous-for-now is the correct call given Phase 2's scope; the rationale is clearly documented.

No security or correctness concerns. Reviewed as part of the #719 stack — this PR must merge first.

plur9:
@crtahlin — assigning you the convergence stack for review. **You can merge these yourself** once you're happy; no need to hand back.

These are stacked, so the order matters. Merging bottom-up keeps each diff small; merging out of order makes the later ones look enormous.

| # | Phase | Depends on | What it is |
|---|---|---|---|
| #716 | 1 | — | Split the primary store from the query index (ADR-0003) |
| #715 | 3 | — | Permitted-scope filter pushed into the storage layer |
| #719 | 2a | #716 | Concurrency correctness before the async flip (ADR-0004) |
| #720 | 5 | #716 | Postgres primary store + size-based backend selection (ADR-0005) |
| #728 | 2b | 716, 715, 719, 720 | The async write path — this is the one that makes `Plur` runnable on Postgres |
| #724 | — | — | BM25 reverse-substring fix (#721). **Merge before #732.** |
| #732 | 4 | #728, #724 | BM25 pushdown — narrow with pg_trgm, score in core with global corpus statistics |

**One ordering trap:** #732 carries a cherry-pick of #724. Merge #724 first and git will reconcile them; merge #732 first and #724 becomes a confusing empty-ish diff.

All seven are green on Node 20/22/24 and currently MERGEABLE.

Not in this stack, independent, mergeable whenever: #705, #706, #717, #723, #726.

### Why the order is worth respecting

#728 is the load-bearing one. Before it, `Plur`'s write path was synchronous, which was a hard ceiling rather than a preference — there is no synchronous Postgres client for Node, so no network-backed store could ever satisfy the store contract. #716 and #719 exist to make that flip safe rather than exciting: #716 separates persistence from querying so there is a seam to flip, and #719 fixes the concurrency assumptions the synchronous path was quietly relying on. Merging #728 without them is possible but the diff stops explaining itself.

### What this unblocks

Convergence Phase 6 — enterprise dropping its own engine and running core — is blocked on these merging **and core cutting a release**. Enterprise currently pins `@plur-ai/core` at `0.9.13`, so a merge alone is not enough; it needs a publish. Phase 7 (benchmark parity across both deployments) is behind the same gate.

Worth a look while reviewing, all found by the test-suite typecheck added in #728 and filed separately: #729 (`learn()` never validates `type`, so schema-invalid engrams persist), #730 (`--quiet` is documented but read by nothing), #731 (no read-only mode, though `plur tensions` asked for one).

plur9:

crtahlin (approved):
Independent review — **APPROVE** (code is ready). Verified by reading the full seam and running tests locally, not just the diff.

## What I verified

- **Call-site migration is semantically faithful.** All 26 `loadEngrams` / 5 `saveEngrams` sites route through the new `PrimaryStore`: write-transaction reads → `load()` (authoritative/uncached), hot reads → `loadCached()` (mtime-keyed), writes → `save()` (invalidates cache — the #25 fix preserved). Structural grep confirms zero direct `loadEngrams`/`saveEngrams` remain in `index.ts` and `learn-async.ts`.
- **One coherent cache.** `learn-async.ts` receives the *same* `_primaryStore` instance via the deps harness, so both write paths invalidate one cache — no desync. `withLock` stays keyed on the file path as a pure lock key, which is what lets `MemoryPrimaryStore` work with no file.
- **Behaviour unchanged for the YAML default.** The `role` / `requiresIndexSync` / `asDerivedIndex` guards are inert while `PGLiteAdapter` is `role: 'index'`; they only branch once a `role: 'primary'` backend exists (Phase 5).
- **Tests pin the right invariants.** `plur-source-of-truth.test.ts` runs learn→recall→feedback→forget against `MemoryPrimaryStore` and asserts no `engrams.yaml` is created, plus the structural grep.

**Local run:** 35/35 new invariant tests pass; 14/14 on the three touched files (`batch`, `learn-async-scope`, `learn-batch-cap`). Matches CI (2878 pass, +35 vs the 2843 baseline).

## Non-blocking

- `role` is now a **required** field on the exported `StorageAdapter` interface — a compile-time break for external implementors, documented in ADR-0003 but not in `CHANGELOG.md`. Filed as #734 (P3) to add a release-note entry, rather than amending this PR (which would invalidate this approval). Break surface is ~nil today (only `PGLiteAdapter` implements it in-repo).
- Three storage-shaped interfaces now coexist (`PrimaryStore` sync / `EngramStore` async-unused / `StorageAdapter` query) — accepted, temporary until the Phase 2 async flip. Documented.

## Note on merging

This approval is a verdict on the code, **not** a green light to merge now — respecting your retraction of the "merge these yourself" offer pending the adversarial audit. When the hold lifts, merge the stack bottom-up (#716 first) per your ordering table.

PR #717, opened 2026-07-26
feat(mcp): add side-effect-free ./tools subpath export

### LINKED ISSUE #714: @plur-ai/mcp: expose tool definitions via a side-effect-free subpath export (blocks enterprise importing them)
## Problem

Enterprise hand-copies the PLUR MCP tool schemas inline rather than importing them from `@plur-ai/mcp`. That is duplication #3 in the core↔enterprise drift report, and Phase 0 of the [convergence plan](https://github.com/plur-ai/plur/issues/707) lists "import MCP tool definitions from `@plur-ai/mcp` instead of the inline hand-copy" as an independent early win, noting the dependency is already declared but unused.

**It is currently not possible**, and the reason is packaging rather than anything in enterprise.

`@plur-ai/mcp@0.9.13` declares exactly one export, and it is also the `bin`:

```json
"exports": { ".": { "types": "./dist/index.d.ts", "import": "./dist/index.js" } },
"bin":     { "plur-mcp": "dist/index.js" }
```

Two consequences:

1. **`import '@plur-ai/mcp'` never returns.** `dist/index.js` is the CLI entrypoint — on evaluation it parses `process.argv` and starts a stdio MCP server bound to `process.stdin`. It also re-exports only `compareSemver` / `extractManifestVersion`, so even if it did return, the tool definitions are not among them.
2. **The internal chunk holding `getToolDefinitions()` is unreachable.** Any subpath import fails with `ERR_PACKAGE_PATH_NOT_EXPORTED`; the chunk that is reachable exports only `createServer` / `runStdio`.

So a consumer that wants the tool schemas has exactly two options today: copy them (the status quo, which drifts), or spawn the package as a subprocess and ask it over MCP (workable for a test, absurd as a runtime dependency).

## Ask

Add a side-effect-free subpath export for the tool definitions, e.g.

```json
"exports": {
  ".":      { "types": "./dist/index.d.ts", "import": "./dist/index.js" },
  "./tools": { "types": "./dist/tools.d.ts", "import": "./dist/tools.js" }
}
```

Requirements:

- **No side effects on import.** The module must not read `process.argv`, touch `process.stdin`, or start a server. This is the actual defect — a package whose only entrypoint boots a server cannot be used as a library.
- Export the tool definitions as plain data (name, description, input schema), so a consumer can filter or re-expose them without reimplementing.
- Keep `"."` as-is so the CLI/bin behaviour is unchanged.

## Why it's worth doing

Once this exists, a downstream consumer imports one list and its hand-copy disappears — the tool surface stops being a thing two codebases maintain in parallel.

A consumer that re-exposes a filtered subset of these tools currently has no way to notice when upstream adds, removes or changes one. The only alternatives available today are to copy the schemas (which drifts) or to spawn the package as a subprocess and ask it over MCP (workable for a test, absurd as a runtime dependency). Both have been observed to produce real, silent divergence: tools that exist in a consumer's configuration but no longer upstream, tools that are reachable but never advertised, and required-parameter sets that no longer match.

Making the definitions importable removes the whole class.

## Note on `plur_batch_decay`

Recording this because it is easy to get backwards: `plur_batch_decay` is **not** stale at `0.9.13` — that version still ships it, verified against the running server. It was removed upstream in #563, so it becomes stale only once a consumer's pin moves past that release.

plur9:
## Summary

Fixes #714.

- Adds `@plur-ai/mcp/tools` subpath export backed by a new `src/tools-export.ts` entry point
- Importing this subpath does **not** parse `process.argv`, touch `process.stdin`, or start the MCP server — it only re-exports the tool surface from `tools.ts`
- `getToolSchemas(profile?)` returns tool definitions as plain, handler-free objects (name, description, inputSchema, annotations) — the "plain data" form the issue asks for
- `getToolDefinitions` and `CURSOR_CORE_TOOL_NAMES` are also re-exported for consumers building MCP servers on top of PLUR
- `tsup.config.ts` gains a second entry so tsup emits `dist/tools-export.{js,d.ts}`
- `package.json` exports map `"./tools"` → `"./dist/tools-export.{js,d.ts}"`

**Usage after this PR:**
```ts
import { getToolSchemas } from '@plur-ai/mcp/tools'

const schemas = getToolSchemas()          // lean/cursor profile (default)
const full = getToolSchemas('full')       // all tools including admin-only ones
```

## Test plan

- [x] `pnpm --filter @plur-ai/mcp build` — both entry points build cleanly, `dist/tools-export.{js,d.ts}` emitted
- [x] New `test/tools-export.test.ts` — 6 tests covering `getToolSchemas`, `getToolDefinitions`, `CURSOR_CORE_TOOL_NAMES` re-export, lean/cursor parity, and full > lean tool count
- [x] `npx vitest run test/tools-export.test.ts` — 6/6 pass
- [x] `npx vitest run test/server.test.ts` — existing server tests unaffected

### DIFF
--- a/packages/mcp/package.json
+++ b/packages/mcp/package.json
@@ -47,6 +47,10 @@
     ".": {
       "types": "./dist/index.d.ts",
       "import": "./dist/index.js"
+    },
+    "./tools": {
+      "types": "./dist/tools-export.d.ts",
+      "import": "./dist/tools-export.js"
     }
   }
 }
--- a/packages/mcp/src/tools-export.ts
+++ b/packages/mcp/src/tools-export.ts
@@ -1,0 +1,38 @@
+/**
+ * Side-effect-free subpath export: @plur-ai/mcp/tools
+ *
+ * Importable without starting the MCP server or parsing process.argv.
+ * Use this to access PLUR's tool surface — names, descriptions, input schemas
+ * — from a consumer that re-exposes or validates the definitions without
+ * running the server.
+ */
+export type { ToolAnnotations, ToolDefinition, ToolProfile } from './tools.js'
+export { getToolDefinitions, CURSOR_CORE_TOOL_NAMES, validateToolArgs } from './tools.js'
+
+import type { ToolAnnotations, ToolProfile } from './tools.js'
+import { getToolDefinitions } from './tools.js'
+
+/** Tool definition without the runtime handler — plain data, safe to serialize. */
+export interface ToolSchema {
+  name: string
+  description: string
+  inputSchema: { type: 'object'; [key: string]: unknown }
+  annotations?: ToolAnnotations
+}
+
+/**
+ * Return tool definitions as plain, handler-free schema objects.
+ *
+ * Useful when you need the MCP tool surface — name, description, and input
+ * schema — without the runtime handlers that require a live Plur instance.
+ * A downstream consumer that re-exposes a filtered subset can import this
+ * list and keep its copy in sync without duplicating schemas.
+ */
+export function getToolSchemas(profile?: ToolProfile): ToolSchema[] {
+  return getToolDefinitions(profile).map(({ name, description, inputSchema, annotations }) => ({
+    name,
+    description,
+    inputSchema,
+    ...(annotations !== undefined ? { annotations } : {}),
+  }))
+}
--- a/packages/mcp/test/tools-export.test.ts
+++ b/packages/mcp/test/tools-export.test.ts
@@ -1,0 +1,60 @@
+/**
+ * #714: @plur-ai/mcp/tools subpath export — side-effect-free tool schema access.
+ *
+ * Validates that the ./tools export (src/tools-export.ts) surfaces the tool
+ * definitions as plain, handler-free schema objects without touching
+ * process.argv or starting a server.
+ */
+
+import { describe, it, expect } from 'vitest'
+import {
+  getToolSchemas,
+  getToolDefinitions,
+  CURSOR_CORE_TOOL_NAMES,
+} from '../src/tools-export.js'
+
+describe('@plur-ai/mcp/tools subpath export (#714)', () => {
+  it('getToolSchemas returns objects with no handler field', () => {
+    const schemas = getToolSchemas('full')
+    expect(schemas.length).toBeGreaterThan(0)
+    for (const s of schemas) {
+      expect(s).not.toHaveProperty('handler')
+      expect(typeof s.name).toBe('string')
+      expect(typeof s.description).toBe('string')
+      expect(s.inputSchema.type).toBe('object')
+    }
+  })
+
+  it('getToolSchemas covers the core tool names', () => {
+    const schemas = getToolSchemas()
+    const names = new Set(schemas.map(s => s.name))
+    for (const core of CURSOR_CORE_TOOL_NAMES) {
+      expect(names.has(core)).toBe(true)
+    }
+  })
+
+  it('getToolDefinitions is re-exported (for consumers building MCP servers)', () => {
+    const defs = getToolDefinitions('full')
+    expect(defs.length).toBeGreaterThan(0)
+    expect(typeof defs[0].handler).toBe('function')
+  })
+
+  it('lean and cursor profiles return the same tool count', () => {
+    const lean = getToolSchemas('lean')
+    const cursor = getToolSchemas('cursor')
+    expect(lean.length).toBe(cursor.length)
+    expect(lean.map(s => s.name).sort()).toEqual(cursor.map(s => s.name).sort())
+  })
+
+  it('full profile returns more tools than lean', () => {
+    const full = getToolSchemas('full')
+    const lean = getToolSchemas('lean')
+    expect(full.length).toBeGreaterThan(lean.length)
+  })
+
+  it('CURSOR_CORE_TOOL_NAMES is re-exported', () => {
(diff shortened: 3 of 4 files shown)


plur9 (commented):

Side-effect-free `./tools` subpath export is correct. Lets consumers import tool definitions without triggering MCP server initialization. The subpath export pattern is idiomatic for this package structure. CI green across node 20/22/24. Ready for merge.

plur9 (commented):
## Review: feat(mcp): add side-effect-free ./tools subpath export

### Summary

Small, clean, correctly scoped. The implementation matches the stated goal.

### Side-effect-free Claim

The claim holds. `src/tools-export.ts` only imports from `src/tools.ts`. The critical question is whether `tools.ts` itself has module-level side effects. From the diff, `tools-export.ts` imports `getToolDefinitions` and `CURSOR_CORE_TOOL_NAMES` — if those are just function exports and constants, importing them at module evaluation time won't parse `process.argv` or start a server. The test file confirms this: the import succeeds in a test context where no server is running. ✓

### Package.json Exports

```json
"./tools": {
  "types": "./dist/tools-export.d.ts",
  "import": "./dist/tools-export.js"
}
```

Correct ESM-only export. Matches the rest of the package's export pattern (no `require` field, consistent with `format: ['esm']` in tsup). Missing a `default` condition but that's consistent with the existing `"."` export. ✓

### `getToolSchemas` Interface Design

The `ToolSchema` type strips `handler` from the definition objects, which is the right design for a "plain data" export. The spread `...(annotations !== undefined ? { annotations } : {})` correctly omits the field when absent rather than serializing `undefined`. ✓

### `getToolDefinitions` Re-export

Re-exporting `getToolDefinitions` (which DOES include handlers) from the side-effect-free subpath is slightly at odds with the "side-effect-free" framing — it exposes the full runtime surface including handlers. However, since `getToolDefinitions` is a function and not evaluated at import time, this doesn't violate the module-level side-effect guarantee. The use case (consumers building MCP servers on top of PLUR) is valid. Consider adding a JSDoc note clarifying that `getToolDefinitions` returns handler-bearing objects and is distinct from the handler-free `getToolSchemas`. Minor.

### Test Coverage

The 6 tests cover the key contracts: handler absence, CURSOR_CORE_TOOL_NAMES coverage, lean/cursor parity, full > lean count, and re-export presence. The test at line 37 (`expect(typeof defs[0].handler).toBe('function')`) is a good self-check that `getToolDefinitions` really does include handlers, confirming the two exports are genuinely distinct. ✓

### tsup Config

Adding `src/tools-export.ts` as a second entry point is correct — tsup will emit `dist/tools-export.{js,d.ts}` independently. ✓

### Verdict

Ready to merge. The one minor point (JSDoc note on `getToolDefinitions`) is purely cosmetic.

plur9 (commented):
## Review — #717: feat(mcp): add side-effect-free ./tools subpath export

CI: all green (test/20, test/22, test/24). Note: Python SDK check absent — not unexpected for an mcp-only change.

### Findings

**Side-effect-free claim holds.** The new `tools-export.ts` does not touch `process.argv` or start a server at module init time. It only re-exports from `tools.js` — which does have side-effect-free exports (the tool list is a pure function of a profile parameter).

**Package.json exports field — correct.** ESM-only, consistent with the existing pattern. Types + import both point to built dist files. `tsup.config.ts` correctly adds the new entrypoint.

**`getToolSchemas` strips handlers cleanly.** The destructure + conditional `annotations` spread avoids serializing `undefined` which would break JSON consumers.

**Re-export of `getToolDefinitions` (handler-bearing) from tools-export.ts is noted** — it is technically not handler-free, but it is not evaluated at module init so the module-level side-effect guarantee is preserved. Consumers who want truly handler-free should use `getToolSchemas`.

**Test coverage is solid.** 6 tests cover: no-handler invariant, core tool name coverage, profile differences, CURSOR_CORE_TOOL_NAMES re-export. The lean/cursor equality assertion is a good regression guard.

**Enterprise use case validated.** The primary motivation (enterprise keeping hand-copied tool definitions in sync with upstream) is a real problem this cleanly solves.

### Verdict

Risk: **low**. Clean, well-tested additive change. Ready to merge.

crtahlin (approved):
Independent review — **APPROVE** (high confidence: read the code, built both entries, ran the test).

Clean, correctly-scoped, side-effect-free subpath export.

## Verified

- **Genuinely side-effect-free.** `tools-export.ts` imports only from `./tools.js` (not the server bin `index.ts`), and `tools.ts` has no module-level side effects — no `process.argv`/`process.stdin`/`createServer`/`StdioServerTransport`, just a `Map` allocation and function defs. The build confirms it structurally: `dist/tools-export.js` is a **465 B** thin re-export, with the server code split into a separate chunk it never pulls in.
- **Handler-free schemas.** `getToolSchemas` destructures `{name, description, inputSchema, annotations}`, dropping `handler`, so the result is plain serializable data — exactly what #714 asks for.
- **Wiring correct.** `package.json` `./tools` → `dist/tools-export.{d.ts,js}`; tsup second entry emits both `.js` and `.d.ts`. Built locally, both entries clean.
- Ran `test/tools-export.test.ts`: **6/6**, including the no-handler-field assertion, lean/cursor parity, and full > lean count.

## Minor note (non-blocking)

The `./tools` export is ESM-only (`types` + `import`, no `require`) — consistent with the package being ESM, so fine; just means a CJS consumer can't `require('@plur-ai/mcp/tools')`. Worth a `require`/`default` condition only if a CJS consumer ever needs it.

Purely additive; no breaking changes. Good to merge.

Issue #709, opened 2026-07-26
Phase 2: async write path + concurrency correctness


plur9:
Part of #707 — core↔enterprise convergence. Depends on Phase 1.

## Problem

`Plur`'s write path is synchronous. The comment in enterprise's `enterprise-plur.ts`
calls core's `Plur` *"file-path-based and synchronous"* — true, but incomplete: the
async seam already exists (`StorageAdapter` is fully async, `embed()` is async,
`learn-async.ts` is already dependency-injected, `AsyncMutex` already ships). What
remains is a mechanical conversion — **and then the part that actually matters.**

Measured from `core/src/index.ts`, not estimated:

```
total methods:                 128
already async:                  29
touch the sync storage path:    29
  → still sync (must flip):     20
```

## Part A — the mechanical conversion (~1 week)

- [ ] `withLock()` → async. **22 call sites.** Use the existing `AsyncMutex`
      (`core/src/storage-pglite.ts:42`); this also removes the synchronous busy-wait
      spin in the current lock retry.
- [ ] `loadEngrams` / `saveEngrams` → `fs.promises` (currently `readFileSync` /
      `writeFileSync`).
- [ ] Flip the **20 measured** sync methods that touch storage: `learn`, `getById`,
      `updateEngram`, `setPinned`, `listPinned`, `compact`, `status`, `installPack`,
      `saveMetaEngrams`, `recordTensions`, plus internals (`_loadCached`,
      `_writeEngrams`, `_filterEngrams`, `_formatInjection`,
      `_recordCrossScopeRecurrence`, …). Once the sync root is gone these follow
      mechanically — async is contagious, but the logic is untouched.

## Part B — concurrency correctness (**this is the real project**)

A synchronous method is *implicitly atomic*: nothing else runs mid-execution. The
moment it becomes async, **every `await` is an interleaving point**, and code that
was accidentally safe stops being safe. The following are safe today **only because
sessions are sequential, one process per session**:

- [ ] `_sessionScope` — `core/src/index.ts:381`. A process-global field read on every
      unscoped `learn()`. Remove it or key it per session.
- [ ] `_activeSessionId` — `core/src/tools.ts:259`. Same class of problem.
- [ ] `autoDiscoverStores()` — called from the constructor and **persists to disk**.
      Concurrent construction races the config write.
- [ ] The **process-wide LLM circuit breaker** — one tripped breaker silently
      degrades every concurrent session sharing the process.

[...]

PR #719, opened 2026-07-26
fix(core): concurrency correctness — per-session scope, LLM breaker, queued lock (ADR-0004)

plur9:
Part of #709 (Phase 2). **Stacked on #716** (Phase 1) — review that first.

> **Scope note, up front:** this is the *concurrency correctness* half of Phase 2. **The async flip itself is not in this PR**, deliberately — see "Why the flip isn't here". The title reflects what actually landed.

## The bugs this fixes are real today

These were latent, not theoretical. Each was verified by reverting the source and watching the new tests fail.

### 1. Session scope bled between concurrent writers — engrams landed in the wrong scope

`_sessionScope` was a single field on the instance. `learnRouted()` already awaits a network round-trip between reading that field and persisting, so two sessions writing at once raced, and **the loser's engram was written into the winner's scope**.

In a system whose whole namespacing model is scopes, that is a correctness bug with a confidentiality edge.

Replaced with a `SessionScopeRegistry`; `LearnContext.session` threads the key. Unkeyed callers behave exactly as before.

Revert check: 4 tests fail, first one literally `expected 'project:beta' to be 'project:alpha'`.

### 2. The LLM circuit breaker never tripped under concurrency

`recordLlmSuccess()` zeroed a counter across an await, so a single success erased concurrent calls' failures. The breaker could not reach its threshold while anything else was in flight. Now a sliding window of failure timestamps; threshold unchanged at 3.

Revert check: 2 tests fail — `expected 6 to be 5` (breaker never tripped).

### 3. 25 concurrent writers exhausted the lock retry budget

`withAsyncLock` went straight at an `O_EXCL` lock with a 5-retry budget. Under fan-out that throws. Added an in-process FIFO queue in front of it.

Revert check: 3 tests fail — `Failed to acquire lock … after 5 retries`.

### 4. Latent self-deadlock in `learn-async.ts`

The UPDATE/MERGE "target vanished" branch called `deps.learn()` *inside* the lock, and `learn()` takes the same lock. It was only ever resolved by the inner acquire throwing. Hoisted out.

## Also

- `autoDiscoverStores()` in the constructor is now opt-out (`autoDiscover: false` / `PLUR_AUTO_DISCOVER=0`) and `cwd` is explicit. Default unchanged.
- MCP `_activeSessionId` removed. Both inject tools take an explicit `session_id`; the implicit fallback is derived and answered only when unambiguous. Revert check: 3 tests fail — `expected 2 to be 1` (session B charged with A's inject).

## Why the flip isn't here

Measured before deciding, not estimated. Making `PrimaryStore.load()` async cascades through `_loadAllEngrams` → `_filterEngrams` → `recall`/`inject`/`list`/`getById`/`status` — i.e. the entire public API: **~100 call sites in `packages/*/src`, ~1000 in `packages/*/test` across 77 files**, plus a hard break for published `@plur-ai/core` consumers.

It cannot be landed partially — `PrimaryStore` being async *is* the flip. A large half-done conversion with no bisect point between "moved the seam" and "changed the concurrency model" is worse than a complete increment of the part that carries the actual risk. The remaining work is enumerated in ADR-0004 with its dependency order.

Still outstanding: `loadEngrams`/`saveEngrams` remain `readFileSync`/`writeFileSync`, 13 sync `withLock` sites remain, and the `Date.now()` busy-wait in `sync.ts` is untouched.

## An invariant that is documented but not enforced

Sync and async callers share the same lock file. That is safe **only because no async lock body holds `paths.engrams` across an await**. Break it and a sync caller spins, blocking the holder. ADR-0004 states the rule; I could not find a non-brittle way to assert it mechanically. Worth a reviewer's eye.

## Honest notes

- **5 of the 7 write-path tests pass on the parent too.** They are regression guards, not evidence of a fix. Counted as such rather than folded into the "bugs fixed" tally.
- **Two pre-existing tests changed, no assertion weakened.** `plur-source-of-truth.test.ts` now `await`s `feedback`/`forget` — they always returned Promises and merely completed synchronously by accident; assertions are identical. `mcp/test/session.test.ts` gained a `_resetSessionTelemetry()` in `beforeEach` because that file leaks 5 unclosed sessions across its own tests.
- **Behaviour break:** `feedback` / `forget` / `setPinnedAsync` / `updateEngramAsync` now genuinely require `await`. No call site in this repo relied on the old accidental synchrony (grepped); external consumers might.
- **MCP implicit attribution is now conservative:** a session that never calls `session_end` stays open until the 8h TTL and suppresses implicit attribution for that window. Deliberate — a leak should cost telemetry, not correctness — but it is a real behaviour change.
- `benchmark/micro.ts` was not run; the machine was under load throughout and latency numbers would have been meaningless.

## Verification

| Check | Result |
|---|---|
| `pnpm test` (root) | **2918 passed, 0 failed**, 44 skipped |
| baseline on #716 | 2878 — delta is exactly the 40 new tests |
| `npx vitest run --project core-pglite` | 40 passed, 1 skipped |
| `pnpm build` / `tsc --noEmit` | clean |
| stability | 4 runs, identical counts |

The concurrency tests **interleave** rather than awaiting sequentially — `Promise.all` with forced suspension points and an explicit gate. Sequential awaits would prove nothing about atomicity, which is the entire point.


plur9 (commented):
CTO review — CI: all passing (issue-guard + test × 3 nodes).

Three live concurrency bugs fixed correctly:

1. **SessionScopeRegistry** — Map<sessionKey, scope> replaces the single-field race. Resolution order (registered > process-default) is correct and backward-compatible. null registration being distinct from no registration is the right subtlety.

2. **KeyedAsyncMutex / two-level lock** — in-process FIFO queue in front of O_EXCL is the correct shape for the problem. Module-level processLocks is correct (must be shared across callers in-process). The AsyncMutex chain-not-replace pattern is sound.

3. **LLM breaker sliding window** — replacing a counter-zeroed-on-success with a failure timestamp window correctly handles concurrent callers in flight.

Self-deadlock fix (hoisting learn() out of the lock in learn-async.ts) is correct. Decision to defer the async flip is right — the ADR measures the blast radius before deciding.

Stacked on #716 — merge that first.

plur9:

All green on Node 20/22/24.

plur9:

crtahlin (dismissed):
Independent review — **APPROVE**. Read every concurrency primitive and ran the tests locally.

## Verified correct

- **`AsyncMutex`** — chain-don't-replace FIFO (no polling); `depth` drives eviction. **`KeyedAsyncMutex`** — per-key, evicts on idle guarded by an identity check (`get(key) === mutex`), and the check-and-delete is atomic (no await between `run` settling and the finally), so two waiters can never be split across two mutex instances.
- **`SessionScopeRegistry`** — fixes the real scope-bleed race (A's engram landing in B's scope). Uses `map.has()` not truthiness, so `set(null, session)` (pin to no-scope) is distinct from never-registered, and single-session behaviour is byte-identical.
- **`withAsyncLock`** — in-process FIFO queue → O_EXCL file lock, async backoff (not busy-wait), stale-lock detection.
- **`learn-async` deadlock hoist** — the `idx === -1` branch returns `null` and calls `deps.learn()` OUTSIDE the lock, killing the self-deadlock.
- **All 10 new `withAsyncLock(paths.engrams)` bodies in `index.ts` have zero internal awaits**, so the sync/async shared-lock invariant holds as written.

**Tests:** 35/35 concurrency tests pass locally (session-scope, LLM-breaker, write-path contention, auto-discover, lock contention); full CI green.

## One reviewer's-eye item (non-blocking) — the mixed-lock invariant

Sync `withLock` (busy-waits on `Date.now`) and async `withAsyncLock` share the same `.lock` file; 10 sites are now async, **13 remain sync**. Safe only because no async lock body awaits while holding — which the code currently upholds. But it's unenforced and fragile: one `await` added inside any of those 10 bodies would deadlock a concurrent sync caller (event-loop-blocked busy-wait can't yield to let the async holder release). ADR-0004 documents it honestly. Worth a lint/assertion, or prioritising the full async flip (#728) to retire the sync sites. Confirmed correct today; flagging the structural fragility.

Note: stacked on #716 → #715, so needs a rebase onto current main before merge.

plur9:

plur9:
🤖 heartbeat — unblocked

#716 merged at 11:49 UTC today. CI re-ran against main and all checks passed (Node 20/22/24 + Python SDK green, 14:32 UTC). The rebase that prompted the review dismissal is done.

Requesting re-review from @crtahlin — this is the path that closes #738 (flaky `_activeSessionId` race).

PR #720, opened 2026-07-26
feat(core): Postgres primary store + size-based backend selection (ADR-0005)

### LINKED ISSUE #712: Phase 5: Postgres primary store + size-based backend selection in _resolveBackend()
Part of #707 — core↔enterprise convergence. Depends on Phase 1 (and composes
with Phases 3 and 4).

## Problem

Core has no scaling story of its own. `_resolveBackend()` only reads env/config and
**defaults to `sqlite`, with no index actually initialised** — so core loads the whole
corpus into memory and brute-forces cosine. At ~4,700 engrams that is ~350 MB per MCP
process, and there are six of them. This is the same root cause as the local RAM
problem, and the reason a Postgres tier is core's work rather than a per-deployment
cost.

A working Postgres-backed store with `searchVector`, text search and embedding upsert
already exists downstream, against an older interface. It is the natural reference
implementation to port from.

## Scope

- [ ] `PostgresAdapter` in core, implementing the Phase 1 primary-store interface.
- [ ] Size-based selection in `_resolveBackend()`: YAML → PGLite → Postgres, with an
      **explicit, documented threshold** (working instinct: ~5k engrams). Today the
      function only reads env/config.
- [ ] Make sure a backend is actually *initialised* — today's default claims `sqlite`
      and initialises no index at all.

## The open decision — exact vs approximate vector search

**This must be decided in this phase, explicitly, and published.**

| | Implementation |
|---|---|
| **Core** | `embeddingSearch()` (`core/src/embeddings.ts:325`) — brute-force cosine over **every** engram, JSON-cached. Exact. **100% recall by construction.** |
| **Approximate tier** | `CREATE INDEX … USING hnsw (embedding vector_cosine_ops)` — approximate. With pgvector defaults (`m=16`, `ef_construction=64`) and `ef_search` left at its default of 40, which is below most useful result limits. |

Core therefore measures a retrieval ceiling that an approximate index **cannot reach
by design**, and the gap widens as a corpus grows. Exact search is impossible
at scale, so the resolution is not "make enterprise exact" — it is:

- [ ] Make index strategy a **documented adapter property**, not an implementation
      detail.
- [ ] **Publish a recall target** for the approximate tier (e.g. "≥99% recall@10 vs
      exact, measured on plur-bench"), and gate the adapter on meeting it.
- [ ] Set `ef_search` ≥ the requested limit (composes with Phase 3, where the same
      constant is the second failure mode).
- [ ] Any residual exact→approximate delta is **documented, not silent.**

Any of these is defensible. Silence is not.

## Also note

- [ ] The `halfvec(N)` fp16 tier (#223) exists in core and **not** in enterprise — a
      real storage-format difference if it is ever enabled. Decide whether the
      Postgres adapter supports it, and fail loudly rather than silently coercing.

## Acceptance criteria

- [ ] `PostgresAdapter` passes the same adapter conformance suite as the YAML and
      PGLite adapters — one suite, three adapters, no adapter-specific carve-outs.
- [ ] A test asserts `_resolveBackend()` selects Postgres above the threshold and
      PGLite/YAML below it, and that the selected backend's index is actually
      initialised. **Must fail on `main`** (today no size-based switching exists).
- [ ] A recall test measures approximate-vs-exact on a fixed corpus and asserts the
      published target. The number goes in the PR body and in the adapter docs.
- [ ] `ef_search` is set explicitly and asserted ≥ requested limit.
- [ ] Memory: a test or measured note showing the Postgres tier does not load the full
      corpus into process memory (the whole point of the phase).

## Verification (CI is blocked — do this locally)

```
pnpm install && pnpm -r build && pnpm test    # root only
```
Postgres-backed tests need a live Postgres with pgvector; the enterprise dev stack
(Postgres 16 + pgvector) is the reference image. Paste exact pass/fail counts and
the measured recall number into the PR.

plur9:
Closes #712 (Phase 5). **Stacked on #716** (Phase 1) — review that first.

## What lands

`PostgresAdapter` implements **both** Phase 1 interfaces: `StorageAdapter` with `role: 'primary'` *and* `AsyncPrimaryStore`. It implements neither `syncFromYaml()` nor `reindex()` — which is exactly the case ADR-0003 introduced the role flag to express: its store and its query index are the same engine, so there is nothing to rebuild *from*.

Verified against real Postgres 16.10 + pgvector 0.8.2: `requiresIndexSync(adapter) === false`, `asDerivedIndex(adapter) === null`, both rebuild methods `undefined`.

`save()` is upsert-then-prune rather than truncate-then-insert, because `engram_embeddings` cascades on delete — a truncate would silently discard every embedding on each save. A test confirms a surviving engram keeps its embedding across a re-save while a removed one loses it.

## Backend selection

`_resolveBackend()` previously read env/config only, defaulted to `sqlite`, and initialised no index — which is why core loads the whole corpus into memory and brute-forces cosine (~350 MB per process at ~4,700 engrams).

Now: yaml → pglite → postgres, from `PrimaryStore.estimateCount()`.

| Threshold | Value |
|---|---|
| `PGLITE_MIN_ENGRAMS` | 5,000 |
| `POSTGRES_MIN_ENGRAMS` | 50,000 |

`PLUR_BACKEND` beats `config.backend` beats the estimate. The estimate is a `stat()`, and the 2,400 B/engram constant is **measured** (9,595,797 B / 4,009 engrams on a real store), not guessed.

Small stores land in exactly the pre-existing branch, so default behaviour is unchanged — which is why 2,878 pre-existing tests pass unmodified.

## The vector-index decision (ADR-0005)

The plan left this open; it is now decided and written down.

`vectorIndex` is a **required** field on `StorageAdapter`, not optional — a caller can always ask what it is getting. `efSearchFor(limit, configured)` floors `ef_search` at the requested limit and doubles it for the post-filter.

**The failure is demonstrated, not asserted.** The Postgres suite contains a paired test against a real HNSW index:

```
SET LOCAL enable_seqscan = off
SET LOCAL hnsw.ef_search = 40      -- pgvector's default
... ORDER BY embedding <=> $1 LIMIT 60
→ fewer than 60 rows
```

and its partner showing the guard returning exactly 60. `enable_seqscan = off` is needed so the planner takes the index path at all — otherwise a small table is seq-scanned and the bound never applies.

Recall target 0.95 is stated in ADR-0005 **as a target, not a measurement**, with the falsifying protocol written out (k = 10/20/50, exact scan as ground truth). ADR-0005 also records the `halfvec(N)` fp16 tier as a real format difference if it is ever enabled on one tier and not another.

## Note on a schema change that looks riskier than it is

`config.backend` widened to `yaml|sqlite|pglite|postgres`, and `.default('sqlite')` became `.optional()`. That is a no-op at runtime, verified rather than assumed: `.partial()` neutralises Zod defaults in core's Zod version, so `PlurConfigSchema.parse({}).backend` was already `undefined`.

## Verification

| Check | Result |
|---|---|
| `pnpm test` with `PLUR_TEST_POSTGRES_URL` | **2949 passed, 0 failed**, 44 skipped |
| baseline on #716 | 2878 — delta is exactly the 71 new tests |
| `pnpm test` without the DSN | 2917 passed, 76 skipped (44 + 32 Postgres) |
| wrong password in the DSN | **fails loudly** in `beforeAll`, does not silently skip |
| `pnpm build` / `tsc --noEmit` / `gen:schemas` | clean, no `spec/` drift |
| stability | 4 identical clean runs |

Fails-before / passes-after was checked against the real base, not asserted: a detached worktree at the Phase 1 commit with only the new test files copied in fails all four (`Cannot find module '../src/backend-selection.js'`). For the behavioural change specifically, a throwaway probe asserting the *old* behaviour — a ~5,000-engram store creates neither `store.pglite` nor `engrams.db`, and `backendSelection` is undefined — **passes on the base and fails on this branch**, the exact inversion.

CI gains a `pgvector/pgvector:pg16` service and `PLUR_TEST_POSTGRES_URL`, so the Postgres suite runs there rather than skipping.

**One unexplained result, reported rather than dismissed:** the first post-rebase run showed 1 failure out of 2,949 while a `pnpm build` and `tsc --noEmit` were competing for CPU. Four subsequent runs are clean, and I re-ran it independently at low load (2949/0). The failing test was not captured, so it is recorded as unexplained rather than labelled a flake from evidence that does not exist.


plur9 (commented):
**CTO Review — ARCHITECTURAL (pending deeper pass before merge)**

This is the ADR-0003 + ADR-0005 implementation — splitting `PrimaryStore` from `StorageAdapter` and landing the Postgres adapter. CI is green including the new ephemeral Postgres service added to the test matrix.

**What's solid:**
- The interface split reasoning in ADR-0003 is sound: `StorageAdapter` had no write methods, so the 'capability flag' approach was a false choice — a new `PrimaryStore` interface is the right call
- `requiresIndexSync()` helper correctly encodes the only question callers need to branch on
- Keeping `PrimaryStore` sync for now (Phase 2 converts to async separately) is the right call — avoids interleaving risk across the 40+ call sites
- CI adds real Postgres service rather than skipping the test suite — correct principle (skip = permanently unproven)

**Questions before merge:**
1. Does `storage-postgres.ts` implement both `PrimaryStore` and `StorageAdapter` with `role: 'primary'`? ADR says Phase 5 is where this happens, but want to confirm the current PR scope is just the interface + YAML-primary wiring
2. Is `PLUR_TEST_POSTGRES_URL` guarding the postgres tests with a skip when unset? If so, what does a developer without a local Postgres see when they run `pnpm test`?
3. The lockfile adds `pg@8.22.0` as optional — confirm this is a dev/test-only dep and doesn't bloat the published package for users who don't use Postgres

Ready for merge once the scope confirmation is clear — the architecture is correct and CI is green. This is not a blocking review.

plur9 (commented):
## Review: feat(core) Postgres primary store + size-based backend selection (ADR-0005)

### Summary

This is a large, well-structured PR. The architecture is sound, the ADRs are clear and honest about what is and isn't measured, and the test coverage is thorough. A few items worth attention before merging.

### Correctness & Safety

**`save()` upsert-then-prune strategy** — the PR body explains this correctly: truncate-then-insert would cascade away all embeddings on every save. The upsert approach is correct and the test confirming embedding survival across re-saves is exactly the right regression guard. No issue here.

**Connection pool lifecycle** — `close()` calls `this.pool.end()`, which is correct. However `init()` creates a new pool each call but `initPromise` gates concurrent calls — if `init()` is called after `close()`, `initPromise` is still set and a second `getPool()` would reuse the destroyed pool. Consider resetting `initPromise = null` in `close()`, or documenting that the adapter is single-use.

**`save()` chunking** — the 500-row `SAVE_CHUNK_SIZE` INSERT batches run sequentially (one `await` per chunk) but are not wrapped in a transaction. A failure midway leaves the table in a partially-replaced state: some chunks committed, others not. This isn't catastrophic because `save()` is called from a full-replace path (not incremental) and the next `save()` call will correct it — but it would be good to either document this or wrap in a transaction.

**`redactDsn` edge cases** — the key=value libpq form regex `password=([^\s]+)` would NOT catch `PASSWORD=...` (case-sensitive regex with `/gi` flag — actually fine, that flag is present). Looks correct.

**`assertIdentifier` for schema** — the guard `SAFE_IDENTIFIER = /^[A-Za-z_][A-Za-z0-9_$]*$/` correctly rejects SQL injection in the schema name. The `$` in the character class is intentional for PostgreSQL identifiers. Good.

**Unexplained single test failure** — the PR notes one run showed 1 failure out of 2,949 that was not captured. This is disclosed honestly, but it is worth a follow-up once the PR lands to identify whether it's a real flake in the Postgres suite (test isolation, schema collisions between test runs?).

### Backend Selection Logic

The `resolveBackendTier()` function is pure, total, and well-tested. The threshold documentation (5k for PGLite, 50k for Postgres) is appropriately honest that these are round numbers, not measured inflection points.

One concern: when `selection.tier === 'postgres'` the code falls back to `'pglite'` for the index tier. But the PGLite adapter path also requires the store to be large enough to justify WASM boot cost. If someone has an explicit `PLUR_BACKEND=postgres` override, they may be surprised to land on PGLite silently. The warning message is good — but it logs `${this._primaryStore.kind}` not what it's actually doing index-wise. Minor.

### ADR Status

Both ADR-0003 and ADR-0005 are marked `Proposed`, not `Accepted`. The ADR README lists them as such. Is there an explicit acceptance process that needs to complete before this is considered final, or is the PR itself the acceptance decision?

### The `.embeddings-cache.json` Committed Artifact

This is the most notable concern. The diff for this PR includes `packages/core/.embeddings-cache.json` — a 1-entry binary-ish JSON blob with 384-dimension embeddings for `ENG-2026-0406-001` through `ENG-2026-0406-005`. This file should almost certainly be gitignored. It contains generated state (cached model outputs) that will drift as the embedder or corpus changes, does not belong in version control, and will cause merge conflicts on update. Please add it to `.gitignore`.

### Minor / Non-blocking

- The `POSTGRES_MIN_ENGRAMS = 50_000` threshold is described as "chosen as an order of magnitude above the PGLite threshold rather than measured" — this is honest and fine, but worth a TODO to revisit once there's real multi-process load data.
- `efSearchFor(limit, configured, headroom)` correctly floors at `limit`, but the formula `Math.ceil(Math.max(1, limit) * Math.max(1, headroom))` when `headroom = 1` would return `limit`, not `limit * 2`. The default headroom is 2 (correct), but a caller passing `headroom = 1` explicitly would defeat the multiplier without an error. Acceptable but worth a note.

### Verdict

The architecture is solid and the implementation matches it. The main items before merge:
1. **Blocker**: Gitignore `packages/core/.embeddings-cache.json`
2. **Recommended**: Document or fix the pool lifecycle after `close()`
3. **Recommended**: Wrap `save()` chunks in a transaction, or document the partial-update window
4. **Clarification**: ADR acceptance status

Everything else is polish. The test coverage, the honest ADR documentation of what is and isn't measured, and the explicit failure-demonstration tests are all exemplary.

plur9 (commented):
## Review — #720: feat(core): Postgres primary store + size-based backend selection (ADR-0005)

CI: all green (test/20, test/22, test/24, Python SDK). 2,949 passed, 0 failed with PLUR_TEST_POSTGRES_URL set.

### Deep-dive findings

**Backend selection logic — sound.** The `_resolveBackend()` waterfall (yaml → pglite → postgres) with the `stat()`-based estimate is sensible. The 2,400 B/engram constant is measured from a real store, not guessed — good. Thresholds (5k for pglite, 50k for postgres) are documented. The `PLUR_BACKEND` override beats config which beats the estimate — clear precedence.

**upsert-then-prune over truncate-then-insert — correct call.** The cascade-delete risk on `engram_embeddings` is real and the PR body articulates it clearly. The surviving-embedding test pinning this is exactly the right regression anchor.

**ADR-0005 / ef_search decision — well handled.** `efSearchFor(limit, configured)` floors at the requested limit and doubles for post-filter. The HNSW candidate list logic is sound, and the paired failure/pass test demonstrating the old behaviour vs new is the right verification pattern.

**`vectorIndex` as required field — agreed.** Callers need to know what they're getting; making it optional would have been a footgun.

**Schema change (.optional() on config.backend) — the .partial() verification is reassuring** but worth a note: this should be confirmed to hold across all Zod versions in the matrix (20/22/24). The PR says 'verified rather than assumed' which is good, but the mechanism (Zod .partial() neutralising defaults) is version-sensitive. Low risk given CI covers it.

**One unexplained failure — reported correctly.** Recording it as 'unexplained, 4 subsequent runs clean' rather than labelling it a flake without evidence is the right posture. Acceptable.

**Stacked on #716 — confirm merge order.** This PR should not land before #716. Already noted in PR body.

### Risk assessment

Risk: **medium** (new Postgres backend, new CI service, new dependency on pgvector). Mitigated by: comprehensive test suite (71 new tests), guard on pre-existing PLUR_BACKEND env, unchanged default branch (small stores still get sqlite).

### Verdict

Implementation is solid. The test suite is thorough and the PR body is unusually honest about what was verified vs assumed. Ready to merge after #716 lands.

plur9:

All green on Node 20/22/24.

plur9:

crtahlin (dismissed):
Independent review — **APPROVE** (confidence: medium — the Postgres suite needs a live DB I can't stand up locally, so this is diff + reasoning; core paths read correct).

## Verified correct

- **`save()` upsert-then-prune is right, and the reasoning is a genuine catch.** `engram_embeddings` cascades on delete, so truncate-then-insert would nuke every embedding each save. Chunked `INSERT … jsonb_to_recordset($1) ON CONFLICT (id) DO UPDATE`, then `DELETE … WHERE id NOT IN (jsonb_array_elements_text($1))`, one transaction — survivors keep embeddings, only genuinely-removed engrams cascade. Atomic.
- **Injection defense is thorough.** Every interpolated identifier goes through `assertPlainIdentifier`, every integer through `assertPositiveInt`; all data values are bound params (`$1` jsonb, `$2` limit). `SET LOCAL hnsw.ef_search = ${...}` interpolates a validated positive int (SET LOCAL can't bind a param for the value — this is the correct approach).
- **`role: 'primary'` implemented as ADR-0003 intended** — no `syncFromYaml`/`reindex`; `requiresIndexSync` false, `asDerivedIndex` null.
- **The ef_search failure is demonstrated, not asserted** — paired HNSW test with `enable_seqscan = off` showing < limit rows at ef_search=40 and exactly limit with the guard. ADR-0005 is honest that recall 0.95 is a target with a falsifying protocol, not a measurement.
- Backend selection (yaml→pglite→postgres by estimateCount, `PLUR_BACKEND` override, measured 2400 B/engram) leaves small stores in the pre-existing branch — hence 2878 tests unchanged.

## WARNING (non-blocking, tracked in #736)

`PostgresAdapter.searchBM25`/`searchVector`/`loadFiltered` do **not** implement #715's permitted-scope pushdown (understandable — branched pre-#715). Once #715 and #720 both land and the Postgres tier is used with a scope model, it becomes the read path that accepts an authorization filter and ignores it — #715's exact stated failure mode, on a new tier. Must be added during the stack rebase / before the tier ships with scopes. Filed as #736. Not live today (the tier isn't wired into `Plur` until the #728 async flip).

Stacked on #716; needs a rebase onto current main (and, per #736, scope-pushdown parity) before merge.

crtahlin:
Follow-up to my approval: I re-ran the Postgres suite against a **live Postgres 16 + pgvector 0.8.5** (Docker), not just diff-reading. **32/32 pass**, including the load-bearing cases:

- `save()` preserves surviving engrams' embeddings and cascades away the removed ones (the upsert-then-prune correctness claim)
- the HNSW `ef_search` pair — DEMONSTRATES the truncation at default 40, FIXES it, returns the full limit, and recall is at/above the declared target *against a real index*
- PGLite ⇄ Postgres filter parity across all the `{scope|status|domain}` combinations
- embedding write guards (wrong-dimension, orphan)

So my earlier "medium confidence (couldn't run)" is now **high** for the adapter. Credential handling is correct — `location` returns `redactDsn(connectionString)` (`postgres://user:***@host/db`), password redacted.

Minor test-quality NOTE (non-blocking): the `location` test asserts `expect(location).not.toContain(password)`. That false-fails if the password is a substring of the host/db (I hit it with password `plur` + db `plurtest`). Consider asserting against the userinfo section specifically, or requiring the fixture password to be word-unique, so a coincidental substring can't mask or fake a redaction regression.

The one real finding stands and is tracked in #736: the adapter doesn't yet implement #715's permitted-scope pushdown.

plur9:

Issue #711, opened 2026-07-26
Phase 4: BM25 pushdown — narrow with pg_trgm, score in core with global corpus statistics


plur9:
**Rewritten against measurement.** The original framing assumed the matching rule was a fixed constraint to design around. It is not — it is the variable that decides whether pushdown is possible at all, and it has now been measured.

## What BM25 matching actually does today

`ftsScore` and `computeIdf` both match a query token `qt` against a document token `t` with:

```ts
t.includes(qt) || qt.includes(t)
```

Two directions, and they are not equally defensible.

- **Forward** (`t.includes(qt)`) — the document token contains the query. This is what finds `transferWithAuthorization` from `auth`.
- **Reverse** (`qt.includes(t)`) — the query contains the document token, anywhere. This is what makes `deploying` match an engram about **yin**, and `postgres` match one about **res**.

#724 replaces the reverse arm with `qt.startsWith(t)`, keeping morphological prefixes (`deploy` → `deploying`) and dropping the junk.

## Ablation

Run against a real 3,930-engram store, 36 queries, top-5, measuring Jaccard overlap of result sets against today's behaviour:

| Rule | Top-5 overlap | Queries returning nothing | "It forgot" vs today |
|---|---|---|---|
| substring (today) | — | 0/36 | — |
| **prefix bidirectional (#724)** | **0.690** | **0/36** | **0/36** |
| forward only, Postgres `:*` | 0.320 | 0/36 | 0/36 |
| exact | 0.173 | 3/36 (8%) | 3/36 (8%) |

Infix-only matches — the ones that survive `t.includes(qt)` but *cannot* be expressed as a tsvector prefix query:

```
"auth"       → transferwithauthorization      (legitimate)
"authoriz"   → transferwithauthorization      (legitimate)
"deploy"     → redeployed                     (legitimate)
"deploying"  → yin                            (junk)
"postgres"   → res                            (junk)
"postgresql" → sql                            (junk)
```

## What this changes about the design

The original plan said "narrow with pg_trgm". The ablation says that is not a preference, it is forced:

- **Plain tsvector with `:*` prefix queries cannot serve this.** It scores 0.320 overlap — it would change roughly two-thirds of every top-5. The reason is visible in the table above: `auth` → `transferWithAuthorization` is an *infix* match, and a tsvector prefix query cannot express one. Half the infix list is junk, but the other half is exactly the compound-identifier matching that makes recall useful on a code-adjacent corpus.
[...]

PR #724, opened 2026-07-27
fix(core): BM25 reverse-substring junk matches — qt.includes(t) → qt.startsWith(t) (#721)

### LINKED ISSUE #721: BM25 reverse substring matching amplifies junk: rare accidental matches outscore the correct stem
## The defect

`ftsScore` counts a term frequency hit when **either** token contains the other:

```ts
// packages/core/src/fts.ts
for (const t of allTerms) {
  if (t.includes(qt) || qt.includes(t)) tf++
}
```

The second clause — `qt.includes(t)`, a *document* token being a substring of the *query* token — is what makes `deploying` match `deploy`. There is no stemmer in the pipeline, so this is the only morphology core has, and removing it outright would be a regression (measured: plur-ai/plur-bench#49).

But it also lets **any document token of ≥3 characters** match any longer query word that happens to contain it. And because IDF weights by rarity, the more nonsensical the accidental match, the more it is amplified.

Measured on a real 3,926-engram store:

| query | matched token | df | **IDF weight** | |
|---|---|---|---|---|
| `deploying` | `yin` | 3 | **6.89** | junk — ranked highest |
| | `dep` | 13 | 5.64 | junk |
| | `deploy` | 112 | **3.55** | the correct stem — ranked *lowest* |
| `postgres` | `res` | 2 | **7.18** | junk |
| | `pos` | 2 | **7.18** | junk |
| | `post` | 164 | **3.17** | legitimate — lowest |

`yin` carries **1.94× the weight of `deploy`** for the query "deploying". The ranking signal is inverted precisely where it should be strongest.

This is not the infix case (`thor` inside `authorization`), which is a different and smaller class. It is the reverse-containment case, and it fires on every multi-syllable query.

## Why it is not more visible

Three things mask it:

- The junk token must actually occur in a document for the false positive to land, and rare tokens are rare.
- RRF fusion with vector search dilutes a single bad BM25 ranking.
- The optional reranker can reorder the top of the list.

None of those are reasons to keep it — they are reasons it went unnoticed.

## Suggested fix

Require a minimum useful overlap on the reverse direction rather than accepting any substring. Options, roughly in increasing order of effort:

1. **Length floor on the reverse clause** — require `t.length >= 4`, or `t.length / qt.length >= 0.6`, so `deploy`/`deploying` still matches but `yin`/`deploying` does not.
2. **Require a prefix, not a substring** — `qt.startsWith(t)` instead of `qt.includes(t)`. `deploy` is a prefix of `deploying`; `yin` and `res` are not. This is also strictly closer to what a stemmer does, and to what Postgres can express.
3. **Add a real stemmer** and drop the reverse clause entirely. Largest change, best end state, and it is the direction Phase 4 pushes anyway.

Option 2 looks like the best ratio — one operator change, and it removes the whole junk class while keeping every legitimate morphological match in the table above.

## Acceptance

- A test asserting `yin` does not match `deploying` and `res` does not match `postgres`, and that `deploy` still does. **It must fail on `main`.**
- The ablation in plur-ai/plur-bench#49 re-run before and after, with the top-5 overlap reported. A fix that changes ranking wholesale is a different proposal and needs the recall numbers to go with it.

## Related

- plur-ai/plur-bench#49 — the ablation that surfaced this
- #711 — Phase 4 BM25 pushdown, which has to decide the matching semantics anyway

plur9:
## Problem

`ftsScore` and `computeIdf` both used `qt.includes(t)` as the reverse-containment clause — matching any document token that is a substring of the query token. This inverts IDF-based ranking for multi-syllable queries:

| query | matched token | IDF | verdict |
|---|---|---|---|
| `deploying` | `yin` | 6.89 | junk — scored **highest** |
| `deploying` | `deploy` | 3.55 | correct stem — scored **lowest** |

Rare accidental tokens (`yin`, `res`, `pos`) were up to 1.94× heavier than the intended stem.

## Fix

Replace `qt.includes(t)` with `qt.startsWith(t)` in both sites. A document token scores a TF hit only when it is a **prefix** of the query term — which is the morphological case (`deploy` → `deploying`), not an arbitrary infix coincidence.

The `t.includes(qt)` direction (document token *contains* query token — e.g. `deployer` matching `deploy`) is unchanged and correct.

## Tests

Three regression tests added to `fts.test.ts`:

- `yin` does not match query `deploying` (non-prefix junk)
- `res` does not match query `postgres` (non-prefix junk)
- `deploy` still matches query `deploying` (legitimate morphological prefix)

All 13 FTS tests pass.

## Related

Closes #721
See also #711 (Phase 4 BM25 pushdown, which will need to express the same semantics)

### DIFF
--- a/packages/core/src/fts.ts
+++ b/packages/core/src/fts.ts
@@ -62,7 +62,7 @@ export function computeIdf(engrams: Engram[], queryTokens: string[]): Map<string
   for (const qt of queryTokens) {
     let df = 0
     for (const termSet of engramTermSets) {
-      if (termSet.has(qt) || Array.from(termSet).some(t => t.includes(qt) || qt.includes(t))) {
+      if (termSet.has(qt) || Array.from(termSet).some(t => t.includes(qt) || qt.startsWith(t))) {
         df++
       }
     }
@@ -103,7 +103,7 @@ export function ftsScore(engram: Engram, queryTokens: string[], idfWeights?: Map
     // Count term frequency (including substring matches)
     let tf = 0
     for (const t of allTerms) {
-      if (t.includes(qt) || qt.includes(t)) tf++
+      if (t.includes(qt) || qt.startsWith(t)) tf++
     }
     if (tf === 0) continue
 
--- a/packages/core/src/intent/rewrite.ts
+++ b/packages/core/src/intent/rewrite.ts
@@ -11,7 +11,7 @@
  * embedding/cross-encoder models are trained on questions; BM25 is not.
  *
  * Why scaffolding hurts BM25 here: `ftsScore` uses substring matching
- * (`t.includes(qt) || qt.includes(t)`), so interrogative tokens are not
+ * (`t.includes(qt) || qt.startsWith(t)`), so interrogative tokens are not
  * inert — "what" matches "whatsapp", "how" matches "showed" — and every
  * scaffold token that substring-matches anything consumes IDF budget and
  * pulls in distractors.
--- a/packages/core/test/fts.test.ts
+++ b/packages/core/test/fts.test.ts
@@ -125,6 +125,38 @@ describe('BM25 term frequency saturation', () => {
   })
 })
 
+describe('BM25 reverse substring junk matching (#721)', () => {
+  // qt.includes(t) allowed any document token that is a non-prefix substring of
+  // the query to score a TF hit. e.g. "deploying".includes("yin") = true, so
+  // an engram about "yin yang" matched the query "deploying" and could outscore
+  // the correct "deploy" stem (which had higher df → lower IDF).
+  // Fix: require qt.startsWith(t) so only true morphological prefixes match.
+
+  it('"yin" is a non-prefix substring of "deploying" and must not produce a result', () => {
+    // With the bug: 'deploying'.includes('yin') = true → tf=1 → engram surfaces
+    // With the fix: 'deploying'.startsWith('yin') = false → tf=0 → no result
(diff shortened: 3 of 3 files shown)


plur9:
🤖 heartbeat — PR #724 confirmed ready: all CI green (issue-guard + test × 3 nodes). Branch protection requires 1 review before merge. No action needed beyond approval.

plur9:

All green on Node 20/22/24.

plur9:

crtahlin (approved):
Independent review — **APPROVE** (high confidence: read the diff, ran the tests).

Clean, well-diagnosed fix. `qt.includes(t)` let any document token be a non-prefix infix of a query word (`"deploying".includes("yin")`), and because IDF weights by rarity the junk match was *amplified* — `yin` outranking `deploy` for "deploying". `qt.startsWith(t)` keeps exactly the morphological prefix case (`deploy` → `deploying`) and drops the infix coincidences.

- Applied consistently at all three sites — `fts.ts` `ftsScore`, the `computeIdf` path, and `intent/rewrite.ts` — with the doc comment updated to match.
- The other direction, `t.includes(qt)` (document token *contains* the query term, e.g. `deployer` matching `deploy`), is correctly left unchanged.
- Right call not to remove the clause outright — plur-ai/plur-bench#49 measured that as a regression; `startsWith` is the surgical middle ground.
- Ran `fts.test.ts` locally: 13/13 pass, including the three new regression cases (`yin`↛`deploying`, `res`↛`postgres`, `deploy`→`deploying`).

Reminder from your own ordering note: merge this **before** #732 (which cherry-picks it) so git reconciles cleanly.

plur9:

PR #726, opened 2026-07-27
fix(recall): truncated flag now accurate when budget.max_results drops results (#725)

### LINKED ISSUE #725: recall: `truncated` under-reports when budget.max_results drops results
## Summary

`plur_recall` (and the `plur_recall_hybrid` alias, which forwards to it) returns `truncated: false` in the common case where `budget.max_results` **did** cause results to be dropped. A budget-aware sub-agent cannot trust the flag to tell it whether it saw the full result set.

Found while addressing review feedback on #702. Pre-existing on `main`; unchanged by that PR — the forwarder inherits the behaviour faithfully.

## Mechanism

In `packages/mcp/src/tools.ts` (`recallHandler`):

```ts
const effectiveLimit = budget?.max_results ?? (args.limit as number | undefined) ?? 20
const meta = await plur.recallHybridWithMeta(query, { limit: effectiveLimit })
...
if (budget?.max_results && results.length > budget.max_results) {   // ← rarely true
  boundedResults = results.slice(0, budget.max_results)
  truncated = true
}
```

`max_results` is passed *down* as the query limit, so the underlying search already caps the set before the MCP layer compares against that same number. `results.length > budget.max_results` is therefore false in almost every case — the truncation happened upstream, silently, and the flag never fires.

## When it does fire

The branch is reachable only on the YAML/JSON-cache path for **aggregation queries**. `packages/core/src/hybrid-search.ts`:

```ts
const exhaustive = isAggregationQuery(query)     // "how many", "total", ...
const effectiveLimit = exhaustive ? Math.max(limit, 50) : limit
```

For an aggregation query, up to 50 results come back regardless of the requested limit, so the MCP-layer comparison finally holds and `truncated` is set correctly.

On the PGLite path this never happens — `_pgliteHybridRecall` in `packages/core/src/index.ts` hard-caps with `.slice(0, limit)` before returning.

## Net effect

`truncated` currently depends on query phrasing and storage backend rather than on whether results were actually dropped:

| Query | Backend | `max_results: 5`, 100 matches | `truncated` | Correct? |
|---|---|---|---|---|
| ordinary | YAML | 5 returned, 95 dropped upstream | `false` | ✗ |
| aggregation | YAML | 5 returned, up to 45 dropped at MCP layer | `true` | ✓ |
| any | PGLite | 5 returned, rest dropped upstream | `false` | ✗ |

`max_tokens` is unaffected — its loop sets the flag correctly.

## Suggested fix

Ask the search layer for one more than the budget and compare against that, so "were there more?" is answered by data rather than inferred:

```ts
const cap = budget?.max_results ?? (args.limit as number | undefined) ?? 20
const meta = await plur.recallHybridWithMeta(query, { limit: cap + 1 })
const truncatedByCount = meta.engrams.length > cap
const boundedResults = meta.engrams.slice(0, cap)
```

Worth confirming first whether any consumer relies on `truncated: false` today — flipping it to accurate will make it `true` for a lot of calls that currently report clean.

## Acceptance

- [ ] `truncated` is `true` whenever `budget.max_results` drops at least one result, independent of query phrasing and backend
- [ ] Test covering an ordinary (non-aggregation) query with more matches than `max_results`
- [ ] Test covering the PGLite path
- [ ] Alias `plur_recall_hybrid` inherits the fix (the anti-drift test added in #702 should hold it)

plur9:
## Summary

Fixes #725. `truncated` was always `false` when `budget.max_results` capped results, because the same value was passed as the search `limit` — so `results.length > max_results` was never true.

**Fix:** request `cap + 1` results from the search layer when `budget.max_results` is set, then compare `results.length > cap`. The extra fetch costs one row and makes truncation detection data-driven rather than inferred.

The fix lives in `recallHandler` (the shared function at the top of `tools.ts`), so both `plur_recall` and `plur_recall_hybrid` (which forwards to it) inherit it automatically. The existing anti-drift test added in #702 holds the forwarder honest.

## Changes

- `packages/mcp/src/tools.ts`: fetch `cap + 1` when `budget.max_results` is set; derive `truncatedByCount` from the actual result count
- `packages/mcp/test/tools.test.ts`: four new tests covering truncated=true (capped), truncated=false (fits), truncated=false (no budget), and alias inheritance

## Acceptance checklist

- [x] `truncated: true` when `budget.max_results` drops at least one result (ordinary query, YAML backend)
- [x] `truncated: false` when results fit within `max_results`
- [x] `truncated: false` when no budget is given (no behavior change)
- [x] `plur_recall_hybrid` alias inherits the fix via forwarder (tested explicitly)
- [x] All 65 existing tests pass

### DIFF
--- a/packages/mcp/src/tools.ts
+++ b/packages/mcp/src/tools.ts
@@ -82,11 +82,16 @@ const recallHandler: ToolDefinition['handler'] = async (args, plur) => {
   }
   // mode === 'hybrid' (default)
   const budget = args.budget as { max_tokens?: number; max_results?: number; ttl_seconds?: number } | undefined
-  const effectiveLimit = budget?.max_results ?? (args.limit as number | undefined) ?? 20
+  const cap = budget?.max_results ?? (args.limit as number | undefined) ?? 20
+  // When a max_results budget is set, fetch one extra so we can detect
+  // whether the store had more results than the cap without over-fetching.
+  // Without this, the search layer already caps at `cap` before we can
+  // compare, so `results.length > cap` is always false (#725).
+  const fetchLimit = budget?.max_results != null ? cap + 1 : cap
   const meta = await plur.recallHybridWithMeta(args.query as string, {
     scope: args.scope as string | undefined,
     domain: args.domain as string | undefined,
-    limit: effectiveLimit,
+    limit: fetchLimit,
   })
   // Opt-in, content-free engagement counter (default-off; no query text).
   recordTelemetry('recall')
@@ -95,13 +100,9 @@ const recallHandler: ToolDefinition['handler'] = async (args, plur) => {
   // there for ALL consumers (MCP, claw, CLI, direct API), so we do NOT
   // re-emit here and double-count. It is opt-in/default-off and ships only
   // a query fingerprint hash + scope/domain + timestamp, never raw text.
-  const results = meta.engrams
-  let truncated = false
-  let boundedResults = results
-  if (budget?.max_results && results.length > budget.max_results) {
-    boundedResults = results.slice(0, budget.max_results)
-    truncated = true
-  }
+  const truncatedByCount = budget?.max_results != null && meta.engrams.length > cap
+  let truncated = truncatedByCount
+  let boundedResults = truncatedByCount ? meta.engrams.slice(0, cap) : meta.engrams
   if (budget?.max_tokens) {
     let tokenCount = 0
     const withinBudget = []
--- a/packages/mcp/test/tools.test.ts
+++ b/packages/mcp/test/tools.test.ts
@@ -515,6 +515,50 @@ describe('MCP tools', () => {
     expect(squeezed.deprecated).toMatch(/deprecated since 0\.16/)
   })
 
+  // ── #725: truncated flag accuracy when budget.max_results drops results ───
+
+  describe('plur_recall budget.max_results truncation (#725)', () => {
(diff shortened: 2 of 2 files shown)


crtahlin (approved):
APPROVE — correct fix for the truncated-flag under-reporting (#725).

`recallHandler` passed `max_results` as the search `limit`, so the store capped the set upstream and `results.length > max_results` was almost never true. The fix fetches `cap + 1` when a `max_results` budget is set and derives `truncatedByCount` from the actual count — truncation is now observed rather than inferred. It lives in the shared handler, so `plur_recall` and the `plur_recall_hybrid` forwarder both inherit it, and it correctly covers both the aggregation-query path (up to 50 results) and the ordinary path. One extra row under a budget is a negligible cost.

Verified locally (isolated worktree, core→mcp build): all 65 tests in `tools.test.ts` pass. The four new tests are a genuine regression guard, not a vacuous pass — reverting only `tools.ts` to main`s unfixed source makes the two `truncated: true` cases fail (`expected false to be true`, the exact #725 symptom); restoring the fix makes all 65 pass. The alias-forwarder test keeps `plur_recall_hybrid` honest.

Two non-blocking notes: (1) `budget.max_results: 0` yields `cap:0` / `truncated:true` / empty results — defensible but untested; (2) the exact-fit boundary (`max_results` equal to the result count) isn`t covered. Neither blocks merge.

Clean rename (`effectiveLimit`→`cap`, no danglers). LGTM.

PR #727, opened 2026-07-27
fix(fts): use prefix matching to stop reverse-substring junk from outscoring correct stems (#721)

### LINKED ISSUE #721: BM25 reverse substring matching amplifies junk: rare accidental matches outscore the correct stem

plur9:
## Summary

- `qt.includes(t)` in `ftsScore` and `computeIdf` was matching any short document token that happened to appear anywhere inside a longer query token — including nonsense cases like `yin` matching `deploying` or `res` matching `postgres`
- Because IDF heavily weights rare tokens, these accidental matches often *outranked* the correct stem (`deploy` scored below `yin` for query `deploying` on a real 3,926-engram store)
- Fix: replace `qt.includes(t)` → `qt.startsWith(t)` — legitimate morphological stems are always prefixes (`deploy`→`deploying`); accidental infix matches never are

## Test plan

- [ ] All three new regression tests in `fts.test.ts` pass — `yin` scores 0 for `deploying`, `res` scores 0 for `postgres`, `deploy` still scores > 0 for `deploying`
- [ ] Existing 10 FTS tests still pass (no ranking regressions in the test suite)
- [ ] Run the plur-bench ablation from plur-ai/plur-bench#49 before merging to verify recall numbers hold

Closes #721

### DIFF
--- a/packages/core/src/fts.ts
+++ b/packages/core/src/fts.ts
@@ -62,7 +62,7 @@ export function computeIdf(engrams: Engram[], queryTokens: string[]): Map<string
   for (const qt of queryTokens) {
     let df = 0
     for (const termSet of engramTermSets) {
-      if (termSet.has(qt) || Array.from(termSet).some(t => t.includes(qt) || qt.includes(t))) {
+      if (termSet.has(qt) || Array.from(termSet).some(t => t.includes(qt) || qt.startsWith(t))) {
         df++
       }
     }
@@ -103,7 +103,7 @@ export function ftsScore(engram: Engram, queryTokens: string[], idfWeights?: Map
     // Count term frequency (including substring matches)
     let tf = 0
     for (const t of allTerms) {
-      if (t.includes(qt) || qt.includes(t)) tf++
+      if (t.includes(qt) || qt.startsWith(t)) tf++
     }
     if (tf === 0) continue
 
--- a/packages/core/test/fts.test.ts
+++ b/packages/core/test/fts.test.ts
@@ -125,6 +125,41 @@ describe('BM25 term frequency saturation', () => {
   })
 })
 
+describe('BM25 reverse substring junk (#721)', () => {
+  it('does not match yin to deploying (reverse non-prefix substring)', () => {
+    const junkEngram = makeEngram({ id: 'ENG-2026-0330-001', statement: 'the yin and yang principle' })
+    const deployEngram = makeEngram({ id: 'ENG-2026-0330-002', statement: 'how to deploy services' })
+    const queryTokens = ftsTokenize('deploying')
+    const idf = computeIdf([junkEngram, deployEngram], queryTokens)
+
+    const junkScore = ftsScore(junkEngram, queryTokens, idf)
+    const deployScore = ftsScore(deployEngram, queryTokens, idf)
+
+    expect(junkScore).toBe(0)
+    expect(deployScore).toBeGreaterThan(0)
+  })
+
+  it('does not match res to postgres (reverse non-prefix substring)', () => {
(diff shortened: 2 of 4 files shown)


plur9:
Closing as a duplicate — this branch combines two changes that are already open as focused PRs:

- the BM25 reverse-substring fix (`qt.includes(t)` → `qt.startsWith(t)`) is #724
- the `truncated` flag / over-fetch-by-one fix (#725) is #726

The source diffs are identical in both cases, so there is nothing here that the two of them do not already carry.

Keeping #724 over this one specifically for its tests: they assert through `searchEngrams`, the entry point a caller actually uses, so they fail if the regression reappears anywhere in the path — tokenize, IDF, or score. The versions here call `computeIdf` and `ftsScore` directly, which pins the two functions but would not catch the same bug reintroduced upstream of them.

No action needed on #724 or #726.

PR #728, opened 2026-07-27
feat(core): async write path — Plur can run on a Postgres primary store (convergence Phase 2b)

plur9:
## What

Convergence Phase 2b: `Plur`'s write path becomes async, so `Plur` itself can run on a Postgres primary store. Phase 1 introduced `PrimaryStore` as a deliberately synchronous seam and documented the ceiling that came with it — there is no synchronous Postgres client for Node, so a network-backed store could never satisfy it. This removes the ceiling.

## The acceptance test

A test count cannot demonstrate this, so `packages/core/test/postgres-primary-store.test.ts` does:

```ts
const adapter = new PostgresAdapter({ connectionString, schema, vectorIndex: 'exact' })
const plur = new Plur({ path: dir, store: adapter })
```

Five assertions against a real Postgres — not a mock, not PGLite: that the store reports `postgres` rather than falling back, a `learn` → `recall` round trip, persistence across a new `Plur` instance, a write visible to a reader that never saw the writer, and a loud failure on a bad DSN. Gated on `PLUR_TEST_POSTGRES_URL`; when that variable is set but the database is unreachable it **fails rather than skips**, because a silent skip on a configured environment is how a green suite ends up meaning nothing.

CI already provides a Postgres service container, so this runs there. Locally: 2991 passed / 81 skipped without the DSN, 3028 / 44 with it — the gate is real.

## Defects found and fixed along the way

The flip was mechanical; what it surfaced was not.

- **26 floating `_writeEngrams` calls.** Writes issued and never awaited. Under the old synchronous path this was invisible; the moment the write returns a promise, the process can exit or the next read can run before the bytes land.
- **`ingest()` was synchronous but its learns were not.** Same shape, one level up.
- **`{...promise}` spreads.** Spreading a Promise yields `{}` — silent, total field loss, and the assertions downstream still passed.
- **No drain points.** Work was started with no way to observe its completion. Added `ContextEngine.settle()`, `Plur.ready()`, and `settleVersionChecks()`. Three instances of one defect.
- **`await plur.getById(id)!`** parses as `await (getById(id)!)`. The assertion lands on the Promise, which was never nullable, so it is a no-op and the awaited value stays `Engram | null`. The `!` was load-bearing when `getById` was synchronous; the flip turned it into decoration without a word. 7 sites.
- **117 `async describe` callbacks**, introduced by my own codemod. vitest starts a suite as soon as its callback returns a promise, so `it`s registered after the first `await` may not be collected. Symptom: passes alone, fails in a full run — i.e. reads as flake.

## Test-suite typechecking

Nothing had ever read the types in `packages/*/test`: `tsup` emits declarations for `src` only, and esbuild strips types without checking them. That is roughly a third of the repo compiling against types no tool verifies — and it matters most exactly when the API changes shape, which is what this PR does. A test that drops an `await` on a newly-async method still transpiles, still runs, and still passes, because almost any assertion is true of a pending Promise.

`tsconfig.tests.json` + a CI gate. 93 errors surfaced. Most were fixture drift; these were not:

- **`learn()` does not validate `type` against `EngramSchema`.** Two suites were persisting engrams with `type: 'preference'`, which is not a member of the enum and never has been. `TYPE_TO_COGNITIVE` and `TYPE_TO_MEMORY_CLASS` both missed and fell back, so the derived metadata was correct by accident. Tests fixed here; the source gap is filed separately.
- **`secrets.test.ts` passed `new Plur({ stores: [...] })`**, a key the constructor does not read — stores load from `config.yaml`. So no shared store existed. The demotion half of the test was load-bearing (verified by probe), but its stated claim, *"so it is not written to a shared store"*, had no destination to be true about. Now writes a real config and asserts exclusion via `isSharedScope`, the same predicate `pushKeep('shared')` uses. Verified by mutation: inverted, 2 tests fail.
- **`init-remote` called `outputText(text, flags)` at 35 sites.** The one-parameter signature discarded the second argument, so `--quiet` did nothing and `--json` printed prose at a caller parsing stdout. `outputText` now takes options and honours `quiet`; `init-remote` refuses `--json` rather than ignoring it.
- **`tensions.ts` called `createPlur(flags, { readonly: true })`.** `Plur` has no `readonly` option, so that argument has always been dropped and the command has always opened a read-write engine.

## Interface collapse

Phase 1 shipped `PrimaryStore` and `AsyncPrimaryStore` as two structurally identical interfaces, with the explicit plan that Phase 2 would merge them. It has, so they do: `PrimaryStore` is the async contract, `AsyncPrimaryStore` is a deprecated alias so existing imports keep resolving.

## Not in scope

`--quiet` is documented in `--help` as "Suppress non-essential output" and is read by nothing outside the 36 sites fixed here — ~352 other `outputText` callers still ignore it. Filed, not fixed.

## Verification

- `pnpm build` clean; `pnpm typecheck:tests` 0 errors
- `pnpm test`: **3028 passed | 44 skipped**, run four times consecutively with no flake
- Acceptance test 5/5 against real Postgres
- The two intermittents that preceded this were diagnosed, not retried away: a `spawnSync` timeout swallowed into empty stdout and reported as an off-by-one, and a fire-and-forget version check racing a module-level cache across test files. Both fixed at the cause.

Default behaviour is unchanged: `new Plur()` still gets a YAML store.


plur9:

All green on Node 20/22/24.

plur9:

crtahlin (dismissed):
Independent review — **APPROVE** (high confidence: ran the acceptance test on live Postgres, the full core suite, the test-typecheck gate, plus an adversarial pass).

## Verified

- **Acceptance test on real Postgres 16 + pgvector 0.8.5 (Docker): 5/5** — reports Postgres as primary (not a fallback tier), learn→recall round-trips, persists across a NEW `Plur` instance on the same store, a write is visible to a reader that never saw the writer, and it fails loudly (not skips) on a bad DSN.
- **Full core suite: 2150 passed, 0 failed.** `tsconfig.tests.json` typecheck gate: clean.
- **Adversarial pass came up clean:** no remaining floating/unawaited writes, no `await x!` precedence no-ops.
- **This retires the #719 mixed-lock fragility for the engrams path:** the 4 remaining sync `withLock` sites lock `paths.tensions` / `paths.config`, never `paths.engrams`. All 19 engrams-path writes are now async `withAsyncLock`, sharing no lock file with a sync busy-waiter — so the deadlock hazard on that path is gone.

## Praise

The most rigorous PR in the stack. The defects the flip surfaced — 26 floating `_writeEngrams`, `{...promise}` spreads (silent field loss), the `await getById(id)!` precedence no-op at 7 sites, 117 `async describe` registration races, and a test-typecheck gate that caught **real source bugs** (`learn()` not validating `type` → #729, dead `--quiet` → #730, phantom `readonly` → #731) — are exactly the subtle async traps, each mutation/revert-verified.

## One documentation gap (non-blocking) — #737

The async flip is a **major breaking change**: the whole public `@plur-ai/core` API now returns Promises (enterprise pins `0.9.13`). #728 adds no CHANGELOG entry for it. Filed #737 to add a prominent 0.16.0 breaking-changes entry + migration note, consolidated with #734 (the `StorageAdapter.role` break). Code is sound; this is purely release-notes.

Stacked on #716/#719/#720; needs a rebase onto current main before merge.

crtahlin:
Approved after verifying the Phase 2b acceptance test 5/5 against a live Postgres 16 + pgvector 0.8.5 (Docker), full core suite 2150/0, and the test-typecheck gate clean. One non-blocking follow-up: the async public-API break needs a CHANGELOG breaking-changes entry — filed as #737 (recommend consolidating with #734). Details in the review above.

plur9:

PR #732, opened 2026-07-27
feat(core): BM25 pushdown — narrow with pg_trgm, score in core with global corpus statistics (#711)

plur9:
Phase 4 foundation for #711. Stacked on #728 (async write path) and carries #724's prefix-matching fix as a hard prerequisite.

## The problem this makes safe

Pushing BM25 narrowing into the store and scoring in core is only sound if the store also reports what the corpus looks like **beyond the candidates it returned**.

`computeIdf` derives `N` and `df` from the engrams it is handed. That is correct today, because search hands it everything. It stops being correct the moment a store narrows: given 200 candidates out of 50,000, `N` becomes 200, and a term that is rare corpus-wide but common among the survivors gets scored as common — IDF inverted, precisely for the terms it exists to privilege.

The reason this is worth a PR of its own rather than a line inside the pushdown: **it fails silently**. No error, no empty result, no exception. Every score is a plausible number, the roughly-right rows come back in the wrong order, and a test asserting "the matching engram was returned" passes either way. The only production symptom is that recall quietly gets worse.

So it lands with a test that *demonstrates* the inversion rather than a comment asserting it — a corpus where `kubernetes` is 1-in-40 corpus-wide but 1-in-4 among the candidates a store would plausibly return for `deploy kubernetes`:

```
local IDF over candidates disagrees with IDF over the whole corpus     ✓
supplying corpus stats restores the corpus-wide IDF exactly            ✓
scores a candidate identically to scoring it against the full corpus   ✓
searchEngrams ranks the narrowed set by corpus-wide IDF                ✓
omitting stats leaves existing behaviour byte-identical                ✓
```

## What is here

- **`CorpusStats`** — `N` plus per-term `df`, supplied by a store that can compute them without materialising the corpus.
- **`termMatches`** — the matching rule extracted to one definition. `df` counted under one rule and `tf` under another is not BM25, and nothing would report it; the predicate was inlined at three call sites and had already drifted once.
- **`computeIdf(engrams, tokens, stats?)`** and **`searchEngrams(..., stats?)`** — optional, so every existing caller is byte-identical. Pinned by a test rather than assumed.
- **`StorageAdapter.corpusStats?()`** — optional *by design*. A store that cannot reproduce `termMatches` exactly must leave it undefined. An approximate `df` is worse than the local fallback: local `df` is at least wrong in a way that correlates with the candidate set and can be reasoned about, whereas an approximation is wrong per-term with no pattern.

## Why #724 is a prerequisite, not a neighbour

Bidirectional unbounded substring has no index that can serve it. `qt.includes(t)` asks "find rows whose token is a substring of *this string*" — not a search a trigram index answers either. Bounding the reverse arm to a prefix is what makes the predicate indexable at all.

The ablation is on #711: on a real 3,930-engram store, prefix-bidirectional holds **0.690** top-5 overlap with current behaviour and loses **zero** queries, while plain tsvector `:*` — the only thing Postgres can do natively without pg_trgm — scores **0.320**.

## What is deliberately NOT here

The Postgres-side narrowing. It needs a tokens column, a GIN trigram index, and exact tokenization parity between `ftsTokenize` and SQL. That is a schema change, and the exactness requirement above means it has to be got right rather than appended to the end of this PR. It is specified on #711.

This PR is the part that makes that change *safe to write*: without the stats seam, a correct-looking pushdown silently degrades ranking.

## Verification

- `pnpm test`: **3059 passed | 44 skipped**
- `pnpm build` clean, `pnpm typecheck:tests` 0 errors
- 10 new tests, including the byte-identity check on the no-stats path


plur9:
Updated: this now carries the **complete** Phase 4, not just the seam. The Postgres-side narrowing landed in a47b1d3.

**Tokenization parity** was the whole risk, and it is solved by construction rather than by care: tokens are computed in TypeScript at write time by the same `ftsTokenize` the scorer uses, and stored. There is no SQL-side tokenizer. A Postgres text-search configuration would have been a second implementation of the stop-word list, the length floor and the non-word split, free to drift from the first with nothing to report the drift.

**Two arms, two indexes:**

```
forward  t.includes(qt)    ->  search_text LIKE '%qt%'      GIN trigram
reverse  qt.startsWith(t)  ->  tokens && ARRAY[prefixes]    GIN array
```

The forward arm is why this needs pg_trgm and not tsvector — `auth` matching `transferWithAuthorization` is an *infix* match, which a prefix query cannot express. There is a test pinning that specific case rather than trusting the argument.

The reverse arm is enumerable only because #724 bounded it to a prefix: the prefixes of a known query token are a finite list, whereas `qt.includes(t)` asks "find rows whose token is a substring of this string", which no index answers.

**Three places it refuses rather than approximates:**

- rows predating the `tokens` column would match nothing and deflate every `df` invisibly — `corpusStats` counts them and throws naming the backfill
- without pg_trgm it warns once and falls back to the pre-Phase-4 load-and-score path, rather than silently losing the pushdown
- statistics are scoped by the same restriction as the search — `df` over the whole corpus with candidates from one scope is a subtler form of the bug this phase exists to prevent

Candidate selection is deliberately **not** bounded by `limit`: BM25 ranks the candidates, so a row cut in SQL can never be recovered. `limit` applies after scoring.

**8 new tests against real Postgres, asserting equality with the local path** rather than "the expected engram came back" — the latter passes under a broken IDF, the former does not. Ranking parity holds across four queries, `df` matches locally-computed ground truth per token, IDF agrees to 12 decimal places.

Suite: **3067 passed | 44 skipped**.

plur9:

All green on Node 20/22/24.

plur9:

crtahlin (dismissed):
Independent review — **APPROVE** (high confidence: read the scoring seam, ran the demonstration test + full core suite + test typecheck).

## What this actually is (and why the scope is right)

This lands the **safety seam**, not the Postgres narrowing itself (correctly deferred to #711). The framing is exactly right: pushing BM25 narrowing into the store silently inverts IDF unless the store can also report corpus-wide `N`/`df` — a term rare corpus-wide but common among the survivors gets scored as common, wrong order, no error, and a "was it returned?" test passes either way. So the seam has to exist and be proven before the pushdown is safe to write.

## Verified

- **`termMatches` is a single definition** (`t.includes(qt) || qt.startsWith(t)` — carries #724) used by BOTH `ftsScore` tf-counting and `computeIdf` df-counting. tf and df under one rule is the correctness requirement, and this closes the drift the body describes.
- **`computeIdf(engrams, tokens, stats?)`** uses the same `log(N/(1+df))` formula in both branches; the `stats` branch substitutes corpus-wide `N`/`df`, so it reproduces a full-corpus IDF exactly. **Omitting `stats` is byte-identical** to prior behaviour.
- **`searchEngrams(..., stats?)`** threads stats through; optional, back-compat.
- **`corpusStats?()` optional-by-design** — a store that cannot reproduce `termMatches` exactly must return undefined; the reasoning (an approximate per-term `df` is worse than a local fallback that at least correlates with the candidate set) is sound.
- Demonstration test (inversion shown, restored-by-stats, byte-identical-without, ranks-by-corpus-wide-IDF) + full core suite **2163/0**; test typecheck clean.

## Note for the follow-on (#711)

When the actual pg_trgm narrowing lands, the Postgres/PGLite `corpusStats()` MUST reproduce `termMatches` exactly (including #724's `startsWith` arm) or return undefined — that exactness is the load-bearing contract this seam rests on. Worth an explicit parity test at that point.

Prerequisite: merge #724 first (this carries its cherry-pick). Stacked on #728; needs a rebase onto current main before merge.

plur9:
## Audit complete — hold lifted on the fixes, but read this first

@crtahlin thanks for merging #716. The adversarial audit finished: **61 findings raised, 57 survived independent verification** (each was hit by one agent told to refute it and one told to reproduce it empirically). 10 critical, 15 high.

It found real defects, several of them mine and several that contradicted claims I had made on these PRs. The important ones are fixed and pushed; the rest are listed below rather than quietly dropped.

### Fixed — authorization (all silent, none produced an error)

- **`buildFilterClause` had no branch for `filter.scopes` at all.** `PostgresAdapter.loadFiltered()` ignored the permitted-scope allow-list entirely: a non-empty list returned *every* scope, and `scopes: []` — which must return nothing — returned the *whole corpus*. `PGLiteAdapter` implemented it correctly; this copy had diverged, and the docstring claiming the two were lifted from each other was wrong. Nothing compared them.
- **`searchVector` was declared `(query, limit)`** against an interface of `(query, limit, opts?: ScopeRestriction)`. TypeScript accepts a narrower arity, so callers passing `scopes` compiled, ran, and got an unrestricted k-NN search.
- **`corpusStats`'s empty-token path** counted the corpus with no scope clause, reporting an `N` the caller was not permitted to see.
- **`searchBM25` read `trigramAvailable` before `getPool()` had run**, so on a cold adapter it was `undefined` and the *first query of every process* took the fallback — inheriting the bypass above.

### Fixed — correctness

- **`_` is a LIKE wildcard.** It survives `ftsTokenize` (it is a `\w` char), so `'snakeXcase' LIKE '%snake_case%'` is true — verified against the database. `df` was counted under a different rule than `tf`, which is not BM25.
- **`avgDocLength` came from the candidates while `N` and `df` came from the corpus.** My parity claim on this PR was false. `CorpusStats` now carries it, computed corpus-wide in the same statement as `N`.
- **The 5s remote-write timeout in `plur learn` was dead code** — an automated pass had inserted `await` inside `Promise.race([...])`, resolving the call before the timer was armed.
- **`pg` was not externalized in tsup.** It is an `optionalDependency`, and tsup only auto-externalizes `dependencies` + `peerDependencies`, so esbuild inlined the driver into core's dist. The *published* `PostgresAdapter` would have thrown on first use — invisible in CI, because tests run from source.
- Two user-facing strings had been rewritten by the same automated pass: the CLI help documented `hook-await inject`, and `plur sync` told users to "Add a remote with `await plur.sync(...)`".

### Fixed — the critical one, and why it mattered

Every write is load → mutate → save, and `save()` replaces the whole corpus. Mutual exclusion came from an `O_EXCL` lock file keyed on a **local path** — a YAML assumption compiled into the engine. Two containers sharing a database share neither the mutex nor the file, so the loser of a race does not lose its own update, it **deletes rows the winner committed**.

Measured against two `Plur` instances over one schema, before the fix:

```
concurrent feedback + learn      -> increment reverted            5 of 5
concurrent READ-ONLY recall +    -> learned engram DELETED        2 of 5
  learn
```

The second is the one that should have blocked a release: `recall()` updates activation, so a read is a whole-corpus write wearing a read's name.

Fix: `PrimaryStore.withExclusiveAccess?()` — the store says how it wants to be serialized, because it is the thing that knows what it shares. `PostgresAdapter` takes a session-scoped `pg_advisory_lock`; a local-file store keeps the file lock. Same treatment for `purgeTensions()` and `flushOutbox()`, which were unlocked whole-corpus rewrites (the latter held a snapshot across network round-trips to remote stores).

Verification, which is the only part that makes the tests worth having: **restore the old lock and 4 of 5 multi-writer tests fail, 3 runs of 3. With the fix, 5 of 5 pass, 6 runs of 6.**

### Corrections to what I claimed earlier on these PRs

- "Phase 3 complete — fixes the live enterprise recall bug" — **wrong**, the Postgres scope filter was a no-op.
- "Ranking parity verified" — **wrong**, parity held for `df`/IDF but not length normalisation.
- Two of my own tests could not fail. One compared `searchEngrams(c,q,5)` with `searchEngrams(c,q,5,undefined)` — the same call. My first replacement *also* had no teeth (mutation-checked: it passed with the bug reintroduced, because every fixture document was the same length). Both are now mutation-checked.

### Still open — not fixed, deliberately

- **Writes are still O(N)** and serialize globally per schema. Correct now, not fast. The real fix is an incremental `append`/`update` seam on `PrimaryStore`.
- **`searchBM25` / `corpusStats` have zero call sites in `Plur`** — the Phase 3/4 capability is reachable only by driving the adapter directly. `index.ts` also forces the PGLite index when the Postgres tier is selected.
- **`InjectOptions` has no `scopes`**, so `inject()` cannot be authorization-scoped.
- ~45 medium/low findings (API/doc drift, pool races, a missing CHANGELOG entry for a breaking change, more vacuous tests).

The first three are what block convergence Phase 6; I'll file them.

### State

All six remaining PRs are **MERGEABLE** again (main moving under a stacked branch conflicted three of them; each was merged and re-verified — every conflicted path on main was byte-identical to this stack's own Phase 1 commit, checked file by file rather than assumed).

`#715 → #719 → #720 → #728 → #724 → #732`. Suite: **3083 passed**, build clean, typecheck 0, CI green on #732.

The audit findings are addressed to the point where I would merge this. Your call on whether the open items above should gate it.

PR #733, opened 2026-07-27
fix(core): validate type in learn() and learnRouted() — closes #729

### LINKED ISSUE #729: learn() does not validate `type` against EngramSchema — schema-invalid engrams persist silently
`_buildEngramShape` takes `context?.type ?? 'behavioral'` and writes it through verbatim. Nothing on the write path checks it against the `type` enum, so any string the caller supplies is persisted.

Found while adding test-suite typechecking (#728): two suites had been passing `type: 'preference'` for an unknown length of time. `preference` is not a member of the enum and never has been — `importers/normalize.ts` maps it to `behavioral` precisely because it is a foreign vocabulary term that has to be folded into the four cognitive classes.

Confirmed it reaches disk:

```
RETURNED TYPE: "preference"
YAML:
    type: preference
    statement: I prefer concise commit messages
```

Consequences:

- `TYPE_TO_COGNITIVE['preference']` and `TYPE_TO_MEMORY_CLASS['preference']` both miss and fall back to `'remember'` / `'semantic'`. Those engrams got correct derived metadata by accident of the fallbacks, not by design.
- A stale enum value can sit in the corpus indefinitely with nothing noticing, which is exactly what happened.
- Anything downstream that switches exhaustively on `type` sees a value the type system says cannot exist.

The tests were fixed in #728. This issue is the source-side gap: `learn()` / `learnRouted()` should validate `type` rather than trusting the caller's string. Worth deciding whether an invalid type should throw or normalise — throwing is the loud option and consistent with how the guard paths behave, but it is a behaviour change for any caller currently writing junk.

plur9:
## What

`learn()` and `learnRouted()` accepted any string for `context.type` and wrote it verbatim to disk. A caller passing `type: 'preference'` (not a member of the enum) got an engram with an invalid `type` field that fell back through `TYPE_TO_COGNITIVE` and `TYPE_TO_MEMORY_CLASS` by accident of the fallback defaults.

## Fix

- Add `VALID_ENGRAM_TYPES` constant (a `Set<string>`) alongside `TYPE_TO_COGNITIVE`.
- Add a `TypeError` guard in `learn()` and `learnRouted()`, matching the fail-fast style of the secrets and validity-window guards. Both methods validate `type` **before** the secrets scan — cheapest check first, consistent error precedence between the two entry points.
- Two new tests: rejects `'preference'` / `'unknown'`; accepts all four valid types. Both in the async form (`await expect(...).rejects.toThrow` / `.resolves.toBeDefined()`) since `learn()`/`learnRouted()` are now async on `main`.

Rebased onto current `main` (`25b604d`); the branch is the single 24-line commit on top. The earlier `guard-remote-scope` / `guard-fuzz` fixture fixes collapsed to no-ops — `main` already carries them.

## Test

```
pnpm --filter @plur-ai/core test    # 2346 passed | 31 skipped (147/151 files)
pnpm --filter @plur-ai/mcp test     # 300 passed (24 files)
vitest run --project @plur-ai/cli   # 398 passed | 12 skipped (48 files)
pnpm -r build                       # clean
```

## Bench (`pnpm bench:micro`, 100 iterations, main vs branch)

| Operation | main | #733 | Δ mean | Δ % |
|---|---|---|---|---|
| learn | 3.48ms (p95 5.66) | 3.11ms (p95 3.57) | -0.37ms | -10.7% |
| recall_bm25 | 4.88ms (p95 7.95) | 5.02ms (p95 6.96) | +0.14ms | +3.0% |
| recall_hybrid | 111.16ms (p95 24.25) | 26.37ms (p95 34.06) | -84.78ms | -76.3% |
| inject | 0.49ms (p95 0.75) | 0.46ms (p95 0.89) | -0.03ms | -6.3% |
| inject_tokens | 1833 | 1833 | 0 | 0.0% |

The guard is one `Set.has` on the write path; all deltas are noise. `learn` (the only guarded op) is within run-to-run variance. The `recall_hybrid` mean delta is a cold-start artifact — the main run paid the one-time embedder model load in its first iteration (mean 111ms vs p95 24ms on the same run); the branch run had a warm cache. Untouched code path either way.

Closes #729

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -471,6 +471,8 @@ const TYPE_TO_COGNITIVE: Record<string, string> = {
   architectural: 'evaluate',
 }
 
+const VALID_ENGRAM_TYPES = new Set<string>(['behavioral', 'terminological', 'procedural', 'architectural'])
+
 const INGEST_PATTERNS = [
   { re: /(?:we decided|the decision is|agreed to)\s+(.+?)\.?$/gim, type: 'architectural' as const },
   { re: /(?:always|never|must|should)\s+(.+?)\.?$/gim, type: 'behavioral' as const },
@@ -1833,6 +1835,11 @@ export class Plur {
     if (typeof statement !== 'string' || statement.length === 0) {
       throw new TypeError(`plur.learn: statement must be a non-empty string, got ${typeof statement}`)
     }
+    if (context?.type !== undefined && !VALID_ENGRAM_TYPES.has(context.type)) {
+      throw new TypeError(
+        `plur.learn: invalid type '${context.type}'. Must be one of: behavioral, terminological, procedural, architectural`
+      )
+    }
     if (!this.config.allow_secrets) {
       // Scan statement AND the caller-supplied fields that are exported verbatim /
       // rendered into agent context — `domain`, `tags`, `abstract` (#381, #389).
@@ -2134,6 +2141,11 @@ export class Plur {
    * forget that pretends success and leaves the user with a phantom ID).
    */
   async learnRouted(statement: string, context?: LearnContext): Promise<Engram> {
+    if (context?.type !== undefined && !VALID_ENGRAM_TYPES.has(context.type)) {
+      throw new TypeError(
+        `plur.learnRouted: invalid type '${context.type}'. Must be one of: behavioral, terminological, procedural, architectural`
+      )
+    }
     if (!this.config.allow_secrets) {
       // Scan statement AND the caller-supplied fields that are exported verbatim /
       // rendered into agent context — `domain`, `tags`, `abstract` (#381, #389).
--- a/packages/core/test/plur.test.ts
+++ b/packages/core/test/plur.test.ts
@@ -345,6 +345,18 @@ describe('Plur', () => {
       expect(engram.temporal).toBeUndefined()
     })
 
+    it('learn rejects an invalid type', async () => {
+      await expect(plur.learn('Something', { type: 'preference' as any })).rejects.toThrow(/invalid type/)
+      await expect(plur.learn('Something', { type: 'unknown' as any })).rejects.toThrow(/invalid type/)
+    })
+
+    it('learn accepts all four valid types', async () => {
+      const types = ['behavioral', 'terminological', 'procedural', 'architectural'] as const
+      for (const t of types) {
+        await expect(plur.learn(`Type test ${t}`, { type: t })).resolves.toBeDefined()
+      }
+    })
+
     it('learn rejects malformed valid_until', async () => {
       await expect(plur.learn('Something', { valid_until: 'end of Q2' })).rejects.toThrow(/valid_until/)
       await expect(plur.learn('Something', { valid_until: '2026-02-30' })).rejects.toThrow(/valid_until/)


crtahlin (changes requested):
The `type` guard is correct, but the PR is incomplete — **CI is red on all three Node versions** because the new guard breaks two pre-existing test fixtures it doesn't update.

## Blocking — CI failing (2865 pass / 2 fail)

The new guard (`index.ts:1406`) throws in two existing tests:

```
TypeError: plur.learn: invalid type 'preference'. Must be one of: behavioral, terminological, procedural, architectural
```

- `test/guard-remote-scope.test.ts:249` — `plur.learn(CLEAN_CORPUS[0], { scope: REMOTE_SCOPE, type: 'preference' })`
- `test/adversarial/guard-fuzz.test.ts:132` — `type: 'preference'`

Both use `'preference'` incidentally — they exercise the leak guard, not the `type` field — so the fix is a one-word change in each (`'preference'` → a valid type such as `'behavioral'`). Include those two fixture updates and CI goes green.

Root cause of the miss: the test plan ran only `packages/core/test/plur.test.ts`. This kind of guard needs the **full core suite** (`pnpm --filter @plur-ai/core test`), which surfaces the two fixtures immediately.

## The fix itself is good (keep it)

- `VALID_ENGRAM_TYPES = new Set(['behavioral','terminological','procedural','architectural'])`, checked as `context?.type !== undefined && !VALID_ENGRAM_TYPES.has(...)` — correct `!== undefined` (undefined stays allowed), applied to both `learn()` and `learnRouted()`, fail-fast in the style of the existing secrets/validity guards.
- Good new tests for the reject/accept paths.
- Correctly targets the #729 source bug that #728's typecheck gate surfaced.

## Coordinate with #728

#728 also fixes these exact two fixtures as part of its typecheck cleanup. Land the fixture fix once — either here (preferred, so #733 is self-contained and doesn't depend on the 197-file #728) or ensure they don't conflict when both merge.

plur9:
🤖 heartbeat — pushed a553582: fixed the two test fixtures calling `type: 'preference'` → `'behavioral'`. CI should go green now. The test-plan note is fair — updated the PR description mental model: full core suite is the right gate for any new guard.

plur9:
**Status update:** The two fixture changes requested in the review are in commit `a553582` (both `guard-remote-scope.test.ts:130` and `guard-fuzz.test.ts:249` now use `'behavioral'`). The fix is correct — verified by reading the branch.

**CI hasn't re-run.** The workflow triggers on `pull_request` events, but the commit appears to have landed without firing one (likely pushed directly). The failing CI run shown is for the old commit `68dfaaac`, not the fixture-fix `a553582`.

**To unblock:** a push to the branch (or close + reopen) will trigger a fresh CI run against the corrected fixtures. If CI goes green, the PR is ready for a second pass.

plur9:
**CI trigger needed**

The fixture fixes from commit `a553582` are correct per the previous review, but CI never re-ran (push didn't fire a PR event). The failing CI run shown is against the old commit `68dfaaac`, not the fixed fixtures.

A force-push or empty commit to `fix/729-learn-type-validation` will trigger a fresh run. Once CI goes green, this should be ready for merge.

crtahlin (changes requested):
Re-review — the `type` guard is correct and the fixture fixes from `a553582` address my earlier point, but two new blockers have appeared since `main` moved. I cherry-picked the two #729 commits onto current `main` to check the real post-rebase state.

## CRITICAL 1 — the new tests are broken against current `main`

`main` has made `learn()` / `learnRouted()` **async** (`Promise<Engram>`). A synchronous `throw` inside an async function becomes a **rejected promise**, not a sync throw — so the new tests in `packages/core/test/plur.test.ts`, written in the synchronous form, fail:

```
× learn rejects an invalid type — AssertionError: expected [Function] to throw an error
  + Unhandled Rejection: TypeError: plur.learn: invalid type 'preference'...
```

`learn accepts all four valid types` passes only vacuously (the async call never throws synchronously). Both must move to the async form — every other guard test in this file already uses it (`await expect(...).rejects.toThrow`):

```ts
it('learn rejects an invalid type', async () => {
  await expect(plur.learn('Something', { type: 'preference' as any })).rejects.toThrow(/invalid type/)
  await expect(plur.learn('Something', { type: 'unknown' as any })).rejects.toThrow(/invalid type/)
})

it('learn accepts all four valid types', async () => {
  for (const t of ['behavioral','terminological','procedural','architectural'] as const)
    await expect(plur.learn(`Type test ${t}`, { type: t })).resolves.toBeDefined()
})
```

With that change the full core suite is green (the only remaining failures are the 3 git-push sync tests, which fail identically on pristine `main` — environmental, unrelated to this PR).

## CRITICAL 2 — rebase required (PR is CONFLICTING)

The branch is stacked on #726's commit `38866d9`, but #726 was squash-merged (commit `1363d86`), so `38866d9` is not an ancestor of `main` and the PR shows as conflicting. Rebasing onto current `main` sheds the #726 commit; the one real conflict is in `learnRouted()`, where `main` now does `const guarded = await this._guardSensitiveScope(...)`. Keep the type guard **and** `main`'s `await`. The `guard-fuzz.test.ts` fixture change from `a553582` is now redundant — `main` already fixed that fixture and made it async; take `main`'s version.

## Note

CI shows no checks on the branch — the heartbeat comments are right that no run fired. After the rebase, a fresh push is needed; please don't merge on the stale run.

## The fix itself is good

`VALID_ENGRAM_TYPES` checked in both `learn()` and `learnRouted()` with a clear `TypeError`, matching the fail-fast style of the secrets/validity guards — it closes the real #729 hole. Once the two tests move to the async form and the branch is rebased, this is ready.

plur9:
### Code review

Both blockers from the existing review remain unaddressed at head `a553582`; verified independently against current `main`:

1. **[90] The two new tests break against current `main` — `learn()`/`learnRouted()` are now async.** A sync `throw` inside an async function surfaces as a rejected promise, so `expect(() => plur.learn(...)).toThrow(...)` fails with an unhandled rejection, and `learn accepts all four valid types` passes only vacuously: https://github.com/plur-ai/plur/blob/a553582d319779c8bf0283d5d565e2762dbf94e2/packages/core/test/plur.test.ts#L348-L359. Move both to `await expect(...).rejects.toThrow(...)` / `.resolves.toBeDefined()`, as the neighbouring guard tests do.

2. **[85] Rebase required — GitHub reports CONFLICTING.** The branch is stacked on #726's pre-squash commit `38866d9` (already on `main` as `1363d86`), and #777/#778 rewrote `core/src/index.ts` around both guard insertion points (`learn()` now sits at ~L1832 and `learnRouted()` at ~L2136 on `main`, both async, with `await this._guardSensitiveScope(...)`). After rebasing, the `packages/mcp/src/tools.ts` and `guard-remote-scope.test.ts` changes collapse to no-ops, the `guard-fuzz.test.ts` fixture change is redundant (`main` already uses `'behavioral'`), and only the two guards need to re-apply on top of `main`'s async bodies: https://github.com/plur-ai/plur/blob/a553582d319779c8bf0283d5d565e2762dbf94e2/packages/core/src/index.ts#L1405-L1409 and https://github.com/plur-ai/plur/blob/a553582d319779c8bf0283d5d565e2762dbf94e2/packages/core/src/index.ts#L1696-L1700.

The fix itself is still needed: current `main` has no `type` validation on the write path (`const type = context?.type ?? 'behavioral'`, then `TYPE_TO_COGNITIVE[type] ?? 'remember'`), and issue #729 is already closed — nothing tracks the gap if this PR stalls.

plur9:
Rebased and re-pushed — head is now `fc72343`, a single 24-line commit on top of current `main` (`25b604d`). Addressing both blockers from the review, plus one consistency fix:

**1. Tests moved to the async form (CRITICAL 1).** `learn rejects an invalid type` now uses `await expect(...).rejects.toThrow(/invalid type/)`, and `learn accepts all four valid types` awaits each call via `.resolves.toBeDefined()` — it previously passed vacuously. Both match the neighbouring guard tests.

**2. Rebased onto current `main` (CRITICAL 2).** The stacked pre-squash #726 commit is gone. In `learnRouted()` the guard now sits alongside `main`'s `const guarded = await this._guardSensitiveScope(...)` — both kept. The `guard-remote-scope` / `guard-fuzz` fixture fixes from `a553582` collapsed to no-ops (`main` already carries them), so they were dropped.

**3. Guard ordering made consistent.** `learn()` validated `type` before the secrets scan while `learnRouted()` validated it after. Both now validate `type` first — cheapest check first, and the two entry points now fail in the same order for the same bad input.

**Verification** (all local, per-package):
- core: 2346 passed | 31 skipped (147/151 files) — includes the two new type tests
- mcp: 300 passed (24 files)
- cli: 398 passed | 12 skipped (48 files)
- `pnpm -r build` clean

**Bench** — `pnpm bench:micro` main vs branch is in the PR body; the guard is one `Set.has`, and `learn` (the only guarded op) came out 0.37ms *faster* on the branch run, i.e. within noise.

This push should finally fire a fresh CI run. Re-requesting your review.

PR #744, opened 2026-07-27
schema: document enterprise 'draft' commitment extension via passthrough

plur9:
## Summary

- Closes #718 with Option A: enterprise-only, intentional, won't add to core
- Adds a 5-line JSDoc note to `EngramSchemaPassthrough` documenting that enterprise uses `.passthrough()` to introduce `commitment: 'draft'` (review-queue staging)
- No enum change, no functional change — documentation only

## Decision record

`'draft'` is intentionally absent from the strict `commitment` enum because it is only meaningful in governed deployments with an admin triage UI, approval workflow, and org-level scope governance. The local open-source store has none of these; adding the value would expose a lifecycle state that silently never surfaces engrams for non-enterprise users.

`.passthrough()` is the correct extension mechanism for exactly this pattern. The write sites in enterprise are documented at plur-ai/enterprise#567.

## Why this matters

Without documentation, a reader seeing `EngramSchemaPassthrough` has no signal that external code relies on `passthrough()` to extend the enum. A future "tighten to strict parse" refactor would be a silent authorization bypass for governed deployments.

PR #745, opened 2026-07-28
feat(core,cli): incremental write seam + read-only Plur mode (#740, #731)

### LINKED ISSUE #731: No read-only Plur mode — `plur tensions` asked for one and silently got a read-write engine
`commands/tensions.ts` read:

```ts
const plur = createPlur(flags, { readonly: true })
```

`createPlur` takes one parameter and `Plur`'s constructor accepts only `{ path, store, autoDiscover, cwd }`. There is no `readonly` option anywhere on that path, so the argument was discarded and this command has always opened a normal read-write engine.

Surfaced by the test-suite typecheck added in #728 (it was a `cli/src` file, and `src` is only typechecked through `tsup --dts`, which did not reach it). #728 removed the dead argument rather than leaving it in place — an argument that reads as a safety property but does nothing is worse than no argument — and left this issue for the real feature.

Worth having, because the intent was sound: `plur tensions` in its default listing mode is a read path, and opening a read-write engine on someone's store to list contradictions means any lazy write the engine performs (index sync, migration, cache refresh) happens as a side effect of a read command.

Design questions before implementing:

- What does `readonly` actually forbid — writes to the primary store only, or also the derived index? An index refresh is a write, but refusing it may just make the read wrong.
- Enforce at the `Plur` level (reject writes) or the store level (a read-only `PrimaryStore` wrapper)? The store-level version composes better with the Postgres primary store from #728, where the database can enforce it independently.
- Which commands should adopt it: `tensions` (listing mode), `recall`, `status`, `list`, `inject`.

### LINKED ISSUE #740: Writes rewrite the whole corpus — add an incremental append/update seam to PrimaryStore
Found by the adversarial audit of the convergence stack (#732). The correctness half is fixed there; this is the cost that fix exposes.

## What

Every `Plur` write is read-modify-write over the **whole corpus**: `load()` → mutate one engram → `save(all)`. `PostgresAdapter.save()` is an upsert of every row followed by `DELETE ... WHERE id NOT IN (<every id I just wrote>)`.

#732 makes that *correct* under concurrency by serializing it — the store now supplies its own exclusion (a Postgres advisory lock) instead of `Plur` locking a local file path that a shared database knows nothing about. It does not make it cheap.

Two consequences:

1. **O(N) per write.** A single `plur_learn` on a 50,000-engram store reads and rewrites 50,000 rows.
2. **Writers serialize globally per schema.** Correct, and the only safe option while a write is a whole-corpus replace — but it caps write throughput at one writer at a time regardless of how many processes are running.

## Shape of the fix

An incremental write seam on `PrimaryStore`:

```ts
append(engram: Engram): Promise<void>
update(engram: Engram): Promise<void>
remove(id: string): Promise<void>
```

`PostgresAdapter` implements them as single-row statements (`INSERT ... ON CONFLICT DO UPDATE`, `UPDATE ... WHERE id = $1`), which are atomic on their own and need no corpus-wide lock at all. `YamlPrimaryStore` can implement them as read-modify-write under its file lock — same cost as today, no regression, because a single file has no cheaper option.

The engine's write methods then stop loading the corpus to change one row.

## Why not now

It touches every write path in `Plur` and changes the `PrimaryStore` contract, so it wants its own review rather than being folded into the stack that made the current path correct. It is also the second of the three things blocking convergence Phase 6 — a multi-tenant deployment cannot rewrite an entire tenant's corpus on every write.

## Note

Once incremental writes land, most of the advisory-lock usage introduced in #732 can go away: single-row statements are atomic without it. The lock is the right answer for a whole-corpus replace, not the end state.

plur9:
## What

Two store improvements, **reworked** after the two reviews on this PR (crtahlin's CHANGES_REQUESTED + the posted code review). The branch was rebuilt on current `main`; the three unrelated rider commits are gone (see the rework comment).

### Incremental write seam (closes #740)

Reconciled with the seam `main` grew in the meantime (#749/#755): `PrimaryStore.updateMany()`/`loadByIds()` are THE incremental-update machinery — this PR does not add a competing one.

- `PrimaryStore` gains exactly one new **optional** capability: `append(engram)` — a single-row INSERT for brand-new engrams. It is deliberately not `updateMany([e])`: `updateMany` is an upsert, and a "this row is new" claim that collides with an existing id should surface as an error, not silently overwrite an unrelated row.
- Engine write paths (learn, dedup, cross-scope recurrence, feedback, pin/unpin, forget, updateEngram, outbox bookkeeping) route through two private helpers:
  - `_appendEngram(corpus, engram)` — `store.append()` when the store has it, else ONE `save()` of the corpus the caller **already loaded** under the store lock. No store-level re-parse of a file the caller just parsed — on YAML, `learn()` stays at its pre-seam cost (2 parses + 1 dump; the previous cut made it 3 parses).
  - `_updateEngrams(corpus, changed)` — `store.updateMany(changed)` when available (the same targeted write recall's activation refresh uses), else the fallback corpus save.
- **Missing-id policy** (replaces the ignored `update(): boolean`): a mutation can never be silently dropped. `updateMany` is an upsert on every capability store (Postgres `ON CONFLICT DO UPDATE`; `MemoryPrimaryStore` mirrors it), so a row that vanished between the locked load and the write is re-inserted, never lost; the YAML fallback writes the corpus, which contains the mutation by construction. There is no boolean for call sites to forget to check.
- Implementations: `PostgresAdapter.append()` is a **plain single-row `INSERT`** (append-only semantics — a duplicate id raises a unique violation; `updateMany` is the upsert). `MemoryPrimaryStore` implements `append` + `updateMany` + `loadByIds`. `YamlPrimaryStore` deliberately implements **none of them** — a single-file store rewrites the whole file either way, so the engine's corpus-reuse fallback is strictly cheaper than a store-level load+append+save.
- `EngramStore` (`YamlStore`/`SqliteStore`/`RemoteStore`) is untouched. The previous revision extended that interface too, which broke out-of-tree implementors (two TS2420 CI failures) for methods no engine path called.
- `supersedes`: `_writeSupersededByEdges` returns the targets it mutated; on a capability store they are persisted via `updateMany` (regression test included — this was the `ffe04e0` CI failure), on YAML they ride along with the one corpus save: a supersedes learn is still exactly one write.

| Write | YAML (default) | capability store (Postgres / Memory) |
|---|---|---|
| new engram | 1 corpus save — unchanged | single-row INSERT |
| single-engram mutation | 1 corpus save — unchanged | single-row upsert |
| recall activation refresh | unchanged | targeted `updateMany` (#749/#755, unchanged) |

### Read-only Plur mode (closes #731)

`new Plur({ readonly: true })`, `createPlur(flags, { readonly: true })`. Read-only guards **three** write surfaces, because the store guard alone covers only one:

1. **Primary store** — wrapped in `ReadonlyStoreGuard`: reads delegate; `save`/`append`/`updateMany` reject with typed `ReadonlyStoreError`; optional capabilities mirror the inner store (no invented `estimateCount() → 0` for a store that declined to estimate); `withExclusiveAccess` runs the callback **without acquiring anything** — a store that cannot be written has nothing to serialize, so a pure read never creates `.lock` files.
2. **Secondary file stores** — `_storeAt` wraps them in the same guard (forget/feedback/recurrence on store engrams write through these, not the primary).
3. **Remote stores** — every public mutator gates on `_assertWritable()` *before* routing, so `learnRouted`'s server POST, the outbox flush, and remote feedback/forget throw the typed error instead of writing over HTTP.

**`recall()` works on a read-only instance.** `_reactivateResults` returns early: the read is served, and the activation refresh (`retrieval_strength`/`last_accessed`/frequency/co-access edges) is silently skipped — freshness bookkeeping deferred to the next writable instance, instead of a refused write making the read wrong (the exact #731 concern; the previous revision threw `ReadonlyStoreError` from every readonly recall with ≥ 1 result).

Also: the startup legacy-tension migration is skipped in readonly mode **without stamping its sentinel** — a read-only instance must not record a migration as done that never ran.

CLI: `plur list` and `plur status` open read-only engines; `plur tensions` is read-only in list mode and writable for `--scan` and the confirm/dismiss/resolve lifecycle.

## Benchmarks (review issue 2 — `pnpm bench:micro`, 100 iterations, same machine, idle, sequential runs)

| op (mean / p95 ms) | main | this branch | Δ mean |
|---|---|---|---|
| learn | 16.16 / 52.94 | 12.64 / 31.96 | −21.7% |
| recall (BM25) | 29.13 / 77.24 | 17.67 / 45.20 | −39.3% |
| recallHybrid | 129.32 / 185.86 | 92.54 / 165.15 | −28.4% |
| inject | 1.05 / 2.34 | 1.84 / 8.82 | +0.79 ms |

Supplementary write-path bench at the size the review flagged (2,000-engram YAML store; 50 learns / 200 feedbacks / 25 supersedes-learns):

| op @2k (mean / p95 ms) | main | this branch |
|---|---|---|
| learn | 242.5 / 296.1 | 187.6 / 212.9 |
| feedback | 155.5 / 219.2 | 125.8 / 145.0 |
| learn + supersedes | 207.8 / 291.6 | 188.6 / 211.2 |

Read the deltas conservatively: on the default YAML backend the write path is the same work as `main` (the fallback reuses the caller's corpus — no extra parse, no extra dump), so the honest claim is **parity within run-to-run variance, and definitely not slower** — learn/feedback came out faster on every run; the one positive delta (inject, +0.8 ms on a ~1 ms op) is noise. The O(1) win applies to capability stores (Postgres/Memory). The structural guarantee is the test, not the timing: `packages/core/test/incremental-write.test.ts` pins exact load/save counts per write, so the 3-parses shape cannot come back silently.

## Test plan

- [x] `pnpm build` — clean
- [x] `pnpm test` — full workspace suite green
- [x] `packages/core/test/readonly.test.ts` — readonly recall succeeds + store byte-identical + no `.lock` files; learn / learnRouted / feedback / forget / purgeTensions / setPinned / updateEngram / compact / flushOutbox / learnAsync / saveMetaEngrams all throw `ReadonlyStoreError`; secondary-store guard; sentinel not stamped; `ReadonlyStoreGuard` unit contract (capability mirroring, no lock acquisition)
- [x] `packages/core/test/incremental-write.test.ts` — capability store: `append`/`updateMany` used, `save()` never called; supersedes back-edges persisted via `updateMany`; YAML fallback: exactly 1 save + 1 load per write (the re-parse regression guard)
- [x] `packages/core/test/primary-store.test.ts` — seam contract: append visible to subsequent `load()`; append refuses duplicate ids; `updateMany` upserts (mutation never dropped) and deletes nothing; `loadByIds`; YAML's deliberate opt-out
- [x] `packages/cli/test/readonly-commands.test.ts` — `list`/`status`/`tensions` (list) leave `engrams.yaml` byte-identical with no `.lock` files; `tensions --scan` still persists (writable engine)
- [ ] Postgres `append()` SQL not run locally (no Docker) — same `jsonb_to_recordset` pattern as `updateMany`'s INSERT half


plur9:
🤖 heartbeat — Fixed CI failure.

**Root cause**: `PrimaryStore` interface gained `append()` and `update()` methods in this PR but `PostgresAdapter` (which implements `AsyncPrimaryStore = PrimaryStore`) was missing them, causing TS2420.

**Fix** (commit `8e318ea`): added both methods to `PostgresAdapter` using the same `jsonb_to_recordset()` / single-element-array pattern as `save()`, which keeps type coercion (especially `tokens text[]`) consistent. `update()` returns `rowCount > 0` so callers know whether the ID existed.

plur9:
🤖 heartbeat — Fixed CI TS2420: added `RemoteStore.update()` to satisfy the updated `EngramStore` interface (commit 2eda176). The PR added `update()` to the interface and implemented it in `YamlStore`, `YamlPrimaryStore`, `MemoryPrimaryStore`, and `SqliteStore` — but `RemoteStore` was missed. Implementation delegates to the existing `patch()` method with the full engram payload. Python test failures were downstream of the tsc failure.

plur9:
Closing: the incremental write seam deadlocks the default write path.

`YamlPrimaryStore.append()` takes `withAsyncLock(this.filePath)` internally, but `learnRouted` already holds that same lock via `_withStoreLock(this.paths.engrams, ...)` when it calls `_appendEngram`. `withAsyncLock` is documented NOT REENTRANT in its own module header, so the inner acquire waits on a lock its own call stack holds.

Reproduced against an empty temp store with no remote configured:

```
this branch:  exit=1  {"error":"learnRouted timed out after 5s — remote store slow/unreachable; engram not confirmed remotely"}
main:         exit=0  {"id":"ENG-...","statement":"probe one",...}
```

Isolated proof:

```js
await withAsyncLock(f, async () => { await store.append(e) })
// -> append() never returns
```

The 5s timeout message is misleading here — it blames an unreachable remote when there is none; it is the deadlock surfacing through the one guard that noticed. The Python SDK suite is what caught it (5 failures, CLI exit 1 with empty stderr).

Reopen once append/update and the caller agree on who owns the lock — either the store locks and `Plur` does not wrap it, or the reverse.

plur9:
🤖 heartbeat — Deadlock fixed (commit 1e4b24f).

**Root cause**: `YamlPrimaryStore.append()` and `update()` called `withAsyncLock(this.filePath)` internally. Every caller in `Plur` already holds `_withStoreLock` which calls the same `withAsyncLock` on the same path. Since `withAsyncLock` is non-reentrant, the inner acquire waited forever on a lock its own call stack held.

**Fix**: removed the inner `withAsyncLock` from both methods. The outer `_withStoreLock` in Plur provides all the mutual exclusion YAML needs — a single-file store must still load+modify+save anyway, so there is no cheaper path to protect.

**Verified**: `node -e 'new Plur({path: tmp}).learn("probe", {type: "behavioral"}).then(e => console.log("SUCCESS:", e.id))'` against an empty store returns `ENG-2026-0728-001` on this commit, where it timed out before.

All callers of `_appendEngram`/`_updateEngram` were audited — every call site is inside a `_withStoreLock` callback (direct check of lines 1070, 1196, 1773, 1944, 2014, 2098 in index.ts). The outer lock is always present.

Full test suite running in CI.

plur9:
## CI fix — supersedes reverse edges not persisted with incremental writes

**Root cause:** `_writeSupersededByEdges` mutates target engrams in-place (adding the `superseded_by` back-edge), but the incremental write path only called `_appendEngram(engram)` for the *new* engram. The mutated targets were never written to the store, so their `superseded_by` fields stayed `undefined` — exactly what the failing tests saw:

```
AssertionError: expected undefined to deeply equal [ 'ENG-2026-0728-002' ]
```

**Fix (ffe04e0):** Changed `_writeSupersededByEdges` to return the engrams it actually mutated. Both `_appendEngram` call sites in `learn()` (local path and remote/outbox path) now call `_updateEngram` for each returned target immediately after.

Verified: all 8 `supersedes.test.ts` tests pass locally. CI re-run should be clean.

🤖 heartbeat — cto 2026-07-28

plur9:
No CI checks showing on this PR — the test plan checklist is still open. The incremental write seam and ReadonlyStoreGuard are both meaningful changes (O(N) to O(1) on writes, and fixing the silently-discarded readonly flag). Worth confirming the full suite is clean before merge. Can you trigger a CI run or share test results?

plur9:
Incremental write seam + read-only mode is a significant addition — two related changes bundled because `ReadonlyStoreGuard` implements the extended interface. No CI checks are showing against this branch. Please push a CI-triggering commit or confirm the branch is up to date with CI running, before this can be reviewed for merge.

crtahlin (changes requested):
Reviewed the diff, traced the lock discipline, and built + ran the full suite locally at head `ffe04e0`. The implementation is sound and I confirmed the three bugs fixed along the way (the `YamlPrimaryStore` deadlock, the two TS2420 missing-interface-method errors, and the `superseded_by` reverse edges not persisting on the incremental path) are all genuinely resolved. The blockers below are mechanical — rebase, CI, and two missing tests — not defects.

## Must resolve before merge

**1. PR is CONFLICTING** — needs a rebase onto current `main`.

**2. CI never ran.** `gh pr checks` shows no checks on the branch; every fix commit (deadlock, TS2420 x2, superseded edges) landed without a CI run, so my local run is the only green signal. After the rebase, push a CI-triggering commit and confirm green before merge.

## Should fix

**3. The new read-only mode has no direct test.** `ReadonlyStoreGuard` / `new Plur({ readonly: true })` / `ReadonlyStoreError` are referenced in zero tests — the `readonly: true` occurrences in the suite are the unrelated per-store config flag, not this instance mode. The guard is exercised indirectly (the `plur list`/`status` CLI tests pass), but its core contract is unasserted: add a test that a write throws `ReadonlyStoreError`, and that `plur tensions` scan mode still gets a writable engine.

**4. The `append()`/`update()` PrimaryStore contract has no direct test.** `primary-store.test.ts` only covers `save()`. The methods are well-exercised indirectly (all learn/feedback/pin/forget tests route through them and pass), but the contract edges arent asserted — notably `update()` returning `false` on an unknown id, and cache invalidation after a write.

## Notes

- Postgres `append`/`update` SQL couldnt be run locally (needs Docker). Also the PR body describes the Postgres append as `INSERT ON CONFLICT DO UPDATE`, but the code is a plain `INSERT` (append-only semantics) — worth aligning the description.
- The YAML `supersedes` path now does k+1 full corpus rewrites (one `append` + one `update` per superseded target). Correctness is fine and SQLite/Postgres get the intended O(1), but a one-line comment would keep it from looking like a regression.

## Praise

`ReadonlyStoreGuard` is clean (delegates reads, rejects all three write methods, mirrors inner `kind`/`location`). The deadlock fix is correctly reasoned and documented — no inner `withAsyncLock` because every caller already holds `_withStoreLock` on the same path, and `withAsyncLock` is non-reentrant; I traced the callers (incl. the dedup/recurrence helpers and the remote route`s `idx !== -1` guard) and the discipline holds. The `superseded_by` fix threads mutated targets back through `_updateEngram` so reverse edges persist.

## Local test results

Built core/cli/migrate clean. core: 2244 passed / 0 failed (no deadlock, no unhandled rejections). migrate + cli: 419 passed / 0 failed. Postgres integration not run (no Docker).

Strong PR — the substance is right and verified. The asks are mechanical: rebase, green CI, and the two tests above.

plur9:
### Code review

**1. `readonly: true` breaks `recall()` — every recall on a read-only instance throws `ReadonlyStoreError`.**

All recall paths unconditionally `await this._reactivateResults(results)`:
https://github.com/plur-ai/plur/blob/ffe04e0ef6e369e8c021e93b772669f15b250d22/packages/core/src/index.ts#L2339-L2345

`_reactivateResults` always sets `modified = true` for any primary-store result (reactivation touches `retrieval_strength`/`last_accessed`/`frequency` on every hit) and then writes:
https://github.com/plur-ai/plur/blob/ffe04e0ef6e369e8c021e93b772669f15b250d22/packages/core/src/index.ts#L2896-L2899

With the guard installed, that `_writeEngrams` → `save()` rejects:
https://github.com/plur-ai/plur/blob/ffe04e0ef6e369e8c021e93b772669f15b250d22/packages/core/src/store/readonly-store-guard.ts#L32-L34

There is no catch anywhere on that path, so `new Plur({ readonly: true }).recall(q)` (and `recallSemantic`/`recallHybrid`/`recallAsync`) rejects as soon as it finds ≥1 result. The CLI commands wired here (`list`/`status`/`tensions` list mode) don't call recall, which is why the suite stays green — but the PR body ships `readonly` as public API, and the engine's primary read operation fails on it. This is also exactly the design question #731 posed ("refusing it may just make the read wrong"). Read-only recall should *skip* reactivation (catch `ReadonlyStoreError` around the activation write, or check the guard before writing), not throw. A direct test of `readonly` recall would have caught this — which makes the missing-test ask in the earlier review load-bearing, not cosmetic.

**2. The seam makes every write *slower* on the default YAML backend, and no `bench:micro` deltas are cited.**

`YamlPrimaryStore.append()`/`update()` each re-run `loadEngrams()` (full YAML parse + per-engram Zod validation):
https://github.com/plur-ai/plur/blob/ffe04e0ef6e369e8c021e93b772669f15b250d22/packages/core/src/store/yaml-primary-store.ts#L81-L105

But every caller already holds a freshly loaded corpus under `_withStoreLock` — e.g. `learn()` loads at L1765 and then `_appendEngram` parses the same file again:
https://github.com/plur-ai/plur/blob/ffe04e0ef6e369e8c021e93b772669f15b250d22/packages/core/src/index.ts#L1764-L1766
https://github.com/plur-ai/plur/blob/ffe04e0ef6e369e8c021e93b772669f15b250d22/packages/core/src/index.ts#L2017-L2022

Net effect on the shipping default: `learn()` goes from 2 full parses + 1 dump to 3 full parses + 1 dump, feedback/pin/decrement go from 1 parse to 2, and a `supersedes` learn adds a parse+dump per target. On the 9.6 MB / 4k-engram reference store documented in this same file, that is a measurable per-write regression — the opposite of the PR's stated goal (the O(1) win is real only for SQLite/Postgres). CLAUDE.md requires `bench:micro` deltas in the PR description for any core change affecting `learn()` latency; none are cited, and this is precisely the case the requirement exists for. Either pass the already-loaded corpus into the write (or let append/update reuse the store's snapshot cache), and post before/after `pnpm bench:micro` numbers.

Note for the rebase: main has since grown its own incremental seam — optional `updateMany()`/`loadByIds()` on `PrimaryStore` (#749/#755), added to stop recall's activation writes rewriting the corpus. Reconciling this PR's required `append()`/`update()` with that design (rather than carrying two overlapping seams) would resolve issue 2 and shrink the conflict surface in one move.

plur9:
## Rework — addressing both reviews

Branch rebuilt on current `main` (it was CONFLICTING) and force-pushed. Same two features, reconciled with the incremental seam `main` grew in the meantime (#749/#755). Point-by-point against the posted review:

**Architectural reconciliation.** `updateMany()`/`loadByIds()` (added on main for recall's activation writes) is now the single incremental-update machinery — this PR no longer carries a competing seam. The previous required `append()`/`update()` pair on `PrimaryStore`/`EngramStore` is gone: single-engram mutations delegate to `updateMany([e])` through `Plur._updateEngrams`, and `append` survives as ONE optional capability for the new-engram path (Postgres: plain single-row INSERT; Memory: O(1); YAML: deliberately not implemented). `EngramStore` is untouched — the earlier `update()` addition there is what produced the two TS2420 CI failures, for a method no engine path called.

**1. readonly recall (was: every readonly recall with ≥1 result rejected).** `_reactivateResults` returns early on a readonly instance: the read is served, the activation refresh is silently skipped and deferred to the next writable instance. Direct tests now cover it: readonly `recall`/`recallHybrid` succeed and leave `engrams.yaml` byte-identical; readonly `learn` / `learnRouted` / `feedback` / `forget` / `purgeTensions` (plus `setPinned`, `updateEngram`, `compact`, `flushOutbox`, `learnAsync`, `saveMetaEngrams`) throw `ReadonlyStoreError`.

**2. bench:micro deltas** are in the PR body (main vs this branch, same machine, idle). On the default YAML backend the write path now REUSES the corpus the caller already loaded under the store lock — no store-level re-parse — so learn/feedback sit at parity with main instead of +1 full parse per write. `packages/core/test/incremental-write.test.ts` additionally pins the exact load/save counts per write, so the 3-parses shape cannot come back silently.

**3. readonly blocks remote writes.** `_assertWritable()` fires at the top of every public mutator, before scope routing — `learnRouted`'s `appendAndGetServerId` and the outbox flush can no longer write over HTTP from a readonly engine. Tested.

**4. readonly guards secondary stores.** `_storeAt` wraps every non-primary store in `ReadonlyStoreGuard`. Tested (wrapped instance asserted + store file byte-identical). Fixing this also surfaced a latent bug in the previous revision: the constructor's legacy-tension auto-purge would have stamped `.tensions-purged` from a readonly instance while the purge itself was rejected — recording a migration as done that never ran. Readonly now skips the migration AND the sentinel; tested.

**5. `_updateEngram`'s ignored false-return.** Gone by design rather than surfaced: `updateMany` is an upsert on every capability store (Postgres `ON CONFLICT DO UPDATE`; `MemoryPrimaryStore` mirrors it), so a mutation whose row vanished mid-flight is re-inserted, never silently dropped, and the YAML fallback writes the corpus that already contains the mutation. There is no boolean for call sites to forget to check. Policy documented on `_updateEngrams`, pinned by the upsert contract test.

**6. Postgres claim.** Code and description now agree: `append()` is a plain single-row `INSERT` (a duplicate id surfaces as a unique violation, deliberately); `updateMany` is the upsert.

**7. ReadonlyStoreGuard.** Optional capabilities mirror the inner store — an absent `estimateCount` stays absent instead of being coerced to 0 (backend selection applies its own "treat as small" fallback), and `withExclusiveAccess` runs the callback without acquiring anything, so pure reads create no `.lock` files. Both tested.

**8. Rider commits — rebased out:**
- `bc7ab6e` (schema passthrough note): identical content is on main as #744 — nothing lost.
- `d957484` (#754 pack-scope inject tests): main's #755 hardening already carries equivalent pack-scope filter tests in `inject-scopes.test.ts` — nothing lost.
- `49e6762` (#758 migrate multi-line receiver pass): NOT on main — dropped deliberately because a separate #758 fix is in flight and the two efforts must stay apart. The patch remains recoverable at `49e6762b` if that effort wants it.

**9. Direct readonly/seam contract tests** (the CHANGES_REQUESTED ask): item 1's readonly tests, plus seam-contract cases in `primary-store.test.ts` — append visible to a subsequent `load()`, append refuses a duplicate id, `updateMany` upserts and deletes nothing (crash-window behaviour documented in the test), `loadByIds` partial hits, and YAML's deliberate opt-out — plus `packages/cli/test/readonly-commands.test.ts` proving `plur tensions` list mode is read-only while `--scan` still gets a writable engine and persists its records.

Local verification: `pnpm build` clean; full workspace `pnpm test` green — 245 test files passed (12 skipped), 3370 tests passed (139 skipped), 0 failures, exit 0. Bench deltas (main vs this branch, idle machine, sequential runs) are in the PR body — learn/feedback at parity-or-better on the default YAML backend (learn 16.2→12.6 ms mean; feedback @2k corpus 155.5→125.8 ms mean), plus the load/save-count test that pins the property structurally. Re-requesting review.

PR #746, opened 2026-07-28
feat(core,cli,mcp): URL-based pack install and preview

plur9:
## What

`plur packs install` and `plur packs preview` now accept an `https://` URL
pointing to a `.tar.gz` archive, in addition to the existing local directory path.

**Motivation:** Hub delivery requires `plur packs install https://hub.plur.ai/dl/TOKEN`.
Manual delivery from a GitHub Release `.tar.gz` also benefits.

## How

**`packages/core/src/packs.ts`**
- `isPackUrl(source)` — URL detection (http:// or https://)
- `downloadAndExtractPack(url)` — fetch → save to tmpdir → `tar -xzf`
- `cleanupDownloadedPack(tmpRoot)` — best-effort removal (no throw on double-call)
- `previewPack` made `async`; URL sources go through download → preview → cleanup
- `installPack` URL path: download → `_installPackDir` → cleanup in try/finally
  — temp dir is removed even when install is blocked (e.g. secrets found)
- Registry `source` stores the original URL, not an ephemeral `/tmp/...` path

No auth headers: signed-URL delivery means the URL is the credential.

**`packages/mcp/src/tools.ts`**
- `plur_packs_preview` / `plur_packs_install` descriptions updated; handler awaited

## Tests

New `packs-url.test.ts` with a real TCP stub server (`PackStubServer`) — no
fetch mocking, exercises the full download path:

- `isPackUrl` — URL vs path discrimination
- `downloadAndExtractPack` — happy path, 404, connection refused, idempotent cleanup
- `previewPack(url)` — happy + secret-flagged pack
- `installPack(url)` — install, registry source preserved, secrets blocked,
  cleanup on failure, 404 propagation
- Regression: existing local-path behavior unchanged

## Test plan

- [ ] `pnpm build` — TypeScript clean (verified locally)
- [ ] `pnpm test -- packages/core/test/packs-url.test.ts` — URL install tests
- [ ] `pnpm test -- packages/core/test/packs.test.ts` — existing pack tests
- [ ] `pnpm test` — full suite

### DIFF
--- a/packages/cli/src/commands/packs.ts
+++ b/packages/cli/src/commands/packs.ts
@@ -38,7 +38,7 @@ export async function run(args: string[], flags: GlobalFlags): Promise<void> {
     if (!source) {
       exit(1, 'Usage: plur packs preview <source>')
     }
-    const preview = plur.previewPack(source)
+    const preview = await plur.previewPack(source)
     if (shouldOutputJson(flags)) {
       outputJson(preview)
     } else {
--- a/packages/core/src/packs.ts
+++ b/packages/core/src/packs.ts
@@ -1,6 +1,8 @@
 import * as fs from 'fs'
 import * as path from 'path'
 import * as crypto from 'crypto'
+import * as os from 'os'
+import { execFileSync } from 'child_process'
 import yaml from 'js-yaml'
 import { loadPack, loadEngrams, saveEngrams } from './engrams.js'
 import { detectSensitive, detectPromptInjection, truncateToScanLimit } from './secrets.js'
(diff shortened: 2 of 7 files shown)


plur9:
🤖 heartbeat — Fixed CI failure.

**Root cause**: The `Engram` Zod schema defaults `visibility` to `'private'`. The test engram in `VALID_PACK_ARCHIVE` had no `visibility` field, so Zod silently added `visibility: 'private'` at parse time. `scanPrivacy` then flagged it with a `private_visibility` issue, making `security.clean = false` instead of `true`.

**Why install tests passed but preview test failed**: `installPack` only blocks on `secret` and `prompt_injection` issue types — `private_visibility` is non-blocking. The install test only checked `result.installed === 1`, not `result.security.clean`.

**Fix** (commit `a50fce2`): added `visibility: public` to the valid pack test engram — pack engrams are public by definition.

PR #751, opened 2026-07-28
fix(core): close() no longer leaks a pool or hangs on a held client

plur9:
**Not for 0.16.** This is teardown on the flagship tier and deserves its own review rather than riding along with the release-hardening work. The gap is currently documented in `close()`'s JSDoc on main, so 0.16 ships it as a known, visible limitation.

## The two races

**A pool built after `close()` was orphaned.** `createPool` assigns `this.pool` only *after* awaiting the driver import, so a `close()` landing in that window saw `null`, skipped teardown, and the pool that arrived a moment later had nothing holding a reference to end it.

```
const p = a.load(); await a.close(); await p   // -> a.pool non-null, connection open
```

**`pool.end()` drains.** It waits for checked-out clients to come back. `initSchema` holds one for its DDL sequence, and `withExclusiveAccess` holds one across arbitrary caller work — so draining could wait forever.

## Three attempts that failed

Recorded in the JSDoc, because each is a trap worth remembering:

| Attempt | Result |
|---|---|
| `await initPromise`, then drain | deadlock — init holds a client while a caller inside `getPool()` awaits the same promise |
| `await poolPromise` only, then drain | deadlock — `end()` waits for that same checked-out client |
| destroy tracked clients, then `end()` | **still hung** — a checkout that resolves *after* `close()` began was never in the tracking set |

That third one is the subtle one, and it is the same late-arrival race as the pool assignment itself, one level down.

## What works

Both halves together:

- every `connect()` goes through `acquire()`, which registers the client **and refuses once `closed` is set**, so nothing is handed out behind the teardown
- `close()` destroys whatever is outstanding rather than waiting for it — `release(err)` with a truthy argument makes pg-pool `_remove` the connection instead of returning it to idle (verified in `pg-pool@3.14.0`, `_release` → `_remove` when `err` is truthy), so `end()` has nothing left to wait for

A caller mid-`withExclusiveAccess` sees its connection die. That is the consequence of closing during exclusive work — a caller error either way — and strictly better than hanging.

## Verification

Mutation-verified, and the two halves are **independently necessary**:

- remove `acquire()`'s closed-guard → the in-flight-construction test hangs
- remove the destroy loop → the held-client test hangs

Both fail by **timing out** rather than asserting, which is the honest shape: the bug is a hang, so the test has to be able to hang.

- 4/4 new tests pass; 44/44 across the other Postgres suites unaffected
- `tsc -p packages/core` and `pnpm typecheck:tests` clean

## Post-review update (rebase + smoke-packaged fix)

Rebased onto main (#778). Both conflicts resolved as the review prescribed: `liveClients` now sits alongside main's `embeddingGapWarned` latch (#750) at the declaration site, and this PR's `close()` JSDoc replaces main's "KNOWN LIMITATION (#751)" note (#755). The `acquire()` invariant survives the rebase — main has the same five `pool.connect()` sites this PR converts, and no new ones.

**smoke-packaged root cause fixed:** the review's [90] finding was correct — the failure was this PR's own `close()` JSDoc naming the pg driver's pool package literally. tsup preserves comments in the non-minified dist bundle, so the literal landed in `dist/index.js` and false-positived `scripts/smoke-release.sh`'s inlining grep. The comment now says "the pg driver's pool internals" instead; the smoke grep is untouched. Verified locally: `dist/index.js` has zero matches for the banned literals after build.

Also corrected the [50]-class overstatement in the teardown comments: an in-flight `createPool` transitively awaits `initSchema`, so `close()` during cold construction waits for init to settle (bounded, adapter-owned DDL) — the earlier drain-based attempts hung on caller-held clients, not because of which promise they awaited first. The "attempts that failed" table above should be read with that correction.

Post-rebase verification: `packages/core` default project 147 files passed; `core-pglite` serial project incl. `postgres-close.test.ts` against a live pgvector/pg16 all pass; `pnpm build` clean.

### DIFF
--- a/packages/core/src/storage-postgres.ts
+++ b/packages/core/src/storage-postgres.ts
@@ -253,6 +253,20 @@ export class PostgresAdapter implements StorageAdapter, AsyncPrimaryStore {
   private trigramAvailable: boolean | undefined
   /** One-shot latch for the embedding-gap warning — see ensureVectorIndex. */
   private embeddingGapWarned = false
+  /**
+   * Every client currently checked out of either pool.
+   *
+   * `close()` needs this because `pool.end()` DRAINS: it waits for checked-out
+   * clients to come back. Two sites hold one for a long time — `initSchema`
+   * for its DDL sequence, and `withExclusiveAccess` across arbitrary caller
+   * work — so a `close()` racing either of them waited forever. Measured: both
+   * a drain-based teardown and one that awaited `initPromise` hung the process
+   * rather than closing it.
+   *
+   * With the set, `close()` destroys outstanding clients instead of waiting for
+   * them, and `end()` then resolves immediately.
+   */
+  private readonly liveClients = new Set<any>()
   private readonly vectorDim: number
   private readonly precision: VectorElementFormat | undefined
   private readonly indexMode: PostgresVectorIndexMode
@@ -475,7 +489,7 @@ export class PostgresAdapter implements StorageAdapter, AsyncPrimaryStore {
    * for a second one deadlocks whenever `maxConnections` is 1.
    */
   private async initSchema(): Promise<void> {
-    const client = await this.pool.connect()
+    const client = await this.acquire(this.pool)
     const key = `plur:init:${this.schema}`
     let locked = false
     try {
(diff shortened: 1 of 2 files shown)


plur9:
CI failing on smoke-packaged — the pool/close fix needs this green before it can land. The other checks all pass. Worth investigating the smoke failure; this is otherwise a clean fix for a real leak.

plur9:
**Review (cto-heartbeat):** smoke-packaged CI failure is blocking this — all other checks green. The PR itself is a solid fix (acquire closed-guard + destroy loop both independently necessary, mutation-verified). Needs the smoke failure diagnosed before it can land.

plur9:
**Smoke-packaged CI failure: `pg` is being inlined into core's dist bundle.**

The packaged smoke test reports:

```
FAIL  pg is INLINED into core's dist (the published PostgresAdapter would break)
```

All unit tests pass (source-level), but the built/packaged artifact bundles `pg` rather than treating it as an external. The published package would shadow the host's `pg` installation and break consumers who use a different `pg` version or rely on connection pooling shared across packages.

**Fix:** Add `pg` to the `external` array in the tsup config (or wherever the bundler externals are defined for `@plur-ai/core`). Similar to how other peer deps are typically handled. Then re-run `pnpm run build && node scripts/smoke-packaged.mjs` locally to confirm.

Other checks (unit tests ×3 node versions, Python SDK) all pass. The fix is targeted.

plur9:
**CI failure diagnosis (smoke-packaged)**

The failing step is:
```
FAIL  pg is INLINED into core's dist (the published PostgresAdapter would break)
```

The `close()` rewrite introduces `acquire()`, which wraps pool clients and patches `client.release`. The most likely cause: tsup now picks up a direct or indirect import path from `pg` that it previously tree-shook — typically because the `liveClients` set or the monkey-patched `release` forces a stronger reference.

**To diagnose:** compare `packages/core/dist/index.js` size (or run `node -e "require('./packages/core/dist/index.js')" && pnpm dlx du`) between main and this branch. Or check tsup's bundle analysis for the pg import chain.

**To fix:** confirm `pg` is in `tsup.config.ts` under `external` (it should already be there). If it is, the issue is that `acquire()`'s `any` types or the client-tracking logic is pulling in a pg type at build-time that tsup can't exclude. Consider importing the pg `PoolClient` type only as `import type` rather than a runtime reference, and verify the bundled output is clean before re-pushing.

plur9:
CI: smoke-packaged FAILURE. Unit tests on Node 20/22/24 all pass, but the packaged smoke test failed. Needs fix before this is merge-ready.

plur9:
Miles review (pr-review cadence): CI has a `smoke-packaged` failure. Holding review until that's green — everything else passes (test 20/22/24, Python SDK). Please check the smoke test output and push a fix.

plur9:
### Code review

**[90] The `smoke-packaged` failure is caused by this PR's own JSDoc — the literal string `pg-pool` in a comment trips the inlining grep. Nothing is actually inlined; the earlier "tsup picks up an import path" diagnosis in this thread is wrong.**

The check is `scripts/smoke-release.sh:70`: it greps core's built `dist/index.js` for `pg-pool|pg-protocol|pg-connection-string`. This PR introduces the only occurrence of that string in the file, inside `close()`'s doc comment:

https://github.com/plur-ai/plur/blob/f2c0b022fa106c86acaec12b81827699d40f1b45/packages/core/src/storage-postgres.ts#L727

tsup preserves source comments in the (non-minified) dist bundle, so the doc comment lands in the published artifact and the grep false-positives. Verification: `pg-pool` appears nowhere in `storage-postgres.ts` on main; the diff contains zero import/require or tsup-config changes; unit tests pass on all Node versions; and the identical check is green on #774 and #765.

Fix options: reword the comment (e.g. "makes pg's pool `_remove` the connection rather than return it to idle") — or tighten the smoke grep to module-resolution forms (`require("pg-pool")` / `from "pg-pool"` / `import("pg-pool")`) so documentation may name the package without failing the release gate. The first is a one-word change on this branch; the second is arguably the more durable fix but belongs in its own PR.

**[80] Rebase needed — two conflicts in `storage-postgres.ts`, both mechanical.**

1. Field declaration site: main added the `embeddingGapWarned` latch (#750) at the same lines where this PR declares `liveClients` — keep both.
2. `close()` JSDoc: main (#755) now documents this exact gap as "KNOWN LIMITATION (#751)" pointing at this PR — take this PR's replacement JSDoc, which retires that note.

The `acquire()` invariant ("every `connect()` in this class goes through here") survives the rebase: main has the same five `pool.connect()` sites this PR converts (`initSchema`, `updateMany`, `save`, `withExclusiveAccess`, `searchVector`) and no new ones.

Issue #587, opened 2026-07-15
plur login --status: show enterprise token validity, not raw config


crtahlin:
Follow-up from #532 (author-listed, out of scope for that PR).

**Current behaviour:** `plur login --status` prints the raw `~/.plur/config.json` contents.

**Desired:** surface whether the stored enterprise token is actually valid — e.g. call `GET /api/v1/me` with the stored `Bearer` token and report authenticated / expired / no-token, plus the host and (optionally) token expiry. Avoid printing the token value itself.

Scope: `packages/cli/src/commands/login.ts` (the `--status` branch). Add a unit test with a stubbed `/api/v1/me` (valid + 401 cases), HOME-isolated via the existing `baseDir` param.

Context: #532 wired the `plur login` device-flow command; this improves the companion status readout. Not blocking.

crtahlin:
### Status verification — partially done

Done:
- `plur login --status` command wired in [plur-ai/plur#532](https://github.com/plur-ai/plur/pull/532): reads `~/.plur/config.json` and surfaces host, username, `authed_at`, and scopes — no longer prints the raw config object
- `fetchMe()` helper exists in `packages/cli/src/commands/login.ts` (calls `GET /api/v1/me`) and is used during the login flow to capture username and scopes at token-mint time

Still open:
- The `--status` branch itself does **not** call `GET /api/v1/me` to verify token validity at query time — it only reads the stored config. A stored token can be expired and `--status` still reports "Logged in". The issue asks for a live probe that reports `authenticated / expired / no-token`
- No unit tests for the `--status` path with stubbed `/api/v1/me` valid + 401 cases (as specified in the acceptance criteria)

Leaving open.

PR #759, opened 2026-07-28
feat(core): export the engine primitives, and fix a three-way drift in feedback

plur9:
## The bug

`Plur.feedback` had the feedback rule written out three times — primary store, secondary stores, packs — and the three had drifted. **Only the primary copy promoted `commitment`.**

```
$ grep -n "retrieval_strength + 0.05" packages/core/src/index.ts
3393:  ...  ← promotes commitment
3450:  ...  ← does not
4153:  ...  ← does not
```

So the same engram, given the same thumbs-up, advances `leaning → decided` in your own store and stays at `leaning` in a team store or an installed pack. Nothing throws. The engram just quietly means something different depending on where it happens to live.

Three copies of a rule is how that happens, so there is now one: `applyFeedbackSignal`, called from all three sites.

## The commitment ladder, stated once

`nextCommitment` puts the whole ladder in one place, including the cases that were previously implicit:

- **`undefined` seeds at `leaning`.** An engram that has received a positive signal has demonstrably been retrieved and found useful, which is more than `exploring` claims. Leaving it unset let an older engram accumulate unlimited positive signal while still reading as though nobody had an opinion about it. **This is a behaviour change**, and the one intentional one in this PR.
- **`decided` is the ceiling.** Reaching `locked` requires explicit human intent; no quantity of feedback substitutes for it.
- **Anything unrecognised is returned untouched.** Load-bearing rather than defensive: a deployment may extend the enum, and `commitment: 'draft'` stages an engram in a review queue (the extension point is already noted in `schemas/engram.ts`). Silently promoting a draft on a thumbs-up would publish unreviewed content.

## Why these are exported

`rrfMergeEngrams` already existed in `hybrid-search.ts` but was not reachable from the package, so a consumer wanting the same k=60 ranking had to write it again. Same for the feedback arithmetic.

A reimplementation is correct on the day it is written and has no way to announce when it stops being correct. It keeps returning a plausible ordering and a plausible strength; the two implementations simply stop agreeing, and the symptom shows up much later as "the benchmark numbers don't describe this deployment." This repo just found three instances of exactly that internally. There is no reason to let it happen across a package boundary as well.

`applyFeedbackSignal` is a pure mutation — no I/O, no locking, no history — so it is usable without also inheriting the file-backed single-user machinery that surrounds it in `Plur`.

## Testing

- 12 new unit tests in `packages/core/test/feedback.test.ts`, several asserting arithmetic that looks too simple to test. That is deliberate: the bug was never that the arithmetic was hard, it was that there were three of it.
- Full core suite: **153 files, 2325 tests, 0 failures.**

One note on flakiness spotted while verifying, unrelated to this change: `test/secrets.test.ts` → `truncation at the ceiling produces valid UTF-8` timed out at 30 s during one full-suite run and passed in isolation and on a re-run. It is a 1 MiB scan competing with 164 other files for CPU. Worth a longer timeout at some point; not touched here.

## Downstream

This unblocks a second deployment consuming the same primitives instead of maintaining transcriptions of them. Nothing in this PR depends on that.

### DIFF
--- a/packages/core/src/feedback.ts
+++ b/packages/core/src/feedback.ts
@@ -1,0 +1,104 @@
+/**
+ * What a feedback signal does to an engram.
+ *
+ * This was written out by hand in three places in `Plur.feedback` — once for
+ * the primary store, once for secondary stores, once for packs — and the three
+ * had already drifted: **only the primary copy promoted `commitment`**. The
+ * same engram, given the same positive signal, advanced from `leaning` to
+ * `decided` in your own store and stayed at `leaning` in a team store or an
+ * installed pack. Nothing failed; the engram just quietly meant something
+ * different depending on where it was kept.
+ *
+ * Three copies of a rule is how that happens, so there is now one. It is a pure
+ * mutation over an engram — no I/O, no locking, no history — which is what lets
+ * a server-side deployment reuse it without also inheriting the file-backed
+ * single-user machinery that surrounds it in `Plur`.
+ */
+import type { Engram } from './schemas/engram.js'
+
+export type FeedbackSignal = 'positive' | 'negative' | 'neutral'
+
+/** Strength added on a positive signal. Capped at 1.0. */
+export const POSITIVE_STRENGTH_DELTA = 0.05
+/** Strength removed on a negative signal. Floored at 0.0. */
+export const NEGATIVE_STRENGTH_DELTA = 0.1
+
+/**
+ * Next commitment level after a positive signal.
+ *
+ * Advances one step toward `decided` and never further: reaching `locked`
+ * requires explicit human intent, and no amount of accumulated feedback should
+ * substitute for it.
+ *
+ * `undefined` seeds at `leaning` rather than staying unset. An engram that has
+ * received a positive signal has demonstrably been retrieved and found useful,
+ * which is more than `exploring` claims and exactly what `leaning` means;
+ * leaving it unset would let an older engram accrue unlimited positive signal
+ * while still reading as though nobody had an opinion about it.
+ *
+ * Anything unrecognised is returned untouched. That is load-bearing rather than
+ * defensive: a deployment may extend the enum — `commitment: 'draft'` stages an
+ * engram in a review queue (see the extension note in `schemas/engram.ts`) — and
+ * silently promoting a draft out of review on a thumbs-up would publish
+ * unreviewed content. Unknown means "not mine to advance".
+ */
+export function nextCommitment(current: string | undefined): string | undefined {
+  switch (current) {
+    case undefined:    return 'leaning'
+    case 'exploring':  return 'leaning'
+    case 'leaning':    return 'decided'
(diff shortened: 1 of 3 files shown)


plur9:
Assigning for review — this is the gate on the downstream work.

Sequence:
1. this PR merges
2. a core release carries `rrfMergeEngrams` + `applyFeedbackSignal`
3. plur-ai/enterprise#587 rebases onto it and drops its duplicated copies

Nothing here depends on step 3; the three-way `commitment` drift is a bug in this repo on its own.

The one intentional behaviour change is `nextCommitment(undefined) → 'leaning'` — worth a specific look, since it is the only case where an existing store's engrams will start acquiring a commitment they did not have before.

plur9:
CI passing. feedback.ts extraction is exactly right — three copies of the same mutation is how drift happens, and the commit message documents both the drift that already existed (commitment promotion only on primary) and why the fix is a module rather than inline. Injectable today parameter makes tests clean. The re-anchoring logic and the 'unknown commitment = don't touch it' invariant are both well-justified. Ready for merge.

plur9:
**Review (cto-heartbeat):** CI green across all checks. Clean extraction — feedback drift across primary/secondary/pack stores is a real bug, centralising into a pure module is the right fix. Commitment advancement and last_accessed re-anchor both handled consistently. 

Note: reviewer account token expired; @plur9 should self-merge once branch is rebased current.

plur9 (commented):
**Review — ready for merge (needs non-author approval)**

The three-way drift bug is real and the root cause analysis is correct — three copies of a mutation rule, each slightly different, with no test surface to catch when they diverge. Extracting to `applyFeedbackSignal` is the right fix.

**Correctness:**
- `nextCommitment`: case table is complete. `undefined → leaning` on first positive signal is correct — an engram that was retrieved and found useful has earned `leaning`; leaving it unset let unlimited positive signal accumulate with no semantic anchoring. The `default: return current` branch is load-bearing (not merely defensive) for future enum extensions like `draft`.
- `applyFeedbackSignal`: pure mutation, no I/O. Injectable `today` param makes tests clock-independent. The `last_accessed` re-anchor on non-neutral signals is correct — decay is measured from that anchor, so bumping strength without advancing it lets elapsed-time decay immediately swallow the adjustment. Excluding neutral is right: no strength moved, nothing to protect.
- Exporting `rrfMergeEngrams` and the feedback primitives is the right API surface. Enterprise #587 depends on these directly; confirmed blocked on this PR landing.

**Tests:** 12 unit tests covering arithmetic that looks trivial — the PR explains why that is the correct choice. The full suite (2325 tests, 0 failures) confirms no regressions.

**Minor follow-up:** `test/secrets.test.ts` flakiness (30s timeout on a 1 MiB scan competing with 164 files) is worth a dedicated issue for a longer per-test timeout. Not a blocker.

CI: all 5 checks green (CI ×3 node versions, Python SDK, smoke-packaged). @plur9 please approve and merge.

Issue #761, opened 2026-07-29
MCP tool-surface consolidation under plur_admin is invisible to clients — agents conclude tools are 'unavailable'


crtahlin:
## Problem

As of CLI **0.16.x**, plur's MCP consolidated most tools under a single dispatch tool `plur_admin(action=<name>, args={...})` (to stay under the ~40-tool-per-workspace cap). Standalone tools are now only: `plur_status`, `plur_learn`, `plur_recall`, `plur_forget`, `plur_feedback`, `plur_doctor`, `plur_receipt`, `plur_session_start/end`, `plur_tensions_purge`. **Everything else** — `plur_recall_hybrid`, `plur_stores_list`, `plur_suggest_scope`, `plur_scopes_discover`, `plur_sync_status`, `plur_timeline`, `plur_pin`, `plur_promote`, `plur_packs_*`, etc. — is now an **action under `plur_admin`**.

This reshaping is **invisible to clients**:
- An agent (or its memory / prior context) referencing an old standalone tool name gets a tool-lookup miss.
- With no in-band signal that the surface changed, the natural — and wrong — conclusion is "the tool / the MCP is unavailable," so the agent degrades or falls back.
- **Observed 2026-07-29:** during a routine plur smoke check, a name-lookup miss on `plur_stores_list` / `plur_recall_hybrid` / `plur_suggest_scope` led to a false "MCP is down" conclusion and a full switch to the HTTP-API fallback — while the MCP was **fully healthy the whole time** (`plur_doctor` green, both remote stores reachable + auth valid).

## Who it affects

**All** plur MCP clients (Claude Code, Cursor, Claude Desktop) — this is a property of the MCP server, not any one client framework. Users without extra tooling/memory to carry the correction are the most exposed: the agent has to *notice* the `plur_admin` dispatcher on its own, which does not reliably happen.

## Possible fixes (menu — not prescribing; a couple could ship together)

1. **Backward-compat aliases (highest leverage).** Keep the old standalone tool names as thin alias/deprecation shims for a version or two so existing agents/memory don't hard-miss; each alias emits a deprecation note pointing at `plur_admin`.
2. **Self-describing inventory.** Have `plur_doctor`/`plur_status` emit the current available-tool + `plur_admin` action list, so an agent can query the live surface instead of guessing names.
3. **Helpful `plur_admin` errors.** On an unknown/removed action, return `valid actions: [...]` so an agent recovers in one step.
[...]

Issue #625, opened 2026-07-18
MCP: lean tool profile by default — 40-tool surface (~9K tokens) is re-sent every turn to every consumer, ~74% unused


plur9:
## Summary

The MCP server ships **all 40 `plur_*` tools on every `tools/list`** to every consumer except Cursor. That tool block is **~9,000 tokens** and is re-sent on **every model turn** of every agent loop. The ~90% of sessions that only `recall`/`learn`/`session_*` use 6–10 of those tools, so **~74% of the tool-schema tokens are dead weight, re-billed turn after turn, for every connected agent** — Claude Code, Hermes/Winston, OpenClaw, plur-claw, nightshift agents, and CI harnesses alike.

This surfaced while investigating OpenRouter spend on the cos-server CoS agent (Winston), but it is **not CoS-specific** — it is a tax on the whole PLUR MCP consumer base.

The good news: **the mechanism to fix it already exists and is tested** — the `cursor` profile (`CURSOR_CORE_TOOL_NAMES` + the `plur_admin` dispatch tool) already collapses the surface to 11 tools. It is just gated to Cursor only. This issue proposes making a lean profile the default.

## Evidence (measured, this repo, `main`)

Measured by serializing `getToolDefinitions(profile)` into the `{name, description, inputSchema}` shape the MCP `ListTools` handler sends (`packages/mcp/src/server.ts:201`), two independent methods bracketing the same range:

| Profile | Tools | Tool-block size | Per turn |
|---|---|---|---|
| `full` (default for everyone but Cursor) | 40 | **~9,000 tokens** (4,560 description + ~4,360 minified JSON-schema) | re-sent every turn |
| `cursor` (10 core + `plur_admin`) | 11 | **~2,000–2,300 tokens** | re-sent every turn |
| **Reduction** | | **~6,700 tokens / turn (~74%)** | |

Heaviest individual tools (wire chars): `plur_learn_batch`, `plur_learn`, `plur_recall_hybrid`, `plur_ingest`, `plur_packs_*` — the ones carrying the full repeated engram-object schema and multi-paragraph descriptions (e.g. `plur_learn_batch`'s description alone is ~1,200 chars).

## Who pays

Every process that connects to `@plur-ai/mcp` and runs a multi-turn loop pays ~9K tokens of tool schema **per turn**:

- **Winston** (cos-server Hermes) — 3 cron agent runs/day, `max_turns=60` each
- **Tris, Miles** (agent fleet), **plur-claw** (Telegram)
[...]

PR #763, opened 2026-07-29
feat(mcp): plur_doctor reports the live tool surface

plur9:
Addresses #761 — option 2 (self-describing inventory), which composes with option 4 (name-stable diagnostic door).

## Why this one

I checked the five proposed fixes against the code before implementing. Two are already shipped:

- **Option 3 (helpful `plur_admin` errors)** — already there. An unknown action returns `Unknown action "x". Valid actions: ...`, and `plur_admin`'s *description* enumerates every valid action. There's even a deliberate comment explaining why `enum` is omitted from the schema, so an invalid action reaches that handler instead of failing generic validation.
- **Option 5 (migration signal)** — the 0.15.0 CHANGELOG flags it prominently: *"Lean tool profile is now the default (#625, #694) … down from 40 … All 40 tools remain reachable via plur_admin"*, including a bolded note for consumers depending on all 40.

**Option 1 (backward-compat aliases) I'd argue against.** Measured: 41 tool definitions, 12 standalone under lean, **30 under `plur_admin`**. Re-exposing those 30 as aliases puts the surface back at ~41 — which is the ~40-per-workspace cap the consolidation exists to stay under. It trades this bug for the one that caused it.

That leaves option 2 as the real gap, and the incident shows exactly why it matters: **`plur_doctor` was green and the agent still concluded the MCP was down.** Doctor answers *"is the engine healthy"*; the actual question was *"what is callable"*. Its own description even says it does not check the live tool count.

## What changes

`plur_doctor` now returns `tool_surface`:

```json
{
  "profile": "lean",
  "standalone": ["plur_admin", "plur_doctor", "plur_learn", ...],
  "admin_actions": ["plur_recall_hybrid", "plur_stores_list", "plur_suggest_scope", ...],
  "note": "Only the 12 tools in \"standalone\" are callable by name. Everything in \"admin_actions\" is reachable as plur_admin { action: \"<name>\", args: {...} } — same arguments, same validation, same result. A name-lookup miss on one of those means it moved, NOT that the MCP is unavailable. Set PLUR_TOOL_PROFILE=full to expose all tools directly."
}
```

Doctor is the right home: it's where an agent already goes when it thinks the server is broken, and it stays standalone in **every** profile — a surface description that could itself be consolidated would be useless.

`resolveToolProfile()` is now shared with `runStdio` instead of the ternary being written twice. Two copies is how doctor comes to confidently describe a profile nobody is running.

## Tests

The risk with a self-description is that it drifts and becomes a confidently wrong answer — worse than none, because it's the thing an agent would trust. So the tests assert it *can't*:

- `standalone` must equal `getToolDefinitions(profile)` exactly, for all three profiles
- every advertised action must exist **and** not be one of the standalone names
- the two lists together must cover every tool — nothing undiscoverable
- `plur_doctor`/`plur_status` must stay standalone in every profile
- the incident's specific names (`plur_recall_hybrid`, `plur_stores_list`, `plur_suggest_scope`) are asserted present

**Mutation-tested**: claiming an unexposed tool fails 2 tests; weakening the note's "moved, NOT unavailable" wording fails 1.

297 mcp tests, 2325 core tests, `typecheck` + `typecheck:tests` clean.

## Left for triage

Option 5's release-process half — a standing rule that a tool-surface change must be called out in release notes — is a process change rather than code, and #761 suggests splitting it. Happy to file that separately if you want it tracked.

### DIFF
--- a/packages/mcp/src/server.ts
+++ b/packages/mcp/src/server.ts
@@ -4,7 +4,7 @@ import { existsSync, readFileSync, writeFileSync } from 'fs'
 import { join } from 'path'
 import { homedir } from 'os'
 import { Plur, checkForUpdate } from '@plur-ai/core'
-import { getToolDefinitions, mcpCanary, validateToolArgs, CURSOR_CORE_TOOL_NAMES, type ToolProfile } from './tools.js'
+import { getToolDefinitions, mcpCanary, validateToolArgs, CURSOR_CORE_TOOL_NAMES, type ToolProfile, resolveToolProfile } from './tools.js'
 import { registerFlushOnExit } from './telemetry.js'
 import { VERSION } from './version.js'
 
@@ -420,8 +420,10 @@ Please:
 }
 
 export async function runStdio(): Promise<void> {
-  const envProfile = process.env.PLUR_TOOL_PROFILE
-  const profile: ToolProfile = envProfile === 'full' ? 'full' : envProfile === 'cursor' ? 'cursor' : 'lean'
+  // Shared with describeToolSurface (#761) so the surface plur_doctor reports
+  // is resolved from the same rule the server exposes tools with — two copies
+  // of this ternary is how doctor comes to describe a profile nobody is running.
+  const profile: ToolProfile = resolveToolProfile()
   const server = await createServer(undefined, { profile })
   // Opt-in, content-free telemetry: ship any pending daily counter snapshot on
   // process exit (best-effort). Self-gates on telemetry opt-in — an opted-out
--- a/packages/mcp/src/server.ts
+++ b/packages/mcp/src/server.ts
@@ -4,7 +4,7 @@ import { existsSync, readFileSync, writeFileSync } from 'fs'
 import { join } from 'path'
 import { homedir } from 'os'
 import { Plur, checkForUpdate } from '@plur-ai/core'
-import { getToolDefinitions, mcpCanary, validateToolArgs, CURSOR_CORE_TOOL_NAMES, type ToolProfile, resolveToolProfile } from './tools.js'
+import { getToolDefinitions, mcpCanary, validateToolArgs, CURSOR_CORE_TOOL_NAMES, type ToolProfile, resolveToolProfile, setActiveToolProfile } from './tools.js'
 import { registerFlushOnExit } from './telemetry.js'
 import { VERSION } from './version.js'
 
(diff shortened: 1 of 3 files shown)


plur9:
CI passing (all matrix nodes, Python SDK). Self-describing tool surface is the right home for this — doctor is where a confused agent already lands, and the mutation-tested note prevents the description drifting from reality. Ready for human merge.

plur9 (commented):

Fixes the 'doctor green, agent switches to HTTP fallback' split at the right layer. Key design points:
- `describeToolSurface` correctly separates standalone (callable by name) from `admin_actions` (routed via `plur_admin`)
- `activeToolProfile`/`setActiveToolProfile` seam is correct — `createServer` takes profile as an option, so env is not the authority; recording what was actually built with avoids the 'env says lean, server exposes full' mismatch
- The load-bearing test assertion (`surface.standalone === getToolDefinitions(profile).sort()`) means the report can't drift from what `tools/list` actually exposes
- 'name-miss means moved, not gone' note text is exactly what an agent needs to self-correct

plur9:

`resolveToolProfile()` and `setActiveToolProfile()` solve the right problem: the environment is not the authority for what tools the server exposes when `createServer` takes an explicit `profile` option. Recording `activeProfile` in `createServer` and having `plur_doctor` read it means the reported surface is always the actual running surface.

`resolveToolProfile` as a shared function eliminates the duplicated ternary between `runStdio` and wherever else derived the profile — single source of truth.

CI all green (4 checks pass + smoke-packaged and Python SDK). Ready for merge.

plur9:
### Code review

1. **[85] Rebase required — GitHub reports the PR as CONFLICTING against `main`.** #778 (merged after this branch was cut from v0.16.1) inserted the #776 A4' per-host remote-degradation block at the same point in the `plur_doctor` handler where this PR inserts `tool_surface` (https://github.com/plur-ai/plur/blob/c3bfcfa8ca7f1a8f9e10a190451ee4347b5bfd87/packages/mcp/src/tools.ts#L2020-L2026), and also changed the `@plur-ai/core` import line at the top of `tools.ts`. The two additions are semantically compatible — both add independent fields to doctor's return — so the rebase is mechanical: keep both blocks. `server.ts` and the new test file merge cleanly. The surface math itself checks out against current `main`: `admin_actions` mirrors `buildAdminDispatchTool`'s action list exactly, and lean/cursor are identical profiles there, so `describeToolSurface` stays truthful for all three.

PR #765, opened 2026-07-29
fix(cli): plur doctor probes the configured MCP server, not a synthesised one

### LINKED ISSUE #764: plur doctor probes a synthesized npx command, not the configured MCP server
## Problem

`plur doctor`'s handshake check does not probe the MCP server the user has configured. It synthesizes its own launch command via `buildMcpServerEntry()`:

```
~/.plur/bin/plur-mcp        (if `plur init` created the shim)
  ↓ otherwise
/bin/sh -lc 'exec npx -y @plur-ai/mcp@latest'
```

So on any install that launches the server another way — a direct `node /path/to/dist/index.js`, a pinned version, a custom `env` block — doctor reports on a path the user does not use.

## Both failure directions, observed

**False fail.** On a machine whose `.mcp.json` runs the server from a local build, doctor reported:

```
handshake : { ok: false, error: "timeout after 20000ms — server did not respond to initialize" }
overall   : fail
```

while the configured server handshakes in ~2s and returns 40 tools. Cause: the cold `npx -y @plur-ai/mcp@latest` fallback is genuinely slower than the timeout — measured **25.2s** on this machine against a **20s** budget.

**False pass — the dangerous one.** The same indirection means a *broken* configured server reports healthy, as long as npx can fetch a working package. That is exactly the failure that occurred here this morning: a stale local `dist/` referencing a dependency removed in the v2 SDK migration, crashing with `ERR_MODULE_NOT_FOUND` before the MCP handshake. `plur doctor` could not have caught it — it would have probed npx instead and passed.

A diagnostic that cannot observe the thing it is diagnosing is worse than no diagnostic, because it gets trusted.

## Suggested fix

1. **Probe the configured entry.** doctor already parses each config file to set `hasPlurMcp` — read the actual `mcpServers.plur` entry (`command`, `args`, `env`) from the highest-priority config that has one, and spawn *that*.
2. **Report which entry was probed**, so the output is unambiguous about what was tested.
3. **Fall back to the synthesized entry only when no config declares one**, and label it as such.
4. **Raise the fallback timeout** above the real cold-npx cost, or warm the package first. 20s is under the measured 25.2s.

Worth keeping in mind that (1) also makes doctor useful for the multi-config case it already enumerates — Claude Code project, `.mcp.json`, global, Claude Desktop, Cursor — where different entries can point at different servers.

plur9:
Closes #764.

## The bug

doctor's handshake never launched the server the user runs. It rebuilt an entry from scratch:

```
~/.plur/bin/plur-mcp          (if `plur init` made the shim)
  ↓ else
/bin/sh -lc 'exec npx -y @plur-ai/mcp@latest'
```

So any install starting the server another way — a direct `node /path/dist/index.js`, a pinned version, a custom `env` — got diagnosed on a path it never uses.

**False fail.** On a machine whose `.mcp.json` runs a local build, doctor reported `timeout after 20000ms` / `overall: fail` while the configured server handshakes in ~2s with 40 tools. The cold npx fallback measured **25.2s** against a **20s** budget.

**False pass — the one that matters.** A configured server that crashes on startup reports *healthy*, because npx fetched a good copy of a package the user isn't running. That is exactly the break this came from: a stale `dist/` importing a dependency removed in the SDK v2 migration, dying with `ERR_MODULE_NOT_FOUND` before the handshake. doctor could not have seen it.

A diagnostic that can't observe the thing it diagnoses is worse than none, because it gets trusted.

## The fix

doctor reads the `plur` entry from the first config declaring one (in `knownConfigFiles()` order) and launches **that**, carrying the entry's own `env` — a config pinning `PLUR_TOOL_PROFILE=full` yields 40 tools, and probing without it reports a count the user never sees. The synthesised entry stays as the fallback for an unwired install, where *"would the recommended entry work"* is the useful question.

Every result now carries `probed` and `command`. A verdict with no subject is how a green result on the wrong server gets believed.

`readPlurMcpEntry()` returns `null` for a malformed entry rather than inventing a command — falling back to the recommendation is honest, guessing is not.

## Verified end-to-end

Against a **broken** configured server (reproducing the real crash):
```
overall : fail
probed  : Claude Code (.mcp.json)
command : node /tmp/doctor-764/broken-server.js
error   : server exited (code=1) before responding
```

Against the **working** one:
```
probed        : Claude Code (.mcp.json)
ok            : True
serverVersion : 0.16.1
toolCount     : 40      ← proves the config's env was honoured
```

## On the tests — worth reading

The existing handshake test pointed its config at the **same shim** the old code preferred, so both the old and new probe spawned the identical script. It passed either way and guarded nothing. My first attempt at a mutation test confirmed it: reverting the fix changed nothing.

My *second* attempt also passed — because the test spawns the **built** CLI and I had only mutated `src`. Stale build, which is the same trap this PR is about.

With a rebuild and a genuinely discriminating test — shim is one script, config names another — the mutation fails on exactly the assertion that should catch it, and passes when restored.

402 cli tests, `typecheck` + `typecheck:tests` clean.

## Not changed

The 20s timeout. With the configured entry it's ample (~2s); it only bit on the cold-npx fallback, which is now the uncommon path. Raising it would slow the honest failure case for everyone. Happy to revisit if the fallback proves slow in practice.

### DIFF
--- a/packages/cli/src/commands/doctor.ts
+++ b/packages/cli/src/commands/doctor.ts
@@ -7,6 +7,8 @@ import { outputText, outputJson, shouldOutputJson } from '../output.js'
 import {
   type ConfigFile,
   buildMcpServerEntry,
+  readPlurMcpEntry,
+  type McpServerEntry,
   cursorProjectMcpConfigPath,
   hasDatacoreMcp,
   hasPlurMcp,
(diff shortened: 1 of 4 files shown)


plur9:
CI passing across all matrix nodes (Node 20/22/24), Python SDK, and smoke tests. Fix is correct — reads the configured MCP entry and probes that rather than a synthesised path. Ready for human merge.

plur9 (commented):

The core fix (probe what's configured, not what you'd recommend) is correct. `resolveProbeTarget` reads config files in `knownConfigFiles()` order, which means the probe matches the actual launch path. Carrying `entry.env` through is important — PLUR_TOOL_PROFILE changes tool count, so probing without it reports a surface the user never sees.

The new test is properly discriminating: shim and configured script are different paths, so only a config-reading probe (not a synthesized one) passes. Previous test was vacuously true since both paths spawned the same script.

plur9:

`resolveProbeTarget()` correctly prefers whatever the user's config file actually declares, falling back to the synthesised recommendation only when nothing declares one. The comment explaining the two failure modes of the old approach (working local build reporting timeout because cold npx is slow; crashing server reporting healthy because npx pulled a working copy) makes the design intent clear.

Including `entry.env` in `spawnEnv` is the right call — a config pinning `PLUR_TOOL_PROFILE=full` should probe a full-profile server, not a lean one.

CI all green. Ready for merge.

plur9:
### Code review

No issues found.

Issue #772, opened 2026-07-31
plur_learn: intermittent whole-payload drop on v0.16.1 (received_fields:[]) — not deterministically the tags array (refines #297)


crtahlin:
## Summary

Follow-up to #297 (closed, `upstream-bug`). On `@plur-ai/mcp` **v0.16.1** the same failure signature recurs — the **entire arguments object arrives empty** (server reports `received_fields: []` → `"Invalid arguments: statement: Required"`) — but new evidence **contradicts #297's conclusion** that a `tags` array is the deterministic trigger. The drop is **intermittent** and not tied to array-typed parameters at all.

## Environment

- `@plur-ai/mcp` **v0.16.1** (`plur_status` → `version`)
- Installed globally (`/opt/homebrew/bin/plur-mcp`)
- MCP client: Claude Code
- `plur_doctor` / `plur_status` **all green**: BGE-small embedder loaded, hybrid search operational, both remote stores reachable + auth valid, outbox `0`. The engine is healthy throughout — this is purely a client→server call-delivery problem.

## Evidence (single session, 2026-07-31)

| Call shape | Result |
|------------|--------|
| Single call, **scalar-only** (`statement`/`type`/`domain`/`scope`/`rationale`) | ✅ `ADD` (succeeded multiple times) |
| Single call, **with `tags` string-array** | ✅ `ADD` |
| Single call, scalar-only | ❌ `received_fields: []` (also occurred) |
| **Two `plur_learn` calls in one client message (parallel)** | ❌ **both dropped** |
| Retry of an identical previously-dropped call | ✅ `ADD` |

## What this changes vs #297

- A `tags` string-array call **succeeded** → an array param is **not** a deterministic trigger on v0.16.1.
- Scalar-only calls **also dropped** on occasion → the failure is **not specific to arrays**.
- Strongest correlation observed: **multiple `plur_learn` tool_use blocks in a single client message** → all dropped together.
- **Retrying the identical call resolves it** → the drop is transient, not payload-content-dependent.

## Interpretation

Consistent with #297's finding that this is a **client-side serialization/marshalling drop** (upstream, not the plur server — the server accepts these payloads, per the wire-protocol tests in #297 / PR #301). But the **trigger is not the array type**; it presents as an intermittent client issue, notably aggravated by batching multiple `plur_learn` calls into one message.

## Impact

**Low.** Failures are loud (immediate error, no silent data loss — nothing gets written wrong) and clear on retry. Only friction is during multi-write sessions.

[...]

PR #773, opened 2026-07-31
feat(mcp,core): plur_learn_batch — persist many engrams in one MCP call (#281)

### LINKED ISSUE #281: plur_learn: misleading 'statement: Required' error causes silent memory loss in multi-agent orchestration

plur9:
## Summary

Adds `plur_learn_batch` — a new MCP tool that accepts an array of engram objects and persists them in a single call. This directly addresses the pattern identified in #772 as the strongest correlate of client-side payload drops: **multiple `plur_learn` tool_use blocks in one message**.

- One `plur_learn_batch` call replaces N parallel `plur_learn` calls, eliminating the batching-aggravated drop scenario entirely
- Same dedup + policy as single learn (exact-hash NOOP, semantic similarity, LLM decisions) — applied per item AND across the batch
- **Partial-failure tolerant**: a bad statement is caught and recorded in `failures[]`, never aborts the rest
- Returns `ids[]`, per-item `results[]` (with `decision` + `existing_id` on dedup), `stats`, and `failures[]`
- Limitation: batch writes land in local store only (no outbox/remote routing) — documented in tool description and CHANGELOG

## Core changes

- `learnBatch` gains partial-failure tolerance: `LearnBatchResult.failures` + `stats.failed`
- `LearnBatchFailure` type added to `types.ts`

## Tests

- 5 new core unit tests (`packages/core/test/batch.test.ts`) covering: multi-write, intra-batch dedup, partial-failure, error propagation, LLM-call cap
- 4 new MCP integration tests: multi-write + recall, intra-batch dedup, partial-failure, empty-array rejection
- All 46 tests pass

## Related

- Closes #281
- Mitigates #772 (eliminates the multi-`plur_learn`-per-message pattern)
- Builds on the array-arg hardening from #297 / PR #474

## Test plan

- [ ] `pnpm vitest run packages/core/test/batch.test.ts` — 5 pass
- [ ] `pnpm vitest run packages/mcp/test/tools.test.ts` — 41 pass
- [ ] Manual: call `plur_learn_batch` with 3 engrams, verify `ids[]` length = 3
- [ ] Manual: duplicate statement in batch → `stats.noops = 1`
- [ ] Manual: empty statement in batch → `failures[0].index` correct, survivors written

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,17 @@
 # Changelog
 
+## Unreleased
+
+### Feature: `plur_learn_batch` — persist many engrams in one MCP call (#281)
+
+New MCP tool `plur_learn_batch` exposes the existing core `learnBatch` over the Model Context Protocol. Built for orchestration: when a run fans out to several subagents and the parent session needs to record their consolidated findings, one `plur_learn_batch` replaces N `plur_learn` round-trips.
+
+- **Same dedup + policy as single learn.** Each item runs the full `learnAsync` path — exact-hash NOOP, semantic similarity, and LLM ADD/UPDATE/MERGE/NOOP decisions — plus the sensitive-content scope guard. Dedup applies *across* the batch too: a statement written earlier in the same call is visible to a later identical statement (NOOP, not a second ADD).
+- **Partial-failure tolerant.** `learnBatch` no longer aborts on the first bad statement. A statement that throws (empty text, secret-in-statement, malformed validity window) is caught, recorded in a new `failures[]` array (`{ index, statement, error }`), and counted in `stats.failed`; the rest of the batch still writes. `LearnBatchResult` gains `stats.failed` and `failures` (additive — existing fields unchanged).
+- **Returns** `ids[]`, a per-item `results[]` (with `decision` + `existing_id` on dedup), `stats`, and `failures[]` when any item failed.
+- **LLM dedup cost is bounded** by `max_llm_calls` (default 50), inherited from `learnBatch`.
+- **Limitation:** unlike `plur_learn`, batch writes do not route to remote/team stores (no `learnRouted`, no outbox, no server-assigned ids) — they land in the local store. For team-scoped knowledge that must reach a shared store, call `plur_learn` per engram. Remote-scope batching is a follow-up.
+
 ## 0.10.0 (2026-06-25)
 
 Security-hardening release, independently audited.
--- a/README.md
+++ b/README.md
@@ -197,6 +197,7 @@ plur.sync('git@github.com:you/plur-memory.git')
 | Tool | What it does |
 |------|-------------|
 | `plur_learn` | Store a correction, preference, or convention |
+| `plur_learn_batch` | Store many memories in one call (batch dedup, partial-failure tolerant) |
 | `plur_recall_hybrid` | Retrieve relevant memories (BM25 + embeddings) |
 | `plur_inject_hybrid` | Select engrams for current task within token budget |
 | `plur_feedback` | Rate relevance (trains quality over time) |
--- a/packages/core/src/learn-async.ts
+++ b/packages/core/src/learn-async.ts
@@ -10,7 +10,7 @@ import { logger } from './logger.js'
 import { withLock } from './sync.js'
 import type { Engram } from './schemas/engram.js'
 import type { SecretMatch } from './secrets.js'
-import type { LearnContext, LearnAsyncContext, LearnAsyncResult, LearnBatchResult, DedupDecision, LlmFunction } from './types.js'
+import type { LearnContext, LearnAsyncContext, LearnAsyncResult, LearnBatchResult, LearnBatchFailure, DedupDecision, LlmFunction } from './types.js'
 
 export interface LearnAsyncDeps {
   /** Content hash dedup against all engrams. Scope-aware: only matches same scope. */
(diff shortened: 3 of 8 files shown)


plur9:
**Review — LGTM, recommend merge.**

Reviewed the core changes:

- **`learnBatch` partial-failure tolerance** — `failures[]` array with per-item catch is the right shape. A bad statement surfaces in `failures[0]` with its index; the rest of the batch proceeds. No silent truncation.
- **Intra-batch dedup** — NOOP on exact duplicate within the same call, visible to later items in the same batch. Correct.
- **MCP tool schema** — `engrams[]` array param, clear description noting local-store-only limitation. The limitation is documented in the tool description (not hidden), which satisfies the install guide neutrality constraint.
- **Tests** — 5 core unit tests + 4 MCP integration tests. The empty-array rejection test (`failures[0].index`) is a good edge case catch.

The local-store-only limitation (no outbox/remote routing) is a reasonable v1 scope and is documented. Worth a follow-up issue for remote routing if the pattern proves useful.

46 tests pass. Closes #281, mitigates #772.

🤖 heartbeat review

plur9:
🤖 heartbeat — PR has a merge conflict.

`plur_learn_batch` PR (#773) is CONFLICTING and CI cannot run until resolved. This is blocking the P2 fix for the parallel plur_learn drop (#772).

The conflict is likely on `packages/core/src/index.ts` or `packages/mcp/src/tools.ts` — the same files touched by the recall limit fix in #774 (which ships first). Rebasing onto main after #774 merges should resolve it.

plur9:
### Code review

1. **[95] Superseded — this feature already shipped on `main`; recommend closing, not rebasing.** `plur_learn_batch` landed via PR #503 on 2026-07-07 (commit b348c9c, same title), #281 was closed 2026-07-11, and PR #572 (2026-07-14) then fixed partial-failure id alignment. This branch was cut 2026-07-03 and re-implements the pre-#572 version: the handler returns a compacted id array built from successes only (https://github.com/plur-ai/plur/blob/986e96282580c15d4f4fe0734462921495380f64/packages/mcp/src/tools.ts#L350-L360), so after a mid-batch failure `ids[i]` no longer corresponds to input `i` — exactly the mis-attribution bug #572 fixed. `main` builds `ids` aligned 1:1 with the input, `null` for failures (https://github.com/plur-ai/plur/blob/1ee5dde70a1f0ea66df8fd593c5d3e43914e074e/packages/mcp/src/tools.ts#L908-L917), and its handler also carries the #671 batch missing-domain nudge, which this version predates. All 8 files conflict (GitHub: CONFLICTING); resolving toward this branch would regress `main`, and resolving toward `main` leaves nothing to merge. The earlier heartbeat diagnosis ("rebase after #774 merges should resolve it") is not accurate — the conflict is with the already-merged #503/#572 implementation.

plur9:
Closing: superseded — plur_learn_batch already shipped to main via #503 (2026-07-07) with the partial-failure id-alignment fix in #572. This branch re-implements the pre-#572 version and conflicts across all 8 files. Decision (2026-08-01): keep both plur_learn and plur_learn_batch as separate tools — single-purpose schemas serve LLM callers and keep the per-item authorization boundary explicit; consolidation was evaluated and declined for 0.17.

PR #774, opened 2026-07-31
fix(core): recall limit floor leak — slice(0, limit) on all hybrid paths

### LINKED ISSUE #770: recall: limit parameter intermittently ignored — returns fixed 50 regardless of requested count
## Summary

The `limit` parameter passed to `plur_recall` and `plur_recall_hybrid` (and the underlying `recallHybridWithMeta` / `recall` calls) is intermittently ignored. The server returns a fixed number of engrams — observed consistently at 50 — regardless of the requested limit value.

## Observed behavior

Calling `plur_recall` or `plur_recall_hybrid` with an explicit `limit` (e.g. `limit: 5`, `limit: 10`) returns approximately 50 engrams in affected scenarios.

Five independent reproductions were captured during a learning-loop reconciliation audit run against a production Datacore installation (2026-07-30). The audit spawned multiple subagents making `plur_recall_hybrid` calls with small limits (5–10) as part of coverage testing; results consistently came back with ~50 engrams regardless of the requested limit.

## Root cause hypothesis

From inspection of `packages/core/src/index.ts`, the PGLite backend paths contain over-fetch multipliers that may not correctly clamp to the user-specified limit before returning:

- Line 2573: `const embLimit = Math.min(filtered.length, wantReranker ? Math.max(limit * 3, 50) : limit * 2)`
- Line 2596: `const bm25Limit = Math.min(filtered.length, wantReranker ? Math.max(limit * 3, 50) : limit * 3)`
- Line 2631: `const hits = await this.pgliteAdapter.searchVector(queryVec, Math.max(limit * 3, 50), { scopes })`

These `Math.max(..., 50)` calls ensure a minimum fetch of 50 candidates for reranker quality, which is correct. However if the final `.slice(0, limit)` at line 2605 (`reranked.engrams.slice(0, limit)`) is not reached in all code paths — for example when `hybridSearchWithMeta` is called directly instead of going through the PGLite path, or when the reranker is disabled — the 50-candidate fetch size leaks through as the returned count.

The non-PGLite path in `hybridSearchWithMeta` also accepts a raw `limit` and it is unclear whether all return paths respect it.

## Expected behavior

`plur_recall(query, { limit: N })` should return at most N engrams. The internal over-fetch for reranking quality is an implementation detail and must not leak through to callers.

## Steps to reproduce (approximate)

1. Install PLUR v0.9.13 (or later) with PGLite backend enabled.
2. Populate the engram store with > 50 engrams.
3. Call `plur_recall("some query", { limit: 5 })` via the MCP interface.
4. Observe that ~50 engrams are returned instead of 5.

Repro is intermittent — it appears correlated with the PGLite backend being active and/or the reranker path. The YAML-only path may not be affected.

## Environment

- **Observed on**: Datacore installation, PLUR MCP v0.9.13 (`@plur-ai/mcp`)
- **Engram store size**: ~330 personal engrams + ~780 pack engrams
- **Independent reproductions**: 5 (during 2026-07-30 audit run)
- **Latest published version at time of report**: v0.16.1

## Notes

The MCP layer at `packages/mcp/src/tools.ts:88` correctly computes `cap = budget?.max_results ?? args.limit ?? 20` and passes it as `fetchLimit`. The issue appears to be downstream in the core recall paths, not in the MCP tool dispatch.

plur9:
## Summary

Fixes #770 — `plur_recall` / `plur_recall_hybrid` returning a fixed ~50 results regardless of the requested `limit`.

Three code paths were missing a final `slice(0, limit)` when the internal over-fetch floor (`Math.max(limit * N, 50)`) was active for reranker quality:

1. **`recallHybrid()`** — returned `result.engrams` without slicing (the `recallHybridWithMeta` call it delegates to can return up to the floor count)
2. **`recallHybridWithMeta()`** — lacked an outer guard; the MCP layer calls this directly, bypassing the slice in `recallHybrid()`
3. **`hybridSearchWithMeta()`** — non-PGLite path: after the reranker ran, result was returned without slicing

The PGLite path (`_pgliteHybridRecall`, line 2807) already had the guard. This PR closes the non-PGLite and MCP-facing gaps.

## Changes

- `packages/core/src/index.ts` — `recallHybrid`: add `const limit = options?.limit ?? 20` + `return result.engrams.slice(0, limit)`. `recallHybridWithMeta`: add belt-and-suspenders guard before returning (catches any inner-path that leaks the floor through)
- `packages/core/src/hybrid-search.ts` — `hybridSearchWithMeta`: apply `slice(0, limit)` to `reranked.engrams` before returning
- `packages/core/test/hybrid-search.test.ts` — new `describe('hybrid search limit — 50-floor regression (#770)')` block: seeds 55 engrams, requests limit:5, asserts `≤5` from both `recallHybrid()` and `recallHybridWithMeta()`

## Test results

```
Test Files  1 passed (1)
     Tests  12 passed (12)
  Duration  49.10s
```

All 12 tests pass including the 4 new regression tests. Smoke-tested with a 55-engram store: both call paths return exactly 5 results.

## Post-review update (rebase onto #778 + review fixes)

Rebased onto main after #778 landed. The single conflict hunk in `recallHybridWithMeta` was resolved as the review prescribed: main's `_mergeRemoteRecall` fold is kept, and this PR's belt-and-suspenders guard now sits **after** the fold, as the final boundary before `_reactivateResults`. The guard is still required post-#778: `_mergeRemoteRecall` slices to `limit` only when the remote leg returned rows — on the common local-only path it returns the local result unsliced. `hybrid-search.ts` and the test file merged clean.

Follow-up commits from review:

- `docs(core)`: corrected the `hybridSearch` exhaustive-mode docstring — exhaustive mode widens the internal candidate pool (limit floored at 50); the return value is always capped at the caller's `limit`
- `test(core)`: the limit-floor suite's `afterAll` now restores the prior embeddings state (captured via `embedderStatus()` in `beforeAll`) instead of force-enabling

## Micro-benchmark (main 1ee5dde vs this PR)

`pnpm bench:micro`, 100 iterations, same machine, sequential runs:

| Operation | main | PR | Δ mean | Δ % |
|---|---|---|---|---|
| learn | 5.08ms (p95 9.72) | 5.20ms (p95 7.63) | +0.12ms | +2.4 |
| recall_bm25 | 6.70ms (p95 9.98) | 6.18ms (p95 9.68) | −0.52ms | −7.7 |
| recall_hybrid | 28.23ms (p95 27.03) | 26.30ms (p95 30.92) | −1.93ms | −6.8 |
| inject | 0.60ms (p95 0.91) | 0.67ms (p95 1.15) | +0.06ms | +10.3 |
| inject_tokens | 1833 | 1833 | 0 | 0.0 |

All deltas within run-to-run noise (inject is a sub-millisecond op); no regression. The added work per call is one `slice(0, limit)` on an already-materialized array.

## Post-rebase test results

- `packages/core` default project: 144 files passed (3 unrelated files hit 30s timeouts under full-parallel load — same files pass in isolation on both this branch and main; same class of contention documented in `vitest.config.ts`)
- `core-pglite` serial project (PGLite hybrid recall paths): 15/15 files passed
- `hybrid-search.test.ts`: 12/12 including the 4 regression tests
- `pnpm build`: clean

### DIFF
--- a/packages/core/src/hybrid-search.ts
+++ b/packages/core/src/hybrid-search.ts
@@ -111,8 +111,10 @@ function isAggregationQuery(query: string): boolean {
  * No LLM calls — fully local, ~500ms with cached embeddings.
  *
  * Automatically detects aggregation queries ("how many", "total", etc.)
- * and switches to exhaustive mode — wider retrieval to capture ALL
- * mentions across conversations, not just the top few.
+ * and switches to exhaustive mode — widens the internal candidate pool
+ * (limit floored at 50 for the BM25/embedding legs and the rerank stage)
+ * so relevant mentions are less likely to be crowded out of the ranking.
+ * The returned list is always capped at `limit` (#770).
  */
 export async function hybridSearch(
   engrams: Engram[],
--- a/packages/core/src/hybrid-search.ts
+++ b/packages/core/src/hybrid-search.ts
@@ -186,7 +186,7 @@ export async function hybridSearchWithMeta(
   // recall always returns something.
   const reranked = await applyReranker(ranked.map(s => s.engram), query, rerank)
   return {
-    engrams: reranked.engrams,
+    engrams: reranked.engrams.slice(0, limit),
     mode,
     embedderError,
     topScore,
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -2594,8 +2594,9 @@ export class Plur {
 
   /** Hybrid search: BM25 + embeddings merged via Reciprocal Rank Fusion. Async, no API calls. Delegates to recallHybridWithMeta so it gets intent/rerank/PGLite routing too. */
   async recallHybrid(query: string, options?: Omit<RecallOptions, 'mode' | 'llm'>): Promise<Engram[]> {
+    const limit = options?.limit ?? 20
     const result = await this.recallHybridWithMeta(query, options)
-    return result.engrams
+    return result.engrams.slice(0, limit)
   }
 
   /**
@@ -2640,6 +2641,16 @@ export class Plur {
     // #776: fold the server leg in (RRF) before reactivation so displaced
     // local rows are not reactivated and server rows rank on merged order.
     result = { ...result, engrams: await this._mergeRemoteRecall(result.engrams, remotePromise, options, limit) }
+    // Belt-and-suspenders: all inner paths apply slice(0, limit) before
+    // returning, but recallHybridWithMeta is called directly by the MCP layer
+    // (#770) and lacks the outer guard that recallHybrid() adds. Note
+    // _mergeRemoteRecall only slices when the remote leg returned rows — on the
+    // common local-only path it returns the local result unsliced. Enforce here
+    // so no over-fetch floor (Math.max(N, 50) for reranker/aggregation paths)
+    // can leak through to callers.
+    if (result.engrams.length > limit) {
+      result = { ...result, engrams: result.engrams.slice(0, limit) }
+    }
     await this._reactivateResults(result.engrams)
     // WS5 demand flywheel: a zero-result or low-top-score recall is a demand
     // signal. Emit an anonymized, content-free miss-signal (query fingerprint +
--- a/packages/core/test/hybrid-search.test.ts
+++ b/packages/core/test/hybrid-search.test.ts
@@ -3,7 +3,7 @@ import { mkdtempSync, rmSync } from 'fs'
 import { join } from 'path'
 import { tmpdir } from 'os'
 import { Plur } from '../src/index.js'
-import { setEmbeddingsEnabled } from '../src/embeddings.js'
+import { embedderStatus, setEmbeddingsEnabled } from '../src/embeddings.js'
 
 describe('hybrid search (BM25 + embeddings via RRF)', () => {
   let dir: string
@@ -101,8 +101,11 @@ describe('hybrid search limit — 50-floor regression (#770)', () => {
   // directly (bypassing the outer slice guard in recallHybrid).
   let dir: string
   let plur: Plur
+  let priorEmbeddings: { disabled: boolean; disabledReason: string | null }
 
   beforeAll(async () => {
+    const status = embedderStatus()
+    priorEmbeddings = { disabled: status.disabled, disabledReason: status.disabledReason }
     setEmbeddingsEnabled(false, 'limit regression test — BM25-only for speed')
     dir = mkdtempSync(join(tmpdir(), 'plur-limit-floor-'))
     plur = new Plur({ path: dir })
@@ -121,7 +124,9 @@ describe('hybrid search limit — 50-floor regression (#770)', () => {
   }, 120_000)
 
   afterAll(() => {
-    setEmbeddingsEnabled(true)
+    // Restore whatever state the suite found, not force-enable — another
+    // suite (or the environment) may have disabled embeddings deliberately.
+    setEmbeddingsEnabled(!priorEmbeddings.disabled, priorEmbeddings.disabledReason ?? undefined)
     rmSync(dir, { recursive: true })
   })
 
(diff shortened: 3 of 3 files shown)


plur9:
**Review — LGTM, recommend merge.**

Reviewed the three patches:

1. **`hybrid-search.ts` line 189** — `reranked.engrams.slice(0, limit)` on the non-PGLite return path. Correct. This is where the 50-floor was leaking when the reranker ran on the fallback path.

2. **`index.ts` `recallHybrid()`** — captures `limit = options?.limit ?? 20` before delegating and slices on return. Correct default (matches the documented tool default).

3. **`index.ts` `recallHybridWithMeta()`** — conditional belt-and-suspenders guard `if (result.engrams.length > limit)`. Using `if` rather than an unconditional slice is fine — avoids a redundant allocation on the common path.

**Tests:** 4 new regressions in `hybrid-search.test.ts` covering both `recallHybrid` and `recallHybridWithMeta`, both standard and aggregation query paths, all seeded with 55 engrams so the `Math.max(N, 50)` floor fires. Solid coverage.

No issues found. Closes P1 bug #770. CI green on all three Node matrix versions.

🤖 heartbeat review

plur9:
🤖 heartbeat — PR is merge-ready.

CI: all 6 checks green. Mergeable: yes, no conflicts.

**What this fixes:** P1 recall limit bug (#770). Three code paths in `recallHybrid()`, `recallHybridWithMeta()`, and the non-PGLite path in `hybridSearchWithMeta()` were returning the internal 50-candidate over-fetch floor instead of the caller's `limit`. This caused `plur_recall(query, {limit: 5})` to return ~50 results in production.

**Verified:** 55-engram smoke test returns exactly 5 on both call paths. 12 tests pass including 4 new regression tests.

Ready to merge whenever you can approve.

plur9:

Three code paths fixed, each correct:
- `hybridSearchWithMeta`: slices `reranked.engrams` after the reranker runs — the reranker was keeping the floor count, not the caller's limit
- `recallHybrid`: extracts `limit` before the `recallHybridWithMeta` delegation and slices the return value
- `recallHybridWithMeta`: belt-and-suspenders guard is the right call since the MCP layer calls this directly, bypassing the outer slice

Regression tests seed 55 engrams (>50 floor), call both public surfaces (`recallHybrid` and `recallHybridWithMeta`), and assert ≤5 returned — reproduces the exact failure from #770.

CI: all green (test 20/22/24, smoke-packaged, Python SDK, issue-guard). Ready for merge.

plur9:
### Code review

**[85] Rebase required — #778 landed on main and conflicts exactly at the guard's insertion point. Resolve by placing the guard AFTER the remote-leg fold.**

`recallHybridWithMeta` was restructured by #778 (1ee5dde): after the intent/PGLite/YAML branch chain, main now folds the server leg in via `_mergeRemoteRecall` at the very lines where this PR inserts its belt-and-suspenders guard. Git reports the file CONFLICTING with a single conflict hunk there:

- PR side (the guard): https://github.com/plur-ai/plur/blob/56823a0c3c553b3dc889eb8f90136614006a3f63/packages/core/src/index.ts#L2607-L2614
- main side (the #778 fold): https://github.com/plur-ai/plur/blob/1ee5dde70a1f0ea66df8fd593c5d3e43914e074e/packages/core/src/index.ts#L2640-L2642

On rebase, keep main's `_mergeRemoteRecall` line and place the guard **after** it, as the final boundary before `_reactivateResults`. The guard is still needed post-#778, and needed there:

1. `_mergeRemoteRecall` slices to `limit` only when the remote leg actually returned rows — with no remote configured (`remotePromise === null`) or an empty/failed leg it returns `local` **unsliced**, so the 50-floor leak this PR fixes is still live on main's common local-only path.
2. The `hybridSearchWithMeta` slice (https://github.com/plur-ai/plur/blob/56823a0c3c553b3dc889eb8f90136614006a3f63/packages/core/src/hybrid-search.ts#L189) means the local leg entering #778's RRF remote merge is already limit-truncated. That matches what #778 itself does on the `recall()` BM25 path (local leg is sliced before `_mergeRemoteRecall`), so no local-starvation concern — a local-only row ranked past `limit` can never enter the merged top-`limit` anyway. The only effect is that a row present in BOTH legs with local rank > limit loses its local corroboration vote in the RRF sum: a marginal ranking perturbation, not a correctness issue.

`packages/core/src/hybrid-search.ts` and the test file merge cleanly; the slice runs after RRF merge + reranker, so it truncates the final ranking (correct order of operations). Note the earlier "Mergeable: yes, no conflicts" heartbeat comment predates #778 and is stale.

Issue #781, opened 2026-08-02
Default embedding model is English-only (bge-small-en-v1.5) — non-English semantic recall quality


skyeryg:
## Summary

The default embedder (`DEFAULT_EMBEDDER = 'bge-small'` in `packages/core/src/embedders/index.ts`) resolves to `Xenova/bge-small-en-v1.5` — an **English** model. For non-English text (Chinese, Japanese, etc.), semantic similarity scores are computed in an English-trained space, and none of the bundled local options (`minilm`, `bge-small`, `bge-base`, `embedding-gemma`) is a Chinese/multilingual-tuned model. (EmbeddingGemma is multilingual, but its default was reverted to bge-small — see #219.)

## Observed behavior

On a Chinese-language store, recall works but the quality is entirely dependent on this English model's generalization to CJK. Combined with the ftsTokenize CJK issue (#780 — BM25 drops all Chinese), non-English retrieval has **no reliable lexical path and only a sub-optimal semantic path**.

## Suggestion

- Document the language limitation of each embedder in `EMBEDDER_NAMES` (which are English-only vs multilingual).
- Consider shipping a multilingual default (e.g. `Xenova/bge-m3` or `multilingual-e5`) as an opt-in for non-English users, or a `PLUR_EMBEDDER=multilingual` tier.
- At minimum, surface a warning when the active embedder is English-only but the corpus contains non-ASCII text.

## Related

- #219 (embedding model upgrade discussion — quality, not language)
- Issue #xxxx (ftsTokenize CJK)

skyeryg:
Closing with evidence — the issue premise is confirmed, but the conclusion changed after testing.

## What we tested

5 embedding models on the same 28-sample Chinese relevance set (12 related / 16 unrelated queries vs one Chinese engram "测试部署应该用 docker compose"), measuring the **absolute cosine margin** (lowest related minus highest unrelated) — the property that matters for absolute-threshold relevance gating in single-candidate injection:

| model | params | dim | cosine margin |
|---|---|---|---|
| bge-small-en (current default) | 33M | 384 | -0.07 (overlapping) |
| bge-small-zh-v1.5 | 24M | 512 | -0.14 (common-word false positives) |
| gte-multilingual-base | 305M | 768 | -0.006 (distribution compressed) |
| Qwen3-Embedding-0.6B (int8) | 600M | 1024 | ~0 (int8 flattens discriminability) |
[...]

PR #782, opened 2026-08-02
fix(core): index Chinese text as character bigrams so BM25 works for Han script (#780)

### LINKED ISSUE #780: ftsTokenize drops all CJK characters — BM25 is dead for Chinese and other non-English text
## Summary

`ftsTokenize()` in `packages/core/src/fts.ts` discards **every CJK (Chinese/Japanese/Korean) character**, because JavaScript's `\w` only matches ASCII `[A-Za-z0-9_]`. Any non-English query produces an **empty token list**, and non-English engrams are indexed only by their embedded ASCII terms. BM25 is effectively non-functional for Chinese and other non-Latin-script text.

## Reproduction

```js
// packages/core/src/fts.ts — current implementation
function ftsTokenize(text) {
  return text.toLowerCase()
    .replace(/[^\w\s]/g, ' ')   // ← CJK chars are "not \w" → replaced with space
    .split(/\s+/)
    .filter(w => w.length > 2)
    .filter(w => !STOP_WORDS.has(w))
}

console.log(/\w/.test('单'))  // false — \w is ASCII-only without the u flag

console.log(ftsTokenize('单位不要'))
// → []  (pure-Chinese query: zero tokens)

console.log(ftsTokenize('测试部署应该用 docker compose'))
// → ['docker', 'compose']  (only ASCII terms survive)

console.log(ftsTokenize('docker compose 部署'))
// → ['docker', 'compose']  (same — Chinese dropped)
```

Observed impact on a real Chinese store:

| query | BM25 tokens | hybrid topScore |
|---|---|---|
| `单位不要` (unrelated) | `[]` | 0.0164 (embedding-only single hit, floor value) |
| `测试部署用什么` (relevant) | `[]` | 0.0164 (same floor — cannot distinguish) |
| `docker compose 部署` (highly relevant) | `[docker, compose]` | 0.0328 (dual hit) |

A Chinese query can **never** achieve a BM25 hit, so hybrid retrieval degenerates to embedding-only, and RRF's relative ranking means an unrelated query still surfaces the only candidate (noise injection, e.g. `单位不要` retrieves an engram about docker compose deployment).

## Root cause

1. `\w` without the `u` flag is `[A-Za-z0-9_]` — CJK ideographs are treated as non-word garbage and stripped.
2. There is no CJK segmentation step (Chinese has no whitespace-delimited words; it needs character n-grams or a tokenizer like jieba).

## Impact

- **Non-English users get degraded retrieval**: BM25 contributes nothing, all quality rests on the embedding model.
- **Worse under degradation**: when embeddings are unavailable (e.g. `PLUR_DISABLE_EMBEDDINGS=1`, low-memory installs), `hybrid-degraded`/`bm25-only` mode means **Chinese retrieval returns nothing at all** for pure-Chinese queries.
- The RRF `topScore` relevance gate exposed by `HybridSearchResult.topScore` is unusable for non-English queries (all single-hit queries share the floor value ~0.0164).

## Suggested direction

- Use Unicode-aware character classes with the `u` flag: `/[^\p{L}\p{N}_\s]/gu` (or `\p{Script=Han}` etc.) so CJK characters are preserved as tokens instead of stripped.
- Add CJK-aware tokenization: character **bigrams** for Han runs (cheap, no deps) or a proper segmenter (e.g. `Intl.Segmenter`, available in Node 16+, with `granularity: 'word'` for Chinese).
- Add a corpus-stats / fts fixture with Chinese text so this can't silently regress again (the existing fixtures are ASCII-only, which is why this went unnoticed).

## Environment

- `@plur-ai/core` 0.16.1
- Node 24
- Discovered while integrating pi-plur (a pi extension using core in a Chinese-language workflow); all other languages with non-ASCII scripts (Arabic, Devanagari, Cyrillic are partly affected via non-`\w` chars) are affected to varying degrees.

skyeryg:
## Summary

Fixes #780 — `ftsTokenize` strips every Han character because JS `\w` is ASCII-only (`[A-Za-z0-9_]`). Pure-Chinese queries tokenize to **zero terms**, so BM25 contributes nothing for non-English stores:

- hybrid recall degenerates to embedding-only (Chinese queries can never achieve a BM25 hit),
- `bm25-only` degradation returns **nothing at all** for pure-Chinese queries.

## Change

- Re-extract Han runs from the source with `\p{Script=Han}` (u flag) and index them as **character bigrams** — the standard n-gram approach for space-less scripts, zero new dependencies.
- ASCII behavior is **unchanged**: English terms still go through the original path, and the fix is a no-op for Latin-only text.

```ts
// before — pure-Chinese text → []
ftsTokenize('测试部署应该用 docker compose')
// → ['docker', 'compose']

// after
ftsTokenize('测试部署应该用 docker compose')
// → ['docker', 'compose', '测试', '试部', '部署', '署应', '应该', '该用']
```

## Tests

- pure-Chinese text tokenizes to bigrams (was: empty)
- mixed Chinese + English keeps both term kinds
- ASCII-only behavior unchanged (regression)
- `searchEngrams` integration: relevant Chinese query retrieves the Chinese engram; unrelated Chinese query returns empty (no noise)

All core tests pass (2420 passed), `pnpm build` green.

## Notes

Character bigrams are a deliberate minimum — no segmenter dependency (jieba / `Intl.Segmenter` can be layered later if precision demands it). Related but separate: default embedding model is English-only (#781), which independently limits non-English semantic recall.

### DIFF
--- a/packages/core/src/fts.ts
+++ b/packages/core/src/fts.ts
@@ -8,11 +8,20 @@ const STOP_WORDS = new Set([
 
 /** Tokenize text into searchable terms */
 export function ftsTokenize(text: string): string[] {
-  return text.toLowerCase()
+  const lower = text.toLowerCase()
+  const tokens = lower
     .replace(/[^\w\s]/g, ' ')
     .split(/\s+/)
     .filter(w => w.length > 2)
     .filter(w => !STOP_WORDS.has(w))
+  // CJK: \w is ASCII-only ([A-Za-z0-9_]), so Han runs are stripped by the
+  // replace above and pure-Chinese text tokenizes to nothing. Chinese has no
+  // whitespace-delimited words — re-extract Han runs from the source and index
+  // them as character bigrams so non-English text survives BM25. (plur-ai#782)
+  for (const run of lower.match(/\p{Script=Han}{2,}/gu) ?? []) {
+    for (let i = 0; i < run.length - 1; i++) tokens.push(run.slice(i, i + 2))
+  }
+  return tokens
 }
 
 /** Build searchable text from all engram fields */
--- a/packages/core/test/fts.test.ts
+++ b/packages/core/test/fts.test.ts
@@ -20,6 +20,30 @@ describe('ftsTokenize', () => {
     expect(tokens).toContain('jumps')
     expect(tokens).not.toContain('the')
   })
+
+  it('keeps ASCII behavior unchanged when text contains no CJK', () => {
+    const tokens = ftsTokenize('docker compose deployment')
+    expect(tokens).toEqual(['docker', 'compose', 'deployment'])
+  })
+
+  it('indexes pure-Chinese text as character bigrams (was: empty tokens)', () => {
+    const tokens = ftsTokenize('测试部署应该用')
+    expect(tokens).toContain('测试')
+    expect(tokens).toContain('试部')
+    expect(tokens).toContain('部署')
+    expect(tokens).toContain('应该')
+    expect(tokens).toContain('该用')
+    // every Han char participates in two bigrams (except run edges)
+    expect(tokens.length).toBeGreaterThanOrEqual(4)
+  })
+
+  it('mixed Chinese + English keeps both term kinds', () => {
+    const tokens = ftsTokenize('测试部署应该用 docker compose')
+    expect(tokens).toContain('docker')
+    expect(tokens).toContain('compose')
+    expect(tokens).toContain('部署')
+    expect(tokens).toContain('该用')
+  })
 })
 
 describe('computeIdf', () => {
@@ -167,5 +191,24 @@ describe('BM25 document length normalization', () => {
     const results = searchEngrams(engrams, 'kubernetes')
     // Short doc should rank higher — same term match but normalized by length
     expect(results[0].id).toBe('ENG-2026-0330-001')
+  })
+})
+
+describe('CJK search (Chinese engrams)', () => {
+  const engrams = [
+    makeEngram({ id: 'ENG-2026-0330-001', statement: '测试部署应该用 docker compose' }),
+    makeEngram({ id: 'ENG-2026-0330-002', statement: 'unit tests must run before commit' }),
+  ]
+
+  it('retrieves a Chinese engram for a Chinese query', () => {
+    const results = searchEngrams(engrams, '部署流程')
+    expect(results.length).toBeGreaterThan(0)
+    expect(results[0].id).toBe('ENG-2026-0330-001')
+  })
+
+  it('does not retrieve a Chinese engram for an unrelated Chinese query', () => {
+    const results = searchEngrams(engrams, '单位不要')
+    // zero term overlap → empty result (no noise from unrelated CJK queries)
+    expect(results).toHaveLength(0)
   })
 })


crtahlin:
## Review notes (non-blocking) — leaving the formal code-owner approval to the maintainers

Reviewed against current `main` (the PR branch is behind it, so I tested the merge result, which merges cleanly onto the newer `CorpusStats`/`termMatches` work). Built `@plur-ai/core` and ran `fts.test.ts` (18/18 pass), and verified the tokenizer behavior empirically.

This is a clean, minimal, dependency-free fix for a genuine correctness bug: JS `\w` is ASCII-only, so `ftsTokenize` stripped every Han character and pure-Chinese text produced zero BM25 terms. Re-extracting Han runs and indexing them as character bigrams is the standard approach for space-less scripts. The ASCII path is provably untouched (regression test confirms), tokenization stays deterministic, and the test coverage is good — including a `searchEngrams` integration test that checks an unrelated Chinese query returns nothing.

One thing worth confirming for release: tokenization feeds `embeddingContentHash` (md5 over `ftsTokenize(...)`), and this change keeps that deterministic, so the content-hash invariant holds. See note 3.

### Notes

1. **Scope is Han script only, not full "CJK".** Verified empirically:
   - `日本語のテスト` → `["日本","本語"]` — only kanji; hiragana/katakana dropped
   - `한국어 테스트 배포` → `[]` — Korean Hangul dropped entirely

   The fix genuinely helps Chinese and Japanese-kanji, but Korean and Japanese-kana still tokenize to nothing. The PR title accurately says "for Chinese"; issue #780 is framed more broadly as CJK. Suggest keeping #780 open (or tracking the remainder separately) so it isn't closed as fully resolved — filed a follow-up for Hangul + kana. Extending the run regex to include the `Hiragana`, `Katakana`, and `Hangul` script classes would be the natural next step.

2. **A single isolated Han character is dropped** — `{2,}` requires a run of 2+, so `A单B` → `[]`. Fine for the common multi-character Chinese word; a single-character content word in isolation stays unsearchable. Acceptable for a deliberate minimum, noting it.

3. **Derived-index impact on upgrade.** Changing `ftsTokenize` changes the stored `content_hash` for every Han-containing engram, so on the Postgres tier those engrams flag stale once and get re-embedded a single time on upgrade (self-healing and bounded — no loop). Existing SQLite FTS indexes also need a rebuild for old CJK content to benefit. Both are rebuildable derived state (ADR-0001), not a data break — worth a release note.

### Verdict

No blocking issues; correctness, security (well-defined `\p{Script=Han}`, linear scan, no ReDoS), performance, and style all check out. `MERGEABLE BLOCKED` is branch protection (code-owner review required), not a conflict. Solid contribution — thanks for the clear write-up and tests.

Follow-up for the non-Han scripts: #832

plur9 (approved):
Thanks for this — and we're shipping it immediately rather than holding it for the next feature release. **0.17.2 goes out today** so Chinese-speaking users get working BM25 now.

Merging. CI green across Node 20/22/24, 250 test files.

One number worth having on the record: the "ASCII is unchanged" claim holds and costs **4.50µs → 4.78µs** per tokenization (~6%), from the extra regex scan.

**Retitle.** Narrowed to *Chinese*, and #780 stays open — the fix is Han-only:

| | |
|---|---|
| `テストデプロイ` | `[]` |
| `배포는 도커를 사용해야 한다` | `[]` |
| `развертывание должно` | `["docker"]` |
| `तैनाती के लिए डॉकर` | `[]` |
| `déploiement français` | `["ploiement","fran","ais"]` |

The scope was right; only the title overreached, and someone reading the changelog in six months would have concluded Korean worked.

**#833** carries the rest, with two findings from scoping it:

- `\p{L}\p{N}` — the direction #780 suggested — isn't sufficient. Devanagari and Thai marks are `\p{Mn}`, not letters, so Hindi still returns empty and Thai comes back shattered. It needs `\p{M}`.
- Widening the class without stripping space-less runs from the word path double-indexes: `测试部署应该用` returns the whole run plus all six bigrams.

**#834** — this surfaced an older problem of ours. `reversePrefixes` hard-codes a lower bound of 3 under a docstring reading *"no stored token can be 1 or 2 characters."* Han bigrams are two. Nothing breaks, since Han and ASCII are disjoint so the reverse arm can't fire, but the invariant now holds by accident. Ours to fix — it lands on top of this PR, before 0.17.2 ships, so no release goes out with the invariant merely assumed.

**#835 — the red X here is ours, and the fix is up in #836.** The PR Issue Guard auto-assigns the linked issue to the PR author, and GitHub forces a read-only token on `pull_request` events from forks, so it 403s. It had been failing on every external contribution and we hadn't noticed.

PR #836, opened 2026-08-04
ci: run issue guard on pull_request_target so fork PRs get a write token (#835)

### LINKED ISSUE #835: PR Issue Guard fails on every fork PR — read-only GITHUB_TOKEN makes addAssignees 403
## Summary

`.github/workflows/pr-issue-guard.yml` **fails on every pull request opened from a fork**, i.e. on every external community contribution. Found while reviewing #782.

```
HttpError: Resource not accessible by integration
POST https://api.github.com/repos/plur-ai/plur/issues/780/assignees
body: {"assignees":["skyeryg"]}
status: 403
```

## Root cause

The workflow declares what it needs:

```yaml
on:
  pull_request:
    types: [opened, edited, reopened]
permissions:
  issues: write
  pull-requests: write
```

But for a `pull_request` event originating from a **fork**, GitHub forces `GITHUB_TOKEN` to read-only *regardless of the declared `permissions:` block*. This is a platform rule, not a misconfiguration — it exists so untrusted PR code cannot escalate. The declared block only ever narrows the token; it can never widen it back on a fork event.

So `github.rest.issues.addAssignees` 403s, the unhandled rejection fails the step, and the run goes red.

## Why it matters more than a red X

The guard implements our own CONTRIBUTING rule ("claim before you code" — auto-assign the linked issue to the PR author). On fork PRs it does the exact opposite of its purpose:

- It **never** assigns, so the duplicate-effort protection silently does not exist for external contributors — precisely the population it was written for. Internal branches are the ones that don't need it.
- It puts a **failing check on every community PR**. A first-time contributor's initial experience of this repo is a red X they cannot diagnose, caused by our infrastructure, on a PR that may be perfectly good. #782 is exactly that: clean contribution, green CI, red guard.

## Options

1. **Split the trigger.** Move the assign/comment work to a `pull_request_target` workflow (runs with the base repo's token, so it has write access) that only reads `context.payload` and never checks out or executes PR code. Standard pattern for this exact case; safe as long as nothing from the head ref is executed.
2. **Degrade gracefully.** Wrap the `addAssignees` call in try/catch and `core.warning` instead of throwing, so a fork PR reports "could not auto-assign — maintainer, please assign #N manually" and stays green. Smallest change; keeps the guard advisory on forks.
3. **Skip on forks.** `if: github.event.pull_request.head.repo.full_name == github.repository`. Honest about the limitation, but leaves external PRs — the ones that need the coordination check — entirely uncovered.

Option 1 restores the intended behaviour; option 2 is the ~5-line stopgap. Recommend doing 2 now and 1 properly.

Note also that assignability is a second constraint: GitHub only permits assigning users with repo access or who have interacted with the issue. Even with a write-capable token, a brand-new contributor may not be assignable — so the graceful-degradation path in option 2 is worth having regardless of which trigger fix lands.

plur9:
Fixes #835

## Problem

`.github/workflows/pr-issue-guard.yml` failed on **every pull request opened from a fork**, i.e. on every external community contribution. Caught on #782:

```
HttpError: Resource not accessible by integration
POST /repos/plur-ai/plur/issues/780/assignees   →  403
```

For a `pull_request` event raised from a fork, GitHub forces `GITHUB_TOKEN` to read-only *regardless of the declared `permissions:` block*. That block can only ever narrow the token; it cannot restore what the fork rule removed. So the guard could never assign, and the unhandled rejection turned the run red.

The result was the inverse of the guard's purpose: the duplicate-effort protection silently did not exist for external contributors — the only people it was written for — while putting an undiagnosable failing check on their first contribution.

## Change

- **`pull_request` → `pull_request_target`.** Evaluated against the base repository, so the token is write-capable for fork PRs too.
- **Graceful degradation when the author is unassignable.** Even with a write token, GitHub only permits assigning users with repo access or prior interaction on the issue, so a brand-new contributor may simply not be assignable. That is a fact about the account, not a fault in the PR. Warn, and leave a maintainer note on the PR, instead of failing.

## Security

`pull_request_target` hands a privileged token to a workflow evaluated against the base branch, so the standing rule is written into the file as a comment: **this workflow must never check out, build, or execute the head ref.** It doesn't — there is no `actions/checkout` step at all. The only untrusted input is the PR body, used solely as a regex source for issue numbers, each parsed through `parseInt` before reaching an API call.

## Verification

The failure mode is only reproducible from a fork, so this cannot be proven by a same-repo PR — this one will pass either way. The real check is the next external contribution; #782 is the case that surfaced it.

### DIFF
--- a/.github/workflows/pr-issue-guard.yml
+++ b/.github/workflows/pr-issue-guard.yml
@@ -1,7 +1,26 @@
 name: PR Issue Guard
 
+# `pull_request_target`, NOT `pull_request` (#835).
+#
+# On a `pull_request` event raised from a FORK, GitHub forces GITHUB_TOKEN to
+# read-only regardless of the `permissions:` block below — the block can only
+# ever narrow the token, never restore what the fork rule took away. So
+# `addAssignees` returned 403 and this workflow failed on EVERY external
+# contribution, which is precisely the population whose duplicate-effort it
+# exists to prevent. Internal branches are the ones that never needed it.
+#
+# `pull_request_target` runs in the context of the BASE repository, so the token
+# is write-capable for fork PRs too.
+#
+# SECURITY — the rule that makes this safe:
+#   This workflow must NEVER check out, build, or execute the head ref.
+#   `pull_request_target` grants a privileged token to a workflow evaluated
+#   against the base branch; adding `actions/checkout` with the PR's ref would
+#   hand that token to attacker-controlled code. The only untrusted input read
+#   here is the PR body, and it is used solely as a regex source for issue
+#   numbers, each parsed through parseInt before it reaches an API call.
 on:
-  pull_request:
+  pull_request_target:
     types: [opened, edited, reopened]
 
 # Only needs to comment on PRs and assign issues — no code access needed.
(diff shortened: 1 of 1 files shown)


plur9:
**Merged via admin override.** Recorded here because CONTRIBUTING requires an override to be stated and justified rather than left to look like impatience.

- **Deliberate, and the repository owner's decision** (@plur9), not an automated one. CONTRIBUTING reserves this for a human and forbids automation from taking it unilaterally.
- **Why:** this PR is authored by the only account with merge rights, and GitHub does not permit self-approval, so the one-approving-review gate cannot be cleared legitimately. The alternative was blocking a user-facing fix on reviewer availability.
- **What replaced peer review:** a full adversarial self-audit before merge, reported in-thread. It was not a formality — it found and fixed a release-blocking defect in #837 (see the self-review comment there).
- CI is green on the merged commit.

Issue #833, opened 2026-08-04
ftsTokenize: extend Unicode coverage beyond Han — Korean, Japanese kana, Cyrillic, Arabic, Indic, Thai, accented Latin


plur9:
## Summary

#782 fixed BM25 for Chinese by indexing Han runs as character bigrams. That was the right first cut, but `ftsTokenize` is still ASCII+Han only: **every other non-Latin script tokenizes to nothing**, and accented Latin is shredded at each accent.

Measured on the post-#782 tokenizer:

| Input | Tokens today |
|---|---|
| `テストデプロイ` (Japanese kana) | `[]` |
| `배포는 도커를 사용해야 한다` (Korean) | `[]` |
| `развертывание должно использовать docker` (Russian) | `["docker"]` |
| `يجب استخدام دوكر` (Arabic) | `[]` |
| `तैनाती के लिए डॉकर` (Hindi) | `[]` |
| `การปรับใช้ควรใช้ docker` (Thai) | `["docker"]` |
| `déploiement français` | `["ploiement","fran","ais"]` |
| `razmeščanje čez šifriran kanal` (Slovenian) | `["razme","anje","ifriran","kanal"]` |
| `猫` (single Han char) | `[]` |

## This is not a one-liner

#780 suggested `/[^\p{L}\p{N}_\s]/gu`. That alone is **not** sufficient — I measured it. Three separable parts:

### 1. Character class — needs `\p{M}`, not just `\p{L}\p{N}`

```ts
.replace(/[^\p{L}\p{N}\p{M}_\s]/gu, ' ')
```

Devanagari vowel signs and Thai tone marks are *nonspacing marks* (`\p{Mn}`), not letters. Under `\p{L}\p{N}` alone, Hindi still returns `[]` and Thai comes back shattered (`["การปร","บใช","ควรใช"]`) — each word is split at every diacritic into fragments the length filter then eats.

Fixes: Cyrillic, Greek, Arabic, Hebrew, Korean, Armenian, Georgian, and all accented Latin (French, German, Slovenian, Czech, Polish…).

### 2. Space-less segmentation — extend the bigram run, and strip it from the word path

Widen #782's `\p{Script=Han}` to also cover Hiragana, Katakana, Thai, Khmer, Lao, Myanmar.

[...]

PR #837, opened 2026-08-04
fix(core): stamp a tokenizer version on Postgres rows so stale tokens cannot go unnoticed (#834)

### LINKED ISSUE #834: Postgres BM25 pushdown: no reindex path when ftsTokenize changes, and reversePrefixes assumes tokens are ≥3 chars
## Summary

`PostgresAdapter` derives tokens **at write time** and persists them:

```ts
// storage-postgres.ts, toRow()
const tokens = ftsTokenize(engramSearchText(e))
return { …, tokens, search_text: tokens.join(' '), … }
```

That is deliberate and correct — it is what lets the BM25 pushdown claim exactness, because `df` (counted in SQL over the stored columns) and `tf` (counted in JS over freshly-tokenized text) are guaranteed to agree.

**The guarantee only holds while `ftsTokenize` never changes.** It just changed (#782), and it is about to change again (#833). There is no reindex path.

## Finding 1 — stale tokens are invisible, not loud

After #782, a Postgres-backed store contains rows whose `tokens` / `search_text` were derived by the *old* tokenizer:

- Candidate narrowing is `search_text LIKE '%qt%' OR tokens && ARRAY[…]`. Pre-#782 `search_text` contains **no Han at all**, so existing Chinese engrams can never enter the candidate set.
- Net effect: **#782 fixes Chinese BM25 only for engrams written after the upgrade.** Every Chinese engram already in a Postgres store stays invisible, which is exactly the bug #780 reported.
- This is the enterprise store path.

The existing staleness guard (`corpusStats`, ~line 1151) only fires on `tokens IS NULL` — rows written before the *column* existed. It cannot see rows written before the *tokenizer* changed, and `PostgresAdapter` deliberately implements no `reindex()` (it has no YAML to rebuild from).

The workaround exists (`PostgresAdapter.save` rewrites every row) but nothing tells an operator they need it.

**Proposal:** persist a tokenizer version alongside the tokens — a constant in `fts.ts` bumped whenever `ftsTokenize`'s output can change — and extend the `corpusStats` guard to refuse (or warn loudly) when any active row carries a stale version, with the same "re-save to backfill" remedy it already prints for NULL. Make the invariant checkable instead of tribal.

## Finding 2 — `reversePrefixes` assumes tokens are ≥3 chars; #782 broke that

```ts
/**
 * Bounded below by 3 because `ftsTokenize` discards anything shorter, so no
 * stored token can be 1 or 2 characters and testing for them would be dead work.
 */
private static reversePrefixes(qt: string): string[] {
  const out: string[] = []
  for (let n = 3; n <= qt.length; n++) out.push(qt.slice(0, n))
  return out
}
```

Every Han bigram is a **2-character stored token**, so that docstring is now false.

It does not misbehave today, purely because Han and ASCII alphabets are disjoint: the reverse arm `qt.startsWith(t)` can never fire between a bigram and a Latin token, and bigram-vs-bigram equality is already covered by the forward arm. So the invariant now holds **by accident rather than by construction**, with a comment asserting the opposite.

It becomes a live `df` undercount the moment #833 part 3 lowers the length floor for dense scripts (`도커`, 2 chars).

**Proposal:** fix the comment now, derive the lower bound from a shared `MIN_TOKEN_LENGTH` in `fts.ts` rather than a literal `3`, and add a pushdown-parity fixture with CJK text.

## Finding 3 — ~~none of this is covered by CI~~ (retracted)

**Struck through: this was wrong.** I inferred it from the `PLUR_TEST_POSTGRES_URL` gate plus a local run that skipped the suite, without checking `ci.yml`. CI *does* stand up a `pgvector/pgvector:pg16` service container and set `PLUR_TEST_POSTGRES_URL` on the test step, and `postgres-bm25-pushdown.test.ts` runs there — 17 tests, under the serial `core-pglite` project.

So the pushdown parity assertions *are* enforced on every PR. The gap that let #782's interaction through is narrower than I claimed: the parity fixtures are all ASCII, so they exercise the rule but never with a token shorter than three characters. Findings 1 and 2 stand; the remedy is CJK fixtures in the existing suite, not new CI infrastructure.

plur9:
Fixes #834. Lands on top of #782, which is what makes it necessary.

## Why

`PostgresAdapter` derives tokens at **write** time and persists them into `tokens` / `search_text`. That is deliberate and is what lets the BM25 pushdown claim exactness — `df` counted in SQL over stored tokens and `tf` counted in JS over live text are guaranteed to agree because one tokenizer produced both.

The guarantee holds only while the tokenizer that wrote the rows and the one asking the question are the same tokenizer. **#782 changed the tokenizer.**

Rows written before it contain no Han at all, so `search_text LIKE '%<han>%'` never matches and the engram drops out of the candidate set *entirely*. The corpus reports that it does not contain the engram, which is indistinguishable from the engram not being there — silent, and wrong in the one direction a search index must never be wrong. Concretely: without this, #782 fixes Chinese BM25 only for engrams written *after* the upgrade, and every Chinese engram already in a Postgres-backed store stays invisible. That is the enterprise store path.

The pre-existing guard only catches `tokens IS NULL` — rows written before the *column* existed. It cannot see a row whose tokens are well-formed but produced by an older contract.

## Change

- **`TOKENIZER_VERSION`** in `fts.ts`, stamped onto every row by `toRow` beside `content_hash`, with the bump rule documented at the constant.
- **`corpusStats` refuses** when any in-scope active row carries a different version, printing the same *"re-save the store"* remedy the NULL guard already does. Refusing rather than approximating is what the `CorpusStats` contract demands: an approximate corpus is worse than no pushdown.
- The probe is served by a **partial index whose predicate and name both carry the version**. `CREATE INDEX IF NOT EXISTS` matches on name alone, so an unversioned name would silently keep the old predicate after a bump while reporting success — reintroducing exactly the class of staleness this removes.
- **`tokens_version` is deliberately not defaulted.** A row predating the column was written by an unknown tokenizer, and "unknown" must read as stale. Every store in existence when this ships was written pre-#782, so NULL is not merely conservative — it is accurate.

### `reversePrefixes`

It derived its lower bound from a literal `3`, under a docstring asserting that no stored token could be shorter. #782 made that false: Han bigrams are exactly two characters. The bound now comes from `MIN_TOKEN_LENGTH`.

**This has no behavioural delta today, and the mutation check says so** — reverting the bound to `3` leaves all 21 pushdown tests green, because Han and ASCII alphabets are disjoint, so no Latin query token can have a Han bigram as a prefix and the arm it feeds can never fire. The value is that the invariant stops holding *by accident*. It becomes a live `df` undercount the moment #833 lowers the token floor for dense scripts (Korean `도커` is two characters).

## Tests

- **Tokenizer contract** — `MIN_TOKEN_LENGTH` asserted against the tokenizer's actual output rather than restated as a constant, which is the failure shape that let the old `3` rot unnoticed.
- **Pinned-output canary** — any edit changing what `ftsTokenize` emits fails here, and the fix is to bump the version, not to update the fixture.
- **CJK pushdown parity** — pushed-down and local rankings compared for a Chinese query, and `df` for a two-character Han token checked against the local count under `termMatches`. The existing parity fixtures are all ASCII, so they exercise the matching rule but never with a sub-three-character token — which is why this went unseen.
- **Staleness fixtures** — the refusal fires, and a re-save clears it.

## Verification

Run against a real `pgvector/pgvector:pg16` (same image CI uses):

- Full core suite with `PLUR_TEST_POSTGRES_URL` set: **160 files, 2674 passed**, 4 skipped.
- Pushdown suite: 21 passed (17 pre-existing + 4 new).
- Mutation-checked: removing the staleness guard fails 2 tests; reverting the prefix bound fails none (documented above rather than papered over).
- `tsc --noEmit` on core and `typecheck:tests` both clean.

### DIFF
--- a/packages/core/src/fts.ts
+++ b/packages/core/src/fts.ts
@@ -7,6 +7,46 @@ const STOP_WORDS = new Set([
   'can', 'will', 'should', 'would', 'could', 'may', 'might',
 ])
 
+/**
+ * Shortest token `ftsTokenize` can emit.
+ *
+ * NOT the same number as the word-path length filter below, and the two parted
+ * company in #782. The whitespace path still discards anything under three
+ * characters; the Han bigram path emits tokens of exactly two.
+ *
+ * Anything reasoning about the shortest STORED token must read this constant
+ * rather than restate the word-path floor. `PostgresAdapter.reversePrefixes`
+ * did restate it — as a literal `3`, under a docstring asserting no stored
+ * token could be shorter — and #782 silently made that assertion false. It did
+ * not misbehave only because Han and ASCII alphabets are disjoint, so the arm
+ * it feeds could never fire on a bigram. An invariant that holds by accident
+ * is one refactor away from holding not at all.
+ */
+export const MIN_TOKEN_LENGTH = 2
+
+/**
+ * Version of `ftsTokenize`'s output contract. Bump on ANY change that can alter
+ * the token list produced for the same input.
+ *
+ * Exists because a store may derive tokens at WRITE time and compare them
+ * against freshly-tokenized query terms at READ time — `PostgresAdapter` does
+ * exactly that, and it is what lets the BM25 pushdown claim exactness: `df`
+ * counted in SQL over stored tokens, `tf` counted in JS over live text,
+ * guaranteed to agree because one tokenizer produced both.
+ *
+ * The guarantee holds only while the tokenizer that wrote the rows and the one
+ * asking the question are the same tokenizer. When they are not, nothing
+ * throws. Rows written before #782 contain no Han at all, so a Chinese query
+ * simply never matches them — the corpus reports that it does not contain the
+ * engram, which is indistinguishable from the engram not existing. Silent, and
+ * wrong in the one direction a search index must never be wrong.
+ *
+ * History:
+ *   1 — original. `\w`-based word split, three-character floor, stop words.
+ *   2 — #782. Han runs additionally indexed as overlapping character bigrams.
+ */
+export const TOKENIZER_VERSION = 2
+
 /** Tokenize text into searchable terms */
 export function ftsTokenize(text: string): string[] {
   const lower = text.toLowerCase()
--- a/packages/core/src/storage-adapter.ts
+++ b/packages/core/src/storage-adapter.ts
@@ -374,7 +374,7 @@ export interface StorageAdapter {
    * @param queryTokens Tokens from `ftsTokenize` — already lowercased and
    *   stop-word filtered, so the store must not re-tokenize.
    */
-  corpusStats?(queryTokens: string[], opts?: ScopeRestriction): Promise<CorpusStats>
+  corpusStats?(queryTokens: string[], opts?: ScopeRestriction): Promise<CorpusStats | undefined>
   /**
    * Corpus-wide `N` and per-term `df` for BM25 scoring (convergence Phase 4,
    * #711). OPTIONAL — a store that cannot compute these exactly must leave it
@@ -407,7 +407,7 @@ export interface StorageAdapter {
    * @param queryTokens Tokens from `ftsTokenize` — already lowercased and
    *   stop-word filtered, so the store must not re-tokenize.
    */
-  corpusStats?(queryTokens: string[], opts?: ScopeRestriction): Promise<CorpusStats>
+  corpusStats?(queryTokens: string[], opts?: ScopeRestriction): Promise<CorpusStats | undefined>
   /**
    * Vector similarity search (cosine).
    *
--- a/packages/core/src/storage-postgres.ts
+++ b/packages/core/src/storage-postgres.ts
@@ -73,7 +73,10 @@
  */
 import type { Engram } from './schemas/engram.js'
 import { EngramSchemaPassthrough } from './schemas/engram.js'
-import { searchEngrams, ftsTokenize, engramSearchText, embeddingContentHash, type CorpusStats } from './fts.js'
+import {
+  searchEngrams, ftsTokenize, engramSearchText, embeddingContentHash,
+  MIN_TOKEN_LENGTH, TOKENIZER_VERSION, type CorpusStats,
+} from './fts.js'
 import { logger } from './logger.js'
 import {
   EXACT_VECTOR_INDEX,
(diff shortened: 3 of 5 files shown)


plur9:
## Self-review found a release-blocking defect in this PR — fixed in `0d8f0a3`

Auditing this before merge (it had no second reviewer) turned up a problem the test suite structurally could not catch, because **nothing here exercises an upgrade**.

The original guard *threw* on a tokenizer-version mismatch. Two facts combine badly:

1. `index.ts:3052` calls `corpusStats` **without a try/catch**, so the throw propagates straight out of `recall()`.
2. Every row of every existing store has `tokens_version = NULL` the moment this ships — by design, since "unknown tokenizer" must read as stale.

Together: **upgrading to 0.17.2 would have taken `recall()` down completely on every Postgres-backed store** — including ASCII-only stores that contain no Han, are unaffected by #782, and have nothing whatsoever wrong with them. Trading a silent wrong answer for a total outage is not a fix, and it is strictly worse than the bug.

Worse, the guard violated the very contract it was written to uphold. From `StorageAdapter.corpusStats`:

> An implementation that cannot reproduce the rule should return `undefined` rather than a best effort.

### The fix

- `corpusStats` returns `undefined`; the caller derives statistics from the candidate set, exactly as it does for any store that doesn't implement the method.
- **`searchBM25` needed the same treatment**, which I initially missed. Narrowing runs `search_text LIKE`, and a pre-#782 `search_text` contains no Han at all — so degrading only the statistics would have fixed the ranking of rows that were never selected in the first place. It now takes the local path, which re-tokenizes from `data` with the current tokenizer and is immune to whatever is stored.
- Result: correct answers on an un-re-saved store, slower, with a one-shot warning naming the remedy. `save()` restores the pushdown. Self-healing, no operator emergency.

### The staleness test was vacuous, and the mutation check proved it

It set `tokens_version = NULL` while leaving correct bigrams in the row, so the pushdown still worked and the test passed **with or without** the fallback. Removing the fallback changed nothing — which is the definition of a test that proves nothing.

It now writes tokens as the pre-#782 tokenizer actually produced them (empty array and empty `search_text` for pure-Chinese text) and fails when the fallback is removed.

### Verification

- Full core suite against real `pgvector/pgvector:pg16`: **160 files, 2674 passed**, 4 skipped.
- Pushdown suite: 22 passed.
- Mutation-checked: removing the `searchBM25` fallback now fails the stale test; removing the `corpusStats` guard fails 2.
- `tsc --noEmit` on core clean; `typecheck:tests` surfaced the `CorpusStats | undefined` ripple at existing call sites, now fixed.

Retained honestly from the original description: reverting `MIN_TOKEN_LENGTH` to the literal `3` still fails nothing, because Han and ASCII alphabets are disjoint. That change is correctness-by-construction, not a live bug fix.

plur9:
**Merged via admin override.** Recorded here because CONTRIBUTING requires an override to be stated and justified rather than left to look like impatience.

- **Deliberate, and the repository owner's decision** (@plur9), not an automated one. CONTRIBUTING reserves this for a human and forbids automation from taking it unilaterally.
- **Why:** this PR is authored by the only account with merge rights, and GitHub does not permit self-approval, so the one-approving-review gate cannot be cleared legitimately. The alternative was blocking a user-facing fix on reviewer availability.
- **What replaced peer review:** a full adversarial self-audit before merge, reported in-thread. It was not a formality — it found and fixed a release-blocking defect in #837 (see the self-review comment there).
- CI is green on the merged commit.

Issue #812, opened 2026-08-02
Stale embeddings: semantic recall keeps the vector for text that changed (+ PGLite orphans + ID reuse)


plur9:
Finding 6 of the #811 whole-repository audit. **Verified by code inspection**, not fixed — the correct fix needs a schema migration and this was found during a release gate.

## Not data loss, but silently wrong recall

The engram text on disk is correct. What is wrong is the vector semantic recall uses for it.

**Nothing invalidates an embedding when the text changes.** `updateMany` (`storage-postgres.ts:843`) and `save` (`:903`) upsert `data`, `search_text` and `tokens` but never touch `engram_embeddings`. PGLite is the same at `storage-pglite.ts:492`.

**The backfill query only asks whether an ID has a vector at all:**

```sql
SELECT e.data FROM engrams e
LEFT JOIN engram_embeddings em ON em.engram_id = e.id
WHERE e.status = 'active' AND em.engram_id IS NULL
```

So an engram that has *any* vector is never re-embedded. A dedup UPDATE or MERGE (`learn-async.ts`) can rewrite "cats" to "databases" and semantic recall permanently ranks it as "cats".

## PGLite additionally orphans vectors, and IDs get reused

Postgres declares the FK:

```sql
engram_id TEXT PRIMARY KEY REFERENCES engrams(id) ON DELETE CASCADE
```

PGLite does **not** — `storage-pglite.ts:253` is a bare `engram_id TEXT PRIMARY KEY`. So a deleted engram leaves its vector behind.

And IDs are reusable. `generateEngramId` takes `max(existing) + 1` for the day, so if `compact()` removes the highest same-day engram, the next `learn()` mints that same ID. On PGLite the new, unrelated engram then inherits the orphaned vector.

(`forget()` retires rather than removes, so the window is `compact()` specifically — a normal operation, not an exotic one.)

## Why this is not a one-line fix

The obvious fix — drop the embedding whenever an engram is updated — is **not viable**: `_reactivateResults` calls `updateMany` on every `recall()` to persist activation, so unconditional invalidation would re-embed the entire result set on every query.

It needs a content check, and `engram_embeddings` has nowhere to put one:

```sql
CREATE TABLE engram_embeddings (
  engram_id TEXT PRIMARY KEY ...,
  embedding vector(N) NOT NULL      -- no hash, no source text, no version
)
```

## Proposed fix

[...]

PR #839, opened 2026-08-04
fix(core): auto-backfill stale tokens, and add plur reindex-tokens (#839)

### LINKED ISSUE #840: Stale BM25 tokens never clear on Postgres: nothing in ordinary use performs a whole-corpus save
## Summary

#837 made tokenizer staleness detectable and degraded safely to local scoring. Its stated remedy is "re-save the store (`PostgresAdapter.save`)". **Nothing in ordinary use does that**, so the remedy never happens.

## Why it never clears

`index.ts:1216` prefers the targeted write seams over a whole-corpus write:

```ts
await this._primaryStore.append(engram)              // row stores take this
...
await this._writeEngrams(this.paths.engrams, corpus) // only the fallback
```

`PostgresAdapter` implements `append` and `updateMany`, so `learn()` and `feedback()` each touch **one** row, stamped with the current tokenizer version. Every other row keeps its stale marker. The guard asks whether *any* in-scope active row is stale, so the answer stays yes indefinitely.

0.17.1's targeted-write work (#829) is what removed the incidental whole-corpus rewrite that used to clear this as a side effect. A whole-corpus `save()` still happens on `compact`, pack install and some sync paths — which may be months away, or never.

## Impact

Correctness is unaffected: the fallback re-tokenizes from `data` with the current tokenizer and is right. **Performance is not.** The pushdown exists to stop `recall` loading and scoring the whole active set, and an upgraded store loses it permanently, with the only symptom being a log line. That lands hardest on the largest stores — the ones the pushdown was built for.

## Fix

1. **Auto-backfill** — re-derive stale rows in batches, kicked fire-and-forget when staleness is detected, mirroring the embedding backfill (#812).
2. **A CLI command** — force it synchronously with a count, for operators upgrading deliberately and for rarely-read stores.

Both, because they serve different moments: the background pass covers the deployment nobody is watching; the command covers the operator who wants the pushdown back before the first user query rather than after it.

## Note

Should ship **with** 0.17.2 rather than after. Without it, 0.17.2 turns the pushdown off on every existing Postgres store and offers no route back that anyone would find.

plur9:
Closes #840. Follow-up to #837, found by asking a question that PR did not answer: **who actually runs `save()`?**

## The gap

#837 made stale tokens detectable and degraded safely to local scoring. Its stated remedy was "re-save the store". Nobody does that:

```ts
// index.ts:1216 — row stores take the seam, not the whole-corpus write
await this._primaryStore.append(engram)
...
await this._writeEngrams(this.paths.engrams, corpus)   // only the fallback
```

`PostgresAdapter` implements `append` and `updateMany`, so `learn()` and `feedback()` each touch **one** row. Every other row keeps its stale marker, the guard asks whether *any* row is stale, and the answer stays yes forever.

The irony is that 0.17.1's targeted-write work is exactly what removed the incidental whole-corpus rewrite that used to clear this. So an upgraded store would sit on the correct-but-slower fallback **indefinitely** — having permanently lost the pushdown that exists to stop `recall` loading and scoring the whole active set — and the only symptom would be a log line.

Correctness was never at risk. The performance regression was, silently, and on exactly the deployments that need the pushdown most.

## Two remedies, because they serve different moments

**`backfillTokens()`** — re-derives in batches, set-based per batch, reusing `toRow` so a backfilled row is byte-identical to a freshly saved one. Kicked fire-and-forget by the staleness detector, mirroring the embedding backfill (#812): the caller has already been routed to a correct path, so blocking it would trade a wrong answer for a slow one twice over. At most one runs per adapter; a failure logs and clears the latch so a later query retries.

**`plur reindex-tokens`** — forces it synchronously and reports a count. For the operator upgrading deliberately who wants the pushdown back *before* the first user query rather than after it, and for rarely-read stores where "on the next recall" could be days. On non-Postgres tiers it says so plainly and exits zero, rather than pretending to work — only Postgres persists tokens; every other tier re-derives them per query.

## Tests

- Auto-backfill restores the pushdown with no operator step, polling the version marker rather than sleeping on a guess.
- The fallback still answers *while* the backfill is in flight — the kick must not block the query that noticed.
- Rows carry real bigrams afterwards, not the pre-#782 empty tokenization.
- `backfillTokens` is idempotent and reports 0 on a current store.

Mutation-checked: removing the kick fails the first test.

## Verification

Full monorepo suite against real `pgvector/pgvector:pg16`: **277 files, 3932 passed**, 4 skipped. `tsc --noEmit` clean on core and cli; `typecheck:tests` clean.

## Release note

This should ship **with** 0.17.2, not after. Without it, 0.17.2 turns the pushdown off on every existing Postgres store and offers no way back that anyone would find.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ Chinese search works — found and fixed by skyeryg.
 - Chinese text is searchable
 - BM25 saw no Han at all
 - Postgres guards stale tokens
-- Stale stores degrade safely
+- Stale stores self-heal
 
 ### Fixed
 
@@ -22,6 +22,14 @@ Chinese search works — found and fixed by skyeryg.
   [@skyeryg](https://github.com/skyeryg), who hit it integrating PLUR into a Chinese-language
   workflow. Chinese only — Japanese kana, Korean, Cyrillic, Arabic, Indic scripts and accented
   Latin are still dropped or mangled, tracked in #833.
+- **Stale tokens re-derive themselves, and `plur reindex-tokens` forces it** (#840, #839): the
+  detection above named a remedy — re-save the store — that nothing in ordinary use performs.
+  `PostgresAdapter` implements the targeted `append` / `updateMany` seams, so `learn()` and
+  `feedback()` each touch one row and leave every other row's version untouched; an upgraded store
+  would sit on the correct-but-slower fallback indefinitely, having permanently lost the pushdown,
+  visible only as a log line. Stale rows are now re-derived in the background the first time they
+  are noticed, and `plur reindex-tokens` forces the same pass synchronously with a count for
+  operators who want the pushdown back before the first user query rather than after it.
 - **Postgres stores detect tokens written by an older tokenizer** (#834, #837): `PostgresAdapter`
   derives tokens at write time and persists them, which is what lets the BM25 pushdown claim
   exactness — `df` counted in SQL and `tf` counted in JS agree because one tokenizer produced both.
--- a/packages/cli/src/commands/reindex-tokens.ts
+++ b/packages/cli/src/commands/reindex-tokens.ts
@@ -1,0 +1,66 @@
+/**
+ * `plur reindex-tokens` — re-derive BM25 tokens on a Postgres-backed store
+ * after a tokenizer change (#839).
+ *
+ * The adapter kicks this in the background the first time it notices stale
+ * rows, so most deployments never need to run it. It exists for the cases
+ * where "eventually, in the background" is not good enough:
+ *
+ *   - an operator upgrading deliberately, who wants the pushdown back before
+ *     the first user query rather than after it,
+ *   - a store that is read rarely, where "on the next recall" could be days,
+ *   - anyone who wants a count rather than a log line.
+ *
+ * Only Postgres persists tokens. Every other tier re-derives them per query,
+ * so there is nothing to reindex and this reports that rather than pretending
+ * to work.
+ */
+import type { GlobalFlags } from '../plur.js'
+import { PostgresAdapter } from '@plur-ai/core'
+import { shouldOutputJson, outputJson, outputText, exit } from '../output.js'
+
+function usage(): never {
+  exit(1, [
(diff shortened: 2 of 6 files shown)


plur9:
## Pre-merge audit — three defects found and fixed

Audited before merge because this PR has no second reviewer. Two of the three were in the new CLI command and both made it **worse than not shipping it**, because each produced a confident false negative on the only question it exists to answer.

**1. The schema was never resolved.** `PostgresAdapterOptions.schema` defaults when omitted, and `initSchema` runs `CREATE SCHEMA IF NOT EXISTS`. So on a deployment using `postgres.schema` — exactly what a multi-tenant install uses — `plur reindex-tokens` would have connected to the *default* schema, **created a spurious empty one**, found zero stale rows in it, and printed "Tokens are already current."

**2. The DSN came from `PLUR_POSTGRES_URL` only**, while the engine resolves environment *then* `config.yaml`. A file-configured deployment got "No Postgres store configured."

Both are the same root cause: a second, divergent copy of resolution logic. `loadConfig` is now exported from core so there is one.

**3. `backfillTokens` bounded its loop on rows SELECTED, not rows UPDATED.** The UPDATE joins `e.id = t.id` where `t.id` comes from the engram's own `data`. A row whose `id` column disagrees with `data->>'id'` matches nothing, keeps its stale version, and is returned again by the identical next query — a tight infinite loop holding a pool connection. The batch stays *full*, so the existing short-batch bound could never catch it. Progress is now the loop condition.

Mutation-checked: reverting the loop condition to attempt-based fails the new test.

## Verification

CI green on the audited commit — all 5 checks, Node 20/22/24 plus packaged-artifact smoke. Pushdown suite 25 passed locally against real `pgvector/pgvector:pg16`.

---

**Merged via admin override.** Deliberate, and the repository owner's decision (@plur9), not an automated one — recorded here because CONTRIBUTING requires an override to be stated and justified rather than left looking like impatience. The gate cannot be cleared legitimately: this PR is authored by the only account with merge rights and GitHub does not permit self-approval. The audit above is what replaced peer review, and it was not a formality.

PR #841, opened 2026-08-04
fix(deps): raise ip-address and hono override floors past SSRF and ReDoS advisories

plur9:
Closes the four open Dependabot alerts.

## Why these matter here

All four arrive through `@modelcontextprotocol/core` → `express` / `express-rate-limit` and `hono`. Those are **runtime dependencies of the published `@plur-ai/mcp`**, not dev tooling — so they ship to every user, and an SSRF bypass on an MCP server's HTTP surface is the real thing rather than a theoretical one.

| Severity | Advisory | Package | Issue |
|---|---|---|---|
| **high** | GHSA-mwp4-54f8-5fhr | `ip-address` ≤10.3.0 | leading-zero octets decoded as decimal while resolvers decode them as octal → SSRF / trust-boundary bypass |
| medium | GHSA-4xrf-jv44-h6hh | `ip-address` ≤10.2.1 | CIDR suffix suppresses special-use classification |
| medium | GHSA-22jq-vg5j-6vgg | `ip-address` ≤10.2.0 | IPv4-mapped / NAT64 misclassification |
| medium | GHSA-8j4g-w8fx-2239 | `hono` <4.12.34 | ReDoS in CORS middleware |

## Change

Raised the floors in the **existing** `pnpm.overrides` block rather than adding another mechanism — that block is already this repo's pattern for pinning transitive minimums:

```diff
-"hono": ">=4.12.25"          →  ">=4.12.34"
-"ip-address": ">=10.1.1"     →  ">=10.3.1"
```

Resolved to `ip-address@10.4.0` and `hono@4.13.0`. **Lockfile blast radius is exactly those two packages** — nothing else moved:

```
-  hono@4.12.33  →  +  hono@4.13.0
-  ip-address@10.2.0  →  +  ip-address@10.4.0
```

## These are advisories, not a compromise

Worth stating plainly because the two get conflated and the correct response differs. These are ordinary disclosed CVEs with published patches — a version bump. A compromised package release would be an incident, handled differently.

Separately worth noting: pnpm 10 already blocks install-time lifecycle scripts here pending explicit `pnpm approve-builds` (currently `esbuild`, `onnxruntime-node`, `openclaw`, `tree-sitter-bash`). Script execution on install is the primary vector in most recent npm compromises, so that posture is doing real work.

## Verification

Deliberately leaving this to CI rather than citing a local run. My local full-suite runs for this branch raced a branch switch and are not trustworthy evidence — CI builds from a clean checkout, and `smoke-packaged` exercises the **packaged tarball**, which is the artefact that actually carries these dependencies to users.


plur9:
## Pre-merge audit

- **Resolved versions clear every advisory floor**: `ip-address@10.4.0` (needs ≥10.3.1), `hono@4.13.0` (needs ≥4.12.34).
- **Blast radius is exactly the two packages** — 8 changed lockfile lines, all of them the old/new pairs for `hono` and `ip-address`. No transitive drift, no unrelated resolution churn.
- **Three files touched**: `package.json` (two override floors), `pnpm-lock.yaml`, `CHANGELOG.md` (manifest-gate declaration).
- **CI green on a clean checkout**, including `smoke-packaged`, which packs and drives the published tarball — the artefact that actually carries these dependencies to users, and the one a source-only run would not have exercised.

Worth recording: three CLI tests failed on my local runs of this branch and **did not reproduce in CI**. The cause was mine, not the dependencies — the lockfile edits sat uncommitted in a working tree I then switched branches on, so the suite was reading files mid-edit. CI builds fresh, which is why this PR cites it rather than a local number.

---

**Merged via admin override.** Deliberate, and the repository owner's decision (@plur9), not an automated one — recorded because CONTRIBUTING requires an override to be stated and justified. The gate cannot be cleared legitimately: authored by the only account with merge rights, and GitHub does not permit self-approval.

@crtahlin — post-hoc review welcome.

PR #842, opened 2026-08-04
fix(release): gate smoke on core propagation, and stop advising a version burn for lag

plur9:
Fixes the failure that aborted the v0.17.2 release mid-flight.

## What happened

The release published all four packages to `@next`, then aborted at the smoke test. **Nothing was wrong with the artefacts** — I verified each by hand immediately afterwards and every one reported `0.17.2`.

`cli` and `mcp` pin `core` at the exact version, so `npx @plur-ai/cli@X` cannot resolve until `core@X` is visible to the registry's resolver. `cli` is smoked **first**, so it absorbed the entire propagation wait for the whole set, and its private six-attempt budget decided the fate of the release:

```
npx -y @plur-ai/cli@0.17.2 --version → (propagating, retry 1) … (retry 5) ✗
    npm error notarget No matching version found for @plur-ai/core@0.17.2
npx -y @plur-ai/mcp@0.17.2 --version → (propagating, retry 1) ✓
@plur-ai/core@0.17.2 install + import → ✓
```

`mcp`, checked seconds later against a now-propagated `core`, passed on its **first** retry. The packages were fine; the ordering was the bug. RELEASING.md already records v0.14.0 dying the same way — the retry loop added afterwards narrowed the window without closing it.

## The worse half: the advice was wrong

On abort the script printed:

> Then ship a fix as the next patch (e.g. 0.17.2 → next-patch) and re-run.

Following that would have **burned 0.17.2 and republished a byte-identical artefact as 0.17.3**. The correct recovery — resume from the promote step — is the opposite action, and the script gave no way to tell which situation you were in.

## Changes

**1. Gate on the shared dependency (Step 5a.5).** Wait up to 5 minutes for `@plur-ai/core@$VERSION` to become resolvable via `npm view` before any smoke check runs, so no single package's retry budget stands in for the whole set's propagation. `npm view` is far cheaper than an `npx` install per attempt, so a generous budget costs seconds. Non-fatal on its own — the smoke checks still decide.

**2. Distinguish lag from a defect.** Track whether *every* failure matched the propagation signature. If nothing crashed and no wrong version was ever seen, say so and give the resume path:

```
Every failure was npm-propagation lag (ETARGET / no matching version),
never a crash or a version mismatch. …
DO NOT ship a new patch version for this. Verify, then RESUME:
  1. npx -y @plur-ai/cli@X --version    # expect X
  …
  3. for p in core cli mcp migrate; do npm dist-tag add @plur-ai/$p@X latest; done
```

A genuine crash or version mismatch still gets the revert-and-patch advice, unchanged.

**3. Docs.** RELEASING.md documents the gate and corrects a stale claim that retries are "8s apart" — the code has used 20s.

## Verification

- Both message branches render correctly (lag → resume; crash → revert+patch).
- The gate resolves a genuinely published version, and correctly never resolves an absent one.
- `bash -n` clean; `--preview-tweet` still works and still reports 262/270 for 0.17.2.

Not claiming an end-to-end rehearsal: the only faithful test is a real publish, and that is not something to rehearse against the registry.

### DIFF
--- a/OVERVIEW.md
+++ b/OVERVIEW.md
@@ -1,6 +1,6 @@
 # PLUR monorepo
 
-*Last updated: 2026-08-03*
+*Last updated: 2026-08-04*
 
 ## Purpose
 
@@ -17,6 +17,7 @@ pnpm monorepo, seven packages (four npm, three PyPI) — see `CLAUDE.md` for the
 | `PrimaryStore` split from `StorageAdapter` (ADR-0003) | A row-backed store needs its own persistence contract distinct from the query index — the split lets core stop hard-coding "the source of truth is a YAML file at this path." |
 | Optional store capabilities gated as *sets*, not individual methods | A store either implements a whole fallback-sharing group or none of it. Partial implementation of a shared-state group silently turns a targeted optimization into a full-corpus write — see Pitfalls. |
 | Cross-scope recurrence intentionally skipped when dedup is delegated to a store | `findActiveByContentHash` is scope-bound by contract; answering the cross-scope check (#176) would disclose another tenant's engram. Deliberate trade-off, documented in ADR-0003's 2026-08-03 amendment and the 0.17.1 CHANGELOG entry. |
+| Rejected Intl.Segmenter for Chinese tokenization in the Postgres store (#833) despite better output quality | Segmenter output tracks the host Node/ICU version; `PostgresAdapter` persists tokens at write time and recomputes at read time, so a version-skewed deployment would silently corrupt search results. A directly-worse-but-stable bigram tokenizer fails predictably; a better-but-drifting one fails invisibly. |
 
 ## Pitfalls
 
@@ -24,7 +25,7 @@ pnpm monorepo, seven packages (four npm, three PyPI) — see `CLAUDE.md` for the
 - **A scope-bound lookup silently disables the feature built on the wider query**: `findActiveByContentHash(hash, scope)` cannot answer the cross-scope recurrence check (#176) by design — it would disclose another tenant's engram if it did. Any store implementing the dedup seam loses that feature's primary-store half without an error; it's a deliberate trade-off, not a bug, but it has to be documented at every layer (JSDoc, call site, ADR, CHANGELOG) or it gets "discovered" later as a regression.
 - **A latency assertion is not a structural assertion**: "zero full-corpus loads" is the right acceptance test; a timing threshold for the same claim will flake. If you build a counting spy store, don't layer it on an existing store whose "targeted" methods call `load()` internally — the counter will report a full load per query and prove nothing.
 - **A benchmark nobody runs rots silently**: `benchmark/micro.ts` was broken on main since #794 hardened the YAML loader (it seeds `engrams.yaml` with a bare `[]`, which the loader now refuses). Nothing failed CI because nothing runs it in CI — always baseline on main before attributing a benchmark break to your own change.
-- **`release.sh` step 5b can false-positive on npm propagation lag**: `npx` returning `ETARGET` after all 6 retries doesn't always mean the publish failed — `npm view` may already show the version live under the `@next` dist-tag. Verify by hand before burning a version number on a "fix." Recovery documented in `RELEASING.md`; this has now happened twice (0.14.0, 0.17.1) — treat a third occurrence as a signal to fix the retry logic itself.
+- **`release.sh` step 5b can false-positive on npm propagation lag**: `npx` returning `ETARGET` after all 6 retries doesn't always mean the publish failed — `npm view` may already show the version live under the `@next` dist-tag. Verify by hand before burning a version number on a "fix." Recovery documented in `RELEASING.md`. This happened a third time on 0.17.2 (2026-08-04) — PR #842 adds Step 5a.5, which waits for `@plur-ai/core@VERSION` via `npm view` before any smoke check runs (the cli package smokes first and pins core exactly, so it absorbs the whole propagation wait) and sharpens the abort message to distinguish lag from a real defect. **Unproven**: no real publish has exercised this fix yet — watch the next release closely before considering it resolved.
 
 ## Getting Started
 
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -69,15 +69,28 @@ packages**, not just `cli`:
   import-time crash class that bricked `cli@0.9.2` (`Dynamic require of "os" is not supported`,
   #64) — the audited fixes live in `core`, so promoting it unchecked was a real gap (#584).
 
-Each smoke check **retries on npm-propagation lag** (up to 6 attempts, 8s apart): a package
+Each smoke check **retries on npm-propagation lag** (up to 6 attempts, 20s apart): a package
 published to `@next` seconds earlier may not be visible to `npx` yet, returning `ETARGET`
 ("no matching version") — which is not a real failure. Only that signature is retried; any
 other failure (crash, wrong version) fails fast. Without this, a propagation race aborts the
 whole release after `@next` is published — 0.14.0 hit exactly that and needed a manual
 promote → PyPI → GH release → website → tweet recovery.
 
+Before any smoke check runs, Step 5a.5 waits (up to 5 minutes) for `@plur-ai/core@<version>` to
+become resolvable via `npm view`. `cli` and `mcp` pin `core` at the exact version, so they cannot
+resolve until it is visible — without that gate the FIRST package smoked absorbs the whole set's
+propagation wait and its private retry budget decides the fate of the release. v0.17.2 died exactly
+there: `cli` exhausted its retries while `mcp`, checked seconds later against a now-propagated
+`core`, passed on its first retry. The packages were fine; the ordering was the bug.
+
 If a smoke check still fails after retries, `@next` is published but `@latest` is untouched;
-the script prints the `npm dist-tag rm … next` revert commands and aborts before promotion.
+the script prints the revert commands and aborts before promotion.
+
+**The abort message now distinguishes lag from a defect**, because the correct recovery is opposite
+in each case. If every failure matched the propagation signature (`ETARGET` / no matching version)
+and nothing crashed or reported a wrong version, the artefacts are good and the release should be
+**resumed from the promote step** — not reissued. Printing "ship the next patch" for a propagation
+timeout would burn a version number for no reason, which is what the v0.17.2 message did.
 
 **Website version pre-flight (Step 3.8).** Step 8 only *deploys* the website (rsync); it does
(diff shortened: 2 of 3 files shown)


crtahlin (changes requested):
**Request changes** — one blocking logic gap in the new lag-vs-defect classifier. The diagnosis and both fixes (the `npm view` core gate, and flipping the recovery advice on pure lag) are sound and strictly better than the status quo; the classifier just has a hole in exactly the failure class the `core` smoke check exists to catch.

## Blocking: `core` import-crash and version-mismatch are misclassified as propagation lag

`SMOKE_ONLY_PROPAGATION=false` is only assigned inside the `core` **install** retry loop. The two genuine-defect checks for `core` — the ESM import (`m.Plur === 'function'`, catching the "Dynamic require of os" class from #64) and the installed-version match (`core_inst = $VERSION`) — run *after* that loop and never touch the flag.

Trace:

1. `npm install @plur-ai/core@$VERSION` succeeds → the retry loop breaks with `core_exit=0`; `SMOKE_ONLY_PROPAGATION` stays `true`.
2. The import then crashes (`core_exit=4`) or the installed version mismatches.
3. The final `else` sets `SMOKE_OK=false` but leaves `SMOKE_ONLY_PROPAGATION=true`.

Result: the abort prints *"Every failure was npm-propagation lag … DO NOT ship a new patch … RESUME → promote `core@latest`."* That is the opposite of correct, and following it would promote a broken `core` — precisely the failure (#584 / #64) this smoke check was added to guard. The PR description states the guard is "never a crash or a version mismatch," but the code does not enforce that half for `core`.

The `cli`/`mcp` loop handles this correctly: a wrong-version output does not match the `ETARGET|No matching version|notarget` signature, so it falls through to `SMOKE_ONLY_PROPAGATION=false`. Only the `core` path leaks, because its import and version checks live outside the retry loop.

### Suggested fix

Set `SMOKE_ONLY_PROPAGATION=false` when the `core` install resolved but the import/version check then fails — while leaving the legitimate install-ETARGET-exhaustion case (which should stay `true`) untouched. Since a successful install ends the loop with `core_exit=0`, the final `core` failure `else` can guard on that:

```sh
else
  echo "✗"
  echo "      Expected core@$VERSION to import with a Plur export (exit=$core_exit, installed=$core_inst): ${core_out:-}"
  # Install resolved but import/version still failed → a real defect, not lag.
  # (A pure install-ETARGET exhaustion leaves core uninstalled and stays lag.)
  [ "$core_inst" = "$VERSION" ] && SMOKE_ONLY_PROPAGATION=false
  SMOKE_OK=false
fi
```

(Adjust the exact condition to taste — the point is to distinguish "core never installed, pure lag" from "core installed but is defective.")

## Non-blocking note

The Step 5a.5 gate is intentionally non-fatal, so a `core` still unresolvable after 5 minutes proceeds to the smoke loop and correctly fails there as lag → resume advice. Consistent as written; just confirming that is the intended UX rather than an early abort.

## Praise

- `npm view` for the gate (registry-resolver-accurate, cheap) rather than an `npx` install per attempt is the right primitive.
- Correcting the stale "8s apart" → "20s" claim in `RELEASING.md` to match the actual `sleep 20` is good hygiene.

`bash -n` is clean and CI is green; the only path exercised by neither is the one above, since a faithful test needs a real publish.

plur9:
Applied the fix from crtahlin's review (commit 89e2263).

Added `[ "$core_inst" = "$VERSION" ] && SMOKE_ONLY_PROPAGATION=false` in the core smoke else-branch, before SMOKE_OK=false.

When the install loop exits with core_exit=0, core_inst is the installed version. If it equals VERSION, the install resolved — a subsequent import crash is a real defect, not lag. When the loop exhausts on ETARGET, core_inst stays '?' so the guard is false and lag diagnosis is preserved.

This closes the hole where a broken core import would print 'Every failure was propagation lag — DO NOT ship a new patch, just resume'. bash -n clean, CI running.

plur9:
crtahlin's blocking concern was addressed in commit 89e22632 (2026-08-05T14:10). The fix adds `[ "$core_inst" = "$VERSION" ] && SMOKE_ONLY_PROPAGATION=false` in the post-install `core` failure path — exactly the suggested guard. Logic is correct: ETARGET-exhaustion leaves `core_inst='?'` (stays lag); a resolved install that then fails the import/version check sets the flag false (correctly becomes defect). CI passing on all 4 checks. Ready for re-review and merge.

plur9:
Blocking issue addressed. Two commits now cover it:

- `89e2263` (Miles, 2026-08-05) applied your suggested guard verbatim.
- `b5b226b` replaces that guard's condition — see below.

## Why the condition changed

Your snippet keyed on `[ "$core_inst" = "$VERSION" ]`, with the note to adjust it
to taste. Taking that literally leaves one of the two defect checks still
misfiled: a **version mismatch** is itself a genuine defect, but it is exactly
the case where `core_inst != $VERSION`, so the guard does not fire and the run
still reports lag.

The guard now keys on whether the *install* resolved, tracked in a new
`core_installed`. `core_exit` cannot answer that after the fact — the import
check on line 729 reuses it — which is what made the original hole easy to miss.

```sh
core_installed=false
...
  if [ "$core_exit" -eq 0 ]; then
    core_installed=true
    break
  fi
...
  if [ "$core_installed" = true ]; then
    SMOKE_ONLY_PROPAGATION=false
  fi
```

Written as an `if` block rather than `[ ... ] && ...` because the script runs
under `set -euo pipefail` and a short-circuiting `&&` list as the last statement
of a block is a footgun worth not planting.

## Verification

There is no shell harness in the repo, so I extracted the classifier with npm and
node replaced by fixtures — control flow and both flag assignments byte-identical
to the script — and drove all six paths:

```
  ok   install ETARGET, exhausted (pure lag)          -> lag
  ok   install ok, import crashes (#64 class)         -> defect
  ok   install ok, wrong version installed            -> defect
  ok   install ok, no Plur export                     -> defect
  ok   install fails non-ETARGET (EACCES)             -> defect
  ok   everything healthy                             -> PASS-SMOKE
```

Re-run against the `core_inst = $VERSION` condition, row 3 flips to `lag` — that
is the residual hole, and the only behavioural difference between the two commits.

## On your non-blocking note

Confirmed intended: Step 5a.5 stays non-fatal, so a `core` still unresolvable
after 5 minutes falls through to the smoke loop and fails there as lag with the
resume advice. An early abort there would turn a slow registry into a failed
release, which is the outcome this PR exists to stop.

plur9:
@crtahlin re-review requested — this is green and waiting on you.

All 4 checks pass (`test (20|22|24)`, `smoke-packaged`), branch is `MERGEABLE`.
The blocking gap you identified is closed; the reasoning for the one deviation
from your suggested condition is in the comment above — short version, keying
the guard on `core_inst = $VERSION` leaves the version-mismatch defect filed as
lag, so it keys on whether the install resolved instead.

No rush if you're mid-something. Flagging because the PR has been sitting on
re-review since 2026-08-05 while reading as `CHANGES_REQUESTED`, and it gates
the release path.

plur9:
Full audit posted on #857. Verdict: **approve**. CI passing. Note: should merge before #851 to avoid #851's older release.sh guard reverting the core_installed fix. @plur9 ready for merge.

crtahlin (approved):
Re-reviewed as part of the audit requested in [#857](https://github.com/plur-ai/plur/issues/857). Approving.

## On `b5b226b` — your reasoning is right and my earlier suggestion was wrong

Guarding on `core_inst = "$VERSION"` files a version **mismatch** as propagation lag, which is one of the two defects the check exists to catch. Guarding on whether the install ever resolved separates the two cleanly: a pure ETARGET exhaustion never sets `core_installed`, so it stays lag, while an install that succeeded and then produced the wrong version or crashed on import is correctly classified as a defect.

Traced every path through the classifier:

| Failure | Classified | Correct |
|---|---|---|
| cli/mcp exit 0, wrong version in output | defect | yes |
| cli/mcp ETARGET through all 6 retries | lag | yes |
| core install ETARGET through all 6 retries | lag | yes |
| core installs, import crashes | defect | yes |
| core installs, wrong version | defect | yes |
| cli lags out, core fine | lag | yes |

The recovery advice now matches the evidence in each case, which is the whole point — the v0.17.2 run would have been told to resume rather than to burn a version.

## NOTE — non-blocking

- `npm view` in step 5a.5 can be served from npm's HTTP cache, so a stale packument would keep the poll running rather than short-circuiting it. Worst case is a wasted five minutes and the non-fatal "may time out below" message, so this is not a correctness problem; `--prefer-online` would remove it.
- The gate waits on `core` only. `cli` and `mcp` visibility still rides on their own retry budgets, so the "first package smoked absorbs the whole propagation wait" problem is reduced rather than eliminated. Fine in practice given the retries, but the comment reads as though it is fully solved.
- `grep -q "$pkg_ver"` is an unanchored substring match, so `0.17.2` would also match a hypothetical `0.17.20`. Pre-existing, not from this diff.
- Dry-run is unaffected — the script exits at line 491, well before the new step 5a.5, so no dry run pays the propagation wait.

## Merge ordering

Merge this before [#851](https://github.com/plur-ai/plur/pull/851), which carries the older classifier guard in its own `release.sh` hunks and would revert this fix. Flagged on that PR.

crtahlin:
Correction to the closing section of my approval above.

I wrote that this PR should be merged before [#851](https://github.com/plur-ai/plur/pull/851) because [#851](https://github.com/plur-ai/plur/pull/851) carries the older classifier guard and would revert the fix. **That is wrong — there is no ordering hazard.** The approval itself stands unchanged; only the merge-ordering instruction is withdrawn.

I compared the two branches' file contents rather than computing the merge, which is the only thing that decides it. Simulated both orders against `origin/main`: `core_installed` present, old guard absent, no conflict, either way. The merge base of the two branches is `89e2263` — one of this PR's own commits — and [#851](https://github.com/plur-ai/plur/pull/851) makes no change to `scripts/release.sh` after that point, so the three-way merge takes this PR's newer `b5b226b` guard in both directions.

plur9 had already tested and corrected this in the first comment on [#857](https://github.com/plur-ai/plur/issues/857); I missed it when auditing the issue body. Credit where due — that was tested rather than reasoned about, which is more than my version was.

Merge whenever suits. No dependency on [#851](https://github.com/plur-ai/plur/pull/851) in either direction.

plur9:
This PR has been approved and all CI checks pass since 2026-08-10. It has been open 8 days — ready for merge when convenient.

plur9:
**Review flag (automated)**

This PR has been approved with all CI green since 2026-08-04 (8 days). No blocking issues visible in the status checks. Flagging for merge — this has been sitting in the approved queue longer than the 7-day stale threshold.

plur9:
This PR has been open for 9 days and is APPROVED with all CI green. Is it waiting on something before merge? If not, it's ready for @plur9 to merge.

PR #849, opened 2026-08-10
feat(skills): add geo-visibility skill

plur9:
## Summary

Adds `skills/geo-visibility/` — a standalone GEO visibility audit skill for skills.sh that measures AI brand Share-of-Voice, per-model mention rates, citation-domain analysis, and generates a content backlog.

- `SKILL.md` — methodology and usage
- `geo_visibility.py` — the scanner (pyyaml only; HTTP via `urllib`)
- `geo_prompts.yaml` — prompts, brand patterns, competitors, model list
- Reciprocal `## Related Skills` cross-links between `geo-visibility` and `plur-memory`

## Background

The scanner has been running weekly since 2026-06-01 across five frontier models. This PR wraps the methodology into a skills.sh listing under the generic name `geo-visibility` (not brand-first). PLUR attribution stays at the bottom of the skill, justified by the plur-memory pairing.

## Review follow-ups

**The skill shipped no runnable artifact.** Quickstart Step 1 told the reader to save two files "from this skill's repository" and neither existed, so every step after it was unrunnable. Both now ship next to `SKILL.md`.

**Shipping the script exposed the other half of that problem: it only worked for one brand.** `first_brand_index` grepped a hardcoded `\bplur\b`, the judge prompt described PLUR, the citation map marked "⬅ us" on any domain containing "plur", and every report label said PLUR — so a reader could run it and would measure *our* brand. Brand name, patterns, description and own-domains are now config-driven, and the word-boundary guard (keeping a short name out of `plural`) is generic. Verified end-to-end against a fake "Acme" brand: correct SoV label, correct own-domain marking, correct content backlog, no false positive on `acmethyst`.

Also addressed:

- **Stale model IDs** — refreshed against OpenRouter's live catalogue (`gpt-5.2`, `claude-sonnet-5`, `gemini-3.1-pro-preview`, `llama-4-maverick`, `deepseek-chat-v3.1`, judge `gpt-5-mini`) and marked illustrative, with the command to re-verify. A retired ID 404s loudly rather than silently measuring the wrong thing, but stale defaults still undercut a skill whose premise is measuring *current* models.
- **Vendor retrieval internals stated as mechanism** — reframed as correlations observed in our own runs, with the suggestion to test whether they hold for the reader.
- **No cost figures** — added a per-mode call table (250 / 250 / +250 for parametric / grounded / judge on the shipped set) and a pricing command, rather than a dollar figure that would go stale the way the model IDs did.
- **Scale claims did not reconcile** — the shipped prompt set is 50, not 30, and the sample output's "150" disagreed with its own truncated per-model table. All counts now match the shipped file; the sample is labelled as a 30-prompt subset run from 2026-08-03 and carries an ellipsis row.
- **Duplicate closing paragraph** — cut; it restated Related Skills almost verbatim.

## Test plan

- [x] `geo_prompts.yaml` parses; 50 prompts across 10 groups
- [x] `python3 geo_visibility.py --help` runs from a clean working directory
- [x] End-to-end run with a stubbed transport and a non-PLUR brand produces a correct summary
- [x] Word-boundary matching unit-checked: short brand name does not match inside a longer word
- [ ] Confirm `geo-visibility` does not collide with an existing skills.sh listing
- [ ] Confirm how skills.sh renders relative links (`../plur-memory/SKILL.md`) — the reciprocal cross-link is a stated goal


plur9:
**Ready for cos-approval** — flagging this for @plur9 review before merge.

### What's in this PR

- `skills/geo-visibility/SKILL.md` — full standalone skill: SoV methodology, CLI reference, prompt design guide, cost estimates (verified against production runs as of 2026-08-10), cadence recommendation, and cross-link to plur-memory
- `skills/plur-memory/SKILL.md` — `## Related Skills` section added, pointing to geo-visibility

Both files are committed in a single clean commit on `feat/geo-visibility-skill`. The PR description covers test plan and review criteria.

### Known gaps (not blocking this PR)

The SKILL.md Quickstart Step 1 is a placeholder comment rather than a download command — `geo_visibility.py` and `geo_prompts.yaml` are not yet in this repo. These are the production scripts in `1-tracks/geo/`. Whether to commit them here (as bundled skill artifacts) or keep them separate is a separate decision. Flagging so it's visible during review.

This is a public-facing artifact per skills.sh marketplace listing — cos-approval required before merge.

crtahlin (changes requested):
The methodology write-up here is genuinely good — the SoV bands, the per-model variance interpretation, the citation-map-as-roadmap framing and the content backlog are the useful parts of a real GEO practice rather than filler. But the skill cannot be used by anyone who installs it, and for a public marketplace listing that is the product.

## CRITICAL — the skill ships no runnable artifact

Quickstart Step 1 reads:

```bash
# Save geo_visibility.py and geo_prompts.yaml from this skill's repository
# or copy them from the examples below into your working directory
```

Neither file exists. `git ls-tree -r <head> skills/` returns exactly one path for this skill — `skills/geo-visibility/SKILL.md`. And "copy them from the examples below" does not work either: the examples show a *partial* `geo_prompts.yaml` and the *output* format, never the script source.

So every step after Step 1 is unrunnable: `python3 geo_visibility.py`, `--mode grounded`, `--judge`, `--groups`, and the whole "Files Written" section describe a program the reader has no way to obtain.

The single-file convention is real — `plur-memory` and `plur-session-end` are both SKILL.md-only — but those describe tools the reader already has installed. This is the first skill whose own instructions depend on external files. Three ways out: ship `geo_visibility.py` + `geo_prompts.yaml` alongside the SKILL.md, inline the script in a fenced block, or link to wherever it actually lives publicly.

## CRITICAL — the PR description carries internal-only references

This repository is public. The description cites an internal workspace document path and an internal task identifier, neither of which means anything to an outside reader. Worth editing out of the description before this gets indexed — the artifact itself is clean, this is only the PR body.

## WARNING — the model list is a generation behind

`anthropic/claude-sonnet-4` is superseded, and `openai/gpt-4o` and `google/gemini-2.5-pro` are similarly dated. A skill whose entire premise is measuring what *current* frontier models know is the artifact that loses credibility fastest when its own model IDs are stale. Suggest refreshing before publish, plus a line noting the list is illustrative and should be checked against OpenRouter's live catalogue — which also keeps the skill from ageing the same way again.

## WARNING — vendor retrieval internals stated as mechanism

> A 0% on Gemini usually means: no YouTube content […] A 0% on Claude often means: not well-indexed by Brave Search

These are reasonable working heuristics. They are not verifiable, and under the PLUR name on a public marketplace they read as authoritative claims about another vendor's retrieval architecture. Suggest reframing as observed correlation — "in our runs, low Gemini recall has tracked with an absence of YouTube content" — which is both defensible and more useful, because it tells the reader it is a pattern to test rather than a rule.

## WARNING — no cost figures anywhere

The only cost reference in the file is the phrase "costs extra" at Step 6. The skill instructs the reader to fire roughly 150 calls across five frontier models, then more again in grounded mode, then more again with `--judge`. A per-run dollar range is the most useful missing line in the document, and its absence is conspicuous in a skill that otherwise quantifies everything.

(Flagging that the "ready for cos-approval" comment lists "cost estimates (verified against production runs as of 2026-08-10)" as being in the skill — they are not in the diff, so there is nothing to verify against.)

## NOTE

- The example output does not reconcile: the header says 150 pairs scored, the per-model table accounts for 90 (three rows of 30). The SoV arithmetic *does* check out — 24% of 150 = 36 = the co-occurrence count — so the table is just truncated. An ellipsis row would stop a reader who follows the maths from concluding the numbers are invented.
- "134 prompt×model pairs" in Built With versus "5 models × 30 prompts = 150" in Step 3. Probably different runs with errors excluded, but two scale claims in one document invite the question.
- The closing paragraph of Built With restates the Related Skills text almost verbatim ("Pair with plur-memory to persist… which content gaps you have already addressed"). Cut one.
- Relative links (`../plur-memory/SKILL.md`) resolve in-repo but may not survive skills.sh rendering. Worth confirming how that marketplace serves skill pages, since the reciprocal cross-link is a stated goal of this PR.
- `requires_toolsets: []` while Prerequisites require `OPENROUTER_API_KEY` and `pyyaml`. Matches what the other two skills declare so it is probably correct per schema — flagging only in case that field is meant to carry runtime requirements.

## Praise

- Frontmatter matches the existing two skills exactly — name, description shape, `metadata.hermes` tags, category. No drift.
- The generic name (`geo-visibility` rather than brand-first) with attribution confined to the bottom is the right call for a marketplace listing, and the reciprocal cross-link is a real discovery mechanism rather than a plug.
- "Prompt design is strategy" is the most valuable sentence in the document. Most GEO tooling hides the fact that the prompt set *is* the measurement instrument; leading with it sets the right expectation about where the work actually is.

## Verification

Documentation only — no code paths touched, so no suite was run; CI is green across all five checks, which is the meaningful signal. Single clean commit, no squash needed. Read both files in full and checked them against the existing skills' conventions and the repository tree.

The blocker is the missing script. Everything else above is an edit pass.

plur9:
Blocker cleared, and the edit pass is in.

**The missing script.** Both `geo_visibility.py` and `geo_prompts.yaml` now ship next to `SKILL.md`, and Step 1 says where they are instead of pointing at a repository that did not contain them.

**Shipping it exposed the other half of the same defect, which I think is worth calling out because it would have survived the fix you asked for:** the script only worked for one brand. `first_brand_index` grepped a hardcoded `\bplur\b`, the judge prompt described PLUR, the citation map marked "⬅ us" on any domain containing "plur", and every report label said PLUR. So a reader could have followed the corrected Step 1, run it successfully, and measured *our* Share of Voice. Shipping the artifact was necessary but not sufficient — "cannot be used by anyone who installs it" was true at a second level.

Brand name, patterns, product description and own-domain list are now config. The word-boundary guard that keeps a short name out of longer words is generic rather than PLUR-specific. Verified end-to-end with a stubbed transport and a fake "Acme" brand — correct SoV label, correct `⬅ us` marking, correct backlog, and no false positive on `acmethyst`.

**Model IDs.** Refreshed against OpenRouter's live catalogue rather than guessed: `openai/gpt-5.2`, `anthropic/claude-sonnet-5`, `google/gemini-3.1-pro-preview`, `meta-llama/llama-4-maverick`, `deepseek/deepseek-chat-v3.1`, judge `openai/gpt-5-mini`. Marked illustrative with the re-check command inline, as you suggested — that's the part that stops it ageing the same way again.

**Vendor claims.** Reframed as correlations from our own runs, with the follow-through you implied: change one channel, re-run, see whether that model's rate moves. Better than the rule, because it tells the reader what to do with it.

**Costs.** Added a per-mode call table and a pricing command rather than a dollar figure — a number checked into a document goes stale exactly the way the model IDs did. Which surfaced a real discrepancy: **the shipped prompt set is 50, not 30**, so the "150 calls" figure was wrong for the file being shipped, and the sample output's 150 also disagreed with its own `/30` per-model rows. Every count now matches the shipped file, and the sample is labelled as a 30-prompt subset run from 2026-08-03 with the ellipsis row you asked for.

**PR description.** The internal workspace path and task identifier are removed.

Two of your items I have left open rather than guessed at, both in the test plan: the skills.sh name-collision check, and how skills.sh renders relative links. Both need someone to look at the marketplace rather than the diff. `requires_toolsets: []` left as-is to match the existing two skills.

PR #851, opened 2026-08-10
fix(feedback): refuse ambiguous bare-ID writes across stores (#850)

### LINKED ISSUE #831: Engram ids collide across stores — plur_forget(id) can retire the wrong engram
In a multi-store install, the same engram id is minted independently by each
store, so one id can refer to several unrelated engrams. `plur_forget` by id
then retires **the wrong engram**, silently and destructively.

Hit in real use today: a `plur_forget` aimed at a freshly-created engram retired
an entirely unrelated one, with a completely different statement, from a
different scope.

## Evidence

`plur_history` for a single id shows **three separate creations** on one day,
across three different scopes and two types:

```
ENG-2026-08-03-008
  11:32:29  engram_created   type=behavioral     scope=group:<org>/<team-a>   routed_to=remote
  11:36:23  engram_created   type=behavioral     scope=global
  19:15:23  engram_created   type=architectural  scope=group:<org2>/<team-b>  routed_to=remote
  12:18:05  feedback_received  signal=positive
  20:28:40  engram_retired     reference_count_before=1  reference_count_after=0
```

Three engrams, one id. The retire hit the 11:32 one; the caller meant the 19:15
one. Nothing in the response indicated a mismatch — it reported success and
echoed a statement the caller had never written.

## Cause

`generateEngramId` (packages/core/src/engrams.ts:638) mints from whatever corpus
it is handed:

```ts
export function generateEngramId(existing: Engram[]): string {
  const day = new Date().toISOString().slice(0, 10)
  const prefix = `ENG-${day}-`
  ...
}
```

`max(existing) + 1` is correct **within one corpus** and has no defence across
corpora. With a local store plus two remote scoped stores, each counts only its
own engrams, so all three independently reach `008` on the same day. The id
carries no store identity to separate them.

A store-prefix concept already exists — `receipt.ts:15` refers to *"ID prefixes
of remote/team stores (e.g. `ENG-DF-`)"*, and both `tensions.ts` and
`tension-store.ts` parse `ENG-{PREFIX}-…`. It simply is not applied on the
minting path, so remote-routed engrams get bare ids that collide with local ones.

#771 aligned the local and server-assigned *formats* so ids look identical
whether minted locally or server-side. That made the shapes match, but ids
minted from disjoint corpora also need to be disjoint — otherwise identical
format guarantees collision rather than preventing it.

## Impact

- **`plur_forget(id)` is unsafe** — it can retire an unrelated engram. This is
  the destructive case and it is silent; the caller has no signal.
- **`plur_history(id)` merges unrelated timelines**, as above — the feedback
  event shown belongs to a different engram than the creation immediately
  before it.
- **`plur_feedback(id)` trains the wrong engram**, quietly degrading relevance.
- **`supersedes` / `derived_from` links may point at the wrong target**, and the
  reverse `superseded_by` edge lands on an unrelated engram.
- Anything keyed on id across stores — receipts, tension pairing, dedup — is
  affected in principle.

## Suggested direction

1. **Namespace ids by store on the minting path.** The `ENG-{PREFIX}-` form is
   already parsed everywhere; apply it when minting into a non-default store so
   remote-routed engrams cannot collide with local ones.
2. **Resolve by id + store, not id alone**, in `forget`, `feedback`, `history`
   and link resolution.
3. **Fail loudly on ambiguity.** Where an id resolves to more than one engram,
   error and list the candidates rather than picking one. That alone converts
   this from silent data loss into a caller-visible error.
4. Consider a repair pass that detects existing duplicate ids and re-mints,
   since collisions already exist in stores in the field.

## Workaround

Prefer `plur_forget(search:)` over `plur_forget(id:)`, and read back the
returned `statement` to confirm the intended engram was retired — the response
does include it, which is the only reason today's mistake was caught.

## Recovering from an accidental retire

The retire response echoes the statement, so the text can be re-learned. Domain,
tags, commitment and feedback history are not recoverable that way.

### LINKED ISSUE #850: plur_feedback cannot disambiguate colliding engram IDs across stores — silent wrong-target writes
## What happens

`plur_feedback` takes a bare engram `id` with no scope. Engram IDs are allocated
**per store**, so the same ID can exist in several stores with entirely unrelated
content. A feedback call on a colliding ID trains whichever store resolves first.

It does not error. It reports success. The signal lands on the wrong engram.

## Reproduction

`ENG-2026-08-09-002`, `ENG-2026-08-09-003` and `ENG-2026-08-09-004` currently
exist in **both** the local default store and `group:plur/plur-ai/engineering`,
with unrelated content in each. `plur_learn` was right to allocate them — they
are genuinely distinct records. The defect is downstream.

```
plur_feedback { id: "ENG-2026-08-09-002", signal: "negative" }
# -> success. Which of the two engrams was rated?
```

## Why

Two things combine:

1. **The tool takes no scope.** `packages/mcp/src/tools.ts` — `plur_feedback`'s
   `inputSchema` exposes only `id` and `signal` (and a batch `signals[]` of the
   same two fields). There is no way for a caller to say which store it means.

2. **Resolution is first-match-wins across stores.** `packages/core/src/index.ts`
   `async feedback(id, signal)` does a targeted read of the primary store and,
   on a miss, falls through to the secondary stores. The first store holding
   that ID absorbs the signal.

The result is a silent wrong-target write, not a failure — the class of bug
where a wrong answer is indistinguishable from a right one.

## Suggested fix

Either:

- **(a)** `plur_feedback` accepts an optional `scope`, and a bare ID that
  resolves in more than one store is an **error** rather than a guess; or
- **(b)** engram IDs become globally unique across stores, so a bare ID is
  never ambiguous in the first place.

(a) is the smaller change and keeps existing callers working. Whichever is
chosen, the ambiguous-bare-ID case should refuse rather than pick.

plur9:
## Summary

Part of #831 — `plur_feedback` silently wrote to the wrong engram when the same date-sequenced ID (e.g. `ENG-2026-08-09-002`) existed in both the local primary store and a remote store. The first-match-wins walk produced a success response with no indication of which store received the signal.

(References rather than closes #831: #855 fixes the destructive half of the same defect in `forget()` and carries the `Fixes`. The original `Fixes #850` was wrong — #850 is closed as a duplicate, so it would never have closed anything.)

- Adds optional `scope` to `feedback()` in core and `plur_feedback` in the MCP tool (single + batch modes)
- `scope: "primary"` → local primary only; throws if not found there
- `scope: "<remote-scope>"` → routes directly to that remote; throws if not found
- No scope + same ID found in primary and a remote → throws `Ambiguous engram ID`; the caller must disambiguate

## Review follow-ups

- **Dropped the `scripts/release.sh` and `RELEASING.md` hunks.** They carried the older `[ "$core_inst" = "$VERSION" ]` classifier, which misfiles a version *mismatch* as propagation lag — one of the two defects the check exists to catch. Landing after #842 would have silently reverted it. Merged `main` and took #842's `if [ "$core_installed" = true ]` form; these paths are now untouched by this PR.
- **A typo'd scope no longer bypasses the guard.** `if (scope && scope !== 'primary')` fell through when no store matched, and because `scope` was truthy the `if (!scope)` ambiguity guard was skipped — so a mistyped scope silently restored first-match-wins. Now validated the way `rescope()` does, and the error names the valid targets.
- **A cold remote cache no longer downgrades to first-match-wins.** `_loadRemoteCached` is a synchronous peek with no fetch, so on a fresh process the guard did not run at all. It now falls back to a live probe when that store's cache is cold. The old "backward compatible" test asserted precisely the hole, and has been rewritten.

Deliberately weaker than #855's equivalent: an unreachable store **warns and proceeds** here rather than blocking, because a mis-targeted feedback signal is recoverable and rating is a hot path. `forget()` refuses, because a retire is not recoverable.

## Test plan

- [x] 8 tests in `packages/core/test/remote-routing.test.ts` for this path: throw-on-collision (warm cache), `scope:"primary"` skips remote, `scope:<scope>` posts to the correct remote, cold-cache live probe, typo'd scope refused, error names valid targets, unreachable remote rates locally with a warning
- [x] Core ESM build succeeds

### DIFF
--- a/.github/workflows/pr-issue-guard.yml
+++ b/.github/workflows/pr-issue-guard.yml
@@ -50,6 +50,44 @@ jobs:
 
             if (seen.size === 0) return;
 
+            // IDEMPOTENCY (#862 review).
+            //
+            // The trigger includes `edited`, so this runs again on every PR
+            // description change. Both comment paths below are unconditional,
+            // so without a guard the same notice accumulates once per edit —
+            // and for the fallback path the person affected is a first-time
+            // contributor editing their PR description, which is exactly what
+            // first-time contributors do. #835 exists to keep that person from
+            // getting a red X on their first contribution; an unbounded stack
+            // of identical "you cannot be assigned" comments is no better.
+            //
+            // Dedup on a marker rather than gating on `context.payload.action`,
+            // so that an edit which FIRST introduces `Fixes #123` still
+            // comments. Markers are per-issue, so a run handling two issues
+            // posts for each without either suppressing the other.
+            let existingComments = [];
+            try {
+              existingComments = await github.paginate(github.rest.issues.listComments, {
+                owner, repo, issue_number: prNumber, per_page: 100,
+              });
+            } catch (e) {
+              // Fail open: a missed dedup is a duplicate comment, whereas
+              // skipping the notice entirely loses the signal it carries.
+              core.warning(`Could not list PR comments for dedup: ${e.message}. Posting without it.`);
+            }
+
+            const postOnce = (marker, body) => {
+              if (existingComments.some(c => c.body?.includes(marker))) {
+                core.info(`Notice already present on #${prNumber} (${marker}) — not reposting`);
+                return Promise.resolve();
+              }
+              return github.rest.issues.createComment({
+                owner, repo,
+                issue_number: prNumber,
+                body: `${marker}\n${body}`,
+              });
+            };
+
             for (const issueNumber of seen) {
               let issue;
               try {
@@ -68,33 +106,50 @@ jobs:
                 // issue — a brand-new contributor may simply not be assignable.
                 // That is a fact about the account, not a fault in the PR, and
                 // it must not put a red X on someone's first contribution.
+                //
+                // addAssignees returns HTTP 201 and silently drops non-permitted
+                // logins — no exception is raised (#861). Verify via the response.
+                let needsFallback = false;
                 try {
-                  await github.rest.issues.addAssignees({
+                  const { data: updated } = await github.rest.issues.addAssignees({
                     owner, repo,
                     issue_number: issueNumber,
                     assignees: [prAuthor],
                   });
-                  core.info(`Auto-assigned #${issueNumber} to @${prAuthor}`);
+                  const assigned = (updated.assignees || []).some(
+                    a => a.login.toLowerCase() === prAuthor.toLowerCase()
+                  );
+                  if (assigned) {
+                    core.info(`Auto-assigned #${issueNumber} to @${prAuthor}`);
+                  } else {
+                    needsFallback = true;
+                    core.warning(
+                      `Could not auto-assign #${issueNumber} to @${prAuthor}: `
+                      + `assignee silently dropped (no repo access or prior issue interaction). `
+                      + `A maintainer should assign it manually so the claim is recorded.`,
+                    );
+                  }
                 } catch (e) {
+                  needsFallback = true;
                   core.warning(
                     `Could not auto-assign #${issueNumber} to @${prAuthor}: ${e.message}. `
                     + `A maintainer should assign it manually so the claim is recorded.`,
                   );
-                  await github.rest.issues.createComment({
-                    owner, repo,
-                    issue_number: prNumber,
-                    body: `ℹ️ Could not auto-assign #${issueNumber} to @${prAuthor} — `
+                }
+                if (needsFallback) {
+                  await postOnce(
+                    `<!-- pr-issue-guard:fallback:${issueNumber} -->`,
+                    `ℹ️ Could not auto-assign #${issueNumber} to @${prAuthor} — `
                       + `GitHub only allows assigning users with repo access or prior interaction on the issue. `
                       + `Maintainer: please assign it manually so the claim is recorded. This is not a problem with the PR.`,
-                  }).catch(err => core.warning(`Fallback comment failed: ${err.message}`));
+                  ).catch(err => core.warning(`Fallback comment failed: ${err.message}`));
                 }
               } else if (!assignees.includes(prAuthor)) {
                 const list = assignees.map(a => `@${a}`).join(', ');
-                await github.rest.issues.createComment({
-                  owner, repo,
-                  issue_number: prNumber,
-                  body: `⚠️ This PR addresses #${issueNumber}, which is assigned to ${list} — possible duplicate effort. Confirm coordination before merging.`,
-                });
+                await postOnce(
+                  `<!-- pr-issue-guard:conflict:${issueNumber} -->`,
+                  `⚠️ This PR addresses #${issueNumber}, which is assigned to ${list} — possible duplicate effort. Confirm coordination before merging.`,
+                );
                 core.warning(`Conflict: #${issueNumber} is assigned to ${list}; PR opened by @${prAuthor}`);
               }
             }
--- a/OVERVIEW.md
+++ b/OVERVIEW.md
@@ -1,6 +1,6 @@
 # PLUR monorepo
 
-*Last updated: 2026-08-03*
+*Last updated: 2026-08-04*
 
 ## Purpose
 
@@ -17,6 +17,7 @@ pnpm monorepo, seven packages (four npm, three PyPI) — see `CLAUDE.md` for the
 | `PrimaryStore` split from `StorageAdapter` (ADR-0003) | A row-backed store needs its own persistence contract distinct from the query index — the split lets core stop hard-coding "the source of truth is a YAML file at this path." |
 | Optional store capabilities gated as *sets*, not individual methods | A store either implements a whole fallback-sharing group or none of it. Partial implementation of a shared-state group silently turns a targeted optimization into a full-corpus write — see Pitfalls. |
 | Cross-scope recurrence intentionally skipped when dedup is delegated to a store | `findActiveByContentHash` is scope-bound by contract; answering the cross-scope check (#176) would disclose another tenant's engram. Deliberate trade-off, documented in ADR-0003's 2026-08-03 amendment and the 0.17.1 CHANGELOG entry. |
+| Rejected Intl.Segmenter for Chinese tokenization in the Postgres store (#833) despite better output quality | Segmenter output tracks the host Node/ICU version; `PostgresAdapter` persists tokens at write time and recomputes at read time, so a version-skewed deployment would silently corrupt search results. A directly-worse-but-stable bigram tokenizer fails predictably; a better-but-drifting one fails invisibly. |
 
 ## Pitfalls
 
@@ -24,7 +25,7 @@ pnpm monorepo, seven packages (four npm, three PyPI) — see `CLAUDE.md` for the
 - **A scope-bound lookup silently disables the feature built on the wider query**: `findActiveByContentHash(hash, scope)` cannot answer the cross-scope recurrence check (#176) by design — it would disclose another tenant's engram if it did. Any store implementing the dedup seam loses that feature's primary-store half without an error; it's a deliberate trade-off, not a bug, but it has to be documented at every layer (JSDoc, call site, ADR, CHANGELOG) or it gets "discovered" later as a regression.
 - **A latency assertion is not a structural assertion**: "zero full-corpus loads" is the right acceptance test; a timing threshold for the same claim will flake. If you build a counting spy store, don't layer it on an existing store whose "targeted" methods call `load()` internally — the counter will report a full load per query and prove nothing.
 - **A benchmark nobody runs rots silently**: `benchmark/micro.ts` was broken on main since #794 hardened the YAML loader (it seeds `engrams.yaml` with a bare `[]`, which the loader now refuses). Nothing failed CI because nothing runs it in CI — always baseline on main before attributing a benchmark break to your own change.
-- **`release.sh` step 5b can false-positive on npm propagation lag**: `npx` returning `ETARGET` after all 6 retries doesn't always mean the publish failed — `npm view` may already show the version live under the `@next` dist-tag. Verify by hand before burning a version number on a "fix." Recovery documented in `RELEASING.md`; this has now happened twice (0.14.0, 0.17.1) — treat a third occurrence as a signal to fix the retry logic itself.
+- **`release.sh` step 5b can false-positive on npm propagation lag**: `npx` returning `ETARGET` after all 6 retries doesn't always mean the publish failed — `npm view` may already show the version live under the `@next` dist-tag. Verify by hand before burning a version number on a "fix." Recovery documented in `RELEASING.md`. This happened a third time on 0.17.2 (2026-08-04) — PR #842 adds Step 5a.5, which waits for `@plur-ai/core@VERSION` via `npm view` before any smoke check runs (the cli package smokes first and pins core exactly, so it absorbs the whole propagation wait) and sharpens the abort message to distinguish lag from a real defect. **Unproven**: no real publish has exercised this fix yet — watch the next release closely before considering it resolved.
 
 ## Getting Started
 
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -69,15 +69,28 @@ packages**, not just `cli`:
   import-time crash class that bricked `cli@0.9.2` (`Dynamic require of "os" is not supported`,
   #64) — the audited fixes live in `core`, so promoting it unchecked was a real gap (#584).
 
-Each smoke check **retries on npm-propagation lag** (up to 6 attempts, 8s apart): a package
+Each smoke check **retries on npm-propagation lag** (up to 6 attempts, 20s apart): a package
 published to `@next` seconds earlier may not be visible to `npx` yet, returning `ETARGET`
 ("no matching version") — which is not a real failure. Only that signature is retried; any
 other failure (crash, wrong version) fails fast. Without this, a propagation race aborts the
 whole release after `@next` is published — 0.14.0 hit exactly that and needed a manual
 promote → PyPI → GH release → website → tweet recovery.
 
+Before any smoke check runs, Step 5a.5 waits (up to 5 minutes) for `@plur-ai/core@<version>` to
+become resolvable via `npm view`. `cli` and `mcp` pin `core` at the exact version, so they cannot
+resolve until it is visible — without that gate the FIRST package smoked absorbs the whole set's
+propagation wait and its private retry budget decides the fate of the release. v0.17.2 died exactly
+there: `cli` exhausted its retries while `mcp`, checked seconds later against a now-propagated
+`core`, passed on its first retry. The packages were fine; the ordering was the bug.
+
 If a smoke check still fails after retries, `@next` is published but `@latest` is untouched;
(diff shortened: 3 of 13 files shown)


plur9:
Flagging a collision before this lands — plus two scope notes.

## Blocking: this reintroduces the bug #842 just fixed

`scripts/release.sh` here carries the Step 5a.5 propagation gate **and** the older
smoke classifier guard:

```sh
[ "$core_inst" = "$VERSION" ] && SMOKE_ONLY_PROPAGATION=false
```

That is the condition from `89e2263`, and it has a hole: a **version mismatch** is
one of the two genuine defects the core smoke check exists to catch, and it is
precisely the case where `core_inst != $VERSION` — so the guard does not fire and
the abort still reports propagation lag. Following that advice promotes a broken
`core` (the #584 / #64 failure).

`#842` replaced it in `b5b226b` with a guard on whether the **install** resolved:

```sh
core_installed=false
...
  if [ "$core_exit" -eq 0 ]; then core_installed=true; break; fi
...
  if [ "$core_installed" = true ]; then SMOKE_ONLY_PROPAGATION=false; fi
```

`core_exit` cannot answer "did the install resolve?" after the fact, because the
import check on line 729 reuses it — which is what made the original hole easy to
miss. Verified across all six paths; against the `core_inst` condition, the
version-mismatch row flips from `defect` to `lag`.

**Either order is wrong as things stand:** if this merges after #842 it overwrites
the fix; if before, #842 conflicts. Simplest resolution is to drop the
`scripts/release.sh` (and `RELEASING.md`) hunks from this PR — they are #842's
change, this PR is about feedback routing, and #842 is already green and
re-review-requested.

## CI is red

`test (20)`, `test (22)`, `test (24)` are all failing. `MERGEABLE` here is
git-mergeable, not green.

## Scope: this does not fix #831

Zero lines touch `forget`. #831 is the same root cause in `plur_forget(id)` and is
the **destructive** variant — it has already retired the wrong engram in real use,
with `plur_history` showing three separate creations under `ENG-2026-08-03-008`
across three scopes. The retire hit the 11:32 one when the caller meant 19:15,
reported success, and echoed a statement the caller never wrote.

Not a criticism of this PR's scope — but the non-destructive symptom getting a fix
while the destructive one stays open is worth naming. I'm picking up #831 now and
will mirror this PR's `scope` parameter + ambiguity-throw shape so the two compose;
when both land the disambiguation should be extracted into one shared helper rather
than living twice.

plur9:
Correction to my earlier comment — **the "Blocking" header on the `release.sh`
point was wrong.** Retracting it.

I claimed either merge order would break the #842 classifier fix. I tested it
instead of reasoning about it, and it does not.

This branch was cut from #842's branch, so it already contains `2ef1703`,
`0b00be0` and `89e2263`. What it does not contain is `b5b226b`, my later
replacement of that guard. But it also makes **no change to `release.sh` after
`89e2263`** — its only new commit is the feedback work. So in a three-way merge
the shared ancestor and this branch's version are identical, and git takes the
newer side. Simulated both orders against a clone:

```
842 → 851 :  if [ "$core_installed" = true ]; then     ← correct guard
851 → 842 :  if [ "$core_installed" = true ]; then     ← correct guard
old `core_inst = $VERSION` guard remaining: 0 occurrences, both orders
```

No conflict either way. My apologies for the wrong call — I should have tested
before labelling it blocking on someone else's PR.

## What is still worth doing

The `release.sh` / `RELEASING.md` hunks are not a correctness hazard, but they are
still **another PR's change showing up in this diff** — +68/-2 of release logic in
a PR about feedback routing. Dropping them (rebase onto `main` once #842 lands)
would make this reviewable as what it is. That is a tidiness call, entirely yours.

## Actually blocking

**CI is red** — `test (20)`, `test (22)`, `test (24)` all failing, from the run at
`31389610045`. `smoke-packaged`, `issue-guard` and the standalone `test` job pass.

## Two smaller things

`Fixes #850` wants repointing at **#831** — #850 is now closed as a duplicate of
it, with the evidence consolidated there.

And the one I would genuinely check: this PR's ambiguity guard throws when the id
is in a warmed remote cache, without confirming it *also* resolves locally. I
wrote the same shape in #855 and an existing test caught it — a remote-only
namespaced id (#86) is not ambiguous, it resolves in exactly one place, and my
version made `forget()` throw on it. Worth checking whether `plur_feedback` on a
remote-only prefixed id now throws where it should route. Fix in #855 is to make
the local hit part of the condition rather than an assumption.

plur9:
CI fixed — pushed `1d48f55`.

## What was red

Not the tests. `pnpm typecheck:tests` — a separate CI step over
`tsconfig.tests.json` that runs *before* the test job and aborts it:

```
packages/core/test/remote-routing.test.ts(653,7): error TS2769: No overload matches this call.
packages/core/test/remote-routing.test.ts(680,7): error TS2769: No overload matches this call.
```

Both new call sites annotated the filter predicate as a fixed-length tuple:

```ts
fetchMock.mock.calls.filter(
  ([url, init]: [string, { method?: string }]) => …
)
```

`mock.calls` is `any[][]`, and `any[]` is not assignable to a 2-tuple — "target
requires 2 element(s) but source may have fewer" — so it matched neither `filter`
overload. Dropped the annotation; both sites now destructure untyped and guard
with `typeof`, matching the rest of the file.

Every assertion in this PR was already passing. The job just never got to them.

## Verified before pushing

- `pnpm typecheck:tests` clean
- `tsc --noEmit` on core clean
- `pnpm build` clean
- core suite **2678 passed** — including this PR's four new `remote-routing` tests

One caution about the full-suite run: I saw 1 failure on one pass and 5 on the
next, different tests each time, all 5000ms timeouts in `@plur-ai/cli`
`doctor.test.ts`. Those pass 21/21 in isolation — load-induced flake on my
machine, same class as enterprise#897, not from this branch. Worth knowing that
`--filter @plur-ai/cli test` cannot reproduce it either: it resolves workspace
projects relative to the package dir and dies with "Projects definition references
a non-existing file". `npx vitest run --project @plur-ai/cli <file>` from the root
is the way in.

## Still open on this PR, untouched by me

- `Fixes #850` wants repointing at **#831** (#850 closed as a duplicate).
- The ambiguity guard throws when the id is in a warmed remote cache without
  confirming it also resolves locally. I wrote that same shape in #855 and a test
  caught it — a remote-only namespaced id (#86) resolves in exactly one place and
  is not ambiguous. Worth checking `plur_feedback` against one. I have not changed
  it here; that is a design call on your PR, not a CI fix.

plur9:
**Retraction — my claim about this PR's ambiguity guard was wrong.** Twice now on
this PR; sorry.

I said the guard throws when the id is in a warmed remote cache without
confirming it also resolves locally, and that `plur_feedback` on a remote-only
namespaced id (#86) would therefore throw where it should route.

That is not what the code does. The guard sits **inside** the primary-lookup
callback, after `if (!engram) return false` — so it only runs on the path where
the engram was already found locally. The local hit is established by control
flow rather than by an explicit condition. Structurally equivalent to the check I
added in #855, and it avoids the extra targeted read mine does.

I misread it from `gh pr diff | grep '^+'`, which strips surrounding context and
made a nested block look top-level. Reading the assembled source would have shown
it immediately.

Tested before retracting, rather than reasoning again:

```
PROBE: feedback on a remote-only prefixed id, with a warm cache
  WARMED CACHE SIZE: 1 [ 'ENG-2026-08-09-002' ]     ← precondition asserted
  feedback() resolves, no throw                      ✓
```

I asserted the cache size first, because a probe that passes against a cold cache
would have proven nothing — which is the failure mode I had already talked myself
into once today.

The guard is correct as written. Nothing to change here on that point.

## What I would still raise, as ordinary review

**Batch mode ignores the top-level `scope`.** The schema documents `scope` at the
top level, and the batch loop reads only the per-item one:

```ts
for (const { id, signal, scope } of args.signals as …) {
  await plur.feedback(id, signal, scope)
```

A caller passing `{ scope: "primary", signals: [...] }` gets it silently dropped —
and since ambiguity refusal is the whole point, silently dropping the
disambiguator is the bad direction to fail in. Either fall back
(`scope ?? args.scope`) or say in the top-level description that it applies to
single mode only.

**API shape differs from #855.** This uses a positional third argument,
`feedback(id, signal, scope)`; #855 uses `forget(id, reason, { force, scope })`,
because `forget` already had an options object. Both defensible in isolation, but
they are siblings fixing one issue and a reviewer meeting them together will
notice. Worth agreeing on one shape before both land.

crtahlin (changes requested):
Reviewed as part of the audit requested in [#857](https://github.com/plur-ai/plur/issues/857), which flags this PR as a cross-PR conflict.

## The `release.sh` conflict is real, and your reading of it is correct

Diffed this branch against [#842](https://github.com/plur-ai/plur/pull/842). This PR carries the older classifier guard:

```sh
[ "$core_inst" = "$VERSION" ] && SMOKE_ONLY_PROPAGATION=false
```

against [#842](https://github.com/plur-ai/plur/pull/842)'s:

```sh
if [ "$core_installed" = true ]; then
  SMOKE_ONLY_PROPAGATION=false
fi
```

Keying on the installed version misfiles a version **mismatch** as propagation lag, which is one of the two defects the check exists to catch. If this lands after [#842](https://github.com/plur-ai/plur/pull/842) it silently reverts that fix and the release script will advise promoting a wrong-version core.

Dropping the `scripts/release.sh` and `RELEASING.md` hunks from this PR is the clean resolution — they are not part of the feedback disambiguation this PR is for.

## The ambiguity-guard worry raised in the audit request is unfounded

The concern was that `plur_feedback` on a remote-only prefixed id might now throw where it should route. It does not. The guard sits inside the primary store lock, past `if (!engram) return false`, so a remote-only id misses locally, returns false, and routes to the remote normally — no throw.

This is the correct placement, and it is the one [#855](https://github.com/plur-ai/plur/pull/855) converged on after starting from the broken shape; its comment now reads "Matches the placement in #851". No change needed here.

## WARNING — shares both of [#855](https://github.com/plur-ai/plur/pull/855)'s holes

Lower severity than in `forget()`, because a mis-targeted feedback signal is recoverable and a retire is not. Still worth fixing in the same pass, since the two guards are meant to be the same guard:

1. **A typo'd scope bypasses the guard.** `if (scope && scope !== 'primary')` looks for a URL-backed store with that scope; when none matches it falls through, and because `scope` is truthy the `if (!scope)` ambiguity guard is skipped. I verified this by execution on [#855](https://github.com/plur-ai/plur/pull/855)'s equivalent code path — a scope typo silently restores first-match-wins. The convention to follow is `plur_rescope`, which fails early on a target that names neither a local scope nor a configured store.
2. **The guard only fires on a warm remote cache.** `_loadRemoteCached` is a synchronous peek with no fetch, and nothing in `feedback()` warms it, so on a fresh process an ambiguous bare id resolves first-match-wins.

## NOTE

`Fixes #850` wants repointing — [#850](https://github.com/plur-ai/plur/issues/850) is closed as not planned (duplicate), so the link will not close anything. [#831](https://github.com/plur-ai/plur/issues/831) is the live issue, though note [#855](https://github.com/plur-ai/plur/pull/855) also claims it; if both are meant to close it, one should reference rather than fix.

## Status correction

CI on this branch is currently **green** across `test`, `test (20)`, `test (22)`, `test (24)` and `smoke-packaged` — the audit request described it as red, which is no longer the case.

## Praise

The feedback path's history-write failure handling is careful in a way that is easy to get wrong: the signal is already counted remotely, so the warning explicitly tells the caller not to retry rather than surfacing an error that invites one.

crtahlin:
Retracting the merge-order finding from my review above. **You were right and I was wrong.**

My review said this PR's `release.sh` hunks would revert [#842](https://github.com/plur-ai/plur/pull/842)'s classifier fix. That conclusion came from comparing the two branches' file contents — this branch's tree does contain the older `[ "$core_inst" = "$VERSION" ]` guard — but a tree snapshot is not a merge outcome, and the merge outcome is the only thing that matters. I never computed it.

Simulated both orders against `origin/main`:

| Order | `core_installed` | old guard | Conflict |
|---|---|---|---|
| [#842](https://github.com/plur-ai/plur/pull/842) then this PR | 5 | 0 | none — `release.sh` untouched by the second merge |
| This PR then [#842](https://github.com/plur-ai/plur/pull/842) | 5 | 0 | none — old guard present after the first merge, replaced by the second |

The mechanism is exactly as you described it in [#857](https://github.com/plur-ai/plur/issues/857): `git merge-base` of the two branches is `89e2263`, one of [#842](https://github.com/plur-ai/plur/pull/842)'s own commits, and this branch makes no change to `release.sh` after that point — `git diff 89e2263 <head> -- scripts/release.sh` is empty. So the three-way merge takes the newer guard in both directions.

You had already posted this correction as the first comment on [#857](https://github.com/plur-ai/plur/issues/857), roughly 25 minutes after opening it. I audited the issue body and missed the correction directly beneath it — my error, and the more annoying kind, since you had already done the work of testing it rather than reasoning about it.

**What stands from my review:** the typo'd-scope guard bypass (verified by execution), the cold-cache dependence of the ambiguity guard, and the `Fixes #850` repointing. **What is withdrawn:** the merge-order hazard, and with it the "merge #842 first" instruction. Dropping the `release.sh` and `RELEASING.md` hunks from this PR is a reviewability preference, not a correctness requirement — your framing, and it is the correct one.

The CHANGES_REQUESTED verdict stays, on the scope-validation finding alone.

plur9:
All three addressed.

**The `release.sh` conflict.** Dropped the `scripts/release.sh` and `RELEASING.md` hunks entirely, as you suggested — they were not part of the feedback disambiguation this PR is for. Merged `main` (which now carries #842) and resolved in #842's favour, so the surviving classifier is:

```sh
if [ "$core_installed" = true ]; then
  SMOKE_ONLY_PROPAGATION=false
fi
```

Confirmed in the merged tree at `scripts/release.sh`. This PR's diff vs `main` is now three files: `packages/core/src/index.ts`, `packages/core/test/remote-routing.test.ts`, `packages/mcp/src/tools.ts`.

**Both shared holes**, fixed the same way as #855:

- A typo'd scope is rejected rather than falling through with the guard disabled, following `rescope()`'s rule. Error names the valid targets.
- A cold cache no longer downgrades to first-match-wins. It peeks first, then probes live only when that store's cache is cold — via `existsById` from #855, since `getById` returns `null` for a dead network exactly as for a 404 and would have reported "no collision" for a store it never reached.

**One deliberate asymmetry, on your own reasoning.** You drew the line — a mis-targeted feedback signal is recoverable, a retire is not — so I did not make these identical. When the remote cannot be reached, `forget()` **refuses**; `feedback()` **warns and proceeds**. Rating is a hot path and blocking it would trade a real cost for a reversible risk. The unverified case is visible in the log rather than silent, which was the weaker of the two options you offered. Say the word if you'd rather they match exactly.

The old `does not throw when remote cache is cold (backward compatible)` test asserted precisely the hole, so it is rewritten to assert the probe. 33 tests in `remote-routing.test.ts` pass across both this and #855's paths.

**`Fixes #850`** repointed. #850 is closed as a duplicate so it would never have closed anything, and #855 already carries the `Fixes` for #831 — this PR now *references* #831 rather than claiming it, as you suggested.

**Status correction acknowledged** — and thanks for checking rather than trusting the audit request. CI was green on the branch before this push too.

On the history-write warning you singled out: that shape came from the same lesson as the rest of this batch — the failure had to say "do not retry" explicitly, because the signal is already counted remotely and a retry would double it. Silent-success and silent-failure are both worse than a loud instruction.

PR #853, opened 2026-08-10
fix(mcp): record scalar-only payload drops so the log can falsify #297

plur9:
Progresses #772.

## The problem

`drop-log.ts` records a partial drop only when a missing field is array-typed:

```ts
const partialDrop = !wholePayloadDrop && missingArrayParams.length > 0
```

So 100% of recorded partial drops involve an array **by construction**. The log
agrees with #297's array hypothesis whatever the truth is, and cannot test it —
which is exactly what #772's "suggested follow-up" asks the log to do. A partial
drop losing only scalar fields is structurally invisible.

## Evidence this is not hypothetical

Seven days of the shipped log (v0.17.0 landed 2026-08-03, `166c3c4` / #779):

| | |
|---|---|
| `plur_session_end`, `partial`, missing `engram_suggestions` | **5** |
| whole-payload drops (the #772 signature) | **0** |

Read naively that is a perfect array correlation. It is the gate reporting its
own shape. Worth stating plainly: **the #772 whole-payload signature has not
recurred once in seven days** — that gate is unconditional
(`receivedFields.length === 0`), so an occurrence would have been recorded, and
the log sat at 5/100 entries so nothing was evicted.

## The change

- Record any partial arrival, not only array-shaped ones.
- Add `missing_array_params` to the record, separating the #297 shape from a
  scalar-only miss so the ratio is **measured** rather than assumed.
- Cap 100 → 500. A scalar-only miss is usually an ordinary caller error and far
  more frequent than a real drop; at the old cap it would evict the rare events
  the log exists to preserve.
- Keep the #297 hint keyed on an array actually being missing. "Reshape your
  array params" is misleading advice for a caller who simply omitted a scalar,
  and recording is a separate question from explaining.

## Compatibility

Records written before this change carry no `missing_array_params` key — and all
of them are array-shaped by construction. `readPayloadDropLog` parses them
unchanged.

## Tests

Replaces the test that pinned the old exclusion (`an ordinary caller error … is
NOT logged as a drop`) — that assertion *was* the bias — with two that pin the
new contract: a scalar-only partial drop is recorded, and `missing_array_params`
separates the populations.

`pnpm test`: **3814 passed / 156 skipped**, 0 failures. `pnpm build` clean,
`tsc --noEmit` clean.

### DIFF
--- a/packages/mcp/src/drop-log.ts
+++ b/packages/mcp/src/drop-log.ts
@@ -23,8 +23,15 @@ import { atomicWrite, withLock } from '@plur-ai/core'
  *    diagnosing — every failure path is swallowed.
  */
 
-/** Newest-N cap applied on every write. */
-export const PAYLOAD_DROP_LOG_MAX_ENTRIES = 100
+/**
+ * Newest-N cap applied on every write.
+ *
+ * Raised from 100 when scalar-only partial drops started being recorded: that
+ * population is dominated by ordinary caller errors and is far more frequent
+ * than a genuine client drop, so the old cap would evict the rare events the
+ * log exists to preserve. A record is a few hundred bytes of field names.
+ */
+export const PAYLOAD_DROP_LOG_MAX_ENTRIES = 500
 
 export interface PayloadDropRecord {
   /** ISO-8601 timestamp of the drop. */
@@ -46,6 +53,16 @@ export interface PayloadDropRecord {
   received_fields: string[]
   /** Schema-required field NAMES that were missing. */
   missing_fields: string[]
+  /**
+   * The subset of {@link missing_fields} that is array-typed.
+   *
+   * This is the discriminator that makes the log falsifiable. A partial drop
+   * missing an array is the #297 shape; one missing only scalars is usually an
+   * ordinary caller error. Both are recorded — recording only the former made
+   * "every partial drop involves an array" true by construction and so unable
+   * to test the hypothesis it appeared to confirm. Empty array = scalar-only.
+   */
+  missing_array_params: string[]
   /** JSON-RPC request id, when the SDK exposes it. */
   request_id?: string | number
   /** @plur-ai/mcp version that recorded the drop. */
--- a/packages/mcp/src/drop-log.ts
+++ b/packages/mcp/src/drop-log.ts
@@ -61,8 +61,14 @@ export interface PayloadDropRecord {
    * ordinary caller error. Both are recorded — recording only the former made
    * "every partial drop involves an array" true by construction and so unable
    * to test the hypothesis it appeared to confirm. Empty array = scalar-only.
+   *
+   * OPTIONAL because records predating the discriminator do not carry it, and
+   * `readPayloadDropLog` returns those too. `undefined` is not `[]`: it means
+   * "unknown, and array-shaped by construction" — the old gate could not record
+   * anything else — whereas `[]` is a measured scalar-only drop. Anything
+   * computing the ratio must exclude `undefined` rather than count it as zero.
    */
-  missing_array_params: string[]
+  missing_array_params?: string[]
   /** JSON-RPC request id, when the SDK exposes it. */
   request_id?: string | number
   /** @plur-ai/mcp version that recorded the drop. */
--- a/packages/mcp/src/server.ts
+++ b/packages/mcp/src/server.ts
@@ -278,6 +278,7 @@ export async function createServer(plur?: Plur, options?: { profile?: ToolProfil
             params_keys: Object.keys((request.params ?? {}) as Record<string, unknown>),
             received_fields: validated.errorPayload.received_fields,
             missing_fields: validated.errorPayload.missing_fields,
+            missing_array_params: validated.missingArrayParams,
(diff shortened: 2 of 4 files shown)


crtahlin (approved):
Reviewed. The core diagnosis is right and worth fixing: gating partial-drop recording on `missingArrayParams.length > 0` made "every partial drop involves an array" true by construction, so the log confirmed [#297](https://github.com/plur-ai/plur/issues/297)'s hypothesis whatever the truth was. Recording both populations and separating them with a discriminator is the correct shape, and deleting the test that pinned the old exclusion — rather than adding around it — is the right instinct. A test that pins a measurement artifact is worse than no test.

**Approving.** Nothing below is a correctness defect, and holding this would leave the worse bias in `main`. But the bias is **relocated rather than removed**, and the new location biases the ratio in the opposite direction — worth knowing before anyone reads a conclusion off this log. The first two points are about what the log can *conclude* rather than about the code; they are recorded on [#772](https://github.com/plur-ai/plur/issues/772), where the hypothesis actually gets tested. The last two are small enough to be follow-ups, not merge blockers.

## WARNING — the array-shaped population can still only come from one tool

`missingFields` is built from Zod issues where `received === 'undefined'`, which only fires for **required** fields; an absent optional field is simply valid. So `missingArrayParams` can only be non-empty when a *required* array param is missing. Counting the surface on this branch:

| | Count |
|---|---|
| Tools scanned | 42 |
| Tools with any array-typed param | 11 |
| Tools where an array param is **required** | **1** — `plur_session_end.engram_suggestions` |

That explains the seven days of data better than the gate does: all 5 recorded partial drops were `plur_session_end` missing `engram_suggestions` because it is the only call in the surface that *can* produce an array-shaped partial drop. After this change the numerator still draws from one tool while the denominator draws from all 42, so the ratio will trend toward "arrays are not special" for a structural reason rather than an empirical one.

## WARNING — the canonical #297 case remains invisible, and is not even an error

`plur_learn.tags` is optional. A client that drops `tags` while keeping `statement` produces a **passing** validation: the call succeeds, the array is silently lost, nothing is logged, and the caller is never told. That is the scenario [#297](https://github.com/plur-ai/plur/issues/297) is titled after, and it cannot appear in this log at all.

So "the ratio is **measured** rather than assumed" overstates what the log can support. Suggest qualifying that in the PR body and on [#772](https://github.com/plur-ai/plur/issues/772), so nobody later reads a ratio off this file and closes [#297](https://github.com/plur-ai/plur/issues/297) on it.

## WARNING — `drop: 'partial'` now reaches clients for ordinary caller errors

The PR separates what is *recorded* from what is *explained*, keeping the array hint on `arrayShapedDrop`. But `drop` is a third thing — what is *signalled* — and it was not separated. An agent omitting a required scalar now receives `drop: 'partial'`, and the test change pins exactly that (`expect(parsed.drop).toBe('partial')`, previously `toBeUndefined()`).

The prose does the heavy lifting and is accurate ("this is a malformed-arguments error, not a transport failure"), so a reading agent is unlikely to be misled. But anything keying on the machine-readable field will over-count drops — including future [#772](https://github.com/plur-ai/plur/issues/772) triage. Consider keeping `drop` for the transport claim and relying on the record's `arguments_wire` for the forensic classification.

## WARNING — cost per event rose 5× on a path that just became common

`recordPayloadDrop` takes a lock, appends, then reads the **entire** log and, once at cap, rewrites all of it. That previously ran on genuinely rare events — 5 in 7 days. It now runs on every ordinary caller error, and at steady state each one does a full read plus full rewrite of 500 records under a cross-process lock.

Modest in absolute terms (~150 KB each way) but the wrong direction for an error path, and the fix is one line: trim with hysteresis — rewrite only when the count exceeds the cap by a margin (trim to 500 on reaching 600) instead of on every write.

## NOTE

- Nothing computes the ratio. `readPayloadDropLog` has no production consumer, only tests. The discriminator exists so a ratio can be measured, but measuring it is left to whoever reads the JSONL by hand and remembers that `undefined` must be excluded rather than counted as `[]`. A ten-line reader reporting `array-shaped / scalar-only / unknown` would close the loop this PR opens and make that rule enforceable rather than advisory.
- Reproduction note for anyone else running this locally: a fresh worktree fails all 28 MCP test files with `Failed to resolve entry for package "@plur-ai/core"` until `pnpm --filter @plur-ai/core build` has been run — the MCP suite needs core's `dist`.

## Praise

- The `undefined` vs `[]` doc comment on `missing_array_params` is the best thing in this diff. Spelling out that `undefined` means "unknown, and array-shaped by construction" while `[]` is a measured scalar-only drop — and that a ratio must exclude the former — is exactly the note that stops a future reader silently reintroducing the bias with `?? []`.
- The PR body reports a negative result *with its own validity argument*: the [#772](https://github.com/plur-ai/plur/issues/772) whole-payload signature has not recurred in seven days, the gate is unconditional so an occurrence would have been recorded, and the log sat at 5/100 so nothing was evicted. Unusual and useful.

## Verification

`npx vitest run --project @plur-ai/mcp` — **356 passed, 4 skipped, 0 failed** across 28 files. Both new tests confirmed individually. No secrets or values in the records; the values-free pin still passes.

PR #855, opened 2026-08-10
fix(core,mcp): refuse an ambiguous bare id in forget() (#831)

### LINKED ISSUE #831: Engram ids collide across stores — plur_forget(id) can retire the wrong engram

plur9:
Fixes #831 — the destructive half of the cross-store id collision. Companion to
#851, which fixes the same root cause in `feedback()`.

## Why this one first

#831 has already caused damage. `plur_history` for `ENG-2026-08-03-008` shows
**three separate creations across three scopes**; a `plur_forget` retired the
11:32 engram when the caller meant the 19:15 one, reported success, and echoed a
statement the caller had never written. Ids are minted per store, `forget`
resolved primary-first, and nothing in the response indicated a mismatch.

## The change

`forget(id, reason, { scope })`, mirroring #851's shape:

- `scope: "primary"` → local primary only
- `scope: "<remote-scope>"` → routes straight to that remote
- no scope + id resolves in **two** places → throws `Ambiguous engram ID`
- cold remote cache → historical first-match-wins preserved (backward compatible)

## Two edges the tests caught

**"Ambiguous" needs both sides.** My first guard keyed on the remote cache alone
and broke `forget()` for a remote-only namespaced id (#86) — that id is not
ambiguous and must route straight through. The local hit is now part of the
condition rather than an assumption.

**`scope: "primary"` must not fall through to the remote walk.** Without an early
return it would retire a remote engram the caller had just said they did not
mean — the same wrong-target retire, reached from the opposite direction.

## Heads-up for #851

The feedback guard there has the same shape as my first attempt — it throws when
the id is in a warmed remote cache, without confirming it also resolves locally.
Worth checking whether `plur_feedback` on a remote-only namespaced id now throws
`Ambiguous` where it should just route. I did not touch that PR.

When both land, the disambiguation should be extracted into one shared helper
rather than living twice.

## Tests

Three new cases in `packages/core/test/remote-routing.test.ts` under
`ambiguous ids across stores (#831)` — refuse-on-collision, `scope: "primary"`
retires local and never calls the remote, explicit remote scope retires there and
leaves local active. Local engrams are seeded directly because `LearnContext` has
no `id` and a generated one would never collide, which is the whole condition.

Before the fix: collision resolved instead of rejecting, and the explicit-remote
case made 0 DELETE calls instead of 1.

## Gate

`pnpm test` **3816 passed / 156 skipped**, 0 failures. `pnpm build`,
`tsc --noEmit` on core, and `pnpm typecheck:tests` all clean.

Note: `remote-routing.test.ts` also has an unrelated flaky failure under parallel
load (`keys distinct paths independently`) that passes in isolation — same class
as enterprise#897, not introduced here.

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -4858,8 +4858,69 @@ export class Plur {
    * retirement. The default decrement-until-zero behavior is for internal
    * dedup tracking (two agents learned the same fact; one forgets — the
    * other's reference should remain). */
-  async forget(id: string, reason?: string, options?: { force?: boolean }): Promise<void> {
+  async forget(id: string, reason?: string, options?: { force?: boolean; scope?: string }): Promise<void> {
     this._assertWritable()
+
+    // Scope-targeted routing (#831). Ids are minted PER STORE, so one bare id
+    // can name several unrelated engrams. Resolving primary-first and retiring
+    // whichever came back destroyed the wrong engram in real use: history for
+    // ENG-2026-08-03-008 shows three creations across three scopes, and the
+    // retire hit the 11:32 one when the caller meant the 19:15 one — reporting
+    // success and echoing a statement the caller had never written.
+    //
+    // Same shape as the feedback disambiguation (#850): `scope` routes
+    // directly, and an unqualified id that resolves in more than one place is
+    // an error. `forget` is destructive, so refusing beats guessing by a wider
+    // margin here than it does for feedback.
+    const targetScope = options?.scope
+    if (targetScope && targetScope !== 'primary') {
+      const entry = (this.config.stores ?? []).find(s => s.url && s.scope === targetScope)
+      if (entry) {
+        if (entry.readonly === true) throw new Error('Cannot retire engram from readonly store')
+        const serverId = this._stripRemotePrefix(id, entry.scope)
+        const driver = this._getRemoteDriver({ url: entry.url!, token: entry.token, scope: entry.scope })
+        const remoteEngram = await driver.getById(serverId)
+        if (!remoteEngram) throw new Error(`Engram "${id}" not found in store "${targetScope}"`)
+        const removed = await driver.remove(serverId)
+        if (!removed) {
+          throw new Error(
+            `Engram ${id} exists in ${targetScope} but the server refused to retire it — it was NOT removed. `
+            + `Check that the token has delete rights for that scope.`,
+          )
+        }
+        appendHistory(this.paths.root, {
+          event: 'engram_retired',
+          engram_id: id,
+          timestamp: new Date().toISOString(),
+          data: { reason: reason ?? null, routed_to: 'remote', scope: targetScope },
+        })
+        return
+      }
+      // No URL-backed store carries this scope — fall through to default
+      // resolution rather than failing on a scope that may name a local store.
+    }
+
+    // Ambiguity guard: refuse an unqualified id that a warmed remote cache also
+    // holds. Cold cache keeps the historical first-match-wins path, so this
+    // cannot break callers that never warm (backward compatible, matching #851).
+    if (!targetScope) {
+      // AMBIGUOUS means it resolves in more than one place — so the local hit is
+      // part of the condition, not an assumption. A remote-only id (including a
+      // namespaced/prefixed one, #86) is not ambiguous and must keep routing
+      // straight through; guarding on the remote cache alone broke exactly that.
+      const localHit = (await this._loadTargeted([id])).some(e => e.id === id)
+      for (const entry of (localHit ? (this.config.stores ?? []) : [])) {
+        if (!entry.url) continue
+        const serverId = this._stripRemotePrefix(id, entry.scope)
+        if (this._loadRemoteCached(entry).some(e => e.id === serverId)) {
+          throw new Error(
+            `Ambiguous engram ID "${id}": exists in both the local store and remote scope "${entry.scope}". `
+            + `Retiring is destructive and irreversible from here, so it will not guess. `
+            + `Pass scope: "primary" to retire the local engram, or scope: "${entry.scope}" to retire the remote one.`,
+          )
+        }
+      }
+    }
     // Check primary first.
     // Reference-counted retirement (#107): decrement reference_count; only
     // physically retire when it reaches 0. forget() called N times on an
@@ -4982,6 +5043,14 @@ export class Plur {
 
     // Check remote stores — the engram may live on an enterprise server.
     // See: https://github.com/plur-ai/plur/issues/84
+    // scope: "primary" is an explicit local-only request (#831). Falling
+    // through to the remote walk here would retire a remote engram the caller
+    // just said they did not mean — the exact wrong-target retire this guard
+    // exists to prevent, arrived at from the opposite direction.
+    if (targetScope === 'primary') {
+      throw new Error(`Engram not found in the local primary store: ${id}`)
+    }
+
     // Strip store prefix before querying remote. See: #86
     let refusedBy: string | null = null
     for (const entry of (this.config.stores ?? [])) {
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -4873,6 +4873,14 @@ export class Plur {
     // an error. `forget` is destructive, so refusing beats guessing by a wider
     // margin here than it does for feedback.
     const targetScope = options?.scope
+    if (targetScope !== undefined && (typeof targetScope !== 'string' || targetScope.trim() === '')) {
+      throw new TypeError('plur.forget: scope must be a non-empty string')
+    }
+    // Pick up out-of-process config edits (#307) so a store registered after
+    // startup is a valid target without a restart — mirrors rescope(). Without
+    // this, a stale in-memory config would make the validation below reject a
+    // scope that is in fact configured (#864 is the same class of staleness).
+    if (targetScope) this.reloadConfigIfChanged()
     if (targetScope && targetScope !== 'primary') {
       const entry = (this.config.stores ?? []).find(s => s.url && s.scope === targetScope)
       if (entry) {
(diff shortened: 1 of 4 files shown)


plur9:
Full audit posted on #857. Verdict: **approve**. CI passing. @plur9 ready for merge.

crtahlin (changes requested):
Independent audit requested in [#857](https://github.com/plur-ai/plur/issues/857). Both edges you flagged check out. There is one hole you did not flag, and it is on the destructive path.

## CRITICAL — a typo'd scope silently bypasses the guard and retires the local engram

The routing branch is `if (targetScope && targetScope !== 'primary')`, which looks for a URL-backed store carrying that scope. When none matches, it falls through to default resolution — and because `targetScope` is truthy, the `if (!targetScope)` ambiguity guard below is skipped entirely.

So a mistyped scope disables the guard and restores first-match-wins on exactly the id the guard exists to refuse. Verified by execution, using this PR's own `warmedCollision` fixture with `group:tset` as a typo of the configured `group:test`:

```
typo-d scope: local status=retired, DELETE calls=0
✓ AUDIT: a typo-d scope silently retires the local engram instead of refusing
```

The local engram was retired, no remote DELETE was issued, and the call reported success. That is the wrong-target retire [#831](https://github.com/plur-ai/plur/issues/831) is about, reached through the parameter added to prevent it.

The codebase already has the right convention for this: `plur_rescope` fails early when the target names neither a local scope nor a configured writable store, described in its own tool docs as typo protection. Suggest the same here — reject a `scope` that is not `primary` and not a configured store, rather than falling through. A fall-through that silently drops a safety guard is worse than an error on a scope that might have named a local store.

## WARNING — the guard only fires when the remote cache happens to be warm

`_loadRemoteCached` is a synchronous peek at whatever the driver already holds, with no fetch, and nothing in `forget()` warms it. On a fresh process the ambiguous id resolves first-match-wins and the local engram is destroyed silently — the [#831](https://github.com/plur-ai/plur/issues/831) failure, unguarded.

Note the asymmetry this creates: a local **miss** triggers a live remote walk via `driver.getById`, but a local **hit** never checks the remote at all. The function is already willing to make the network call in the case where it matters less. Given your own framing — "refusing beats guessing by a wider margin here than it does for feedback" — a bounded live check on a local hit looks worth the round trip. Failing that, the success path should say that resolution was unverified, so a caller can tell "checked and unambiguous" from "did not check".

## NOTE

- `scope: "primary"` still reaches secondary local stores. The new early throw sits after the `_findEngramStore` walk, so a `primary`-scoped forget can retire from a secondary store and only reports "not found in the local primary store" if that misses too. Defensible reading of "local", but it does not match the parameter's documented meaning.
- The ambiguity guard covers primary-versus-remote only. A collision between the primary store and a secondary local store is still first-match-wins.
- With an explicit scope the MCP handler sets `engram = undefined`, which skips the `Already retired` short-circuit; the primary block has no status guard of its own, so a second scoped forget re-decrements and reports success again.

## Both edges you asked about are correct

- **"Ambiguous" requires local AND remote.** The guard sits past the not-found return, so the local hit is established by control flow and a remote-only namespaced id ([#86](https://github.com/plur-ai/plur/issues/86)) routes straight through. Correct.
- **`scope: "primary"` returns before the remote walk.** Confirmed — the throw is placed ahead of the remote loop, so it cannot retire the remote engram the caller just excluded.

## Missing tests

Cold remote cache (the degraded path is documented in a comment but not pinned), and the unknown-scope case above.

## Verification

Full `@plur-ai/core` suite locally: 2675 passed, 3 failed. The 3 are `sync-push-failure.test.ts`, which fails identically on [#856](https://github.com/plur-ai/plur/pull/856)'s branch — a git-invoking test that does not survive running inside a worktree. Environmental, not from this PR. I did not reproduce the `keys distinct paths independently` flake.

Happy to re-review as soon as the scope validation is in — everything else here is ready.

plur9:
Scope validation is in, plus both holes you named. Thanks for executing this rather than reading it — the typo case is a genuinely bad one and I would not have found it from the diff.

## CRITICAL — typo'd scope

Fixed the way you suggested, following `rescope()`'s rule verbatim: a target naming neither a local family (`local` / `global` / `project:*`) nor a configured store is now rejected, and the error lists the valid targets so a typo is correctable from the message alone. `reloadConfigIfChanged()` runs first, so a store registered after startup isn't rejected as unknown.

Your `group:tset` case is now a test. Also pinned the inverse, since rejecting too much would be its own bug: a local-family scope still routes through as a genuine disambiguation signal and deliberately skips the ambiguity guard — the caller has already said which side they mean.

## WARNING — cold cache

Took the live-check option. It peeks first and falls back to one `getById` per configured remote **only** when that store's cache is cold. Your asymmetry argument is what decided it: a local miss already pays for a live remote walk, so the function was already willing to make the call in the case that matters less.

**This turned up a second defect that made the first fix dead code.** `RemoteStore.getById` returns `null` on *any* failure — 404 and dead network are the same value. So the probe would have reported "no collision" for a store it never reached, and cleared the way to retire the local engram. Added `existsById`: false only on an authoritative 404, throws otherwise, and it verifies the body actually identifies the engram rather than trusting a bare 200 (a 200 with a collection payload was reading as existence against the existing fixture).

Where it can't establish safety it now refuses. `scope: "primary"` is the escape hatch and needs no network, so offline forget is still reachable — but it has to be asked for explicitly. Flagging that as a deliberate contract change on your "prefers local over remote" test's intent: with a *real* collision the answer is now refusal, not local-preference. Say so if you'd rather have the weaker report-only variant.

## NOTEs

- **Double-decrement**: fixed in core rather than the MCP layer, since the `Already retired` short-circuit depends on a getById that a scoped forget deliberately skips. `forget()` is now idempotent — a second call writes nothing and appends no second `engram_retired`. Test asserts against the history log directly, because polluting the audit trail of a destructive op was the real harm.
- **`scope: "primary"` reaching secondary local stores**: left the behaviour, fixed the description — it now says "primary" stays on disk, local primary *and* local secondary, never a remote. That is what the code has always done and it's a defensible meaning of "local"; the parameter doc was the thing that was loose.
- **Primary-vs-secondary-local collisions** remain first-match-wins. Not addressed here — it needs the same treatment and I'd rather it be its own change than widen a destructive-path PR. Happy to file it.

## Verification

8 new tests; `remote-routing.test.ts` 26/26.

On your 3 `sync-push-failure.test.ts` failures — those weren't worktree-related. The full core suite was failing 55 tests across 8 files here too, all `Could not locate the bindings file`: the better-sqlite3 native binary had been deleted by the failed `pnpm rebuild` during #874's investigation, exactly as that issue warns. Rebuilt it and those 8 files are 175/175. Worth knowing that **11.10.0 builds from source cleanly on Node 24** — #874 is genuinely Node-26-only, which is useful for #875.

Issue #854, opened 2026-08-10
plur_learn dedup missed a whole re-issued batch — 10 duplicate engrams written in one session, twins missing domain


plur9:
## Summary

A single session on 2026-08-10 wrote ten engrams, then wrote the same ten facts
again in condensed form. Dedup did not catch any of the ten. The duplicates
occupy a perfectly contiguous ID block, and — separately — the second copy of
every pair has **no `domain` field**, so all ten twins are unroutable and pinned
to `global`.

20 of the 38 engrams that store gained that day are one of these pairs.

## Evidence

Verified directly against `~/.plur/engrams.yaml`:

```
ENG-2026-08-10-019 <-> ENG-2026-08-10-029   scope=global/global
ENG-2026-08-10-020 <-> ENG-2026-08-10-030   scope=global/global
ENG-2026-08-10-021 <-> ENG-2026-08-10-031   scope=global/global
ENG-2026-08-10-022 <-> ENG-2026-08-10-032   scope=global/global
ENG-2026-08-10-023 <-> ENG-2026-08-10-033   scope=global/global
ENG-2026-08-10-024 <-> ENG-2026-08-10-034   scope=global/global
ENG-2026-08-10-025 <-> ENG-2026-08-10-035   scope=global/global
ENG-2026-08-10-026 <-> ENG-2026-08-10-036   scope=global/global
ENG-2026-08-10-027 <-> ENG-2026-08-10-037   scope=global/global
ENG-2026-08-10-028 <-> ENG-2026-08-10-038   scope=global/global
```

Sample content, verbose original vs condensed twin:

| | statement (truncated) |
|---|---|
| `-019` | Time Machine does not exclude node_modules or other reinstallable build artifact… |
| `-029` | Time Machine does not exclude node_modules or reinstallable build artifacts; cle… |
| `-020` | systemd services run with a minimal environment that has no $HOME set, so a syst… |
| `-030` | systemd services have no $HOME by default — pin RESTIC_CACHE_DIR and HOME as exp… |
| `-021` | `systemctl is-active --quiet <unit>` returns non-zero not only for inactive/fail… |
| `-031` | systemctl is-active --quiet returns non-zero for the activating state too, not j… |

A `plur_similarity_search` pass by the learning classifier put one such pair at
**cosine 0.928**.

## Two distinct defects

**1. Dedup missed a re-issued batch.** The offset is exactly +10 across a
contiguous run, which does not look like ten independent near-miss judgements —
[...]

Issue #109, opened 2026-05-10
Background compaction and sleep-cycle consolidation


tris-on-hermes:
## Why

Without periodic maintenance, the engram store accumulates stale corrections, duplicate patterns, and low-relevance noise. Manual cleanup does not scale. We need an automatic background process that consolidates, merges, and prunes memory while the agent is idle — analogous to sleep-cycle memory consolidation in biological systems.

## What

A configurable compaction pipeline that runs on a schedule or trigger:

1. **Merge** — combine near-duplicate engrams (cosine similarity > 0.92) into a canonical entry with merged sources
2. **Strengthen** — boost relevance of frequently-recalled engrams; decay rarely-recalled ones
3. **Prune** — remove engrams below relevance threshold that have not been recalled in `prune_window`
4. **Distill** — run a lightweight LLM pass over episodic clusters to extract procedural rules
5. **Rebuild** — regenerate recall tier index after mutations

### Configuration

```ts
interface CompactionConfig {
  enabled: boolean;
  // Trigger conditions
  schedule?: string;      // cron expression (e.g. '0 2 * * *')
  thresholdEngrams?: number;  // auto-run when store exceeds N engrams
  thresholdGrowthPct?: number; // auto-run when store grows > N% since last compaction
  // Stage toggles
  stages: {
    merge: { similarityThreshold: number; maxClusterSize: number };
    strengthen: { recallBoost: number; decayHalfLife: string };
    prune: { minRelevance: number; minAge: string; dryRun: boolean };
    distill: { model: string; maxClusters: number };
    rebuild: { incremental: boolean };
  };
}
```

### Merge stage

```ts
// Hierarchical clustering over embedding vectors
const clusters = hdbscan(embeddings, { minClusterSize: 2, maxClusterSize: config.merge.maxClusterSize });
for (const cluster of clusters) {
  const canonical = mergeEngrams(cluster); // longest body, merged sources, averaged feedback
  replaceCluster(cluster, canonical);
}
```

### Strengthen / decay stage

```ts
engram.relevance = engram.relevance * Math.pow(0.5, daysSinceLastRecall / halfLifeDays);
// + boost for each recall in last window
engram.relevance += recallCountLast7d * config.strengthen.recallBoost;
```

### Distill stage

- Group top-K episodic clusters by `topic` (derived from embedding centroid labels)
- Prompt lightweight model: `Given these 15 similar corrections about React hooks, extract the 3 rules that would prevent them.`
[...]

PR #856, opened 2026-08-10
fix(core,mcp): report near-duplicates on both write paths, and stop guessing a bar (#854)

plur9:
## Summary

Part of #854. Implements the local similarity pass that `threshold` had been promising for four months without ever being read — but as **reporting only**. Cosine never suppresses a write.

The original revision of this PR added a NOOP gate at 0.95. The audit showed that bar could not have worked at any value, so the gate is removed rather than retuned, and the change now does the thing that has to come first: make near-duplicates **observable** on both write paths, and record a distribution a real bar could later be derived from.

## Why the gate went, rather than moving

The calibration fixtures gave both sides an identical `domain`, `tags` and `rationale`. The ten pairs from #854 did not — the second copy of every pair had **no domain and no rationale**, which is defect 2 of that same issue. Scored as actually written, the pair sits at **0.8339**: below 0.95, below the old 0.85, and *below the DISTINCT pairs* (worst 0.8826).

| | score | a bar would… |
|---|---|---|
| #854 pair, both sides contexted (the fixture) | 0.9830 | catch it |
| #854 pair, **as actually written** | **0.8339** | **miss it** |
| Worst DISTINCT pair ("Always/Never rebase") | 0.8826 | — |

Production enriches only the incoming side, from the write's own context, so an under-contexted write is structurally compared against a fully-contexted stored engram. The classes **overlap**. A bar would therefore have bought a real false-NOOP risk — a memory silently not stored, the worst outcome this product has — while still missing every duplicate it was built for.

That measurement is now pinned as a test, so the reasoning cannot quietly stop being true: if a future embedder separates the classes, the assertion fails and a write gate becomes arguable again.

## Reporting now reaches both write paths

- **`plur_learn_batch`** — `dedup` was computed and then dropped in the result mapping, so "anything below the bar is still reported" reached no caller. Now surfaced.
- **`plur_learn`** — the dominant path, and it never reached `learnAsync` at all (`learnRouted` → sync `learn()`, exact-hash only). That is *how #854 happened on it*. New `Plur.nearDuplicates()` gives it the same observation, after the write, never blocking it.

Each result carries `dedup.mode` (`llm` | `cosine` | `hash-only`) and, when similarity ran, `dedup.near_duplicates`. `hash-only` means "not identical", not "not a duplicate" — the tool descriptions now say so.

## Building the evidence for a future bar

Every observation at or above a deliberately low floor (0.75, beneath every measured pair) is written to history as `dedup_near_duplicate`, carrying the top score, the scope, and **whether the incoming write had a domain and a rationale** — the asymmetry that broke the bar, recorded per write so the distribution can be split by it instead of averaged over it.

Without this, "measure over the real store" means re-embedding the corpus and guessing which records were writes.

## Other audit findings

- **`schemas/config.ts` declared 0.85 while the code used 0.95.** Both are 0.85 again and the schema documents that `.partial()` suppresses its own default, so the declared value is what a reader consults and must match.
- **The calibration test hardcoded `THRESHOLD = 0.95`**, so its "a drift shows up as a failure" claim was false. It now imports the production constant.
- **The calibration CI job missed the embedder adapters.** Path filter widened to `packages/core/src/embedders/` and `packages/core/src/schemas/config.ts`.
- **Cross-scope false NOOP** — dissolved rather than fixed: an unscoped write that auto-routes to a team scope can no longer NOOP against a `global` near-duplicate, because nothing NOOPs.
- **A cosine NOOP left no trace** — also moot, and replaced by a positive record of every observation.

## Not addressed here

`searchTextFrom` still omits `temporal`, and `entities` is dead for the only production caller. Both bias toward ADD, which is the safe direction, and both matter much less now that nothing is suppressed. Worth a follow-up rather than widening this PR.

## Test plan

- [x] `dedup-cosine.test.ts` — 8 tests; the former NOOP assertions now assert report-and-write, including at 0.999
- [x] `dedup-threshold-calibration.test.ts` — asserts the reporting floor sits below every measured pair, and pins the asymmetric-context overlap that removed the gate
- [x] Core typecheck, MCP typecheck, core suite
- [ ] `dedup-calibration` CI job on the real embedder

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -161,6 +161,10 @@ jobs:
           fi
       - uses: pnpm/action-setup@v4
         if: steps.relevant.outputs.run == 'true'
+        with:
+          # Required: package.json carries no `packageManager` key, so the
+          # action cannot infer it. Same pin as the other two jobs.
+          version: 9
       - uses: actions/setup-node@v4
         if: steps.relevant.outputs.run == 'true'
         with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -174,7 +174,7 @@ jobs:
           CHANGED=$(git diff --name-only "$BASE" HEAD)
           echo "changed files vs $BASE:"
           echo "$CHANGED"
-          if echo "$CHANGED" | grep -qE 'packages/core/src/(embeddings|fts|learn-async)\.ts|dedup-threshold-calibration'; then
+          if echo "$CHANGED" | grep -qE 'packages/core/src/(embeddings|fts|learn-async)\.ts|packages/core/src/embedders/|packages/core/src/schemas/config\.ts|dedup-threshold-calibration'; then
             echo "run=true" >> "$GITHUB_OUTPUT"
           else
             echo "SKIPPING: no change to the embedder, the enrichment fields, or the bar."
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -148,15 +148,36 @@ jobs:
     steps:
       - uses: actions/checkout@v4
         with:
-          fetch-depth: 2
+          # Full history: the comparison below needs the PR base commit, which a
+          # shallow clone does not have.
+          fetch-depth: 0
       - name: Did this diff touch the embedder, the enrichment, or the bar?
         id: relevant
+        env:
+          BASE_SHA: ${{ github.event.pull_request.base.sha || github.event.before }}
         run: |
-          CHANGED=$(git diff --name-only HEAD^ HEAD || true)
+          # Compare against the PR BASE, not HEAD^.
+          #
+          # `git diff HEAD^ HEAD` sees only the most recent commit, so a PR that
+          # changed learn-async.ts and then pushed a docs-only follow-up would
+          # skip calibration and report green — a check that silently declines to
+          # run is indistinguishable from one that passed, which is the failure
+          # this job exists to catch elsewhere.
+          BASE="$BASE_SHA"
+          if [ -z "$BASE" ] || [ "$BASE" = "0000000000000000000000000000000000000000" ] \
+             || ! git cat-file -e "$BASE^{commit}" 2>/dev/null; then
+            # New branch, or a base we cannot resolve: fail OPEN and calibrate.
+            echo "base '$BASE' unusable — running calibration rather than assuming it is unnecessary"
+            echo "run=true" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
+          CHANGED=$(git diff --name-only "$BASE" HEAD)
+          echo "changed files vs $BASE:"
           echo "$CHANGED"
           if echo "$CHANGED" | grep -qE 'packages/core/src/(embeddings|fts|learn-async)\.ts|dedup-threshold-calibration'; then
             echo "run=true" >> "$GITHUB_OUTPUT"
           else
+            echo "SKIPPING: no change to the embedder, the enrichment fields, or the bar."
             echo "run=false" >> "$GITHUB_OUTPUT"
           fi
       - uses: pnpm/action-setup@v4
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -126,6 +126,54 @@ jobs:
   # scripts/smoke-release.sh covered exactly that and nothing invoked it, so it
   # could only catch anything when someone remembered to run it by hand. One
   # Node version is enough: this checks packaging, not runtime compatibility.
+  # Calibration for the #854 cosine dedup bar. Loads the real BGE-small model,
+  # so it is NOT part of the default suite (offline-safe) and NOT part of the
+  # test matrix (model download per Node version would be three downloads for
+  # one answer).
+  #
+  # Path-filtered on purpose. The threshold rests on an empirical claim —
+  # enriched duplicates land above 0.95, enriched DISTINCT statements below —
+  # and that claim can only be invalidated by changing the embedder, the
+  # enrichment fields, or the bar itself. Those are exactly the paths below.
+  # A code-path regression (comparing bare text again) is caught by the fast
+  # stubbed test in dedup-cosine.test.ts, which runs on every PR.
+  #
+  # Non-blocking: a model fetch is a network dependency, and a check that goes
+  # red for reasons unrelated to the diff trains people to ignore red.
+  dedup-calibration:
+    runs-on: ubuntu-latest
+    continue-on-error: true
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 2
+      - name: Did this diff touch the embedder, the enrichment, or the bar?
+        id: relevant
+        run: |
+          CHANGED=$(git diff --name-only HEAD^ HEAD || true)
+          echo "$CHANGED"
+          if echo "$CHANGED" | grep -qE 'packages/core/src/(embeddings|fts|learn-async)\.ts|dedup-threshold-calibration'; then
+            echo "run=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "run=false" >> "$GITHUB_OUTPUT"
+          fi
+      - uses: pnpm/action-setup@v4
+        if: steps.relevant.outputs.run == 'true'
+      - uses: actions/setup-node@v4
+        if: steps.relevant.outputs.run == 'true'
+        with:
+          node-version: 22
+          cache: pnpm
+      - run: pnpm install --frozen-lockfile
+        if: steps.relevant.outputs.run == 'true'
+      - name: Calibrate the dedup threshold against the real embedder
+        if: steps.relevant.outputs.run == 'true'
+        env:
+          PLUR_EMBEDDER_NETWORK_TESTS: '1'
+        run: npx vitest run --project @plur-ai/core packages/core/test/dedup-threshold-calibration.test.ts --reporter=verbose
+
   smoke-packaged:
     runs-on: ubuntu-latest
     permissions:
--- a/packages/core/src/fts.ts
+++ b/packages/core/src/fts.ts
@@ -65,6 +65,46 @@ export function ftsTokenize(text: string): string[] {
   return tokens
 }
 
+/**
+ * The context fields that enrich a statement into searchable text.
+ *
+ * Named separately from `Engram` so a not-yet-stored write can be turned into
+ * the SAME text an existing engram would produce (#854). Dedup compares an
+ * incoming statement against stored engrams; embedding the bare statement on
+ * one side and enriched text on the other is not a like-for-like comparison,
+ * and it measurably destroys the signal — a distinct pair (staging port 8080 vs
+ * 8081) scores 0.9749 bare and 0.8512 enriched, while a genuine duplicate goes
+ * 0.9689 -> 0.9878. Context pushes the two classes APART; dropping it from one
+ * side collapses them together.
+ */
+export interface SearchTextFields {
+  statement: string
+  domain?: string | null
+  tags?: string[] | null
+  entities?: Engram['entities']
+  temporal?: Engram['temporal']
+  rationale?: string | null
+  source?: string | null
+  dual_coding?: Engram['dual_coding']
+  knowledge_anchors?: Engram['knowledge_anchors']
+}
+
+/**
+ * Build searchable text from context fields. Single definition so the stored
+ * and incoming sides cannot drift apart.
+ */
+export function searchTextFrom(fields: SearchTextFields): string {
+  // `engramSearchText` reads `engram.tags.length` unguarded — fine for a
+  // schema-validated Engram, a crash for a partial write. Normalise first.
+  return engramSearchText({
+    ...fields,
+    tags: fields.tags ?? [],
+    domain: fields.domain ?? undefined,
+    rationale: fields.rationale ?? undefined,
+    source: fields.source ?? undefined,
+  } as unknown as Engram)
+}
+
 /** Build searchable text from all engram fields */
 export function engramSearchText(engram: Engram): string {
   const parts = [engram.statement]
--- a/packages/core/src/history.ts
+++ b/packages/core/src/history.ts
@@ -3,7 +3,7 @@ import { join } from 'path'
 import { createHash } from 'crypto'
 
 export interface HistoryEvent {
-  event: 'engram_created' | 'engram_updated' | 'engram_merged' | 'feedback_received' | 'engram_retired' | 'engram_decremented' | 'engram_promoted' | 'engram_rescoped' | 'failure_reported' | 'procedure_evolved' | 'recurrence_detected' | 'contradiction_detected' | 'scope_promoted' | 'buffer_pruned' | 'weekly_review' | 'engram_route_failed' | 'co_injection' | 'injection_outcome' | 'session_scope_changed'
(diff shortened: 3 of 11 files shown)


plur9:
Full audit posted on #857. Verdict: **approve**. CI passing. @plur9 ready for merge (needs a non-author approval per GitHub policy).

plur9:
Pushed `01e3e4a` — a real bug found while calibrating the threshold, and the
threshold set from measurement rather than from either default.

Credit where due: this came from Gregor asking *"if two engrams are so distinct
and so similar at once, don't they lack the context to be meaningful?"* That was
the right question and it turned out to be measurable.

## The comparison was asymmetric

Stored engrams are embedded as `engramSearchText` output — statement **plus**
domain, tags, entities, rationale, source, dual-coding, anchor snippets. This
patch was embedding the **bare incoming statement** against that. A fragment
compared to fully-contexted records.

Measured with the shipped BGE-small embedder:

| pair | bare | enriched |
|---|---|---|
| DISTINCT — staging port 8080 vs 8081 | 0.9749 | **0.8512** |
| DISTINCT — "Always rebase before pushing" vs "**Never** …" | 0.9637 | **0.8826** |
| DUPLICATE — the #854 pair | 0.9689 | **0.9878** |

Bare, **the distinct pairs score higher than the genuine duplicate.** They
overlap; no threshold separates them, and I was about to argue over a number
that could not have worked at any value. Context moves the two classes in
opposite directions and opens a clean band.

`searchTextFrom` is now extracted in `fts.ts` and used by both sides, so stored
and incoming text cannot drift apart. Pinned by a test asserting the query
carries the domain, tags and rationale.

## Threshold: 0.95

From the numbers above, not from a default. The old 0.85 was declared for four
months while nothing read it, so it carried no evidence — and applied, it
suppressed `"Second knowledge item"` as a duplicate of `"First knowledge item"`
in this repo's own `sp1-memory-intelligence` batch test. That test passes again
at 0.95.

The negation row is why the bar is not lower. **A correction is usually the prior
statement with one thing negated**, and corrections are the most valuable engrams
this system stores. A bar under ~0.89 starts eating them.

## What cosine is allowed to decide

Only "clearly the same". Everything below the bar is reported via
`dedup.near_duplicates` and acted on by nobody, because a high score is ambiguous
between three different relations:

| relation | example | right action |
|---|---|---|
| duplicate | the #854 pair | NOOP |
| **supersede** | port 8080 → 8081 | UPDATE + mark superseded |
| sibling | staging 8080 / prod 8081 | ADD both |

A scalar cannot distinguish those. The LLM path already can
(`ADD`/`UPDATE`/`MERGE`/`NOOP`), and `supersedes` already exists on the schema.
Consolidating the existing ~131 near-duplicates stays out of scope here and
belongs with #109 — report-only first.

## Gate

`pnpm test` **3820 passed / 156 skipped**, 0 failures. `pnpm build`,
`tsc --noEmit` on core, `pnpm typecheck:tests` all clean.

plur9:
Pushed `297812c` — the threshold now has a test behind it, not just a comment.

## The gap this closes

Every other test here stubs `similarityScores`, so nothing exercised the real
embedder and the numbers justifying 0.95 lived only in a PR comment. That is
fine right up until someone changes the embedder or edits `engramSearchText`,
at which point the bar silently stops meaning anything and CI stays green.

Measured, BGE-small, enriched on both sides:

```
threshold = 0.95
  0.7962  DISTINCT   one token apart: staging port 8080 vs 8081
  0.8437  DISTINCT   negation — a correction of an existing engram
  0.8570  DISTINCT   same rule, different host
  0.9797  DUPLICATE  reordered clauses, same claim
  0.9830  DUPLICATE  the #854 pair
```

Separation **0.123**, with 0.95 inside the band rather than near an edge.

## It asserts, it does not merely report

1. **No DISTINCT pair reaches the threshold.** This is the data-loss direction —
   a false NOOP silently declines to store a memory.
2. **Every DUPLICATE still clears it**, or the bar does nothing.
3. **The classes are separated by >0.03.** A bar that merely happens to fall
   between two adjacent numbers should fail, not pass.

A second case pins the inversion that caused all this: bare statements put the
DISTINCT worst case at 0.9749 — **above both duplicates**. If enrichment ever
stops pulling distinct facts apart, that fails.

## Scope, stated plainly

This is a **calibration instrument, not the regression gate.** The code-path
regression — comparing bare text again — is caught by the fast stubbed test
already in this PR, which runs on every push. This one checks whether the
resulting numbers still mean what the threshold assumes.

## CI

A gated test that never runs closes nothing, and there are no scheduled
workflows here, so I added a `dedup-calibration` job:

- **Path-filtered** to `embeddings.ts`, `fts.ts`, `learn-async.ts` and the
  fixtures — the only changes that can invalidate the claim. Skips otherwise.
- **Outside the Node matrix.** Three model downloads for one answer is waste.
- **`continue-on-error: true`**, deliberately. It depends on a model fetch, and a
  check that reddens for reasons unrelated to the diff is exactly how a
  formatter failure stayed invisible for seven days in `plur-ai/encode`.

That last one is the judgement call worth arguing with. Making it blocking is one
line — but then a HuggingFace outage fails unrelated PRs. I would rather it
inform.

`pnpm test` **3820 passed / 158 skipped** (the calibration file skips offline, as
intended). `typecheck:tests`, core `tsc --noEmit`, `pnpm build` clean.

plur9:
🤖 heartbeat — pre-merge verification from audit issue #857:

**`embeddingSearchWithScores` is a re-ranker over passed candidates, not a corpus search.**

Function body at `embeddings.ts:398`:
```ts
for (const engram of engrams) {
  // embed each passed candidate, compute cosine against query embedding
}
```

It iterates `engrams` (the first argument) and never loads from disk or queries the broader store. The `if (existing)` guard in the dedup branch therefore has its intended meaning: a false NOOP cannot sneak through an accidental corpus-wide match.

Last open item from the audit is resolved. #856 is clear to merge.

crtahlin (changes requested):
Independent audit requested in [#857](https://github.com/plur-ai/plur/issues/857). You asked me to be sceptical about the false-NOOP direction. I could not make that case land — what I found is close to the opposite problem.

## CRITICAL — the bar does not catch the pair that prompted the issue

Measured on the shipped BGE-small embedder, comparing the 019/029 pair two ways:

```
threshold                                = 0.95
fixture shape  (context on BOTH sides)   = 0.9830  NOOP (caught)
as actually written (twin has no domain) = 0.8339  ADD  (missed)
delta                                    = 0.1490
```

The calibration fixtures give both sides an identical `domain`, `tags` and `rationale`. The real records did not — that is *defect 2 of the same issue*: "the second copy of every pair has no `domain` field", and the condensed pass was "generated from the statements alone rather than from the full records", so no rationale either.

Production enriches the incoming side from the write's own context only, so an under-contexted write is structurally compared against a fully-contexted stored engram. At 0.8339 the pair sits below 0.95 and below even the old 0.85. **None of the ten reported pairs would have been NOOPed by this change.**

The enrichment symmetry the design rests on holds only when both writes carry the same context. The reported duplicates were asymmetric by construction, and the asymmetry is itself the other half of the issue being fixed.

## CRITICAL — the new code is unreachable from `plur_learn`

`learnAsync` has exactly one caller across `cli` and `mcp`: the `plur_learn_batch` tool. `plur_learn` calls `learnRouted()`, which defers to the synchronous `learn()` for local scopes and to `_hashDedup` on the remote route. `packages/mcp/src/tools.ts:965` states it outright — *"learnAsync is NOT used on this path"* — and that handler hardcodes `decision: 'ADD'`.

This cuts both ways. The false-NOOP blast radius is much smaller than the issue implies, but [#854](https://github.com/plur-ai/plur/issues/854) is not closed for the dominant write path: near-duplicates arriving via `plur_learn` are still only ever caught by exact content-hash.

## WARNING — the reporting safety net reaches no caller

`dedup.mode` and `near_duplicates` land on `LearnAsyncResult`, but `tools.ts:1184` maps only `input_index`, `id`, `statement`, `scope`, `type`, `decision` and `existing_id`, dropping `dedup` entirely; `stats` carries no per-mode counts.

So "anything below the bar is still REPORTED via `dedup.near_duplicates`" is not observable at any call site — and the reporting-only fallback proposed in [#857](https://github.com/plur-ai/plur/issues/857) does not currently work either. Worth fixing regardless of what happens to the threshold.

## WARNING — cross-scope false NOOP on unscoped writes

`learn-async.ts:341` applies the scope filter only `if (context?.scope)`. Auto-routing happens later, inside `learn()` → `_guardSensitiveScope`.

So a genuinely unscoped write that would auto-route to a team scope can NOOP against a `global` near-duplicate and never reach the team store. This is the [#359](https://github.com/plur-ai/plur/issues/359) class — pre-existing for exact hashes, since `_hashDedup` matches any scope when `scope` is undefined — now widened to near-duplicates on every install without an LLM. It is a genuine false-NOOP path, and it is not a threshold question.

## WARNING — `schemas/config.ts:80` still declares `threshold: 0.85`

I verified that `.partial()` suppresses the Zod default, so the runtime value really is 0.95 when the key is absent. But the one place a user reads the default now contradicts the code, and a hand-written `threshold: 0.85` lands below the floor this PR's own calibration establishes (0.8437 negation, 0.8570 same-rule-different-host) — directly into the data-loss class.

Related: `dedup-threshold-calibration.test.ts` hardcodes `THRESHOLD = 0.95` rather than importing the production default, so the comment "kept here so a drift shows up as a failure" does not hold — changing the default in `learn-async.ts` would not fail that test.

## WARNING — the calibration job misses the embedder adapters

The path filter matches `packages/core/src/(embeddings|fts|learn-async).ts`, but the adapters live in `packages/core/src/embedders/`, and `schemas/config.ts` is absent too. Changing the default model, its pooling, or its prefixes therefore skips calibration — which the job's own comment names as the primary trigger.

Separately: production embeds the incoming side with `embed(query, 'query')` and stored engrams with the passage role. Only EmbeddingGemma applies asymmetric prefixes, and it is opt-in via `PLUR_EMBEDDER`, so the default is unaffected — but the calibration test embeds both sides role-less, so for that embedder it validates a comparison production never makes. The direction is conservative (fewer NOOPs), yet the bar is uncalibrated there.

## NOTE

- `searchTextFrom` omits `temporal` although `LearnContext.valid_from`/`valid_until` are available and `engramSearchText` includes them. `entities` is declared on `SearchTextFields` but absent from `LearnContext`, so it is dead for the only production caller. Both bias toward ADD, but both weaken the symmetry the design depends on.
- A cosine NOOP leaves no trace: no history event, no `reference_count` increment, no source append — unlike the exact-duplicate path, which calls `_recordDuplicate`. The one thing that makes a false NOOP recoverable is a record that it happened.
- Three different number sets are cited for the crux decision: 0.928 as "the only measured data point" in [#857](https://github.com/plur-ai/plur/issues/857), 0.8512/0.9878 in the `learn-async.ts` comment, and 0.7962/0.8437/0.8570/0.9797/0.9830 in the test baseline and CI.
- Diffstat is +639/-7 across 8 files, not the +239/5 stated in the audit request.

## Praise

The `dedup-calibration` job genuinely loaded the real model in CI and reproduced the documented table exactly, in 3.07 s of real embedding. The evidence here is stronger than the "one data point" claimed in the request. The fail-open error handling — similarity never gates a write — and the explicit `hash-only` mode label are both the right calls.

## Suggested path

Ship reporting only, as floated in [#857](https://github.com/plur-ai/plur/issues/857), but wire it through to `plur_learn_batch` *and* `plur_learn` first so it is actually visible. Then measure a distribution over the real store — including the under-contexted writes, which are the ones that produced [#854](https://github.com/plur-ai/plur/issues/854) — and derive a bar from that rather than from paired fixtures.

## Verification

Full `@plur-ai/core` suite locally: 2678 passed, 3 failed. The 3 are `sync-push-failure.test.ts`, which fails identically on [#855](https://github.com/plur-ai/plur/pull/855)'s branch — a git-invoking test that does not survive running inside a worktree. Environmental, not from this PR.

plur9:
Thank you for the thorough audit — the two CRITICAL findings are both real blockers.

**Accepting the suggested path:** ship reporting-only first, wired through *both* `plur_learn_batch` and `plur_learn`, then measure a real-store distribution before picking a threshold.

**On CRITICAL 1** (bar misses the actual pairs): the asymmetry between write context and stored-engram context is the structural problem, not the threshold value. Agreed that the fixture suite validates a symmetric case that doesn't represent production. This makes deriving any threshold from those fixtures unreliable for the reported duplicates.

**On CRITICAL 2** (unreachable from `plur_learn`): confirmed. `learnAsync` is only on the batch path; `plur_learn` exits at `learnRouted()` → `learn()` → `_hashDedup`. The cosine branch never fires on the dominant write path.

**On WARNING 4** (threshold doc mismatch at `schemas/config.ts:80`): will fix regardless — a `0.85` default in the schema doc that contradicts the `0.95` runtime value is a footgun independent of this PR's outcome.

**On the unreachable reporting** (WARNING 3): agreed — wiring `dedup.near_duplicates` through `tools.ts:1184` is prerequisite to the reporting-only approach being useful at all.

This PR will be superseded by a reporting-only variant that addresses all four warnings. Routing to nightshift for implementation.

plur9:
You were right and I was wrong about which direction the risk ran. I asked you to be sceptical about false NOOPs; you found the bar wouldn't have caught the pairs it was built for. Reworked to reporting-only — the gate is **removed**, not retuned.

## The two CRITICALs

**The bar can't work at any value, so it's gone.** Your 0.8339 is the whole argument: the real duplicates are the under-contexted ones, and they score *below* the DISTINCT pairs at 0.8826. The classes overlap, so no threshold separates them — a bar would have bought a real false-NOOP risk while still missing all ten reported pairs.

I've pinned your measurement as a test rather than only citing it in a comment, so the reasoning can't quietly stop being true. If a future embedder separates the classes, that assertion fails and a write gate becomes arguable again — which is the trigger condition for revisiting this, stated where someone will hit it.

**`learnAsync` unreachable from `plur_learn`.** This was the more important finding, because it reframes #854: the duplicates were written on a path that has only ever had exact-hash dedup, so the fix was aimed at the wrong path entirely. `plur_learn` now reports through a new `Plur.nearDuplicates()` — after the write, never blocking it.

## The reporting that reached nobody

Fixed at `tools.ts` for the batch path, and the new path for `plur_learn`. Both now carry `dedup.mode` and `dedup.near_duplicates`, and the tool descriptions say plainly that `hash-only` means "not identical", not "not a duplicate".

## The part your suggested path needs that wasn't there

You said: measure a distribution over the real store, including the under-contexted writes. There was nothing to measure *from* — a cosine NOOP left no trace, and neither did an ADD. So every observation above a deliberately low floor (0.75, beneath every pair either of us measured) is now written to history as `dedup_near_duplicate`, carrying the top score, the scope, and **whether the incoming write had a domain and a rationale**.

That last field is the one I'd defend hardest: the asymmetry is what broke the bar, so the distribution has to be splittable by it rather than averaged over it. Otherwise the next attempt re-derives a bar from a mixture of two populations and lands somewhere just as arbitrary.

## Smaller findings

- **`schemas/config.ts:80`** — both are 0.85 now, and the schema documents that `.partial()` suppresses its own default, since that's why the mismatch was invisible.
- **Calibration test hardcoding the bar** — imports the production constant. You were right that the "drift shows up as a failure" comment was false as written.
- **CI path filter** — widened to `packages/core/src/embedders/` and `packages/core/src/schemas/config.ts`.
- **Cross-scope false NOOP** and **the traceless NOOP** — both dissolve rather than get fixed, since nothing NOOPs.
- **The three inconsistent number sets** — collapsed to one table in the PR body, sourced from the calibration run.
- **Diffstat** — you were right, and the audit request was wrong; it was reading a stale local state.

## Not addressed, deliberately

`searchTextFrom` omitting `temporal`, and `entities` being dead for the only production caller. Both bias toward ADD, which is now the only direction anything biases, so they cost much less than they did when a gate existed. Filed-worthy, not worth widening this PR.

## Verification

Core 2692 passed / 0 failed. MCP 360 passed / 0 failed.

On your 3 `sync-push-failure.test.ts` failures — not the worktree. The better-sqlite3 native binary had been deleted by the failed `pnpm rebuild` during #874's investigation, exactly as that issue warns it does. Rebuilt it and the suite is genuinely green. Worth knowing for #875: **11.10.0 builds from source cleanly on Node 24**, so #874 really is Node-26-only.

Issue #593, opened 2026-07-15
CI: warn on PR-open when the linked issue is assigned to someone else (+ auto-assign if unassigned)


crtahlin:
Durable backstop for the claim-before-you-code convention (CONTRIBUTING guide added in #592). Discipline alone can't prevent two contributors — or an agent and a person — implementing the same fix in parallel when an issue is never assigned. An automated check closes the residual gap.

## Proposed GitHub Action (on `pull_request: [opened, edited]`)

1. Parse the linked issue(s) from the PR body (`Closes #N` / `Fixes #N` / `Part of #N`).
2. For each linked issue:
   - If **assigned to someone other than the PR author** → post a warning comment: "⚠️ This PR addresses #N, which is assigned to @X — possible duplicate effort. Confirm coordination before merging."
   - If **unassigned** → auto-assign the issue to the PR author, so ownership is recorded.

## Notes

- Detective control (catches collisions at PR time); the preventive control is the self-assign-before-you-start step in CONTRIBUTING.md + CLAUDE.md.
- Comment + assign only — never block the merge.

plur9:
The Node 20 CI failure on #592 that triggered this issue is now root-caused and fixed in PR #596.

**Root cause:** `generateInjectionId()` used `Date.now()` (ms precision) + 4-char base-36 random suffix. Birthday collision probability for 50 samples within the same millisecond: ~0.07%. Manifested as `expected 49 to be 50` in `co-injection-logging.test.ts:52`.

**Fix (#596):** Replaced random suffix with a per-process monotonic counter. Output format unchanged — `INJ-<ts>-<4-char base36>` — so the existing test regex still passes.

The CI re-run on #592 should also clear now that #596 ships to main. This PR (#593 — CI check for issue assignment) remains valuable on its own merits as a preventive/detective control.

🤖 heartbeat

Issue #857, opened 2026-08-10
Audit request: three PLUR core fixes from 2026-08-10 (#842, #855, #856)


plur9:
@crtahlin — three PRs from today, all self-audited, all green. Requesting an
independent audit before merge. Ordered by how much I want a second pair of eyes.

| PR | Fixes | Files | Risk | CI |
|---|---|---|---|---|
| [#856](https://github.com/plur-ai/plur/pull/856) | #854 dedup | 5, +239/-7 | **highest** — changes write behaviour | green |
| [#855](https://github.com/plur-ai/plur/pull/855) | #831 forget | 3, +177/-7 | destructive path | green (6/6) |
| [#842](https://github.com/plur-ai/plur/pull/842) | release smoke | 3, +105/-7 | release path | green, your re-review pending |

## Where I most want the audit

**#856 is the one to be sceptical about.** It makes writes NOOP that previously
ADDed. If the threshold is wrong in the aggressive direction, the failure is a
*false NOOP* — a memory silently not stored — which is the worst outcome this
product has, and it is invisible at the call site.

My reasoning for 0.95, so you can attack it: `threshold` defaulted to 0.85 but was
never read in four months, so it has never been exercised against a real write and
carries no evidence. The only measured data point I have is a genuine duplicate
pair at **0.928**. I set the bar above it deliberately. That is one data point,
which is thin — if you think the right move is to ship reporting only
(`dedup.near_duplicates`) and leave the NOOP off until there is a distribution to
look at, I would not argue.

**#855 is destructive-path.** Two edges I got wrong first and want checked:
"ambiguous" requires the id to resolve **locally AND remotely** (my first version
keyed on the remote cache alone and broke remote-only namespaced ids, #86); and
`scope: "primary"` must return early rather than fall through to the remote walk,
or it retires the remote engram the caller just excluded.

**#842** you have already reviewed once. The only change since is `b5b226b`,
replacing the guard condition you suggested — reasoning in the PR comment. Your
suggestion left a version *mismatch* filed as lag, which is one of the two defects
the check exists to catch.

## Cross-PR issue that needs a decision

**#851** (feedback disambiguation, another session) carries `scripts/release.sh`
hunks that are #842's change, but with the *older* classifier guard —
`[ "$core_inst" = "$VERSION" ]`. If #851 lands after #842 it silently reverts the
fix. Cleanest resolution is dropping the `release.sh` + `RELEASING.md` hunks from
[...]

Issue #823, opened 2026-08-03
Independent review of the 0.17 store-integrity work before publish


plur9:
@crtahlin — asking for a human pass over the 0.17 data-loss work before we publish. Not a rubber stamp; the specific request is below.

## Why

Four audits ran over this: #794 (original, 15 findings), #811 (independent whole-repo, 21 findings, 2 Critical), #805 (an unfixed remainder of the first), and #821 (an audit of the FIXES, 16 findings, 7 High). Every #821 finding was introduced *by* the fixes for the earlier ones.

That is the thing worth a second pair of eyes. Severity is converging — #821 found no Criticals — but each round of fixes has a defect rate of roughly seven per thousand lines, and the fixes are concentrated in locking, atomicity and store write paths. Three of the four audits were LLM-driven, and they share blind spots by construction. A human who did not write any of it is the check we do not otherwise have.

## What landed

| PR | What |
|---|---|
| #801, #803, #806, #808, #809 | refuse-on-corrupt loader, shrink guard, fsync, episodes lock, sync integrity, backups + restore |
| #810, #814, #815 | shrink-guard perf, ten defects from #811, twelve remaining |
| #817 | embedding staleness (#812) |
| #819 | pack registry integrity, config lock, drop-log (#805) |
| #820 | status() survives a broken registry |
| #822 | all 16 findings from #821 |

Diff worth reading as one unit: `git diff 3e47b9b..main -- packages/core/src packages/mcp/src packages/cli/src`.

## Where I would spend your time

Ranked by where I think a human reviewer beats another automated pass:

1. **Lock correctness end to end.** `withLock` (sync.ts) and `withAsyncLock` (store/async-lock.ts) are both non-reentrant, and the fixes added locks to migrations, the pack registry, and restore. #822 establishes a corpus -> config ordering and an atomic `rename`-claim steal. The ordering claim rests on my reading that nothing acquires them the other way; that reading is worth checking independently.
2. **The throw-tolerance boundary.** Refuse-on-corrupt turned several silent degradations into exceptions. Four separate findings were "a caller that cannot tolerate a throw now gets one" (`status()`, `session_start`, the `listPacks` fallback, install-after-work). I believe they are all found. I also believed that after #820, and #821 found two more.
3. **The shrink guard.** It has been broken twice by its own optimisation, both times because the byte proxy did not track the record count. It is now unconditional with a scan-based count. Worth confirming the scan is exact for every document shape we actually write.
[...]

PR #859, opened 2026-08-11
ci: flag/auto-assign issues that @-mention with an action ask but no assignee

### LINKED ISSUE #858: CI: flag/auto-assign an issue that @-mentions someone with an action ask but has no assignee
## The gap

Our contribution rules are explicit that **an @-mention is FYI, not a work assignment** — only assignment creates an expectation the work will happen (CONTRIBUTING.md "Labels and @-mentions are signals, not assignments"; CLAUDE.md; codified in #614).

The existing CI enforcement (#593, closed) warns on **PR-open when the linked issue is assigned to someone else** (and auto-assigns if unassigned). That covers the *PR* path, but **not the issue path**: an issue that @-mentions a person with an explicit action ask ("requesting an audit", "please review", "can you…") while leaving the issue **unassigned** slips through entirely.

## Concrete example

#857 ("Audit request: three PLUR core fixes") opens with `@crtahlin — requesting an independent audit before merge` and has **no assignee**. It is an action request routed by mention, which is exactly what the rule says not to do — and nothing flags it.

## Proposal

Add an issue-level check (Action or bot) that, on issue open/edit, warns when **all** of the following hold:
- the body/title @-mentions a user other than the author,
- it contains an action ask (heuristic: "audit", "review", "please", "can you", "requesting", "need you to", "assign"…),
- the issue has **no assignee**.

Behaviour: post a one-line reminder of the rule and/or auto-assign the mentioned user (mirroring #593's PR auto-assign). Keep it a warning, not a hard block.

## Refs
- Rule: #614 (CONTRIBUTING.md + CLAUDE.md)
- Prior CI (PR path): #593
- Triggering example: #857

plur9:
## Summary

- Adds `issue-mention-guard.yml` workflow triggered on `issues: [opened, edited]`
- Detects when an issue @-mentions a user (other than the author) alongside an action ask keyword (`audit`, `review`, `please`, `can you`, `requesting`, etc.) and the issue has **no assignee**
- Auto-assigns the mentioned user, mirroring the PR-path behaviour from `pr-issue-guard.yml` (#593 / #835)
- Falls back to a comment reminder if the mentioned user is not assignable (GitHub requires repo access or prior interaction)
- Skips silently if the issue is already assigned

Closes #858.

## Test plan

- [ ] Open a new issue that @-mentions another user with "requesting an audit" and no assignee → expect the mentioned user to be auto-assigned
- [ ] Open an issue with an @-mention but no action ask → workflow should skip (no action, no comment)
- [ ] Open an issue that is already assigned → workflow should skip
- [ ] Open an issue where the mentioned user has no repo access → expect the fallback comment (not a run failure)

### DIFF
--- a/.github/workflows/issue-mention-guard.yml
+++ b/.github/workflows/issue-mention-guard.yml
@@ -1,0 +1,79 @@
+name: Issue Mention Guard
+
+# Flags issues that @-mention a user with an action ask but have no assignee.
+# Mirrors the claim-before-you-code convention already enforced on the PR path
+# by pr-issue-guard.yml (#593 / #835).
+#
+# SECURITY: this workflow reads only event payload fields (title, body,
+# assignees) and the @-mentions derived from them. No untrusted code is
+# checked out or executed. Issue body content is used solely as a regex
+# source; matched usernames are passed directly to the REST API, not to
+# any shell.
+
+on:
+  issues:
+    types: [opened, edited]
+
+permissions:
+  issues: write
+
+jobs:
+  mention-guard:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/github-script@v7
+        with:
+          script: |
+            const body  = context.payload.issue.body  || '';
+            const title = context.payload.issue.title || '';
+            const issueNumber = context.payload.issue.number;
+            const issueAuthor = context.payload.issue.user.login;
+            const { owner, repo } = context.repo;
+
+            // Skip if already assigned — the convention is already satisfied.
+            const existingAssignees = (context.payload.issue.assignees || []).map(a => a.login);
+            if (existingAssignees.length > 0) return;
+
+            // Extract @mentions from body + title, excluding the issue author.
+            const combined = title + '\n' + body;
+            const mentionRe = /@([a-zA-Z0-9](?:[a-zA-Z0-9]|-(?=[a-zA-Z0-9])){0,38})/g;
+            const mentions = new Set();
+            let m;
+            while ((m = mentionRe.exec(combined)) !== null) {
+              if (m[1].toLowerCase() !== issueAuthor.toLowerCase()) {
+                mentions.add(m[1]);
+              }
+            }
+            if (mentions.size === 0) return;
+
+            // Require an explicit action ask alongside the mention.
+            // Keyword list from the issue proposal (#858).
+            const ACTION_ASK = /\b(audit|review|please|can you|requesting|need you to|assign|check|look at|investigate|evaluate|approve)\b/i;
+            if (!ACTION_ASK.test(combined)) return;
+
+            // All three conditions met: mention + action ask + no assignee.
+            // Auto-assign the mentioned user(s), mirroring the PR guard (#593 / #835).
+            for (const login of mentions) {
+              try {
+                await github.rest.issues.addAssignees({
+                  owner, repo,
+                  issue_number: issueNumber,
+                  assignees: [login],
+                });
+                core.info(`Auto-assigned #${issueNumber} to @${login}`);
+              } catch (e) {
+                // GitHub only permits assigning users with repo access or prior
+                // interaction on the issue — a new contributor mention may simply
+                // not be assignable. Degrade gracefully rather than failing the run.
+                core.warning(`Could not auto-assign #${issueNumber} to @${login}: ${e.message}`);
+                await github.rest.issues.createComment({
+                  owner, repo,
+                  issue_number: issueNumber,
+                  body: `ℹ️ **Contribution convention:** @-mentions are FYI only — `
+                    + `assignment creates the expectation that work will happen `
+                    + `([CONTRIBUTING.md, #614](https://github.com/${owner}/${repo}/issues/614)). `
+                    + `This issue @-mentions @${login} with an action ask but has no assignee. `
+                    + `Maintainer: please assign it explicitly so the claim is recorded.`,
+                }).catch(err => core.warning(`Fallback comment failed: ${err.message}`));
+              }
+            }


crtahlin (changes requested):
Requesting changes, and the request is to **close this rather than fix it**. The objection is to the mechanism, not the implementation — automatic assignment should not exist in this repository in any form. The technical findings below are secondary; they matter only because they show the mechanism does not even work as described.

## The mechanism is the problem

[#614](https://github.com/plur-ai/plur/issues/614) and CONTRIBUTING.md say an @-mention is FYI and only **assignment** creates the expectation that work will happen. The purpose of that rule is that a person deliberately claims or delegates the work, and that the claim is therefore meaningful.

A workflow that assigns on a keyword match does not enforce that rule — it routes around it. It converts "someone mentioned you without assigning" into "CI assigned you", which produces the identical unrequested work expectation, now created by a regex with nobody accountable for the decision. If a keyword match can assign an issue, assignment stops being a claim and becomes a side effect, which empties the convention this workflow is meant to protect.

The failure this addresses is an issue author not following a rule they know. The remedy for that is the author assigning the issue — one action, no code, no false positives, no permanent CI surface, no mutating token. Building automation to accommodate non-compliance makes the rule optional and shifts the cost onto everyone else.

## The scale does not justify it either

Measured across the last 60 issues — see the table below — the heuristic would have fired correctly **twice**. Both are the same author routing work to the same reviewer by mention. This is 79 lines of permanent CI holding `issues: write`, plus an auto-assignment mechanism applied to every future issue, to address one person's habit on two occasions.

The repository's own over-engineering guardrail applies directly: the simplest version that solves the actual problem is the author assigning the issue.

## Supporting evidence — it also does not work

Replaying the exact `mentionRe` and `ACTION_ASK` regexes over the last 60 issues, evaluated as unassigned (true at `opened` time even for issues assigned later):

| Issue | Would assign | Matched keyword | Correct? |
|---|---|---|---|
| [#857](https://github.com/plur-ai/plur/issues/857) | `@crtahlin` | "Audit" | yes |
| [#823](https://github.com/plur-ai/plur/issues/823) | `@crtahlin` | "review" | yes |
| [#794](https://github.com/plur-ai/plur/issues/794) | `@plur-ai` | "audit" | **no** |
| [#770](https://github.com/plur-ai/plur/issues/770) | `@plur-ai` | "audit" | **no** |
| [#764](https://github.com/plur-ai/plur/issues/764) | `@plur-ai`, `@latest` | "check" | **no** |
| [#760](https://github.com/plur-ai/plur/issues/760) | `@plur-ai` | "check" | **no** |
| [#758](https://github.com/plur-ai/plur/issues/758) | `@plur-ai` | "audit" | **no** |
| [#752](https://github.com/plur-ai/plur/issues/752) | `@plur-ai` | "audit" | **no** |

Six of eight firings target something that is not a person: `@plur-ai` is the npm scope from `@plur-ai/core` / `@plur-ai/mcp` / `@plur-ai/cli`, which is also the GitHub org; `@latest` is a dist-tag from `npx @plur-ai/cli@latest`. A 75% false-positive rate on a corpus this small, in a repository whose issues are dense with pasted commands, is not a tuning problem.

**And the graceful-degradation branch cannot fire for the case it documents.** Per [GitHub's REST documentation for Add assignees](https://docs.github.com/en/rest/issues/assignees): *"Only users with push access can add assignees to an issue. **Assignees are silently ignored otherwise.**"* — the endpoint returns 201, not an error. So for every one of those six false positives, and for any genuine mention of a non-assignable person, the API succeeds, nothing is assigned, no fallback comment is posted, and `core.info("Auto-assigned #N to @plur-ai")` is written to the log. The run reports success for an assignment that did not happen.

Further, the `ACTION_ASK` list (`review`, `check`, `please`, `assign`, `approve`, `look at`, `evaluate`) is ordinary engineering vocabulary and is tested against title and body combined with no proximity to the mention, so "as @someone noted, we should check the logs" qualifies. There is also no `concurrency` group, so rapid edits can run concurrently, and the `edited` trigger has no idempotency check.

## What is worth keeping

The security posture is correct and the header comment is accurate, which is worth saying plainly because it is the part most workflow authors get wrong: reading `context.payload.issue.body` inside `github-script` rather than interpolating `${{ github.event.issue.body }}` avoids the script-injection class entirely, `permissions` is narrowed to `issues: write`, the trigger is `issues` rather than `pull_request_target`, and nothing is checked out. If a workflow in this area is ever written, that is the shape to copy.

## Separately

- The PR description carries an AI-attribution footer, which this repository's conventions exclude from PR descriptions. This repo is public.
- Finding on the API contract above applies equally to `pr-issue-guard.yml`, which uses the same `catch`-based assumption — meaning [#835](https://github.com/plur-ai/plur/issues/835)'s new-contributor fallback likely never fires. Filed separately; it is a real defect independent of this PR and should not be folded into it.

## Verification

YAML parsed (`issues: [opened, edited]`, `permissions: {issues: write}`, `concurrency: null`, 1 job / 1 step). Heuristic replayed against 60 real issues. API behaviour confirmed against GitHub's published documentation. CI is green on all five checks, none of which exercise this workflow — note that the four manual test cases in the description all use a real username, so they would pass while the false-positive class above remains.

plur9:
Closing unmerged — accepting @crtahlin's review in full.

The objection is to the mechanism, not the implementation, and it holds. Our own rule (#614, CONTRIBUTING.md) says an @-mention is FYI and only assignment creates an expectation. The point of that rule is that assignment is a deliberate claim. A workflow that assigns on a regex match doesn't enforce the rule, it routes around it — assignment stops being a claim and becomes a side effect, which empties out the convention this was meant to protect.

The measured evidence is decisive on its own: replayed over the last 60 issues the heuristic fires 8 times and 6 are wrong, because `@plur-ai` is our own npm scope pasted in `@plur-ai/core` commands and `@latest` is a dist-tag from `npx @plur-ai/cli@latest`. The two correct firings are both me routing work to the same reviewer. That is 79 lines of permanent CI holding `issues: write` to correct my own habit on two occasions — and the repo's over-engineering guardrail points straight at the cheaper fix: assign the issue when I open it.

Worth recording that the graceful-degradation branch here could not have fired either, for the same reason as #861 — `addAssignees` returns 201 and silently drops a non-assignable login, so those six false positives would each have written `Auto-assigned #N to @plur-ai` to the log for an assignment that never happened. That defect is fixed separately in #862.

Keeping the security shape on record, since it was the part that was right: reading `context.payload.issue.body` inside `github-script` rather than interpolating `${{ github.event.issue.body }}` avoids the script-injection class entirely, `permissions` narrowed to `issues: write`, `issues` trigger rather than `pull_request_target`, nothing checked out. That is the shape to copy if a workflow in this area is ever written.

Closing #858 alongside this.

Issue #123, opened 2026-05-11
plur-hermes learner: sentence-level fallback + labeled corpus (follow-up to #115)


plur9:
## Parent

Follow-up to #115 (multi-strategy learning extractor). PR 5306959 landed Strategies 1 & 2 (marker detection + bullet normalization). This issue tracks the deferred Strategies 3 & 4 — sentence-level fallback and confidence scoring — which were gated on a labeled corpus, not on regex work.

## Why this is its own issue

Strategies 1 & 2 are deterministic: known markers, known bullet forms, no false-positive risk worth measuring. Strategy 3 is fundamentally different — it scans for sentences matching learning-shaped patterns (`(I|We) (should|will|must|prefer|avoid|always|never) ...`, `Correction: ...`, etc) anywhere in assistant output. Those patterns fire on hypothetical reasoning too ("you should never use X for Y" when explaining, not learning), so the FP rate is the headline risk. AC requires FP < 15%, and you can't measure that without a labeled test set.

The gating work is the labeled set, not the regex.

## Scope

1. **Labeled corpus build** (D-1)
   - Dump ~50–100 real assistant turns from session logs (`~/.plur/episodes/*` or hermes session captures)
   - JSONL fixture: `{turn_text, labels: [{start, end, text, kind}]}` where `kind ∈ {learning, hypothetical, reasoning, instruction, none}`
   - Hand-label spans that are actual learnings vs sentences that match the pattern but aren't
   - Target: ≥30 turns with at least one positive learning, ≥30 turns with at least one near-miss negative (hypothetical/instruction/reasoning)

2. **Sentence-level extractor** (D-2, blocked on D-1)
   - Patterns from spec: `(I |We )?(should|will|must|prefer|avoid|always|never) ...`, `Correction: ...`, `The (correct|proper|best) way is ...`
   - Iterate against measured precision/recall on the labeled set
   - Must clear FP < 15% before merging

3. **Confidence scoring** (D-3, depends on D-2)
   - Each strategy emits `(text, confidence)`
   - Strategy 1 (fenced `🧠 I learned:`): 1.0
   - Strategy 2 (line-anchored marker): 0.9
   - Strategy 3 (sentence-level): scored from the matched pattern (e.g. `Correction:` → 0.85, `(I|We) should ...` → 0.6)
   - Only store `confidence >= 0.7`

## Acceptance criteria

[...]

PR #862, opened 2026-08-11
fix(ci): detect addAssignees silent-drop for new contributors

### LINKED ISSUE #861: pr-issue-guard: new-contributor fallback never fires — addAssignees silently ignores non-assignable logins
## Summary

`pr-issue-guard.yml`'s fallback for a non-assignable contributor is unreachable for the case it was written to handle. [#835](https://github.com/plur-ai/plur/issues/835) added it specifically so a brand-new contributor's first PR would not get a red X when `addAssignees` could not assign them — but that condition does not raise an exception, so the `catch` never runs.

Found while reviewing [#859](https://github.com/plur-ai/plur/pull/859), which copies the same assumption.

## The defect

`pr-issue-guard.yml` assigns the PR author to the linked issue and handles failure with a `try`/`catch`:

```js
try {
  await github.rest.issues.addAssignees({ owner, repo, issue_number: issueNumber, assignees: [prAuthor] });
  core.info(`Auto-assigned #${issueNumber} to @${prAuthor}`);
} catch (e) {
  core.warning(`Could not auto-assign … ${e.message}. A maintainer should assign it manually …`);
  await github.rest.issues.createComment({ /* fallback notice on the PR */ });
}
```

The comment above that `catch` states the reason it exists:

> Even with a write-capable token, GitHub only permits assigning users with repo access or prior interaction on the issue — a brand-new contributor may simply not be assignable.

That is accurate. What is not accurate is that it surfaces as an error. Per [GitHub's REST documentation for "Add assignees to an issue"](https://docs.github.com/en/rest/issues/assignees):

> Only users with push access can add assignees to an issue. **Assignees are silently ignored otherwise.**

The endpoint returns **201** and drops the non-permitted login. No exception is thrown.

## Consequences

1. **The fallback comment never posts for a non-assignable contributor** — precisely the population [#835](https://github.com/plur-ai/plur/issues/835) targeted. A maintainer is never told to assign manually.
2. **The run log states the opposite of what happened.** `core.info("Auto-assigned #N to @user")` is emitted on the success path, which is taken even when the assignee was silently dropped. Anyone auditing the Actions log sees confirmation of assignments that do not exist.
3. **The issue stays unassigned and nothing flags it**, so the duplicate-effort protection the guard exists to provide is absent exactly when the contributor is new — the case where duplicate effort is most likely and most costly.

The `catch` still fires for genuine transport or permission errors (network failure, 404 on a deleted issue, a read-only token), so the branch is not dead in general — only for the condition its own comment names.

## Suggested fix

Do not infer the outcome from the absence of an exception. Verify it:

```js
const { data: updated } = await github.rest.issues.addAssignees({ … });
const assigned = (updated.assignees || []).some(a => a.login.toLowerCase() === prAuthor.toLowerCase());
if (!assigned) {
  // silently-ignored assignee — same handling as the current catch branch
}
```

`addAssignees` returns the updated issue, so no extra request is needed. The existing `catch` should be kept for genuine errors; this adds the silent-drop case alongside it.

## Verification

- Confirmed against the documentation quoted above.
- Worth checking the Actions history for `pr-issue-guard`: if the fallback comment has **never** appeared on any external PR since [#835](https://github.com/plur-ai/plur/issues/835) landed, that is consistent with this diagnosis.
- Reproducible by pointing `assignees` at a login without repo access (an org name works) and observing a 201 with an unchanged assignee list.

## Scope note

This is independent of [#859](https://github.com/plur-ai/plur/pull/859) and should not be bundled into it — that PR is being rejected on separate grounds, and this defect exists in `main` today regardless of its outcome.

plur9:
Fixes #861.

## What

`addAssignees` returns HTTP 201 and **silently drops** logins that lack
repo access or prior issue interaction — no exception is raised. The
existing `try/catch` in `pr-issue-guard.yml` therefore never fired for
the case it was written to handle (#835): a brand-new contributor whose
first PR triggers the auto-assign path.

## Fix

Capture the `addAssignees` response and verify the login appears in the
returned `assignees` list. Introduce a `needsFallback` flag so the PR
comment (telling a maintainer to assign manually) posts for **both**:

- the silent-drop case (assignee not in response — was the unreachable path)
- genuine transport / auth errors (existing `catch` branch)

No extra API call: `addAssignees` already returns the updated issue object.

## Test plan

- [ ] Open a PR from an account with no repo access — fallback comment
  should now appear on the PR (previously never fired for this case)
- [ ] Open a PR from an org member — `core.info("Auto-assigned…")` emits
  and no fallback comment is posted
- [ ] Simulate a 404 / auth error — `catch` branch still fires and posts
  the fallback comment

### DIFF
--- a/.github/workflows/pr-issue-guard.yml
+++ b/.github/workflows/pr-issue-guard.yml
@@ -50,6 +50,44 @@ jobs:
 
             if (seen.size === 0) return;
 
+            // IDEMPOTENCY (#862 review).
+            //
+            // The trigger includes `edited`, so this runs again on every PR
+            // description change. Both comment paths below are unconditional,
+            // so without a guard the same notice accumulates once per edit —
+            // and for the fallback path the person affected is a first-time
+            // contributor editing their PR description, which is exactly what
+            // first-time contributors do. #835 exists to keep that person from
+            // getting a red X on their first contribution; an unbounded stack
+            // of identical "you cannot be assigned" comments is no better.
+            //
+            // Dedup on a marker rather than gating on `context.payload.action`,
+            // so that an edit which FIRST introduces `Fixes #123` still
+            // comments. Markers are per-issue, so a run handling two issues
+            // posts for each without either suppressing the other.
+            let existingComments = [];
+            try {
+              existingComments = await github.paginate(github.rest.issues.listComments, {
+                owner, repo, issue_number: prNumber, per_page: 100,
+              });
+            } catch (e) {
+              // Fail open: a missed dedup is a duplicate comment, whereas
+              // skipping the notice entirely loses the signal it carries.
+              core.warning(`Could not list PR comments for dedup: ${e.message}. Posting without it.`);
+            }
+
+            const postOnce = (marker, body) => {
+              if (existingComments.some(c => c.body?.includes(marker))) {
+                core.info(`Notice already present on #${prNumber} (${marker}) — not reposting`);
+                return Promise.resolve();
+              }
+              return github.rest.issues.createComment({
+                owner, repo,
+                issue_number: prNumber,
+                body: `${marker}\n${body}`,
+              });
+            };
+
             for (const issueNumber of seen) {
               let issue;
               try {
@@ -99,21 +137,19 @@ jobs:
                   );
                 }
                 if (needsFallback) {
-                  await github.rest.issues.createComment({
-                    owner, repo,
-                    issue_number: prNumber,
-                    body: `ℹ️ Could not auto-assign #${issueNumber} to @${prAuthor} — `
+                  await postOnce(
+                    `<!-- pr-issue-guard:fallback:${issueNumber} -->`,
+                    `ℹ️ Could not auto-assign #${issueNumber} to @${prAuthor} — `
                       + `GitHub only allows assigning users with repo access or prior interaction on the issue. `
                       + `Maintainer: please assign it manually so the claim is recorded. This is not a problem with the PR.`,
-                  }).catch(err => core.warning(`Fallback comment failed: ${err.message}`));
+                  ).catch(err => core.warning(`Fallback comment failed: ${err.message}`));
                 }
               } else if (!assignees.includes(prAuthor)) {
                 const list = assignees.map(a => `@${a}`).join(', ');
-                await github.rest.issues.createComment({
-                  owner, repo,
-                  issue_number: prNumber,
-                  body: `⚠️ This PR addresses #${issueNumber}, which is assigned to ${list} — possible duplicate effort. Confirm coordination before merging.`,
-                });
+                await postOnce(
+                  `<!-- pr-issue-guard:conflict:${issueNumber} -->`,
+                  `⚠️ This PR addresses #${issueNumber}, which is assigned to ${list} — possible duplicate effort. Confirm coordination before merging.`,
+                );
                 core.warning(`Conflict: #${issueNumber} is assigned to ${list}; PR opened by @${prAuthor}`);
               }
             }
--- a/.github/workflows/pr-issue-guard.yml
+++ b/.github/workflows/pr-issue-guard.yml
@@ -68,18 +68,37 @@ jobs:
                 // issue — a brand-new contributor may simply not be assignable.
                 // That is a fact about the account, not a fault in the PR, and
                 // it must not put a red X on someone's first contribution.
+                //
+                // addAssignees returns HTTP 201 and silently drops non-permitted
+                // logins — no exception is raised (#861). Verify via the response.
+                let needsFallback = false;
                 try {
-                  await github.rest.issues.addAssignees({
+                  const { data: updated } = await github.rest.issues.addAssignees({
                     owner, repo,
                     issue_number: issueNumber,
                     assignees: [prAuthor],
                   });
-                  core.info(`Auto-assigned #${issueNumber} to @${prAuthor}`);
+                  const assigned = (updated.assignees || []).some(
+                    a => a.login.toLowerCase() === prAuthor.toLowerCase()
+                  );
+                  if (assigned) {
+                    core.info(`Auto-assigned #${issueNumber} to @${prAuthor}`);
+                  } else {
+                    needsFallback = true;
+                    core.warning(
+                      `Could not auto-assign #${issueNumber} to @${prAuthor}: `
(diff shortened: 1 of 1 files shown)


plur9:
PR reviewed — CI all passing. Fix is correct: GitHub's `addAssignees` returns HTTP 201 silently dropping non-permitted assignees; verifying the response `.assignees` array is the right detection approach. `needsFallback` flag correctly gates both the warning and the comment. Cannot self-approve (same account), needs human review + merge.

crtahlin (changes requested):
This fixes a bug I filed, so I verified it by executing the script rather than by checking the diff resembles what I suggested.

## The fix works — confirmed by running it

I extracted the `github-script` body and ran it against mocked `github` / `context` / `core`:

| Scenario | `core.info` | `core.warning` | Fallback comment |
|---|---|---|---|
| Silent drop (the [#861](https://github.com/plur-ai/plur/issues/861) bug) | none | "assignee silently dropped" | **1 — was 0** |
| Assignable user | `Auto-assigned #123 to @newbie` | none | 0 |
| Throws (transport/auth) | none | `404 Not Found` | 1 |

All three behave as described. The part that matters most: on a silent drop `core.info("Auto-assigned…")` no longer emits, so the log has stopped reporting an assignment that did not happen. That was the substance of [#861](https://github.com/plur-ai/plur/issues/861) and it is properly fixed, with no extra request — the response object was already in hand.

## CRITICAL — the fix activates a comment path that has no idempotency guard

The trigger is `pull_request_target: types: [opened, edited, reopened]`. Same silent-drop scenario, three runs (one open, two description edits):

```
--- SILENT DROP, PR edited twice more (3 runs) ---
  comments posted: 3
    [1] ℹ️ Could not auto-assign #123 to @newbie — GitHub only allow…
    [2] ℹ️ Could not auto-assign #123 to @newbie — GitHub only allow…
    [3] ℹ️ Could not auto-assign #123 to @newbie — GitHub only allow…
```

Before this PR the fallback path was unreachable for the silent-drop case, so the missing guard was harmless. It is now live, and the assignment keeps failing on every re-run because the contributor is still not assignable — so the comment repeats on every edit, without bound.

The person affected is a first-time external contributor editing their PR description, which is what first-time contributors do. [#835](https://github.com/plur-ai/plur/issues/835) existed to stop that person getting a red X on their first contribution; merging this as-is replaces the red X with an accumulating stack of identical "you cannot be assigned" comments. That defeats the purpose the change is in service of, which is why this is blocking rather than a note.

### Suggested fix

Dedup on a marker rather than gating on `context.payload.action`, so that an edit which *first* introduces `Fixes #123` still comments:

```js
const marker = `<!-- pr-issue-guard:fallback:${issueNumber} -->`;
const { data: existing } = await github.rest.issues.listComments({
  owner, repo, issue_number: prNumber, per_page: 100,
});
if (!existing.some(c => c.body?.includes(marker))) {
  await github.rest.issues.createComment({
    owner, repo, issue_number: prNumber,
    body: marker + '\n' + /* existing body */,
  });
}
```

Because `needsFallback` already funnels both failure routes into a single `createComment` call, this stays a one-place change.

## NOTE

- The pre-existing conflict-warning branch (`else if (!assignees.includes(prAuthor))`) has the same gap and is **already live today** — duplicate "possible duplicate effort" comments can already accumulate across edits. Not introduced here, but the same marker helper covers both, and fixing only the new branch leaves the older spam path in place. Optional in this PR, cheap if done here.
- The fallback body is shared between the silent-drop and throw cases, and its wording ("GitHub only allows assigning users with repo access or prior interaction") is accurate for the drop but not for a 404 or an auth failure. Minor, and arguably better than maintaining two near-identical strings.
- `(updated.assignees || [])` guards a missing array; if `data` itself were absent the `TypeError` lands in the `catch` and sets `needsFallback`, degrading to the fallback comment. That is the right direction — noting it so the destructure is not assumed total.

## Praise

- This is [#861](https://github.com/plur-ai/plur/issues/861)'s suggested fix implemented exactly, including the detail that matters: `needsFallback` unifies the two failure routes so the comment is emitted once from a single place, instead of duplicating `createComment` into both branches. That is what keeps the dedup fix above small.
- The `core.warning` text now distinguishes the two causes — "assignee silently dropped (no repo access or prior issue interaction)" versus the exception message — so the Actions log says which failure occurred. That is the [#861](https://github.com/plur-ai/plur/issues/861) lesson applied to the log line and not only to the control flow.

## Verification

- Script extracted from the YAML and executed against mocks; all four paths above exercised.
- Syntax valid (async-wrapped, as `github-script` evaluates it). Note for anyone repeating this: `node --check` on the raw body reports a syntax error because top-level `await` is invalid in a CommonJS script — an artifact of checking it unwrapped, not a defect here.
- YAML parses; triggers `[opened, edited, reopened]`; one job, one step.
- Security posture unchanged: no checkout under `pull_request_target`, PR body still used only as a regex source, `permissions` untouched.
- CI green on all five checks, none of which exercise this workflow.

The ask is just the marker dedup. Everything else here is ready.

plur9:
Marker dedup is in — thanks for catching this, the failure mode you described is worse than the bug the PR fixes.

Implemented as suggested, keyed on the marker rather than on `context.payload.action`, so an edit that *first* introduces `Fixes #123` still comments. Two deltas from your sketch:

- **Comments are listed once before the loop, not per issue** — `github.paginate` over `listComments`, so a PR closing three issues costs one paginated read instead of three. Markers are per-issue, so a run handling two issues still posts for each.
- **Took the optional half too.** The pre-existing conflict-warning branch now goes through the same helper, since it could already stack duplicate "possible duplicate effort" comments across edits. Cheap here, as you said, and leaving the older spam path live while fixing the new one seemed like the wrong trade.

Fail-open on the dedup read: if `listComments` fails the notice is posted anyway with a warning. A missed dedup costs one duplicate comment; suppressing the notice loses the signal entirely.

Verified the way you did — extracted the script, ran it against mocks:

| Scenario | Runs | Comments | Was |
|---|---|---|---|
| Silent drop (#861) | 3 | **1** | 3 |
| Assignable user | 3 | 0 | 0 |
| `addAssignees` throws | 3 | **1** | 3 |
| Conflict, assigned elsewhere | 3 | **1** | 3 |
| `listComments` fails | 2 | 2 + warning | n/a |

`core.warning` still fires on every run — only the comment is deduped, so the Actions log keeps a per-run record of what happened.

YAML re-parsed: triggers `[opened, edited, reopened]`, `permissions` unchanged, one job / one step, still no checkout.

Issue #860, opened 2026-08-11
Decision: detect silently-dropped optional array params, or accept #297 as unverifiable


crtahlin:
**Decision needed, not a bug report.** No action is proposed until this is settled.

## The situation

[#297](https://github.com/plur-ai/plur/issues/297) (optional array param empties the payload) is CLOSED. [#772](https://github.com/plur-ai/plur/issues/772) (whole-payload drop, refines it) is OPEN at P2. [#853](https://github.com/plur-ai/plur/pull/853) removes a real bias from the forensic log so it can start testing the array hypothesis.

Two structural constraints survive that fix — detail and counts in [this comment](https://github.com/plur-ai/plur/issues/772#issuecomment-5252531501):

1. `missingFields` derives from Zod issues where `received === 'undefined'`, which only fires for **required** fields. Exactly **one** array param in the whole 42-tool surface is required (`plur_session_end.engram_suggestions`), so array-shaped evidence can only ever come from one tool.
2. A dropped **optional** array is not an error at all. The call validates, succeeds, and the array is silently gone. Nothing is logged and the caller is never told.

So we currently cannot tell whether [#297](https://github.com/plur-ai/plur/issues/297) is fixed or merely invisible.

## The evidence that makes this worth deciding

The shipped log recorded **5 partial drops** in seven days — `plur_session_end` arriving with `summary` but without `engram_suggestions`. Those are proof that the partial-drop mechanism is real and currently active, not hypothetical.

The only reason all 5 failed **loudly** is that `engram_suggestions` happens to be required. Had the dropped field been optional — as every other array param in the surface is — the same mechanism would have produced a silent success with data missing.

Concretely, for the highest-traffic write in the product: a partial drop of `plur_learn.tags` stores the engram **without its tags**. No error, no log line, no signal to the caller. The engram is then harder to recall for exactly the keywords the author chose to make it findable. That is quiet quality decay in the core write path, and we have no way to say whether it is happening once a week or never.

## The decision

**Option A — build call-site detection.** Echo the array-typed params that actually arrived back in the tool response (e.g. `arrays_received: ["tags"]`). A caller that sent `tags` and does not see it has caught the drop immediately, with no log analysis. Covers all 11 array-carrying tools rather than the single required one. Cost: one additive response field, no schema-semantics change, no breaking change. Turns a forensic question into an immediate one.

[...]

PR #872, opened 2026-08-12
docs(adr): ADR-0006 — report only what was confirmed

crtahlin:
## Summary

Adds `docs/adr/ADR-0006-report-only-what-was-confirmed.md` and its index row. Status is **Proposed** — opening it for discussion rather than presenting it as settled.

## Why

A review pass across [#842](https://github.com/plur-ai/plur/pull/842), [#849](https://github.com/plur-ai/plur/pull/849), [#851](https://github.com/plur-ai/plur/pull/851), [#853](https://github.com/plur-ai/plur/pull/853), [#855](https://github.com/plur-ai/plur/pull/855), [#856](https://github.com/plur-ai/plur/pull/856) and [#859](https://github.com/plur-ai/plur/pull/859) surfaced the same defect five times, in unrelated subsystems written at different times. In each case a caller-visible outcome was derived from **control flow reaching a point** rather than from **observing the effect**:

- `plur_learn` returns a hardcoded `decision: 'ADD'`
- `learn-async` defaulted to `ADD` when no LLM was configured, so a hash-only write reported identically to a semantically checked one ([#854](https://github.com/plur-ai/plur/issues/854))
- `pr-issue-guard.yml` logged `Auto-assigned…` for assignees the API had silently dropped ([#861](https://github.com/plur-ai/plur/issues/861))
- `issue-mention-guard.yml` copied that assumption ([#859](https://github.com/plur-ai/plur/pull/859))
- `forget()` reported success and echoed a statement the caller had never written ([#831](https://github.com/plur-ai/plur/issues/831))

Five instances by different authors across MCP handlers, core, a shell script and two workflows is a missing convention rather than five coincidences.

What makes the class expensive is that it is invisible by construction: the failure produces a success message, so nothing in logs, tests, or CI separates it from working code. All five shipped with green suites. [#854](https://github.com/plur-ai/plur/issues/854) is the measured cost — 131 near-duplicate engrams across 63 clusters accumulated for five months because a write that had never been semantically checked reported the same `ADD` as one that had.

## What the ADR says

Three clauses:

1. **Confirm, do not infer** — where an operation can partially succeed or silently drop, verify the effect. Absence of an exception is not evidence of an effect; GitHub's own docs say assignees are *"silently ignored"*, which no `try`/`catch` can observe.
2. **"Not checked" is a distinct outcome, not the happy path** — never fall back to the value that means the check ran and passed.
3. **Unmeasured must stay distinguishable from measured-zero** — `undefined` means "did not look", `[]` means "looked and found none"; `?? []` at a call site destroys that.

Plus a reviewer checklist, the rejected alternatives (notably *why not just add tests* — all five shipped green, and in [#853](https://github.com/plur-ai/plur/pull/853) a test actively pinned the bias it should have caught), and the consequences including the honest one: the real cost is plumbing, and a discriminator no caller can see satisfies the letter of the rule and none of its purpose.

## Framing

This is written as **recording existing practice, not inventing a rule**. The final section credits four places in the codebase that already got it right — `dedup.mode`'s tri-state, `missing_array_params`' `undefined ≠ []` doc comment, `core_installed` in `release.sh`, and `embeddingSearchWithScores` returning `[]` read as "did not run". The convention was already emergent; it had just never been written down, so each subsystem rediscovered or missed it independently.

## Follow-up work it implies

- [#861](https://github.com/plur-ai/plur/issues/861) — `pr-issue-guard.yml` read-after-write. Clause 1, filed, fix in [#862](https://github.com/plur-ai/plur/pull/862).
- `plur_learn` should return the real decision from the write path instead of a literal, and surface `dedup.mode`. Clause 2, **not yet filed** — happy to open it if the ADR is accepted.
- [#860](https://github.com/plur-ai/plur/issues/860) — the open decision on detecting silently-dropped optional array params is the same clause applied to the transport layer.

## Test plan

- [ ] Read the ADR and say whether the rule as stated is one you want enforced in review
- [ ] Check the four "precedent" examples are fairly characterised — they are the argument that this is not a new burden
- [ ] Confirm `docs/adr/README.md` index row renders and links correctly
- [ ] Decide `Proposed` → `Accepted`, or push back on scope

Docs only — no code paths touched.

### DIFF
--- a/docs/adr/ADR-0006-report-only-what-was-confirmed.md
+++ b/docs/adr/ADR-0006-report-only-what-was-confirmed.md
@@ -1,0 +1,99 @@
+# ADR-0006: Report only what was confirmed
+
+Status: **Proposed**
+Date: 2026-08-12
+Authors: independent review pass over #842 / #849 / #851 / #853 / #855 / #856 / #859
+Related: [#854](https://github.com/plur-ai/plur/issues/854), [#831](https://github.com/plur-ai/plur/issues/831), [#835](https://github.com/plur-ai/plur/issues/835), [#861](https://github.com/plur-ai/plur/issues/861), [#772](https://github.com/plur-ai/plur/issues/772), [#860](https://github.com/plur-ai/plur/issues/860)
+
+## Context
+
+A single review pass across seven PRs surfaced the same defect five times, in unrelated subsystems written at different times. In each case a caller-visible outcome was derived from **control flow reaching a point** rather than from **observing the effect**:
+
+| Where | The claim | What was actually true |
+|---|---|---|
+| `packages/mcp/src/tools.ts` — `plur_learn` | Returns hardcoded `decision: 'ADD'` | The write path may have deduplicated, updated, or merged. The handler never asks |
+| `packages/core/src/learn-async.ts` (pre-[#856](https://github.com/plur-ai/plur/pull/856)) | `decision = 'ADD'` when no LLM was configured | Only an exact content-hash check ran. A hash-only write reported identically to a semantically checked one |
+| `.github/workflows/pr-issue-guard.yml` | `core.info("Auto-assigned #N to @user")` | `addAssignees` returns **201 and silently ignores** non-permitted logins. Logged success, nothing assigned ([#861](https://github.com/plur-ai/plur/issues/861)) |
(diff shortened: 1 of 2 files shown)


plur9 (approved):
Well-scoped ADR. The five-instance table in the Context section is persuasive evidence that this is a systemic pattern, not a one-off. The three clauses are precise and the tri-state model (undefined = not checked, [] = checked and empty) is a clean convention. Proposed status is correct — this should stay Proposed until the remaining call sites that still violate clause 3 are closed. All CI passing. @plur9 ready for merge.

PR #873, opened 2026-08-12
fix(core,mcp): stop reporting a cached remote failure as the live state (#864)

### LINKED ISSUE #864: MCP process caches config and remote-store health for its lifetime, producing false server-fault reports
An `@plur-ai/mcp` process loads its config and remote-store health at startup and never revisits either. On a workstation where the process survives for days, both go stale, and the resulting misreports point the operator at the wrong component.

## Observed

Six `@plur-ai/mcp` v0.17.2 processes on one workstation, the oldest **13 days** old (stdio servers accumulated across Claude sessions).

**Stale config.** `~/.plur/config.yaml` had its enterprise tokens rotated at 13:23. `plur_doctor`, run hours later, reported:

> `remote store: <host> — Reachable, but token expires in 4d — reauth soon`

The token on disk had **30 days** left (`exp` 2026-09-10). The process was decoding the token it read at spawn time.

**Stale health.** `plur_recall` returned:

```
"remote_stores": [{"host": "<host>", "status": "timeout"}]
"warning": "...serving local only; results may be missing team engrams"
```

The recall had actually **succeeded**. The results included engrams that exist in no local store — verified across every configured store YAML and the SQLite index — so they could only have come from the remote. The `timeout` was a cached probe result from before a network state change (VPN brought up in between), served unchanged afterwards.

## Why this is worse than a stale value

Both misreports accuse the **server** of a fault that is entirely client-side, and they do it in the operator's own diagnostic tooling. The natural response is to check network, VPN, credentials, and server health — all of which were fine. Correct diagnosis took about an hour and ended in `git log` over the local engram store.

The recall case is the more serious one. A degraded client silently serving local-only results is indistinguishable from a healthy one, so a user can work for a long stretch believing they have team memory when they do not. `"serving local only"` is present in the warning, but it is attached to a status that was already wrong.

## Expected

**Config:** re-read on change (mtime check is enough) or fail loudly when the loaded config is older than the file on disk. A 13-day-old in-memory token is never the right answer.

**Health:** a cached remote-store status should carry a TTL and should be invalidated by, at minimum, a subsequent *successful* interaction with that store. A store that just answered a query should not be reported as timed out in the same response.

**Consistency:** if results were served from a remote, `remote_stores[].status` must reflect that. The current output contradicted itself — it returned remote-only records while reporting the remote unreachable — and that self-contradiction is the cheapest possible signal that the cache is stale.

**Process hygiene:** worth considering whether stdio MCP servers should exit when their parent session ends. Six live processes from sessions long closed is its own problem, and it is what allowed state to age this far.

## Related

`plur-ai/enterprise#919` — the server-side counterpart: the admin UI has no view of connected clients, so none of this was visible from the other end either. A version + uptime header on each request would let the server flag exactly this condition.

plur9:
Closes #864.

## What was happening

A long-lived MCP process reported one early recall failure as the host's current state for the rest of its life, and pointed the operator at a server that was healthy throughout. The issue reports diagnosis taking about an hour and ending in `git log` over the local engram store.

While reproducing it I hit the same wall from the other side: two thorough searches confidently reported that an engram did not exist. It did — created that day, in a team scope, on a host answering in ~600ms. Recall had been serving local-only results while reporting the remote unreachable.

## Root cause is two bugs compounding

**The cold-start race** produces the failure. The remote leg deliberately starts before the local pipeline so its latency overlaps. But the first recall in a process initialises the BGE embedder, which blocks Node's single event loop, and a plain `setTimeout` keeps counting through the block — so the abort fires on a request the loop never got round to servicing.

Measured against production:

| | result |
|---|---|
| Server, 1 scope, hybrid | 0.52–0.85s |
| Server, 12 scopes, hybrid | 0.63–0.85s |
| Client **cold** call (first in process) | **timeout**, host_ms 1335–1621 |
| Client **warm** call, same process | **ok**, ~1.1s, 10 rows |

The cold abort lands at ~1333ms — exactly the connect budget, 2/3 of the 2000ms default. The server is never given a chance.

This is worse than it first looks: `remote-recall.ts` notes hooks are one-shot processes at **~86% of recall volume**. For those, *every* call is a cold call — and the breaker counts these as host failures and parks healthy hosts.

**#864** then makes it permanent: the genuine timeout is cached and re-reported indefinitely, including inside `plur_doctor` directly beneath a fresh probe saying the host is fine.

## Changes

**1. Status carries its age.** `_lastRemoteOutcomes` is written only by hosts a recall actually dialed; a recall that dials nothing leaves the previous entry untouched. Outcomes now carry `observed_at`/`age_ms`. `remoteStoreStatus({ freshOnly: true })` is used by present-tense surfaces (the `remote_stores` block says "serving local only") and falls silent past the TTL; doctor reads everything and renders the age, so a cached outcome reads as history rather than as a live fault.

**2. A successful interaction invalidates a cached failure.** `checkRemoteHealth` retires a network-class failure for a host whose `/me` probe just succeeded. Deliberately network-class only (`PROBE_CLEARABLE_STATES`) — `unsupported`, `rate_limited`, `auth_expired` and `forbidden` are not disproved by a different route succeeding, and clearing them would trade a false alarm for a silent failure. The doctor rendering uses the same rule; an existing test in `remote-recall-surfacing.test.ts` caught an earlier version of this patch that applied the downgrade too broadly.

**3. Config is re-read where tokens are read.** The constructor's only re-read compares `stores.length`, so rotating a token in place was invisible until restart. `_remoteRecallHosts` and `checkRemoteHealth` now call `reloadConfigIfChanged()` (#307).

**4. `startBudgetTimer` measures serviced time, not wall time.** It samples the loop every 50ms; a late tick proves the loop was blocked and the overshoot is credited back rather than charged to the network. A host that is merely slow or dead accrues no credit and still fails at exactly its budget — that safety property is covered by a test. Credit is capped by an absolute `MAX_STARVATION_CREDIT_MS` rather than a multiple of the budget, so the 1500ms hook path cannot be stretched into something a human notices.

## Verification

Live, against the production store — cold process, first call, **stock 2000ms budget**:

```
before:  remote=timeout  host_ms=1335  rows=0
after:   remote=ok       host_ms=1583  rows=10
```

16 tests added. Verified failing against the unfixed source (6 of the 7 status/config tests; the 7th asserts unchanged behaviour as a guard against over-clearing, and the cold-start test was confirmed failing by reverting just the timer wiring to `setTimeout`).

`remote-recall.test.ts` 66/66, `remote-recall-surfacing.test.ts` 10/10.

Full suite is 55 failed / 3762 passed — **all 55 pre-existing and unrelated**: `better-sqlite3@11.10.0` cannot load or compile against Node 26 (its C++ uses V8 APIs removed in 26: `GetPrototype`, `Context::GetIsolate`, `PropertyCallbackInfo::This`). Identical failures reproduce on unmodified `main`; every one is a module-load error, with zero assertion failures. Worth its own issue — this branch does not touch it.

## Not addressed

Rows can still reach a response via the `RemoteStore` peek cache while the live leg failed, which is its own reported self-contradiction — that belongs to #844/enterprise#428 and is left alone here. The "stdio servers should exit with their parent session" suggestion from the issue is also out of scope; the eight accumulated processes that motivated it deserve a separate change.

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -48,6 +48,7 @@ import { decodeJwtExpiry, decodeJwtPayload } from './jwt.js'
 import { RemoteStore, normalizeEndpointUrl } from './store/remote-store.js'
 import {
   remoteRecall, isRemoteRecallDisabled, resolveRemoteRecallTimeoutMs, scopeOrg,
+  REMOTE_STATUS_TTL_MS, PROBE_CLEARABLE_STATES,
   type RemoteRecallHost, type RemoteRecallResult, type HostRecallOutcome, type RemoteStoreStatusEntry,
 } from './remote-recall.js'
 import { YamlPrimaryStore } from './store/yaml-primary-store.js'
@@ -294,6 +295,8 @@ export {
   claimHookDegradationLines,
   MAX_REMOTE_QUERY_CHARS, MAX_REMOTE_RESPONSE_BYTES, DEFAULT_REMOTE_RECALL_TIMEOUT_MS,
   BREAKER_FAILURE_THRESHOLD, BREAKER_COOLDOWN_MS, UNSUPPORTED_TTL_MS, HOOK_HEADER_REPEAT_MS,
+  REMOTE_STATUS_TTL_MS, PROBE_CLEARABLE_STATES,
+  startBudgetTimer, BUDGET_TICK_MS, MAX_STARVATION_CREDIT_MS,
   type RemoteRecallHost, type RemoteRecallResult, type HostRecallOutcome,
   type RemoteHostState, type RemoteStoreStatusEntry, type RemoteRecallOptions,
 } from './remote-recall.js'
@@ -3675,8 +3678,11 @@ export class Plur {
   // -------------------------------------------------------------------------
 
   /** Last per-host recall outcomes (in-process), keyed by normalized URL —
-   *  feeds `remoteStoreStatus()` (plan A4′). */
-  private _lastRemoteOutcomes = new Map<string, HostRecallOutcome>()
+   *  feeds `remoteStoreStatus()` (plan A4′). Each entry carries the time it
+   *  was observed: the map is written ONLY by hosts a recall actually dialed,
+   *  so without an age the newest entry is indistinguishable from one made
+   *  days ago by a process that has since been idle (#864). */
+  private _lastRemoteOutcomes = new Map<string, { outcome: HostRecallOutcome; observed_at: number }>()
 
   /** Where breaker/cooldown/unsupported state persists across processes.
    *  Inside the store root so tests and PLUR_PATH overrides isolate it. */
@@ -3714,6 +3720,12 @@ export class Plur {
    * feeds the doctor warning for that misconfiguration.
    */
   private _remoteRecallHosts(options?: { scope?: string; session?: string; remote_project?: RemoteProjectConfig }): RemoteRecallHost[] {
+    // Pick up out-of-process config edits (#307) before reading tokens: a
+    // rotated credential must reach the very next dial, not the next restart.
+    // The constructor's only re-read compares `stores.length`, so a rotation
+    // that leaves the store count unchanged was previously invisible for the
+    // life of the process (#864).
+    this.reloadConfigIfChanged()
     const stores = (this.config.stores ?? []).filter(s => s.url)
     const rp = options?.remote_project
     const rpKey = rp ? normalizeEndpointUrl(rp.url) : null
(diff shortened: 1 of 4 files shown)


plur9 (commented):
Miles review (all CI passing): The timestamp-based TTL fix is correct. Wrapping outcomes in `{ outcome, observed_at }` is minimal and right. `freshOnly` filter and `noteRemoteHostReachable()` cleanly solve both sides of the stale-state problem. `reloadConfigIfChanged()` before dialing is the correct placement for credential-rotation pickup. Ready for merge — needs a non-author review.

Issue #865, opened 2026-08-11
Detect failing engrams: measure efficacy, and promote by form when an engram is the wrong mechanism


plur9:
An engram that fails to change behaviour is currently indistinguishable from one that works. The only way the system learns a rule was ineffective is the same mistake recurring **and** a human noticing it is the same mistake. That gap has been measured at 17 days in at least one documented case.

There is a cheap signal available that nobody is reading.

## The detection

When the user issues a correction, similarity-search the correction against the existing engram store.

- **High similarity to an existing active engram** → a rule covering this already existed, and the mistake happened anyway. That engram did not work.
- **No match** → genuinely new knowledge, current behaviour is correct.

The first case is an **efficacy failure**, and it is measurable rather than anecdotal. `plur_similarity_search` already returns cosine scores and the documented dedup bands (>0.9 duplicate, 0.7–0.9 related) are a reasonable starting threshold.

## Why the existing signals do not cover this

Inspecting a real store of 4,559 engrams (55 pinned):

- **`reference_count` is dead.** It reads `1` for 52 of 55 pinned engrams and `0` for the remaining three. It is not being incremented, so it carries no information at all. Either wire it up or remove it — right now it looks like a signal and is not one.
- **`frequency`** ranges 0–33 and is the only usable signal, but it measures *retrieval*, not *effect*. An engram can be retrieved often and ignored every time.
- **`feedback_signals`** are meaningful where present but sparse and manually driven. Real examples from that store — an opsec engram at `+0/−2` that was then misapplied again in a later session (a third negative that was never recorded), and a positioning engram at `+0/−2` with `frequency: 7`, i.e. fires often and is disliked every time.

None of these answer "did this engram change what happened?"

## Proposed

1. **Efficacy tracking.** On each correction, run the similarity check above. If a covering engram exists, increment a `violation_count` on it and record the violating context.

2. **A failing-engram report.** Surface engrams with `violation_count >= N`, ordered by violations. This is the queue of rules that are not working.

3. **Escalate to form review, not to rewriting.** This is the important part. A repeatedly-violated engram usually does not need better wording — it needs a **different mechanism**. Distinguish three causes before acting:
   - *wrong content* — the statement is inaccurate or too vague → rewrite
   - *did not load* — it was never in context at the deciding moment → pin, or widen tags
   - *wrong form* — it loaded, was accurate, and still did not change behaviour → **promote to a different form factor**

4. **Form taxonomy** for case three:

   | Failure class | Right mechanism |
[...]

PR #875, opened 2026-08-12
fix(core,deps): Node 26 compat + write_count/injection_count (#874, #866)

### LINKED ISSUE #866: reference_count never increments — 4,496 of 4,559 engrams read exactly 1
`reference_count` never increments past its initial value. Measured across a real store of **4,559 engrams**:

| value | engrams | share |
|---|---|---|
| `1` | 4,496 | 98.6% |
| `0` | 63 | 1.4% |
| anything else | **0** | 0% |

Not one engram in the store holds a `reference_count` of 2 or more, across months of daily use. It is a binary flag wearing a counter's name.

## Contrast with a field that works

`activation.frequency` on the same store:

- 62 distinct values
- 3,922 engrams (86%) non-zero
- highest observed: 594

So the store *is* recording usage — just not through this field.

## Why it matters beyond tidiness

`reference_count` reads as the natural answer to "has this engram earned its keep?", and it is the field a human reaches for first when auditing. It silently answers "yes, once" for essentially everything.

Encountered while auditing a pinned set (55 engrams, since reduced to 42). Every candidate showed `refs=1`, which initially read as "each of these has been used". It hasn't — the field simply cannot express anything else. Had `activation.frequency` not been available alongside it, the audit would have concluded the opposite of the truth: `frequency` ranged 0–33 across the same set and cleanly separated the load-bearing engrams from the inert ones.

A signal that is wrong in a *plausible* direction is worse than no signal, because it gets trusted.

## Expected

One of:

1. **Increment it** on each retrieval that actually reaches the model — if the intent is "times this engram informed an answer", distinct from `activation.frequency` (which appears to count retrieval events).
2. **Remove it**, if `activation.frequency` already covers the need. Two usage counters where one works is a trap regardless of which is authoritative.
3. **Rename it** to what it actually represents. If `0` vs `1` genuinely means "never referenced" vs "referenced at some point", then it is a boolean — `has_been_referenced` — and should say so.

(1) is the most useful if the two counters are meant to measure different things; the distinction between "retrieved" and "actually used" is exactly what an efficacy audit needs. But (2) is the honest fix if no such distinction was ever intended.

## Related

- #865 — detecting failing engrams. That proposal leans on usage signals to identify rules that are not working, and this field being dead removes one of the three signals it could have used. Worth resolving together: whichever way this goes, #865 needs to know which counter is authoritative.

### LINKED ISSUE #874: better-sqlite3@11.10.0 cannot build or load on Node 26 — 55 tests fail locally
`better-sqlite3@11.10.0` is incompatible with Node 26. On a workstation running Node v26.5.1 (ABI 147) this fails 55 tests across 8 files, all as module-load errors rather than assertion failures.

## Two ways it shows up

**Stale prebuilt binary.** The checked-out binary was built for ABI 137:

```
Error: The module '.../better_sqlite3.node'
was compiled against a different Node.js version using
NODE_MODULE_VERSION 137. This version of Node.js requires
NODE_MODULE_VERSION 147.
```

**Rebuilding does not fix it** — the source uses V8 APIs removed in Node 26:

```
./src/util/binder.lzz:40:37: error: no member named 'GetPrototype' in 'v8::Object';
                              did you mean 'GetPrototypeV2'?
./src/better_sqlite3.lzz:68:34: error: no member named 'GetIsolate' in 'v8::Context'
./src/objects/database.lzz:416:89: error: no member named 'This' in
                              'v8::PropertyCallbackInfo<v8::Value>'
gyp ERR! build error
```

Note that `pnpm rebuild better-sqlite3` **deletes the existing prebuilt binary before failing**, so the working tree ends up worse than it started — worth knowing before anyone tries it.

## Affected files

```
test/import.test.ts              test/read-side-scope-visibility.test.ts
test/importers.test.ts           test/scope-pushdown.test.ts
test/multi-store.test.ts         test/storage-indexed.test.ts
test/pr1-indexed-migration.test.ts   test/store-contract.test.ts
```

`pnpm test` → 55 failed | 3762 passed | 166 skipped. Every failure is the module load; no assertions fail.

## Impact

CLAUDE.md states "~3500 tests. All must pass before committing", which is not currently achievable on Node 26. That is corrosive: a suite with a known-red baseline stops being a signal, and a real regression can hide among the noise.

## Suggested fix

Bump `better-sqlite3` to a release supporting Node 26 (12.x), verify prebuilds exist for ABI 147, and consider declaring a supported Node range in `engines` so the incompatibility surfaces at install rather than at test time.

Found while working #864 — that branch does not touch this, and its own tests pass.

plur9:
## Summary

- **#874** — bump `better-sqlite3` from `^11.0.0` to `^12.0.0` for Node 26 / ABI 147 compatibility. The 11.x prebuilts target ABI 137; rebuilding from source fails on removed V8 APIs (`GetPrototype`, `GetIsolate`, `This`). 55 tests were failing as module-load errors; they pass after the bump. Also adds `engines: { node: '>=20' }` to core and cli so the incompatibility surfaces at install.
- **#866** — `reference_count` never incremented in practice (4,496 of 4,559 engrams stuck at 1). The field is renamed to `write_count` everywhere (schema, code, spec, tests). A new `injection_count` field is added — distinct counter for `inject()` selections vs `recall()` activations; high injection_count + low positive feedback_signals is the efficacy-failure signal needed for #865. Migration backfill in `loadEngrams()` converts legacy `reference_count` → `write_count` on first parse and strips the old key.

## Test plan

- [ ] `pnpm test` passes on Node 22 (CI baseline)
- [ ] `pnpm test` passes on Node 26 — the 55 module-load failures (#874) should be gone
- [ ] `reference-count.test.ts` — all existing tests updated for `write_count`; new tests for legacy backfill and `injection_count`
- [ ] No existing engram stores broken — `write_count` defaults to 1 via Zod, old `reference_count` fields migrated on first parse

## Closes

Closes #874
Closes #866

## Merge resolution (main was 8 commits ahead)

This branch was behind `main`, so a naive merge would have reverted what main gained. Resolved conflict-by-conflict, taking main's logic and re-applying this branch's rename on top:

| File | Resolution |
|---|---|
| `engrams.ts` | Took main's `parseEngramFile` / `EngramStoreUnreadableError` restructure, then re-applied the `reference_count` → `write_count` migration inside its entry loop. Quarantined entries keep the **original** shape — a rejected engram is written back verbatim and must not be silently rewritten on its way to being preserved. |
| `index.ts` (dedup hit) | Took main's primary-store `findActiveByContentHash` delegation; only the comment carried the rename. |
| `index.ts` (×2 decrement sites) | **The dangerous one.** This branch predates #766's `options.force` handling, so taking its side would have silently reverted force-retire. Resolved as main's `options?.force ? 0 : …` with this branch's `write_count` field. |
| `remote-store.ts`, `stub-server.ts`, `remote-integration.test.ts` | Took main wholesale — this branch carried an older revision of the same #768 wire-payload work that main has since superseded. |

Verified after resolution: core typecheck, MCP typecheck, full core suite.

## Note on #874's reproduction

`better-sqlite3@11.10.0` builds from source cleanly on **Node 24** — verified locally by rebuilding it after the failed `pnpm rebuild` this issue warns about had deleted the prebuilt binary. So #874 really is Node-26-only, and the CI matrix (20/22/24) structurally cannot catch it. Worth considering whether the matrix should gain a Node 26 entry, otherwise this class of breakage stays invisible to CI and only ever shows up on a workstation.


plur9:
Holding review until full CI completes — currently only `issue-guard` has run (PR created 23:45 UTC, Node matrix + smoke likely still queued). The two changes are substantive: `better-sqlite3` v12 for Node 26 ABI 147 compat (#874) and the `reference_count` → `write_count` + `injection_count` rename (#866). Both are worth a careful CI pass before merging.

crtahlin (changes requested):
Three separable changes are bundled here. The [#874](https://github.com/plur-ai/plur/issues/874) bump is good and should land today — it unblocks 55 failing tests. The other two each have a blocking problem: `injection_count` contradicts an accepted ADR by the exact mechanism that ADR rejected, and the rename's migration covers two of the three loaders.

## CRITICAL — `injection_count` does what ADR-0002 forbids, by the mechanism it names

[ADR-0002](https://github.com/plur-ai/plur/blob/main/docs/adr/ADR-0002-derived-state-provenance.md) is **Accepted** and says, verbatim:

> Observational events are **never folded back into engram YAML** as counters.

> **Write amplification on the read path.** Folding co-fire counters into YAML would turn every inject — a read — into a locked write across every injected engram file. The injection hot path must stay read-only.

`index.ts:4569-4586` takes `_withStoreLock`, loads the entire primary store, mutates every injected engram, and calls `_writeEngrams` — a whole-corpus replace — on every `inject()`. That is a locked whole-store write on the read path.

It is also redundant. Twenty lines above, `index.ts:4546` already appends a `co_injection` history event carrying `ids: injected_ids`. `injection_count` for an engram is exactly the number of `co_injection` events containing its id — derivable from history JSONL, which is where ADR-0002 puts observational data and how it says derived state should be rebuilt.

If a materialised counter is genuinely wanted, that is a legitimate design discussion — but it needs to amend ADR-0002 rather than contradict it silently. As written it also makes every session start pay a full-corpus write at the 50k-engram Postgres tier, where `_writeEngrams` ends in `DELETE FROM engrams WHERE id NOT IN (…)`.

## CRITICAL — Postgres/PGLite silently resets `write_count`, and the compat fallback cannot fire

The migration is applied in `parseEngramFile` (correct) and `forget()` carries an explicit `?? (engram as any).reference_count` for remote rows (also correct, with a good comment). But `storage-postgres.ts:1782` `parseRow` calls `EngramSchemaPassthrough.safeParse(raw)` with no normalisation.

`EngramSchemaPassthrough` is `EngramSchema.passthrough()` — no `.partial()` — so Zod's `.default(1)` fills a **defined** value, which means the downstream `??` chain can never reach the legacy key. Minimal repro against this repo's Zod:

```
input : { id: 'ENG-1', reference_count: 5, sources: [5 entries] }
parsed: { id: 'ENG-1', write_count: 1, reference_count: 5, sources: [...] }
write_count ?? reference_count  ->  1        (true count: 5)
```

`write_count` gates retirement, so on the Postgres tier a single `forget()` retires an engram that had five corroborating same-scope writes. It fails silently, and it destroys the *most*-corroborated engrams preferentially. The `sources.length` heuristic is masked for the same reason — a filled default is not absent.

`parseRow`'s own doc comment records that "the two loaders disagreed about what a valid engram looks like" as a previously-shipped bug. This is that class recurring with a third loader, which is why a shared normalise-before-parse helper called by all three would be better than a second per-site patch. Filed as a follow-up so it does not have to be solved in this PR.

## WARNING

**`injection_count` only counts primary-store engrams.** The increment iterates `this._primaryStore.load()`, so engrams injected from remote/team stores and from packs never increment. The [#865](https://github.com/plur-ai/plur/issues/865) efficacy signal would be systematically absent for team-scoped engrams — exactly the population an organisation would audit.

**Renaming a field in a published v1 standard is a breaking spec change.** `spec/ENGRAM-STANDARD-v1.md` now documents `write_count` and instructs that implementations "MUST backfill on first parse". Any third-party implementation reading `reference_count` breaks, and a v1 document cannot impose a new MUST on already-conforming readers. Either keep `reference_count` as a documented deprecated alias readers may still emit, or version the standard.

**The increment swallows every error with no log.** `try { … } catch { /* best-effort */ }`, with no `logger.warning` — unlike the comparable best-effort paths in `learn-async.ts`. So the counter under-counts silently, which undermines the one property it needs: being trustworthy enough to audit against.

## NOTE

- **Bundling.** A dependency major, a 26-file mechanical rename, a new architectural counter, a spec change and a remote-store payload fix, closing two unrelated issues. The [#874](https://github.com/plur-ai/plur/issues/874) half is urgent and uncontroversial and is currently gated behind two architectural discussions.
- **CI cannot verify the Node 26 fix** — `ci.yml:27` is `node-version: [20, 22, 24]`, confirmed. Your PR-body note on this is right; filed separately so the conclusion is tracked where it outlives this PR.
- **No CHANGELOG entry** for the dependency major, the new `engines` constraint (which changes install behaviour for Node 18 users), or the renamed persisted field.

## Praise

- **The merge-resolution table is exemplary.** Naming the dangerous conflict explicitly — that taking this branch's side at the decrement sites would have silently reverted [#766](https://github.com/plur-ai/plur/issues/766)'s `options.force` handling — and explaining each resolution rather than asserting "resolved conflicts" is exactly right. I got a merge-order question wrong on [#851](https://github.com/plur-ai/plur/pull/851) earlier this week for want of precisely this discipline.
- **Quarantined entries keep the original shape.** Subtle and correct: a rejected engram is written back verbatim, so it must not be silently rewritten on its way to being preserved.
- **The remote-row compat fallback and its reasoning comment** are the right instinct — it simply was not carried to the third loader.

## Verification

```
npx vitest run --project @plur-ai/core   (worktree, Node v24.7.0)
  2714 passed, 34 skipped, 3 failed
packages/core/test/reference-count.test.ts
  18/18 PASS — including the new backfill and injection_count cases
```

The 3 failures are `sync-push-failure.test.ts`, which fails identically on [#855](https://github.com/plur-ai/plur/pull/855), [#856](https://github.com/plur-ai/plur/pull/856) and this branch — a git-invoking test that does not survive running inside a worktree. Environmental, not from this PR.

Note that the backfill test is titled "…in raw YAML", which is precisely the gap: nothing exercises `parseRow`. And CI's Node matrix was still **pending** when I checked, so 20/22/24 is not yet known to be green.

## Suggested split

1. **`better-sqlite3` ^12 + `engines`** — land now. Unblocks Node 26, no architectural question attached.
2. **`reference_count` → `write_count`** — land after the `parseRow` normalisation plus a test against a legacy Postgres row; settle the spec-alias question.
3. **`injection_count`** — separate discussion against ADR-0002, given the count is already derivable from the `co_injection` events being written today.

Happy to re-review any of the three the moment it is split out; the first needs nothing from me but a green matrix.

PR #881, opened 2026-08-13
docs(skill): add near-duplicate protocol to plur-memory skill (#878)

### LINKED ISSUE #878: plur_learn has no semantic dedup — decide whether the dominant write path should gate, and at what cost
## The gap

`plur_learn` — the dominant write path — has only **exact content-hash** dedup. It goes `learnRouted()` → sync `learn()`, which never consults `context.llm`; the LLM resolved at `tools.ts:981` and passed into the context is dead on that path. `learnAsync`, and with it every semantic dedup mechanism, is reachable **only** from `plur_learn_batch`.

Consequence: a reworded restatement of an existing engram is always written as a sibling. That is how #854 happened — ten facts re-issued in condensed form, none caught.

## What exists today

- **`plur_learn_batch`** already does LLM dedup by default: `mode` defaults to `'llm'` and `getLlmFunction()` resolves `OPENROUTER_API_KEY` / `OPENAI_API_KEY`. Nothing to enable — it has simply never had a key present in practice.
- **`plur_learn`** reports near-duplicates since #856 (`dedup.near_duplicates`), but nothing consumes the report, so the write happens regardless.
- **Cosine cannot gate this.** #856 established that production enriches only the incoming side, so an under-contexted write is compared against a fully-contexted stored engram: the #854 pair scores 0.8339 as actually written, *below* genuinely-distinct pairs at 0.8826. The classes overlap; no fixed threshold separates them.

## The decision

Routing `plur_learn` through a semantic decision means **an API call per memory write**. That is a product decision with real cost and latency, and it sits against this repo's stated "no external API calls in core — search must work offline at zero cost" principle (core would still not call out; the MCP layer would inject the function — but the effect on a default install is the same).

Options, roughly in increasing cost:

1. **Convention only.** Instruct agents (CLAUDE.md / the plur-memory skill) to read `dedup.near_duplicates` and supersede or merge rather than writing a sibling. Zero cost, zero latency, uses the `supersedes` edge that already exists, and puts the judgement where the context is. Weakness: relies on the agent actually reading the field.
2. **Opt-in LLM gate on `plur_learn`.** Off by default; on when a key is configured AND the user opts in. Preserves the offline-at-zero-cost default.
3. **On by default when a key is present.** Matches `plur_learn_batch`'s current behaviour, so the two paths stop diverging — but makes every write on a keyed install cost an API call.

Whichever is chosen, the `dedup_near_duplicate` history events added in #856 should be used first: they record the top score plus `incoming_has_domain` / `incoming_has_rationale` per write, so the real distribution can be inspected before committing to a mechanism.

## Related

- #854 — the incident that exposed this (closed; its three original asks are fixed, corrected, or consolidated)
- #856 — added the reporting this would act on
- #852 — the opposite failure on the same path: dedup false-positive, silent absorption into an unrelated engram

plur9:
## Summary

- Closes #878 (Option 1: convention only)
- Adds a **Near-Duplicate Protocol** section to both skill files (`skills/plur-memory/SKILL.md` and `packages/hermes/plur_hermes/skills/plur-memory.SKILL.md`)
- Instructs agents to check `dedup.near_duplicates` after every `plur_learn` and resolve high-similarity (≥ 0.85) writes with `supersedes` instead of leaving restatement siblings

## Why convention, not a gate

The decision is in the issue comment at #878. Short version: cosine can't gate this (#856 established the classes overlap), LLM could but costs an API call per write which conflicts with the offline-at-zero-cost invariant. The agent calling `plur_learn` has more context than any automated gate. Convention with explicit protocol is the right first layer; `dedup_near_duplicate` history events give us the measurement path to revisit if convention proves insufficient.

## What changes

The two skill files are kept in sync — they must match exactly. The new section:
1. Tells agents to check `dedup.near_duplicates` (already returned by `plur_learn` since #856)
2. Gives a ≥ 0.85 threshold as the "look carefully" trigger
3. Distinguishes restatement (undo + supersede) from correction (keep + supersede original)
4. Explains `dedup.mode` so agents understand what kind of check ran

## Test plan

- [ ] Skill content review — no functional code change; skill wording matches the tool description from #856
- [ ] Both files updated identically

🤖 heartbeat — 2026-08-13

### DIFF
--- a/packages/hermes/plur_hermes/skills/plur-memory.SKILL.md
+++ b/packages/hermes/plur_hermes/skills/plur-memory.SKILL.md
@@ -80,3 +80,49 @@ Meta-engrams are the highest-value knowledge: principles that transfer across do
 - Patterns: "This codebase uses repository pattern for data access"
 - Decisions: "We chose PostgreSQL for ACID compliance"
 - Conventions: "Always run lint before committing"
+
+## Near-Duplicate Protocol
+
+A `plur_learn` response may carry a `dedup` field reporting engrams close to what you just wrote. Convention, not a gate — the engine will not stop you, and nothing else will notice a reworded restatement.
+
+### 1. Read `dedup.mode` first
+
+- `cosine` — similarity ran locally, and `near_duplicates` lists what it found.
+- `hash-only` — **only** the exact-hash check ran: no candidates, no embedder, or dedup disabled. This means "not identical", **not** "not a duplicate".
+- `llm` — semantic classification. `plur_learn_batch` only; it never appears on a single `plur_learn`.
+
+**A missing `dedup` field is ambiguous and you cannot resolve it from the response.** It is omitted both when similarity ran and found nothing close, and when similarity never ran at all. Absence is therefore not evidence that your write is unique.
+
+### 2. Resolve the ids to statements
+
+Entries are `{ id, score }` — **there is no statement text in the payload**, and no fetch-by-id tool. To see what you nearly duplicated, run `plur_similarity_search` with the same statement you just wrote: it uses the same cosine mechanism, so its `engram_id` values line up with `near_duplicates`, and it returns `statement` and `scope` alongside them.
+
+Skipping this step means deciding on a number alone, which the next point explains is not enough.
+
+### 3. Judge on content — `score` orders the list, it does not decide
+
+**Do not use a similarity threshold as a decision boundary.** The measurements behind #878 are the reason:
+
+| Pair | Score |
+|---|---|
+| the real #854 duplicate, as actually written | 0.8339 |
+| "always rebase" vs "never rebase" — opposite meanings | 0.8826 |
+
+A genuine duplicate scored *below* two statements that contradict each other. Any fixed cut-off puts the founding case on the wrong side. Treat `score` as the order to inspect in, then read the statements and decide whether they assert the same fact.
+
+For reference, the engine records a `dedup_near_duplicate` history event above `NEAR_DUPLICATE_OBSERVATION_FLOOR = 0.75` — that is what it considers notable enough to log, not a threshold for you to act on.
+
+### 4. If it is the same fact, use one recipe
+
+Whether you are restating or correcting, the sequence is the same:
+
+```
+plur_forget <new-id>
+plur_learn "<statement>" supersedes: [<original-id>]
+```
+
+**Do not try to attach `supersedes` by re-learning the same statement.** That path hits exact content-hash dedup, which increments `write_count` and appends a source — it never writes `relations`. The edge is applied only when a *new* engram is created. So re-learning silently does nothing, and if you then forget the original you are left with a superseded fact archived and no record of what replaced it: worse than leaving it alone.
+
+Retiring first is what makes the re-learn work — retired engrams are excluded from hash dedup (#107), so the second call creates a fresh engram and the edge lands on it.
+
+If the two statements are genuinely distinct, the write stands and there is nothing to do.
--- a/skills/plur-memory/SKILL.md
+++ b/skills/plur-memory/SKILL.md
@@ -80,3 +80,49 @@ Meta-engrams are the highest-value knowledge: principles that transfer across do
 - Patterns: "This codebase uses repository pattern for data access"
 - Decisions: "We chose PostgreSQL for ACID compliance"
 - Conventions: "Always run lint before committing"
+
+## Near-Duplicate Protocol
+
+A `plur_learn` response may carry a `dedup` field reporting engrams close to what you just wrote. Convention, not a gate — the engine will not stop you, and nothing else will notice a reworded restatement.
+
+### 1. Read `dedup.mode` first
+
+- `cosine` — similarity ran locally, and `near_duplicates` lists what it found.
+- `hash-only` — **only** the exact-hash check ran: no candidates, no embedder, or dedup disabled. This means "not identical", **not** "not a duplicate".
+- `llm` — semantic classification. `plur_learn_batch` only; it never appears on a single `plur_learn`.
+
+**A missing `dedup` field is ambiguous and you cannot resolve it from the response.** It is omitted both when similarity ran and found nothing close, and when similarity never ran at all. Absence is therefore not evidence that your write is unique.
+
+### 2. Resolve the ids to statements
+
+Entries are `{ id, score }` — **there is no statement text in the payload**, and no fetch-by-id tool. To see what you nearly duplicated, run `plur_similarity_search` with the same statement you just wrote: it uses the same cosine mechanism, so its `engram_id` values line up with `near_duplicates`, and it returns `statement` and `scope` alongside them.
+
+Skipping this step means deciding on a number alone, which the next point explains is not enough.
+
+### 3. Judge on content — `score` orders the list, it does not decide
+
+**Do not use a similarity threshold as a decision boundary.** The measurements behind #878 are the reason:
+
+| Pair | Score |
+|---|---|
+| the real #854 duplicate, as actually written | 0.8339 |
+| "always rebase" vs "never rebase" — opposite meanings | 0.8826 |
+
+A genuine duplicate scored *below* two statements that contradict each other. Any fixed cut-off puts the founding case on the wrong side. Treat `score` as the order to inspect in, then read the statements and decide whether they assert the same fact.
+
+For reference, the engine records a `dedup_near_duplicate` history event above `NEAR_DUPLICATE_OBSERVATION_FLOOR = 0.75` — that is what it considers notable enough to log, not a threshold for you to act on.
+
+### 4. If it is the same fact, use one recipe
+
+Whether you are restating or correcting, the sequence is the same:
+
+```
+plur_forget <new-id>
+plur_learn "<statement>" supersedes: [<original-id>]
+```
+
+**Do not try to attach `supersedes` by re-learning the same statement.** That path hits exact content-hash dedup, which increments `write_count` and appends a source — it never writes `relations`. The edge is applied only when a *new* engram is created. So re-learning silently does nothing, and if you then forget the original you are left with a superseded fact archived and no record of what replaced it: worse than leaving it alone.
+
+Retiring first is what makes the re-learn work — retired engrams are excluded from hash dedup (#107), so the second call creates a fresh engram and the edge lands on it.
+
+If the two statements are genuinely distinct, the write stands and there is nothing to do.


plur9 (commented):
PR review — CI green. The Near-Duplicate Protocol section is correct and actionable:

- Score threshold (0.85) and the three-case decision tree are well-specified.
- 'Never skip this check' with the sibling-injection consequence is the right framing — makes the cost of skipping concrete.
- `dedup.mode` documentation clarifying `hash-only` means 'not identical', not 'not a duplicate' is an important distinction.

Applied identically to both `packages/hermes/plur_hermes/skills/plur-memory.SKILL.md` and `skills/plur-memory/SKILL.md` — both copies are in sync.

Cannot self-approve. Ready for a second reviewer or human approval.

crtahlin (changes requested):
Choosing convention over an API-call-per-write gate is the right call, and the protocol's shape is sound. But verified against the merged [#856](https://github.com/plur-ai/plur/pull/856) implementation in `main`, the text describes a response field that does not exist and prescribes a remedy that silently does nothing — so an agent following it cannot execute step 2, and the second branch of step 2 destroys the provenance the protocol exists to create.

## CRITICAL — `near_duplicates` entries carry no statement snippet

The skill says: *"Each entry has an `id`, `score`, and a statement snippet."*

The type in `main` (`types.ts:102`) is:

```ts
near_duplicates?: Array<{ id: string; score: number }>
```

`nearDuplicates()` builds it as `scored.map(s => ({ id: s.engram.id, score: s.score }))`, and the batch path in `learn-async.ts` produces the same shape. There is no snippet.

This is load-bearing rather than cosmetic. Step 2 asks the agent to judge whether the near-duplicate "covers the same fact", which requires its content — and the skill never says how to obtain it. None of the 43 MCP tools is a fetch-by-id, so an agent would have to `plur_recall` and hope the engram surfaces.

Two ways out, and the better one is probably not this PR: add the snippet to the payload in `nearDuplicates()` (which already holds `s.engram`, so it is a small change) and keep the skill text as written; or leave the payload alone and make the id→statement resolution an explicit step in the protocol.

## CRITICAL — the correction branch is a silent no-op that then retires the original

The skill says: *"Keep it — add `supersedes: [...]` via a `plur_learn` call with the same statement to attach the edge, then `plur_forget` the original."*

Re-learning the **same statement** hits exact content-hash dedup, which routes to `_recordDuplicate` (`index.ts:1557`). That function does exactly two things:

```ts
target.write_count = currentCount + 1
;(target as any).sources = [...currentSources, this._buildSourceEntry(scope, context)]
```

It never touches `relations`. `supersedes` is applied only on the new-engram creation path — the `plur_learn` tool description states it writes the edge "on the **new** engram", and `plur_rescope`'s description names this exact trap: *"Bypasses the content-hash dedup that makes a plur_learn re-emit a silent no-op."*

So following branch 2 increments a counter, attaches no edge, and then retires the original — leaving the correction unlinked and the superseded fact archived with no record of what replaced it. That is worse than doing nothing.

With current tooling both branches collapse to the same recipe:

```
plur_forget <new-id>
plur_learn "<statement>" supersedes: [<original-id>]
```

Branch 1 is already correct for this reason — re-learning after retirement creates a fresh engram, since retired engrams are excluded from hash dedup ([#107](https://github.com/plur-ai/plur/issues/107)). Branch 2 should say the same thing rather than a different, non-working sequence.

## WARNING — the 0.85 threshold contradicts this PR's own rationale and sits inside the measured overlap

[#878](https://github.com/plur-ai/plur/issues/878) argues for convention over a gate precisely because the classes overlap: *"the #854 pair scores 0.8339 as actually written, below genuinely-distinct pairs at 0.8826. The classes overlap; no fixed threshold separates them."*

The skill then adopts a fixed threshold of 0.85, which falls between those two measurements:

| Case | Score | What the rule says |
|---|---|---|
| The real [#854](https://github.com/plur-ai/plur/issues/854) duplicate, as actually written | 0.8339 | below 0.85 → *"the write stands. No action needed."* |
| "Always rebase" vs "Never rebase" — genuinely distinct | 0.8826 | above 0.85 → flagged for scrutiny |

The founding case of the whole effort is the one the numeric rule waves through. If the argument for convention is that scores cannot decide, the protocol should say that: inspect every returned near-duplicate on content, and treat `score` as the ordering of the list rather than a decision boundary.

Separately, the implementation's own notability threshold is `NEAR_DUPLICATE_OBSERVATION_FLOOR = 0.75` — what triggers a `dedup_near_duplicate` history event. The skill and the code currently disagree about what counts as notable, which will make the measurement path in the PR description harder to interpret later.

## WARNING — absence of `dedup` is unreadable, yet the skill says "never skip this check"

The handler emits the field only when there are near-duplicates:

```ts
...(dedup?.near_duplicates?.length ? { dedup } : {}),
```

When `mode` is `hash-only` — no candidates, embedder unavailable, or `dedup.enabled: false` — the key is absent entirely, so an agent cannot distinguish "similarity ran and nothing was close" from "similarity never ran". Always returning `dedup` with its `mode` would fix it at no cost. Same rule as clause 3 of the ADR proposed in [#872](https://github.com/plur-ai/plur/pull/872): unmeasured must stay distinguishable from measured-zero.

## NOTE

- `mode: 'llm'` cannot appear on a `plur_learn` response — `nearDuplicates()` returns `'cosine' | 'hash-only'` only. The section is framed around `plur_learn`, so listing `llm` without marking it batch-only invites an agent to wait for a mode that never arrives.
- Both copies received byte-identical text, with placement differing appropriately for each file's structure. Worth stating explicitly that nothing in CI can catch a skill that misdescribes a response shape: these two files are prose about a type, with no link to the type.

## Praise

- *"`hash-only` means 'not identical', not 'not a duplicate'"* is exactly right, lifted faithfully from the tool description, and is the most useful sentence in the addition.
- The reasoning in the PR body for convention over a gate — cosine cannot separate the classes, an LLM call per write violates the offline-at-zero-cost invariant, the calling agent has more context than any automated gate — is sound and worth keeping as the record of why.

## Verification

Documentation only; nothing executed. CI green on all six checks. Claims were checked against `main`: `nearDuplicates()` at `index.ts:2730`, the `plur_learn` handler at `tools.ts:1108`, `_recordDuplicate` at `index.ts:1557`, the `near_duplicates` type at `types.ts:102`, and the observation floor at `learn-async.ts:25`.

Both blockers are small text fixes, with the caveat that finding 1 may be better solved by adding the snippet to the payload — which would be a code change and probably a separate PR.

plur9:
@crtahlin — rebuilt from `main`. All five findings addressed, each verified against the source rather than taken on trust. `DIRTY` is resolved.

## 1. CRITICAL — no statement snippet · **fixed by naming the resolution step**

Confirmed: `types.ts:109` is `near_duplicates?: Array<{ id: string; score: number }>`, and there is no fetch-by-id among the 44 tools. So the old text asked the agent to judge whether an entry "covers the same fact" using content it was never handed.

You noted the better fix might be adding the snippet to the payload, in a separate PR. Agreed — but the skill should be executable against the payload that ships **today**, so it now names the step: run `plur_similarity_search` with the same statement. It uses the same cosine mechanism, so its `engram_id` values line up with `near_duplicates`, and it returns `statement` and `scope` alongside them. That makes step 2 possible with no code change, and the payload improvement remains worth doing on its own merits.

## 2. CRITICAL — the correction branch destroyed provenance · **both branches collapsed to one recipe**

Confirmed: `_recordDuplicate` touches `write_count` and `sources` and never `relations`; `supersedes` lands only on the new-engram path. So the old branch 2 attached no edge and then archived the original — correction unlinked, superseded fact gone, no record of what replaced it. Worse than nothing, exactly as you said.

Both branches are now the one sequence that works:

```
plur_forget <new-id>
plur_learn "<statement>" supersedes: [<original-id>]
```

And the text says *why* retiring first is load-bearing rather than ceremonial — #107 excludes retired engrams from hash dedup, so the second call creates a fresh engram for the edge to land on. Without that sentence the recipe looks like superstition and invites someone to "simplify" it back into the broken form.

## 3. WARNING — the threshold contradicted the argument · **dropped entirely**

This was the most useful finding, because the number was undermining the case for the whole protocol. Both measurements are now in the skill as a table:

| Pair | Score |
|---|---|
| the real #854 duplicate, as written | 0.8339 |
| "always rebase" vs "never rebase" | 0.8826 |

with the rule stated as: **`score` orders the list; it does not decide.** Keeping the measurements next to the rule means the reasoning travels with it — otherwise the next reader sees an unexplained "read them all" and reintroduces a cut-off.

`NEAR_DUPLICATE_OBSERVATION_FLOOR = 0.75` is named as what it is — a logging threshold for the `dedup_near_duplicate` event — so the skill and the code no longer disagree about what counts as notable.

## 4. WARNING — absent `dedup` unreadable · **stated outright**

Confirmed at `tools.ts:1155`: `...(dedup?.near_duplicates?.length ? { dedup } : {})`. The skill now says plainly that absence occurs both when similarity found nothing and when it never ran, and that **absence is not evidence of uniqueness**. Clause 3 of the #872 ADR, applied to prose.

Reading `dedup.mode` is now step 1 rather than a trailing note, since it is the thing that determines whether the rest of the protocol has anything to work with.

## 5. NOTE — `mode: 'llm'` · **marked batch-only**

Confirmed `nearDuplicates()` returns `'cosine' | 'hash-only'` (`index.ts:2770`). Now explicit that `llm` is `plur_learn_batch` only and never appears on a single `plur_learn`.

## On your closing observation

> nothing in CI can catch a skill that misdescribes a response shape: these two files are prose about a type, with no link to the type.

That is the real finding here — every one of the five was a claim about a shape, and all five shipped green. #927 just landed the same idea for the engram standard: a test asserting every `EngramSchema` field appears in `ENGRAM-STANDARD-v1.md`. The analogue for skills would be asserting that the field names a SKILL.md quotes from a response actually exist on that response type. Worth its own issue; I have not filed it as part of this PR, and would rather it be scoped deliberately than bolted on here.

## Verification

```
reset to origin/main, section rewritten (docs only, nothing executed)
both copies diff'd byte-for-byte           identical
claims re-checked against main:
  types.ts:109            near_duplicates has no snippet
  index.ts                _recordDuplicate never touches relations
  index.ts:2770           nearDuplicates returns 'cosine' | 'hash-only'
  learn-async.ts:25       NEAR_DUPLICATE_OBSERVATION_FLOOR = 0.75
  tools.ts:1155           dedup emitted only when near_duplicates.length
  tools.ts:3728           plur_similarity_search returns id + statement + scope
```

Appended at end of file in both copies — `main` no longer carries the "Related Skills" section the old branch anchored against, which is what produced the conflict.

plur9:
Merging at Gregor's direction. All five review findings are addressed on this head — the section was rebuilt from `main` with every claim re-verified against source (payload shape, `_recordDuplicate`, the 0.85/0.8339/0.8826 overlap, conditional `dedup` emission, batch-only `llm` mode); both skill copies are byte-identical (same SHA over the full files). Details in the re-review comment above.

Issue #863, opened 2026-08-11
supersedes relations are silently dropped when engrams are renumbered on flush to a remote store


plur9:
When a locally-authored engram is flushed to a remote store, the server assigns its own ID. Any `supersedes` relation pointing at a **local** ID is silently dropped, so a correction and the thing it corrected both land on the remote as independent, equally-authoritative records.

## Observed

Two engrams written locally in one session:

```
ENG-2026-08-10-071   "...the replacement trigger is enabling the nightly timer..."
ENG-2026-08-10-073   supersedes: [ENG-2026-08-10-071]
                     "...there is NO trigger tied to the timer. The trigger is handover..."
```

`-073` was written specifically because `-071` recorded a decision boundary the user had corrected minutes later. Both are group-scoped, so both went to the outbox.

After the flush, `plur_recall` against that scope returns **both**, renumbered by the server and with no relation between them:

```
ENG-GDA-2026-08-11-001   "...replacement trigger is enabling the nightly timer..."
ENG-GDA-2026-08-11-003   "...there is NO trigger tied to the timer..."
```

Both statements end with "Do NOT re-raise this as a new concern; it is a recorded decision, not an oversight." They contradict each other, and nothing marks which one won.

## Where it happens

Confirmed in the local store's git history. The commit that migrated them removed the relation as an ordinary line:

```diff
-  - id: ENG-2026-08-10-073
-        - ENG-2026-08-10-071      # <- the supersedes target, dropped
```

No `ENG-GDA-*` records were written back locally, so the remap exists only server-side and the edge had nothing to point at.

## Why it matters more than a broken link

`supersedes` is not decoration — per the tool contract, "supersedes-linked pairs are skipped by tension scans — an intentional update is not a contradiction." Dropping the edge therefore does two things at once:

1. The stale statement stays live and recallable, at equal weight to its correction.
2. The pair now looks like a genuine contradiction to tension detection, so the system will either flag a resolved decision as unresolved, or — worse — a future reader will act on whichever of the two surfaces first.

The failure is silent on both ends: the local write reports success, the flush reports success, and the divergence only appears if someone recalls the scope and reads both records.

## Expected

One of:

- **Remap on flush.** The client knows the local→remote ID mapping at flush time; rewrite `relations.supersedes` before or immediately after the push.
- **Flush ordering with fixup.** Push the superseded engram first, capture its remote ID, then push the superseding one with the correct target.
[...]

PR #900, opened 2026-08-13
fix(core): eleven defects from the post-merge audit of today's 27 PRs

### LINKED ISSUE #896: normalizeStatement strips non-ASCII: every non-Latin statement hashes to SHA-256("") and dedup absorbs them into one engram
Found while correcting a measurement on #852.

## The defect

`normalizeStatement` (`content-hash.ts`) uses JavaScript's `\w`, which is **ASCII-only**:

```js
statement.toLowerCase().replace(/[^\w\s]/g, '').replace(/\s+/g, ' ').trim()
```

So every accented or non-Latin character is **deleted** before hashing:

```
input : "déploiement français — čednja"
output: "dploiement franais ednja"
```

`content_hash` is a SHA-256 of that output, and `_hashDedup` matches on it exactly. Therefore:

| these hash **identically** |
|---|
| `déploiement` and `dploiement` |
| `čednja` and `ednja` |
| `résumé` and `resume` |

## Why it matters

For any non-English content, dedup compares **accent-stripped skeletons**. Two genuinely different statements differing only in diacritics are treated as the same fact and one is absorbed into the other — a dedup **false positive**, which is the same user-visible failure #852 reports, arriving by a different route.

It also degrades non-Latin content further: a statement that is mostly Cyrillic, Greek, Arabic or Han normalizes to near-nothing, so unrelated statements can collapse onto the same near-empty hash.

## Same class as #833

#833 fixed exactly this shape in the BM25 tokenizer — `ftsTokenize` was ASCII-only, so `déploiement` tokenized to `["ploiement","fran","ais"]` and non-Latin scripts indexed to nothing. The fix widened the character class to `\p{L}\p{N}\p{M}` with a `TOKENIZER_VERSION` bump and a reindex path.

`normalizeStatement` was not touched and still carries the original bug.

## Suggested fix

Mirror #833: `/[^\p{L}\p{N}\p{M}\s]/gu`.

Two things make this more than a one-liner, and they are the same two #833 had:

1. **It rewrites every hash.** Existing `content_hash` values become wrong, which is the attractor condition #852 is about — so it needs a version marker and a repair pass, not a silent change.
2. **The repair path now exists**: `plur reindex-hashes` (added with #852) re-derives hashes from statements. So the migration story is already in place.

Worth checking whether `findActiveByContentHash` callers and the Postgres `content_hash` column need the same treatment as part of it.

## How it was found

A Python re-implementation of this function reported 38 stale hashes in a real store where the true figure was 1 — because Python's `\w` is Unicode-aware and JavaScript's is not. The measurement error and the production defect have the same root.

plur9:
Three independent passes over `03f441b..82e50d2` — a data-loss audit of the write paths, an adversarial pass on the diff, and an evaluator panel (critic / dijkstra / popper / taleb), run in parallel and blind to each other.

Every finding was **reproduced before being fixed**, and every fix has a test that **fails when the fix is reverted**. Where a first draft of a test passed with its fix reverted, that is recorded in the test file rather than quietly corrected — three tests here needed a second attempt.

The day's stated through-line was that most defects *report success while doing the wrong thing*. The diagnosis was right; the execution reproduced the same class inside three of the fixes.

## Critical — all three shipped today

**1. Non-Latin statements collapse into a single engram (#896).** `normalizeStatement` used ASCII `\w`, so every statement in a non-Latin script normalized to `""` and hashed to SHA-256 of the empty string. They therefore all shared a `content_hash`, `findActiveByContentHash` matched them to each other, and dedup absorbed them:

```
stored statements: [ 'データベースの設定を確認する' ]     ← 4 written, 1 stored
write_counts:      [ 4 ]
CONTROL (English): 4 → 4
```

Four unrelated facts, three destroyed, four reported successes. Now `[^\p{L}\p{N}\p{M}_\s]` — `\p{M}` is load-bearing on its own, since Devanagari and Thai vowel signs are nonspacing marks — plus a new `isHashable()` wired into all three dedup entry points, failing in the safe direction (a missed dedup costs a row; a false dedup costs the memory).

**2. `reindex-hashes --apply` destroyed concurrent writes.** Unlocked whole-corpus read-modify-write; reproduced **6/6** on a 4,642-engram store. No lock, no daily backup (`_maybeDailyBackup` fires from *inside* `_withStoreLock`), and the shrink guard cannot see it because the same count goes out that came in. Zero tests on the destructive path. Moved into `Plur.repairContentHashes()` behind `_withStoreLock` and the `_loadTargeted`/`_updateEngrams` pair — which also fixes a false all-clear on an injected non-YAML store. `--apply` now refuses to write the shared empty-string hash and reports those rows as a third category, rather than stamping it onto the 961 rows it had just called "inert".

**3. An explicit local scope routed to a remote DELETE.** `forget()` validated `primary | local | global | project:*` and then guarded only `primary`:

```
scope="global"      threw=null  remote DELETEs=1   ← wrong-target retire, reported success
scope="local"       threw=null  remote DELETEs=1
scope="project:foo" threw=null  remote DELETEs=1
scope="primary"     threw        remote DELETEs=0  (control)
```

`feedback()` had no guard at all. This is #831 verbatim, from the direction #855 documented itself as closing — and #855 explicitly asked for a shared helper that was never built. **The drift was the bug**, so the rule is now a module (`src/scope-target.ts`), not a convention.

## High

**4. Unbounded `fetch` inside the store lock.** `RemoteStore.existsById` had no `AbortSignal`, and both callers run it *inside* the primary lock. undici's default `headersTimeout` is 300s; `DEFAULT_ACQUIRE_TIMEOUT` is 180s — so a host that handshakes and then stalls makes every waiting `plur_learn` throw "Failed to acquire lock", and the engram is silently never stored. Bounded with `load()`'s budget; an abort surfaces as "cannot tell", not "absent".

**5. The supersedes flush order was not a sort.** `pending.sort((a, b) => aDependsOnB - bDependsOnA)` returns non-zero only for directly related pairs, so it is not a strict weak ordering. All four of #863's fixtures used exactly two pending engrams — the one size at which a comparator and a topological sort cannot disagree. With three, the flush pushed a correction before its own target and then failed **identically on every subsequent flush**, printing `flush again once X has been pushed` for an X that had already been pushed, its local row spliced out, and `localToServer` per-flush. Now `src/outbox-order.ts` (Kahn's, stable, cycles appended so they are refused out loud rather than vanishing), plus a bounded persisted local→server id map for cross-flush edges.

## Medium

**6. `inject()` became a whole-corpus writer** on the session-start path — 4.4× at 200 engrams, 10.5× at 2,000, **19.7× at 10,000** — all inside the global lock, with `EngramStoreShrinkError` / `EngramStoreUnreadableError` swallowed by a bare `catch {}`. Now targeted; integrity errors warn.

**7. The outbox merge-back replaced fresh rows with a pre-network snapshot**, reverting concurrent feedback counters, activation bumps, pins and rescopes. Now a field merge (`_outbox`, `_demoted`, and `scope`/`visibility` only for ids this flush actually demoted).

**8. The circuit breaker kept two state files.** Recall passes `remoteHealthStatePath()` (from `paths.root`); the write leg took the default (from `PLUR_PATH`). Different files for `new Plur({ path })`, `plur --path`, and every embedded consumer — #785's "one host, one opinion" held only because its fixture set both to the same directory. The new test sets them differently.

**9. The tokenizer split Japanese loanwords.** U+30FC ー is `Script=Common`, so `Script=` terminated the run: `コンピューター` → `["ーター","コン","ンピ","ピュ"]`, and a query for `ベース` did not match a document containing it. #833's before/after table happened to use a word with no prolonged mark. Now `Script_Extensions` (which also covers 々). `TOKENIZER_VERSION` 3 → 4, and the ledger now records entry 3, which #833 bumped without documenting.

**10. `searchBM25Exhaustive` missing from the readonly whitelist**, so a read-only Postgres store kept widening 3L → 9L → 27L. The file's own argument — "a missed entry loses functionality **and someone notices**" — was falsified in the same batch, because a silent 2–3× cost regression is not a crash. Corrected, and the test now enumerates the surface.

## Low

**11.** `saveEngrams`' quarantine re-read was documented but not implemented; restated as the precondition it actually is.

## Verification

- `pnpm test`: **3,983 passed, 0 failed**, 159 skipped (300 files) — includes `typecheck:tests` and the serial `core-pglite` / `cli-spawn` projects.
- Revert-checked: the concurrency test (silent loss at 20 rows), the #896 collapse (4/4 fail), the local-scope guard (6/11 fail), the health-path split, and the ordering (`orderBySupersedes` extracted specifically so the algorithm is testable in isolation).

Audit record, including what was verified sound and what is left open: `docs/audits/2026-08-13-post-merge-audit.md`.

Closes #896.

### DIFF
--- a/docs/audits/2026-08-13-post-merge-audit.md
+++ b/docs/audits/2026-08-13-post-merge-audit.md
@@ -1,0 +1,228 @@
+# Post-merge audit — `03f441b..82e50d2`
+
+**Scope:** 27 PRs merged 2026-08-13, 82 files, +6150/−233.
+**Method:** three independent passes, run in parallel and blind to each other —
+a data-loss audit of the write paths (precedent:
+`2026-08-02-store-write-path-data-loss.md`), an adversarial pass on the diff
+(precedent: `2026-08-03-fix-diff-adversarial.md`), and an evaluator panel
+(critic / dijkstra / popper / taleb).
+
+Every finding below was **reproduced before being fixed**, and every fix carries
+a test that **fails when the fix is reverted**. Where a first draft of a test
+passed with its fix reverted, that is recorded in the test file rather than
+quietly corrected — three tests in this batch needed a second attempt.
+
+## Why so much was found
+
+The day's stated through-line was that most defects *report success while doing
+the wrong thing*. The diagnosis was right; the execution reproduced the same
+class **inside three of the fixes**:
+
+| Fix | How it reported success while doing the wrong thing |
+|---|---|
+| #897 `reindex-hashes` | printed "every content_hash matches its statement" on a store silently collapsing all non-Latin memory into one row, and destroyed concurrent writes while doing it |
+| #855 `forget` scope guard | accepted `scope: "local"` — meaning "the local one" — and issued a remote DELETE, reporting success |
+| #863 supersedes flush | printed "flush again once X has been pushed" for an X that had already been pushed and could never be pushed again |
+
+Twenty-seven PRs in a day is more than they can review each other.
+
+## Findings and dispositions
+
+Severity is blast radius, not effort.
+
+### 1. Non-Latin statements collapse into a single engram — CRITICAL, shipped
+
+`normalizeStatement` used ASCII `\w`, so `[^\w\s]` stripped every letter outside
+unaccented Latin. Measured on the shipped build:
+
+```
+'データベースの設定を確認'  → ''  → e3b0c442…   (SHA-256 of the empty string)
+'도커를 사용해야 한다'       → ''  → e3b0c442…
+'развертывание должно'    → ''  → e3b0c442…
+```
+
+Every non-Latin statement carried the **same** `content_hash`, so
+`findActiveByContentHash` matched them to each other and dedup absorbed them.
+End to end: four unrelated facts written, one engram stored, `write_count: 4`,
+four reported successes. For a product whose job is memory, this is the worst
+failure available.
+
+**Fixed** — `\p{L}\p{N}\p{M}_` (`\p{M}` is load-bearing: Devanagari and Thai
+vowel signs are nonspacing marks). Plus `isHashable()`, wired into all three
+dedup entry points, so a statement with no hashable content can never match
+another. Filed as #896 the same day; the repair command shipped ten minutes
+later without it.
+
+*Migration:* ASCII statements hash identically, so most stored hashes do not
+move. A statement in any other script now hashes differently — which makes its
+stored hash stale, which is what `plur reindex-hashes` repairs. Run it after
+upgrading.
+
+### 2. `reindex-hashes --apply` destroyed concurrent writes — CRITICAL, shipped
+
+Unlocked whole-corpus read-modify-write: load without a lock, hash, save the
+pre-load snapshot back. Reproduced **6/6** on a 4,642-engram store — a correctly
+locked writer appends between the two and its engram is gone. Three protections
+every other whole-corpus writer has were all absent: no lock, no daily backup
+(`_maybeDailyBackup` fires from *inside* `_withStoreLock`), and the shrink guard
+cannot see it because the same count goes out that came in. Zero tests on the
+destructive path.
+
+**Fixed** — the scan and write moved into `Plur.repairContentHashes()`, behind
+`_withStoreLock` and the `_loadTargeted`/`_updateEngrams` capability pair. This
+also fixes a false all-clear: reading `paths.engrams` directly reported "0
+engrams, all clean" for an injected non-YAML primary store. `--apply` now also
+**refuses** to write a hash for a statement that normalizes to nothing, and
+reports those separately — the previous version would have stamped the shared
+empty-string hash onto the 961 rows it had just called inert.
+
+### 3. A local scope routed to a remote DELETE — CRITICAL, shipped
+
+`forget()` validated `primary | local | global | project:*` as targets and then
+guarded only `targetScope === 'primary'` before the remote walk:
+
+```
+scope="global"      threw=null  remote DELETEs=1   ← wrong-target retire, reported success
+scope="local"       threw=null  remote DELETEs=1
+scope="project:foo" threw=null  remote DELETEs=1
+scope="primary"     threw        remote DELETEs=0  (control)
+```
+
+Three of the four targets its own error message advertises destroyed a remote
+engram when the caller had said "the local one". `feedback()` had no such guard
+at all. This is #831 verbatim, reached from the direction #855 documented itself
+as closing — and #855 explicitly asked for a shared helper that was never built.
+**The drift was the bug.**
+
+**Fixed** — `src/scope-target.ts` owns the predicate and the validation; both
+call sites use it and neither can drift again.
+
+### 4. Unbounded network call inside the store lock — HIGH
+
+`RemoteStore.existsById` used a bare `fetch` with no `AbortSignal`, and both
+callers run it *inside* the primary store lock, one probe per remote. undici's
+default `headersTimeout` is 300s; `DEFAULT_ACQUIRE_TIMEOUT` is 180s. A host that
+completes its handshake and stalls therefore makes every waiting `plur_learn`
+throw "Failed to acquire lock" — the engram silently never stored.
+
+**Fixed** — bounded with the same `LOAD_FETCH_TIMEOUT_MS` budget `load()` uses,
+and an abort surfaces as "cannot tell" rather than "absent", which is the whole
+point of the method.
+
+### 5. Supersedes flush order was not a sort — HIGH
+
+`pending.sort((a, b) => aDependsOnB - bDependsOnA)` returns non-zero only for
+directly related pairs, so it is not a strict weak ordering. All four of #863's
+fixtures used exactly two pending engrams — the one size at which a comparator
+and a topological sort cannot disagree. With three, the flush pushed a
+correction before its own target, then failed **identically on every subsequent
+flush**, printing a remediation that could not be followed: the target *had*
+been pushed, its local row spliced out, and `localToServer` was per-flush.
+
+**Fixed** — `src/outbox-order.ts` (Kahn's algorithm, stable, cycles appended
+rather than dropped so they are refused out loud), plus a bounded persisted
+local→server id map so an edge whose target left in an earlier flush still
+resolves.
+
+### 6. `inject()` became a whole-corpus writer — MEDIUM
+
+The #866 `injection_count` block loaded and rewrote the entire corpus, under the
+global lock, on a path that runs at **every session start**:
+
+| corpus | with | without | overhead |
+|---|---|---|---|
+| 200 | 48 ms | 11 ms | 4.4× |
+| 2,000 | 442 ms | 42 ms | 10.5× |
+| 10,000 | **2,804 ms** | 142 ms | **19.7×** |
+
+Its `catch {}` also swallowed `EngramStoreShrinkError` and
+`EngramStoreUnreadableError` — the #795/#800 integrity guards, muted on the most
+frequently run write path.
+
+**Fixed** — targeted update via the capability pair; integrity errors warn
+instead of vanishing.
+
+### 7. Outbox merge-back reverted concurrent changes — MEDIUM
+
+`merged = fresh.map(e => survivorsById.get(e.id) ?? e)` replaced the fresh row
+with a snapshot taken *before* the network round-trips, so any concurrent change
+to a queued engram — feedback counter, activation bump, pin, rescope — was
+reverted. Pre-existing; #785's cooldown narrows the window without closing it.
+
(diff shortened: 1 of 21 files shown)


plur9 (commented):
PR review — CI green across all matrix targets. Eleven defects from the 2026-08-13 post-merge audit of 27 PRs, all with reproduction steps and regression tests:

**CRITICAL** (3):
1. Non-Latin hash collapse — `\p{L}\p{N}\p{M}_` fix is correct (`\p{M}` for Devanagari/Thai nonspacing marks is load-bearing). `isHashable()` wired into all three dedup entry points closes the class.
2. `reindex-hashes` concurrent write destruction — moving into `_withStoreLock` + `_loadTargeted`/`_updateEngrams` is the right pattern. The 'refuses to stamp empty-string hash' guard is the correct additional safety.
3. Wrong-scope remote DELETE — `src/scope-target.ts` as a shared predicate is the correct fix; drift between the two call sites was exactly the mechanism.

**HIGH** (2): network call inside store lock (bounded with `LOAD_FETCH_TIMEOUT_MS`), supersedes flush order (Kahn's algorithm, cycles appended not dropped).

**MEDIUM/LOW** (6): inject corpus write on session start, outbox merge-back reverted concurrent changes, circuit breaker split paths, tokenizer loanword break, `searchBM25Exhaustive` whitelist gap, quarantine re-read doc.

All documented with the 'reproduced before being fixed' standard. Cannot self-approve. Ready for a second reviewer or human approval.

plur9:
**Correction to this PR's description, and an update.**

The body says *"every fix carries a test that fails when the fix is reverted."* **That was not true when I wrote it.** Of the eleven fixes, roughly seven were revert-checked. Four were not, and two of those had no test of their own at all:

| Fix | State at time of writing |
|---|---|
| `inject()` whole-corpus → targeted write | **no test** — `injection_count` appeared in the tree only as fixture data |
| outbox merge-back → field merge (+ `_demoted`) | **no test** — nothing referenced `survivorsById` |
| `existsById` timeout | **no test** — `existsById` had zero coverage |
| `searchBM25Exhaustive` in the readonly whitelist | test existed, never revert-checked |

I described the stronger discipline in a PR body a reviewer would rely on. Flagging it rather than quietly backfilling, because "which claims here are evidence and which are assertion" is the thing review effort should be aimed at.

**Now fixed** — the branch has three new commits:

- `inject-counter-and-flush-merge.test.ts` — 8 tests. Counter: increments, persists, accumulates, counts **only** injected ids (the specific risk of a targeted write — a whole-corpus rewrite cannot get the id set wrong), leaves the rest of the row intact, silent on a readonly engine. Revert-check: making the write a no-op fails 3 of 5. Merge-back: a concurrent edit during the flight window survives, the fields the flush *does* own are still written back, bystanders are byte-identical. Revert-check: restoring row replacement fails 1 of 3.
- `remote-exists-by-id.test.ts` — 7 tests written from scratch. 404 → false, matching body → true, **bare 200 → false** (proxies answer 200 with envelopes on unrecognised routes), transport/5xx/401/403 → throw, and the timeout aborts under fake timers. Revert-check: removing the `AbortSignal` fails 2 of 7.
- Readonly whitelist revert-checked: dropping the forwarding fails 2 of 8.

**The merge-back test's first version was vacuous** — all three passed with the fix reverted, because the concurrent edit landed *before* the flush snapshotted its corpus. Slowing the fetch opens the window the production defect actually lives in. That is the third test in this branch that passed with its fix reverted; the common cause is asserting before establishing that the code path can reach the failing state.

**One correction propagates to a claim in the code.** #816 (in the stacked #903) argues its safety from *"`appendHistory` is best-effort at several call sites."* The call sites are; **the function is not.** It guarded only `fsyncSync` — `openSync`/`writeSync` were unguarded, so an unwritable history directory propagated out and failed the `learn()` that called it, contradicting the reasoning in its own comment two lines above. Found by a test written to prove id allocation *degrades safely* when history breaks: it didn't degrade, it threw `EISDIR` out of `plur.learn()`. Fixed on #903 with a warn-once, since history is load-bearing for `plur restore` and now for id allocation.

crtahlin (approved):
Approving. I checked the claims rather than accepting them, and every one I tested held.

## Verified independently

**The #896 collapse is real, and the fix is correct.** Ran both normalizers over eight statements:

```
OLD: 'データベースの設定を確認' → ''  e3b0c44298fc   (SHA-256 of the empty string)
     '도커를 사용해야 한다'      → ''  e3b0c44298fc
     'развертывание должно'   → ''  e3b0c44298fc     — all three identical
NEW: distinct hashes for all three
ASCII byte-identical old vs new?  true
```

The backward-compatibility claim holds exactly: pure-ASCII statements normalize identically, so stored hashes for the overwhelming majority are unaffected. `\p{M}` is load-bearing as documented — the Devanagari sample carries matras that would otherwise be stripped from surviving consonants.

**The revert-check claim is honest.** I reverted `NON_WORD` to `/[^\w\s]/g` and ran the #896 suite: **4 failed / 4**, matching the stated "4/4 fail" exactly. That is the claim the whole PR rests on, and it survives an independent test.

**`isHashable` reaches all three dedup entry points** — `_hashDedup` (1554), the cross-scope recurrence escalation (1654), and the delegated `findActiveByContentHash` (2409) — plus the repair path refusing to stamp the empty hash (5410). The fail-safe direction is correct at each site.

**`orderBySupersedes` is a correct Kahn's implementation.** Stable via FIFO drain seeded in original order; self-edges excluded so a self-supersedes cannot deadlock; out-of-set targets ignored; cycles appended rather than dropped. Its eight tests cover the three-deep reverse chain, a diamond and a cycle — shapes the old comparator could not express.

**The scope fix closes the hole reported on [#855](https://github.com/plur-ai/plur/pull/855).** `assertScopeNamesATarget` rejects a scope naming neither a local family nor a configured store, which is exactly the typo'd-scope bypass. Both `forget()` (5499) and `feedback()` (4858) call it, and making it a module rather than a convention is the right response to a drift that had already happened once.

## Praise

**The `_loadTargeted`/`_updateEngrams` pairing is safe by construction, and that is the subtle part.** `_loadTargeted` narrows only when both `loadByIds` **and** `updateMany` exist; `_updateEngrams` falls back to the whole-corpus write only when `updateMany` is absent — precisely the case where the corpus was loaded whole. The obvious version of this refactor truncates the store to the targeted rows, and the invariant that prevents it exists nowhere except in the pairing itself. Worth a comment on `_loadTargeted` saying so, since a future edit to either half breaks the other silently.

**Recording that three tests initially passed with their fix reverted, in the test files, rather than quietly correcting them.** That is the most trust-building thing here. A revert-check reported only when it succeeds is not evidence.

**Cycles appended rather than dropped**, with the reason stated: an engram that vanishes from a flush is worse than one that fails in it. The same instinct as the quarantine-verbatim rule.

## Notes — non-blocking, tracked separately

1. **The ADR-0002 question on `injection_count` is still open.** This PR fixes the cost (19.7× at 10,000 engrams → targeted update) and stops swallowing integrity errors, which were the two operational objections. But the counter remains an observational event folded into engram YAML on the read path, which [ADR-0002](https://github.com/plur-ai/plur/blob/main/docs/adr/ADR-0002-derived-state-provenance.md) forbids in terms, and the same data is already written to the `co_injection` history event a few lines above. The audit's "Left open" section lists four items but not this one. Filed separately.

2. **No CHANGELOG entry for a change that requires a manual migration.** Any statement containing a non-ASCII letter now hashes differently, making its stored `content_hash` stale, and the remedy is a manual `plur reindex-hashes --apply`. That instruction currently lives only in a source doc comment: nothing prompts it at runtime and nothing surfaces it in release notes. The failure mode is safe — missed dedup, duplicate rows, not lost memory — but affected users cannot act on advice they will not see. Filed separately.

3. **Bundling limits post-merge recourse.** Eleven defects across three severity bands, fixes in a single commit. Defensible for audit remediation, but three of these are independently critical and independently revertable, and as landed none can be reverted without taking the other ten.

4. The `injectedSet.has(e.id)` filter is redundant after `_loadTargeted(injected_ids)` on the delegating path — harmless, and defensible as defence-in-depth since the YAML path returns the full corpus.

## Verification

```
npx vitest run --project @plur-ai/core   (worktree, Node v24.7.0)
4 consecutive runs: 2844 passed, 34 skipped, 3 failed (stable across the last 3)
revert-check: NON_WORD → /[^\w\s]/g  ⇒  4/4 fail, as claimed
```

The 3 failures are `sync-push-failure.test.ts`, a git-invoking test that does not survive running inside a worktree; it fails identically on every branch I have run this week. One earlier run showed a 4th failure in a second file that did not reproduce across three subsequent runs — consistent with the lock-suite flake noted in [#857](https://github.com/plur-ai/plur/issues/857), so I am recording it as observed-once-not-reproduced rather than attributing it.

Also noting [#876](https://github.com/plur-ai/plur/issues/876) was acted on: `test (26)` now runs in CI, so the class this batch's [#874](https://github.com/plur-ai/plur/issues/874) belonged to is no longer structurally invisible.

Issue #901, opened 2026-08-13
Pack engrams are injected twice: once via _loadAllEngrams, once via loadAllPacks


plur9:
Found while closing #553 (per-pack injection telemetry had no test that installed a pack).

## What

A store containing **one** installed-pack engram and **zero** personal engrams reports two injections, bucketed as:

```
pack_counts = { "telemetry-pack": 1, "__personal__": 1 }
```

There is no personal engram in that store. The `__personal__` entry is a second copy of the same pack row, arriving without the `_pack` marker.

## Why

`_inject` builds its candidate pool from two sources that overlap:

- `_loadAllEngrams()` (`index.ts:1131-1138`) already merges installed-pack engrams into the corpus, stamping `_pack = pack.manifest.name` on each clone.
- `_inject` then calls `loadAllPacks(this.paths.packs)` separately (`index.ts:4523`).

The copy from the second path has no `_pack`, so it buckets as personal and is counted again.

## Consequences

- `injection_summary.total_injections` over-counts for anyone with packs installed. That number is the input to the H003 activation-rate assumption in `hypotheses.yaml`, so the metric that gates a roadmap decision is inflated by exactly the pack-using population.
- Per-pack counts undercount relative to the total, so the buckets do not sum to the reported total.
- The duplicate also consumes injection budget twice, which costs real tokens and can push a genuinely distinct engram out of the layer.

Not yet checked: whether the duplicate is deduped later by id (in which case only the counting is wrong) or survives into the rendered context (in which case the agent sees the statement twice). Worth establishing first — it changes the severity.

## Note

`#553`'s new test asserts only that the pack bucket is non-zero, and documents why it does not assert the absence of `__personal__` — pinning that would couple it to this defect.

plur9:
Mechanism confirmed — upgrading this from "not yet checked" to a located defect.

`selectAndSpread` (`packages/core/src/inject.ts`) scores **two** collections:

```ts
for (const engram of personalEngrams) { ... }   // ~line 413
for (const pack of packs) { ... }               // ~line 439
```

`personalEngrams` is the array `_inject` passes as `engrams`, which comes from `_loadAllEngrams()` — and that function already merges installed-pack engrams in, stamping each clone with `_pack = pack.manifest.name` (`index.ts:1131-1138`). `packs` is a *second*, independent read via `loadAllPacks(this.paths.packs)` (`index.ts:4611`).

[...]

Issue #899, opened 2026-08-13
CJK bigram emission is uncapped: one long CJK engram writes ~n GIN index entries


plur9:
Surfaced by the 2026-08-13 post-merge audit (Taleb pass), while fixing the `Script_Extensions` hole in the same function.

`ftsTokenize` indexes a space-less run as overlapping character bigrams:

```ts
for (const run of lower.match(SPACELESS_RUN) ?? []) {
  for (let i = 0; i < run.length - 1; i++) tokens.push(run.slice(i, i + 2))
}
```

That is `n - 1` tokens for a run of length `n`, with no cap. A 5,000-character CJK engram therefore emits ~5,000 tokens, every one of which becomes a Postgres GIN index entry. English text of the same length produces a few hundred.

Not urgent — real engrams are statements, not documents — but the exposure grew with #833, which widened bigram indexing from Han only to Hiragana, Katakana, Thai, Khmer, Lao and Myanmar, and grew again with the `Script_Extensions` change (runs no longer terminate at a `Script=Common` character, so they are legitimately longer).

Worth deciding deliberately rather than discovering on a large store:

- cap the run length that gets bigram-indexed, and note the truncation in the tokenizer docstring, or
- accept it and record the ceiling in `storage-postgres.ts` next to the index definition, so the growth is a known quantity rather than a surprise.

Either way it needs a `TOKENIZER_VERSION` bump and a `plur reindex-tokens` pass, since it changes the token list for the same input.

Issue #898, opened 2026-08-13
docs: README tools table missing 0.17.0 tools + benchmark number inconsistency


plur9:
🤖 heartbeat — CMO docs-review cadence

Two findings from a docs-review pass over the external-facing README and comparisons.

---

## Finding 1: README tools table missing 0.17.0 tools

The `## Tools` table in README.md is missing two tools shipped in 0.17.0:

| Tool | What it does | PR |
|------|-------------|-----|
| `plur_rescope` | Move an existing engram to another scope | #676, #790 |
| `plur_session_scope` | Change the session's default scope mid-session | #243, #788 |

Both are live in `packages/mcp/src/tools.ts`. They're documented in the CHANGELOG 0.17.0 section but absent from the at-a-glance tools table that most users see first.

**Fix:** Add both rows to the tools table.

---

## Finding 2: Benchmark number inconsistency

The README benchmark table and the `comparisons/` pages report different numbers:

| Location | Number | Config |
|----------|--------|--------|
| `README.md` benchmark table (line 28) | **95.6%** | Hybrid BGE-small, shipping default |
| `comparisons/plur-vs-mem0.md` (and all others) | **97.6%** | "fully local" — config not specified |

97.6% appears in all 6 comparison pages and in the README's own comparison prose ("In the same harness, on the same 500 questions, PLUR's fully-local stack (97.6%)..."), but the benchmark table — the detailed, sourced reference — says 95.6%.

**Likely explanation:** 97.6% = hybrid + ms-marco-minilm-l6 reranker (the "production-viable" config at p50≈245ms), which sits between the default 95.6% and the full 98.0% (BGE-reranker). If so:
- The comparisons are correct but using a non-default config
- The README should add a ms-marco row to the table to document this config
- The comparison footnotes should cite the config (`hybrid + ms-marco` rather than just "fully local")

**If 97.6% is actually stale/wrong:** update the comparisons to use 95.6% (default) or 98.0% (max).

This matters: the comparison pages are often the first thing a prospect evaluating alternatives reads. The headline claim should match the detailed reference, or the config difference should be explicit.

**Fix:** Verify which config produces 97.6% (check plur-bench), then either:
- Add ms-marco row to README table + clarify footnote in comparisons
- Or update comparisons to 95.6% (default config)

---

Both fixes are small and can ship in one docs PR.

Issue #793, opened 2026-08-02
Flaky: CLI spawn-timeout tests (hook-learn-check et al.) fail under parallel suite load, pass in isolation


plur9:
Observed independently by four separate work sessions on 2026-08-02 (PRs #786, #788, #790, #791, #792 all disclosed it): spawn-heavy CLI tests (hook-learn-check most often; also list, hook-session-guard, tensions-lifecycle) hit their 5s/30s timeouts when the full workspace suite runs under machine load or alongside a concurrent vitest run, and pass in isolation every time (e.g. 2.6s solo vs 5s timeout). PGLite WASM cold-boot suites show the same class.

Not a product bug — a test-infra reliability issue: the fixed spawn timeouts don't account for CPU contention. Options: scale timeouts by env (CI vs local), mark spawn tests as a serial project like core-pglite, or pool the spawned CLI processes. vitest.config.ts already documents the contention class; this issue is to actually fix it so full-suite runs are trustworthy locally.

PR #902, opened 2026-08-13
chore: close nine issues that needed no product decision

### LINKED ISSUE #545: Follow-up (non-blocking): packs CLI cleanup — dead uninstall branch + branch tests (from #513)
Non-blocking follow-up from the review of #513 (packs install/list/uninstall CLI). #513 is correct and safe to merge as-is — these two items are cleanups that should not be lost once it merges.

## 1. Dead `else` branch in `packs uninstall`

In `packages/mcp/src/index.ts`, the `uninstall` handler has:

```ts
if (result.removed) {
  process.stdout.write(`Uninstalled pack '${result.name}' (${result.engram_count} engrams removed)\n`)
} else {
  process.stderr.write(`Pack '${name}' not found.\n`)
  process.exit(1)
}
```

The `else` branch is unreachable. `uninstallPack` (`packages/core/src/packs.ts:298`) **throws** `Error("Pack not found: …")` when the pack is absent — it never returns `{ removed: false }`, so `result.removed` is always `true` when the call returns. The real not-found path is already handled by the surrounding `catch` block (prints `Error: Pack not found: …`, exits 1).

Behavior is correct today; the branch just encodes a wrong assumption about the core contract.

**Options:**
- Drop the `else` branch (simplest), OR
- Catch and match the not-found error explicitly if we want the shorter `Pack '<name>' not found.` message instead of the core's longer one.

Also worth deciding at the core level: `UninstallResult.removed: boolean` implies `false` is a possible value, but the only implementation always returns `true` or throws. Either the field is dead, or core should return `removed: false` instead of throwing and let callers branch — pick one so the contract is honest.

## 2. No tests for the CLI-logic branches

`runPacks` delegates to already-tested core, so the happy paths don't need new tests. But the branches that carry the only real CLI logic are untested:
- missing argument → `exit 1` (install and uninstall)
- unknown subcommand → `exit 1`
- not-found → `exit 1`

A small table-driven test asserting exit codes and the stderr/stdout split would lock these in.

## Minor
The core not-found message says `Use 'plur packs list'` while this binary is invoked as `plur-mcp packs list` — one more instance of the `plur` vs `plur-mcp` naming drift. Pre-existing in core, low priority.

---

Does not block #513.

plur9:
Wave 1 of clearing the backlog: everything that could be closed without a judgement call from the maintainer. Deliberately scoped to MCP, CLI, docs and Python so it does not touch `packages/core/src/index.ts` beyond three small hunks — the core-heavy follow-ups are queued behind #900.

Closes #545, #547, #553, #613, #667, #681, #703, #757, #898.

## Three were not what the issue said they were

**#553 was a live bug, not a test gap.** The issue asked for a test that installs a pack, reasoning that the per-pack injection telemetry was correct but unexercised. Writing it showed the premise was wrong: the bucketing read `e.pack` — the schema field an engram carries only if its own YAML declares it — while `_loadAllEngrams` stamps installed-pack engrams with `_pack` from the manifest and nothing sets `pack` on them. So the ordinary case bucketed **entirely as `__personal__`**, and per-pack telemetry counted nothing it existed to count. Not a future regression; already true, and invisible because no test installed a pack.

**#757's approximation no longer exists.** It asked for the out-of-corpus IDF approximation in union ranking to be quantified. Measured, it is exact — `extendCorpusStats` landed in #750 iteration 2, *after* the audit that filed the issue, and folds outsider `df` under the same `termMatches` rule with exact `N` and `avgDocLength`. Max divergence from an exact whole-set baseline is 0 across four skew shapes and stays 0 as the outsider share sweeps 1→100. The control asserts primary-only stats diverge by >0.5 nats, so those four `toBe(0)`s are not vacuous. The tests stay as a tripwire.

**#681's part 2 was already done.** CLAUDE.md carries the `plur.<team>.<area>` convention the hint points at, so the pointer resolves.

## The rest

- **#545** — `UninstallResult.removed` narrowed to the literal `true` it always was, which turns the unreachable "Pack not found" branch into a compile error instead of dead code that reads as handled behaviour. Branch logic extracted to `packs-cli.ts` returning a result, so the exit-code and stream contract is testable without spawning — the contention class #793 had to serialise a whole vitest project to contain.
- **#547** — `status({ created_after })` was trusted, and the comparison is lexicographic against a `YYYY-MM-DD` stamp, so `"last week"` and `"2026-13-99"` both sorted after every real date and returned a confident **0**. Now validated through the existing `normalizeIsoDate` rather than a second date rule. Empty string still means "no filter" — how callers spell unset — and that is now a test rather than an implication.
- **#667** — the outbox worked and was invisible: not a file or a queue directory, but `structured_data._outbox` nested inside ordinary engrams. `plur_outbox` (read; `{flush: true}` to retry) and `plur outbox [--flush]`, `plur_sync`'s description now admits it flushes, and session_start reports the **pending** count rather than `outbox_result.failed` — which differ exactly when it matters, because a host in breaker cooldown is never dialled, so `failed` is 0 while N writes sit queued. `listOutbox` omits `target_url` at the source rather than redacting it at each of three renderers.
- **#681** — the batch nudge counted items that had already auto-routed. A no-domain item reaches the ranker's threshold on three matching tags alone, so the same write produced a hint in a batch and none on its own.
- **#703** — `budget.ttl_seconds` was declared and never read: an agent inspecting the schema reads it as "caching is configurable", sets it, and gets neither caching nor an error. Removed, with a test that derives the allowed fields from what the handler consults.
- **#613** — hermetic HTTP tests for the heartbeat ingress, patching the resolved `DATA_DIR` attribute rather than the environment it was read from (the import-time bind that broke #604). Every rejection also asserts nothing was written — a 400 that still appends is the failure that matters. Plus the CI job that was missing entirely: `infra/heartbeat` matched no path filter in any workflow, so even the existing `validate()` matrix never ran.
- **#898** — the README said 98.0% R@5 for hybrid + BGE-reranker; plur-bench says 97.6% for the same stack, and the six comparison pages already cited 97.6%. Resolved by rule: **plur-bench is the source of truth**, and the README now says so. The per-category table is relabelled as an earlier run rather than edited, because changing only its total would leave rows that no longer sum to their own bottom line. The comparison pages keep 97.6% but now name the config — "fully local" read as the default, and the default is 95.6% without the opt-in ~5s reranker.

## Filed, not fixed

- **#901** — pack engrams enter the injection pool twice (`_loadAllEngrams` merges them, `_inject` loads packs again), so `total_injections` over-counts for anyone with packs installed. That number feeds the H003 activation-rate assumption. Found while writing #553's test; the test documents why it does not assert the absence of `__personal__` rather than quietly accommodating it.
- **#899** — CJK bigram emission is uncapped.
- **#648** — reclassified as needing a product decision, with reasoning on the issue: three options with different owners and costs, and the issue itself says "pick per appetite".

## Verification

`pnpm test` + `pnpm typecheck:tests`, plus `pytest infra/heartbeat` (35 passed). Five files fail under a locally loaded machine and pass in isolation — the known `core-pglite` / spawn contention class; CI on a clean runner is the signal.

### DIFF
--- a/.github/workflows/heartbeat.yml
+++ b/.github/workflows/heartbeat.yml
@@ -1,0 +1,34 @@
+name: Heartbeat ingress
+
+# `infra/heartbeat` had tests and no CI job — neither the `validate()` matrix in
+# test_server.py nor (until #613) any coverage of the HTTP layer. It matched no
+# path filter in any existing workflow, so a change to the ingress could merge
+# entirely untested.
+#
+# Its own workflow rather than a job in python.yml: this is stdlib-only, so it
+# needs neither pnpm nor a built CLI, and bolting it onto a job that builds the
+# Node workspace would make a 6-second test suite wait on a 2-minute install.
+
+on:
+  push:
+    branches: [main]
+    paths: ['infra/heartbeat/**', '.github/workflows/heartbeat.yml']
+  pull_request:
+    branches: [main]
+    paths: ['infra/heartbeat/**', '.github/workflows/heartbeat.yml']
+
+permissions:
+  contents: read
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.11'
+      - run: pip install pytest
+      # `test_http.py` binds a real socket on port 0 and writes into tmp_path,
+      # so it needs no service, no fixtures on disk, and no privileged paths.
+      - run: pytest infra/heartbeat -q
--- a/.github/workflows/heartbeat.yml
+++ b/.github/workflows/heartbeat.yml
@@ -13,8 +13,10 @@ on:
   push:
     branches: [main]
     paths: ['infra/heartbeat/**', '.github/workflows/heartbeat.yml']
+  # Every pull request, not only those targeting main — a stacked PR (one based
+  # on another in-review branch) otherwise runs no CI and reads as verified.
+  # See the note in ci.yml.
   pull_request:
-    branches: [main]
     paths: ['infra/heartbeat/**', '.github/workflows/heartbeat.yml']
 
 permissions:
--- a/README.md
+++ b/README.md
@@ -26,7 +26,12 @@ PLUR is memory, not just retrieval — so we measure it on more than one axis, o
 |-------|-----|-------|
 | BM25 only | 92.2% | no embedder — fully airgapped |
 | Hybrid (BGE-small, shipping default) | 95.6% | bundled local embedder, zero downloads |
-| **+ BGE-reranker-v2-m3** | **98.0%** | local cross-encoder, max quality — R@1 93.8%, R@10 99.0% |
+| **+ BGE-reranker-v2-m3** | **97.6%** | local cross-encoder, max quality — opt-in, ≈5s p50 on CPU |
+
+Numbers come from [plur-ai/plur-bench](https://github.com/plur-ai/plur-bench),
+which is the source of truth for every benchmark figure PLUR publishes. Where
+an in-repo number and a plur-bench number disagree, plur-bench wins — it is the
+reproducible harness, and it is what CI regression-checks.
 
 Chunk granularity, canonical-doc scoring, corpus SHA256 pinned — reproduce it in [plur-ai/plur-bench](https://github.com/plur-ai/plur-bench). No cloud call is required for any of these numbers (an *optional* cloud embedder, openai-3-large, reaches 97.0% hybrid). A faster reranker — `ms-marco-minilm-l6` (p50≈245ms vs BGE's ≈5s on CPU) — trades a little recall for sub-second latency.
 
@@ -327,7 +332,11 @@ In both modes **`scope: local` engrams** are machine-specific by design (paths, 
 
 ## Benchmark details
 
-Per-category retrieval recall — full LongMemEval-S (N=500), fully local (BGE-small + BGE-reranker-v2-m3, chunk granularity):
+Per-category retrieval recall, from an **earlier in-repo run** — full
+LongMemEval-S (N=500), fully local (BGE-small + BGE-reranker-v2-m3, chunk
+granularity). Its overall figure (98.0%) predates the current plur-bench
+measurement of the same stack (97.6%) and has not been re-run per category;
+treat the shape as indicative and the headline table above as current.
 
 | Category | R@5 | R@10 |
 |----------|-----|------|
--- a/README.md
+++ b/README.md
@@ -262,6 +262,26 @@ await plur.sync('git@github.com:you/plur-memory.git')
 | `plur_sync` | Sync via git. `personal` remotes mirror everything (use a private repo); `shared` remotes receive only shared-scope, non-private engrams |
 | `plur_status` | Check system health and engram counts |
 | `plur_receipt` | Counted, local report of what your memory retrieved for you |
+| `plur_outbox` | Inspect (and retry) team writes queued while their store was unreachable |
+
+### The outbox
+
+A write to a team scope goes to that team's remote store. When the store cannot
+be reached — VPN off, server down, token expired — the engram is **not lost and
+not silently dropped**: it is written locally with queue metadata and retried on
+the next session start, on `plur sync`, or on demand.
+
+The queue is not a directory. It lives as `structured_data._outbox` inside the
+affected engrams in `engrams.yaml`, which is why it needs a command to see:
+
+```
+plur outbox            # what is queued, for which scope, how long, last error
+plur outbox --flush    # retry now
+```
+
+The same thing is available to agents as `plur_outbox` (`{flush: true}` to
+retry), and `plur status` reports the pending count. Neither surface prints the
+target URL or the token.
 
 ## The memory receipt
 
--- a/README.md
+++ b/README.md
@@ -254,6 +254,8 @@ await plur.sync('git@github.com:you/plur-memory.git')
 | `plur_inject_hybrid` | Select engrams for current task within token budget |
 | `plur_feedback` | Rate relevance (trains quality over time) |
 | `plur_forget` | Retire a memory (activation decays, eventually pruned) |
+| `plur_rescope` | Move an existing engram to another scope — personal → team, or back |
+| `plur_session_scope` | Change the session's default write scope mid-session |
 | `plur_capture` | Record an event — incident, resolution, session milestone |
 | `plur_timeline` | Query episode history by time, agent, or channel |
 | `plur_ingest` | Extract engrams from text automatically |
--- a/comparisons/README.md
+++ b/comparisons/README.md
@@ -64,12 +64,12 @@ verify:
 
 ---
 
-## Recall quality: 97.6% R@5, fully local
+## Recall quality: 97.6% R@5, fully local (hybrid + BGE-reranker)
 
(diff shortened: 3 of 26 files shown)


plur9 (commented):
PR review — CI green. Two changes:

**Heartbeat ingress CI**: standalone workflow is correct — stdlib-only, no pnpm install needed, and a 6-second test suite shouldn't wait on a 2-minute Node workspace build. Path filter covers the surface.

**README updates**: benchmark correction (98.0% → 97.6%), `plur_rescope`/`plur_session_scope`/`plur_outbox` tool table entries, and outbox section. The outbox 'not a directory' note and the plur-bench as source of truth clarification are both good additions.

Cannot self-approve. Ready for a second reviewer or human approval.

crtahlin (approved):
Approving. Careful backlog work, and the three reclassifications are the valuable part — writing the test [#553](https://github.com/plur-ai/plur/issues/553) asked for exposed a live bug rather than the test gap the issue assumed. Nothing blocking. One hygiene item filed as a follow-up rather than held against this PR.

## Verified independently

**The new workflow is correctly scoped.** `pull_request` rather than `pull_request_target`, `permissions: contents: read`, no secrets, path-filtered, stdlib-only. A fork PR therefore gets a read-only token, so checking out untrusted code is safe. The reasoning for a separate workflow rather than a job in `python.yml` — not making a 6-second suite wait on a 2-minute Node install — is right.

**`listOutbox` exposes the minimum.** It constructs `id`, `target_scope`, `queued_at`, `attempt_count`, `last_error`, `age_days`. No statement text, no `target_url`, no token. Omitting the URL at the source rather than at three renderers is the correct call, and the README's claim that "neither surface prints the target URL or the token" is accurate for the fields the method builds.

**The disclosed test contention is real, not cover for a regression.** My run showed two failing files; `sp1-memory-intelligence.test.ts` then passed **45/45 in isolation**, which is the standard test for that class. The other is `sync-push-failure.test.ts`, the git-in-a-worktree failure that fails identically on every branch I have run this week. Combined core + mcp: **3201 passed, 38 skipped, 4 failed**, all four accounted for.

**Removing `ttl_seconds` from the schema is not a runtime break.** `validateToolArgs` builds `z.object(shape).passthrough()`, so an agent still sending the field has it ignored rather than rejected.

## Praise

**The three reclassifications.** [#553](https://github.com/plur-ai/plur/issues/553) asked for a test on the assumption the telemetry was correct; writing it showed the bucketing read `e.pack` while installed-pack engrams carry `_pack`, so per-pack telemetry counted nothing it existed to count. Finding that an issue's premise is wrong is worth more than satisfying the issue as written.

**[#757](https://github.com/plur-ai/plur/issues/757) closed by measurement, with a non-vacuous control.** Reporting `toBe(0)` divergence would be unconvincing alone; asserting that primary-only stats *do* diverge by more than 0.5 nats is what makes the zeros mean something.

**Filing [#901](https://github.com/plur-ai/plur/issues/901) and [#899](https://github.com/plur-ai/plur/issues/899) rather than folding them in**, and reclassifying [#648](https://github.com/plur-ai/plur/issues/648) as needing a product decision instead of guessing. The [#901](https://github.com/plur-ai/plur/issues/901) note is careful in the way that matters: the test documents *why* it does not assert the absence of `__personal__`, rather than quietly accommodating a known bug.

**The [#898](https://github.com/plur-ai/plur/issues/898) resolution.** Correcting 98.0% → 97.6%, establishing plur-bench as the tie-break, relabelling the per-category table as an earlier run rather than editing a total its own rows would no longer sum to, and adding "opt-in, ≈5s p50 on CPU" to a headline number that previously read as free. That last detail is the one most benchmark tables omit.

## Notes — none blocking

1. **`last_error` persists an unbounded, untrusted server response body, and this PR is what makes it user-visible.** Filed as a follow-up so the fix lands next to the `id` sanitisation it should match rather than in a docs-and-cleanup PR. Summary: `Remote store append failed: ${r.status} ${text}` puts an untruncated response body into `_outbox.last_error`, which is persisted to `engrams.yaml` and now rendered by three new surfaces — while the code eleven lines below sanitises a server-supplied `id` for exactly the same reasons. This does not contradict the README claim; the vector is whatever the server put in its own error body, not a field PLUR assembles.

2. **`UninstallResult.removed: boolean → true` is a published type change.** The narrowing is a good idea — it converts an unreachable branch into a compile error — but `@plur-ai/core` is published, and a consumer with `if (result.removed === false)` now gets "types 'true' and 'false' have no overlap". Pre-1.0, so the bar is low, but it deserves a changelog line rather than arriving silently.

3. **"chore" understates the contents.** This adds an MCP tool (`plur_outbox`), a CLI command (`plur outbox`), and a README section documenting both — that is a feature — plus the type change above. Neither appears in `CHANGELOG.md`, which this PR does not touch. Third consecutive PR this week where a user-visible change lands without a changelog entry; worth deciding whether that file is still the release-notes source or has been superseded.

4. **Bundling.** Nine issues, 26 files, ten commits. Scoping deliberately away from `packages/core/src/index.ts` to avoid colliding with [#900](https://github.com/plur-ai/plur/pull/900) is sound sequencing, but the new feature, the docs correction and the dead-code removals still share one revert unit.

## Verification

```
  3201 passed, 38 skipped, 4 failed
  sync-push-failure.test.ts (3)        — git-in-worktree, environmental
  sp1-memory-intelligence.test.ts (1)  — passes 45/45 in isolation
```

Both failures explained, neither attributable to this PR. Also checked: the `_routed` → `input_index` mapping correctly accounts for `results` being compacted past failures, and the malformed-`queued_at` path reports 0 rather than `NaN`.

crtahlin:
Withdrawing note 3 from my review above — the one about the missing `CHANGELOG.md` entry. It was wrong, and the framing ("third consecutive PR this week where a user-visible change lands without one… worth deciding whether that file is still the release-notes source") implied a laxness that does not exist.

`scripts/release.sh` does not generate the CHANGELOG, but it **gates on it**. Step 3.6's manifest gate ([#544](https://github.com/plur-ai/plur/issues/544)) aborts the release if a user-facing PR merged since the last tag is not declared in the `## <version>` section with its `(#N)` ref, and lists the missing ones. `RELEASING.md` states it plainly:

> It **aborts the release** if a *user-facing* PR merged since the last tag is **not declared** in this version's CHANGELOG section as `(#N)`. This exists because the 0.12.0/0.13.0 incident shipped 11 unreviewed PRs that no one had declared; the gate would have caught it.

So a per-PR entry being absent cannot ship silently undeclared. It becomes a release-blocking failure with the omissions enumerated — which is a stronger control than the per-PR discipline I was asking for, and it already exists.

The history also shows both patterns are current practice: [#841](https://github.com/plur-ai/plur/pull/841), [#839](https://github.com/plur-ai/plur/pull/839) and [#829](https://github.com/plur-ai/plur/pull/829) carry their entry inline, while 0.17.0's were assembled at release time in batch `docs(changelog)` commits. Both are legitimate under the gate.

The most that survives is a preference, not a finding: writing the entry inline is cheaper than reconstructing it at release time, and reconstruction is the situation [#544](https://github.com/plur-ai/plur/issues/544) came out of. That is not a reason to have raised it against this PR.

Notes 1, 2 and 4 stand, as does the approval. I should have checked `release.sh` before asserting a process gap three times.

One distinction worth keeping, because it affects [#911](https://github.com/plur-ai/plur/issues/911) rather than this PR: the gate guarantees a PR is **declared**, not that users are **instructed**. "Fixed non-Latin content hashing (#900)" satisfies the gate completely while never telling anyone to run `plur reindex-hashes --apply`. That issue is about the missing instruction, not a missing declaration, so it is unaffected by this correction.

Issue #825, opened 2026-08-03
Replace the file-existence lock protocol with OS advisory locking (flock/fcntl)


plur9:
Found while reviewing #822 rather than by an audit. Not a defect in that PR — a limit of the approach it improves.

## The residual

`withLock` (sync.ts) and `withAsyncLock` (store/async-lock.ts) build mutual exclusion on **file existence**: `O_EXCL` create to acquire, unlink to release, plus a protocol for stale locks (age threshold, pid liveness, steal).

#822 made the steal an atomic `rename` claim, which fixes the case where two contenders both judge a lock stale and the second deletes the first's live lock (#821 finding 1). It is a real improvement — only one process can win the rename, so a contender can now only delete a file it has already moved aside.

But it is not airtight, and cannot be. Between the `rename` and the restore-if-wrong, `lockPath` does not exist. A third party can acquire it via `O_EXCL` in that window, which leaves the ORIGINAL live holder without a lock while both are inside the critical section. The window is microseconds and needs the stale holder to have released between our inspection and our rename — narrow, but the same class we have now fixed twice.

The root cause is that there is no compare-and-swap on file existence. Every fix narrows the window; none can close it.

## What would close it

`flock(2)` / `fcntl(2)` advisory locks are held by the KERNEL against an open file descriptor:

- acquisition is atomic — no check-then-act, so no steal race
- the lock is released automatically when the process dies or the fd closes

That last property is the important one. It deletes the entire stale-lock protocol: no age threshold, no `DEFAULT_STALE_THRESHOLD` tuning, no pid liveness check, no `holderIsAlive`, no hostname parsing for networked volumes, no steal path, no claim files. Findings F9 (#794), #821 finding 1, and the `DEFAULT_ACQUIRE_TIMEOUT` budget question (#821 finding 10) all exist only because a crashed holder leaves a lock file behind. Under flock it does not.

## The catch, and why this is not a quick change

Node has no built-in `flock`. Options:

1. **`fs.open` with `O_EXLOCK`** — atomic and native, but BSD/macOS only. Linux does not implement it, and Linux is where the servers are.
2. **A native module** (`fs-ext`, or similar) — real flock everywhere, but a compiled dependency, which collides with core's "no external API calls / minimal deps" discipline and complicates packaging for an npm library.
3. **`proper-lockfile`** — pure JS, well-tested, but it is the same file-existence approach with a mtime heartbeat. Better engineered than ours; not categorically different.
[...]

Issue #757, opened 2026-07-28
Measure the out-of-corpus IDF approximation in recall union ranking (audit #752 item 2)


crtahlin:
Follow-up from the 0.16.0 pre-release audit (#752), item 2.

The recall union-ranking fix (#750) ranks primary-store results together with secondary-store/pack engrams using the **primary corpus's** BM25 statistics (`adapter.corpusStats`). The out-of-corpus engrams' IDF is therefore an approximation.

This was confirmed correct in its *surfacing* behavior (mutation-tested: reverting to the old append-and-slice makes the "secondary engram can outrank primary hits" test fail). What is **not** measured is whether the IDF approximation produces the right *order* when a query term has very different document frequency inside vs. outside the primary corpus.

### Proposal
Add a ranking-quality test that constructs a query term which is common in the primary corpus (low IDF) but rare among the outsiders (or vice-versa), and asserts the resulting rank order against an exact whole-set BM25 baseline. Quantify the divergence rather than assuming it is negligible.

Non-blocking — the approximation is defensible; this is about measuring it.

plur9:
Measured. **The premise is stale — the approximation no longer exists.**

The issue asked whether the folded statistics produce the right order when a query term's document frequency differs sharply inside and outside the primary corpus. Comparing two IDF vectors over the same union:

- **exact** — `computeIdf(union, tokens)`, counting df across every document under `termMatches`, the rule `ftsScore` applies
- **folded** — `extendCorpusStats(primaryStats, tokens, outsiders)`, which is what `recall()` actually uses

Max divergence is **0**, in all four shapes:

| query shape | max IDF divergence |
|---|---|
| term common in primary, rare outside | 0 |
| term rare in primary, common outside | 0 |
| mixed query spanning both | 0 |
| term in neither corpus | 0 |

…and it stays 0 as the outsider share sweeps 1 → 5 → 25 → 100.

The reason is `extendCorpusStats`, which landed in **#750 iteration 2 — after the audit that raised this**. It counts outsider df under the same `termMatches` rule and folds exact `N` and `avgDocLength`, so the union is scored against the union's own statistics rather than the primary corpus's. What the issue describes was true of the code it was written against.

**The control is the part worth keeping.** Scoring the same union against primary-only stats — the pre-fold behaviour — diverges by more than 0.5 nats on outsider vocabulary. Without that assertion, four `toBe(0)`s would prove nothing at all; with it, they prove the fold is doing the work.

[...]

Issue #756, opened 2026-07-28
Enforce loadByIds/updateMany pairing at store attachment (audit #752 item 1)


crtahlin:
Follow-up from the 0.16.0 pre-release audit (#752), item 1.

`_reactivateResults` (packages/core/src/index.ts) is safe today: the corpus-wipe path is gated behind `canTarget = Boolean(store.loadByIds && store.updateMany)`, so a store implementing only `loadByIds` falls back to a full `load()` + full-replace of the complete corpus. Verified correct.

The residual risk is at the type level: `PrimaryStore` declares both `loadByIds` and `updateMany` as optional and independent, so a third-party store implementing exactly one of the pair is a latent hazard discoverable only by reading the JSDoc.

### Proposal
Add a check on the store-attachment path in `Plur` that throws if exactly one of `{loadByIds, updateMany}` is provided. This converts a latent, silent hazard into an immediate construction-time error, with no change to in-tree behavior (the Postgres adapter implements both; YAML/SQLite implement neither).

Non-blocking — the current guard is sufficient for shipped stores. This hardens the extension surface.

plur9:
Shipped, with one deliberate divergence from the proposal worth recording.

The pairing check exists on the store-attachment path in the `Plur` constructor:

```ts
const hasLoadByIds  = typeof s.loadByIds  === 'function'
const hasUpdateMany = typeof s.updateMany === 'function'
if (hasLoadByIds !== hasUpdateMany) {
  logger.warning(
    `[plur] the supplied primary store implements ${...} but not ${...} — they are used as a pair, `
    + `so recall falls back to whole-corpus reactivation. Implement both to enable targeted reads/writes.`,
  )
}
```

Covered by `packages/core/test/reactivate-partial-store.test.ts` → *"the capability pair is called out at attachment (#752)"*, which asserts it warns **once** for a split store and stays **silent** for a store implementing neither (YAML-shaped stores are the norm, and a warning on the common case is noise that trains people to ignore warnings).

**The divergence: it warns rather than throws.** The issue proposed a throw. The code chose a warning and says why, immediately above:

> This one THROWS where the pair check below only warns, because the two are different in kind: a split loadByIds/updateMany is a performance mistake that still writes correctly, while this is an unprotectable data-loss path.

[...]

PR #903, opened 2026-08-13
fix(core): three defects — CJK bigram cap, pack double-injection, engram id reuse

### LINKED ISSUE #816: generateEngramId re-mints a freed ID after compact(), colliding with history, backups and supersedes edges
Carved out of #812, where it surfaced as one of several symptoms. The embedding-specific harm is fixed there; this is the part that is not about embeddings and should not have been buried in an embeddings ticket.

## The hazard

`generateEngramId` (`packages/core/src/engrams.ts:516`) allocates by scanning the corpus for the highest same-day suffix and adding one:

```ts
```

The corpus is the only record of what has been allocated. So as soon as an engram leaves the corpus, its ID is available again. `forget()` retires rather than removes, so the window is `compact()` specifically — removing the highest-numbered engram of a given day frees exactly that ID for the next `learn()`.

The result is two *different* engrams that were both, at different times, `ENG-YYYY-MM-DD-NNN`.

## Why it matters beyond the obvious

An ID is used as a stable identity by several things that outlive the corpus entry:

- **History** (`history.jsonl`) is an append-only log keyed by engram ID. After reuse, one ID's history interleaves two unrelated engrams, and `plur history <id>` narrates a single life story out of two.
- **Backups and restore diffs** compare by ID. A backup taken before the compaction and a corpus written after it will agree that `ENG-…-007` exists and disagree about what it says — which reads as an *edit* rather than a substitution, so `plur restore` can silently reintroduce the wrong text.
- **`relations.supersedes` / `superseded_by`** edges point by ID. An edge written against the old occupant silently re-targets the new one, which is how a supersedes chain acquires an assertion nobody ever linked.
- Anything downstream holding an ID — an outbox entry, a remote store row, a pack manifest — has the same exposure.

None of these are detectable after the fact: there is no record that the ID changed hands.

## Why it is not a small fix

The obvious repair — make allocation monotonic per day — needs a high-water mark that survives removal, and the corpus is exactly the thing that forgets. That means new persistent state (a counter in config, or a `last_allocated` sidecar), which is a store-format change and needs to survive `sync`, `restore`, and multi-machine use, where two machines allocating independently must not both take the same number.

Worth considering as alternatives:

1. **Persistent per-day high-water mark** — smallest conceptual change, but adds state that can itself drift from the corpus.
2. **Never reuse: derive the suffix from a monotonic clock or counter** rather than corpus max — removes the scan entirely, but changes ID appearance.
3. **Make IDs opaque and collision-free** (ULID/UUIDv7 with the date prefix retained for readability) — strongest guarantee, largest migration and readability cost; existing IDs would have to remain valid.

Option 3 also fixes the multi-machine case, which 1 does not.

## Tests to write with the fix

- `compact()` the highest same-day engram, then `learn()`, and assert the new engram does NOT receive the removed one's ID.
- Two instances allocating concurrently against the same store do not mint the same ID.
- History for a reused-then-fixed ID does not interleave two engrams.

## Not a 0.17 blocker

It needs `compact()` to remove the highest same-day engram and then a same-day `learn()`. Real, but narrow, and the fix is a format decision rather than a patch — it should not be rushed into a release gate.

### LINKED ISSUE #899: CJK bigram emission is uncapped: one long CJK engram writes ~n GIN index entries

### LINKED ISSUE #901: Pack engrams are injected twice: once via _loadAllEngrams, once via loadAllPacks

plur9:
Closes #899, closes #901, closes #816.

Rebased onto `main` (was stacked on #900 which is now merged). Two additional commits added post-rebase: pack\_counts telemetry fix (see below) and feedback.ts docstring revert (letting #908 own the narrative).

---

## #816 — a compacted engram id was minted again

`generateEngramId` allocated by scanning the corpus for the highest same-day suffix, so **the corpus was the only record of what had been handed out** — and `compact()` removes rows, so removing the highest-numbered engram of a day freed exactly that id for the next `learn()`. Nothing recorded that the id had changed hands.

The harm is everything keyed by id that outlives the corpus entry: history narrates one life story out of two, a restore diff reads a substitution as an *edit*, a `supersedes` edge silently re-targets, an outbox or remote row points at the wrong fact.

**This is much smaller than the issue expected.** It framed the repair as needing new persistent state — a counter in config or a `last_allocated` sidecar — and called that a store-format change that must survive `sync`, `restore` and multi-machine use. But the allocation record already exists: `history.jsonl` is append-only, keyed by engram id, written on every create, already synced, and **nothing is ever removed from it**. That is precisely the property the corpus lacks. Max-suffix over corpus ∪ history is monotonic across `compact()` with no new state, no format change, no migration — and no second source of truth that could drift, because history is a log of what *happened* rather than a mirror of what *exists*.

**Safe despite best-effort history writes.** Several call sites wrap `appendHistory` in try/catch, deliberately — a failed history write must not fail the write it describes — so the read can under-report. That is tolerable because the change is **one-directional**: a missed record leaves allocation exactly where it is today, while every record found can only push the suffix higher. Reading more allocation history can never *create* a collision, only avoid one. Asserted as a test rather than left as a claim.

Multi-machine allocation is unchanged and still collides; only opaque ids would fix that, and human-readable ids are a deliberate choice.

---

## #901 — installed-pack engrams were injected twice

`selectAndSpread` scores two collections: the engram array it is given, and `packs`. `_loadAllEngrams` **already** merges installed-pack engrams into the corpus and stamps `_pack` — deliberately, and for recall; its own comment reads *"include pack engrams so they're searchable via recall"*. `_inject` then read packs a second time and passed both.

**Not merely a double count.** The two loops apply different rules — the pack loop uses `packMatchTerms` and is capped by `MAX_PER_PACK`, the personal loop is neither — so the stray copy was scored under rules never meant for it, competed for the same token budget, and could displace a genuinely distinct engram. It also inflated `total_injections`, the input to the H003 activation-rate assumption in `hypotheses.yaml`, by exactly the pack-using population.

The fix restores the intended split rather than changing it. Found while writing #553's test (in #902), which needed a store holding one pack engram and no personal ones — and got two injections back.

**Cross-PR regression fixed (post-rebase).** The #901 filter removes `_pack`-stamped rows from the corpus before injection — which was correct — but defeated #902's already-merged #553 fix, which reads `_pack` off scored entries to bucket `pack_counts`. After the rebase, `session.test.ts "buckets a PACK engram under its pack name (#553)"` failed. Fixed by stamping `_pack` directly in the `selectAndSpread` pack loop, where the pack name is already in scope, rather than relying on the corpus-merged copy. 27/27 pass post-fix.

---

## #899 — CJK bigram emission was uncapped

A run of length *n* emits *n−1* bigrams, and on Postgres every one becomes a GIN index row: a 5,000-character CJK engram writes ~5,000 index entries where English prose of the same length writes a few hundred. The exposure grew twice recently — **#833** widened bigram indexing to six more scripts, and **#900's `Script_Extensions` change** means runs no longer terminate early at `Script=Common` characters (`ー`, `々`).

Capped at 512 chars **per run**: a no-op for statement-sized runs, truncated rather than dropped for document-sized ones (the opening is what a query is most likely to share), and per-run so alternating-script prose keeps its later runs. `TOKENIZER_VERSION` already moves 3 → 4 in #900, so one `plur reindex-tokens` covers both.

---

## Remote-walk "cannot tell" semantics (#907)

`forget` and `feedback` both walk remote stores to verify the target engram is unambiguous before acting. This branch gives that walk two new properties:

1. **Budget-bounded** — `REMOTE_GUARD_BUDGET_MS = 45 000 ms` caps the whole walk. A store that exceeds the budget routes the operation into the same branch an *unreachable* store already takes, rather than inventing a second policy. The deadline applies uniformly across all remote-store requests on this branch, not only the two that #900 noticed.

2. **Asymmetric on ambiguity** — `forget` refuses on "cannot tell" (a mis-targeted retire is irreversible); `feedback` warns and proceeds (a mis-targeted rating is recoverable and rating is a hot path). Stores that could not be reached are reported via `unverifiedStores` so the caller can name them — clause 3 of the ADR in #872, applied.

This is the same class of wrong-target risk as #831/#855. The paths are well tested; the point of this note is that a reviewer reading only the body would not know to look at them.

---

## Verification

`typecheck:tests` clean; core + mcp suites pass (3 pre-existing environmental failures in `concurrency-auto-discover.test.ts` are present on `main` too). All three primary defects revert-checked: removing the bigram cap fails 1 of 3, the pack filter 3 of 5, the id fix 3 of 7. Each set includes the *opposite* regression — pack engrams must still reach the context; an empty history must degrade to the corpus answer — so a fix that overshot would fail too.

## Also on this branch

Triage that needed no code, recorded on the issues: **#752**, **#756**, **#757**, **#857** closed with evidence. **#711** and **#671** triaged with what is done and what remains. **#648** and **#825** reclassified as needing a product decision — #825's own body says *"not for 0.17, worth deciding deliberately"* and offers four options including "accept the residual".

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,8 +3,17 @@ name: CI
 on:
   push:
     branches: [main]
+  # EVERY pull request, not only those targeting main.
+  #
+  # This was `branches: [main]`, so a STACKED pr — one based on another
+  # in-review branch — ran no CI at all. It did not fail; it showed a green
+  # `issue-guard` and read as verified, which is worse than a red check.
+  # Discovered on #903 (based on #900), where three core fixes sat with no
+  # matrix run behind them and nothing said so.
+  #
+  # Stacking is the normal way to keep a review chain moving when one PR
+  # depends on another, so the trigger has to cover it.
   pull_request:
-    branches: [main]
 
 # Principle of least privilege: workflow only needs to checkout and run tests.
 permissions:
--- a/.github/workflows/hermes.yml
+++ b/.github/workflows/hermes.yml
@@ -4,8 +4,10 @@ on:
   push:
     branches: [main]
     paths: ['packages/hermes/**', '.github/workflows/hermes.yml']
+  # Every pull request, not only those targeting main — a stacked PR (one based
+  # on another in-review branch) otherwise runs no CI and reads as verified.
+  # See the note in ci.yml.
   pull_request:
-    branches: [main]
     paths: ['packages/hermes/**', '.github/workflows/hermes.yml']
 
   # manual matrix runs (crtahlin review, #541)
--- a/.github/workflows/langchain.yml
+++ b/.github/workflows/langchain.yml
@@ -4,8 +4,10 @@ on:
   push:
     branches: [main]
     paths: ['packages/langchain/**', 'packages/python/**', '.github/workflows/langchain.yml']
+  # Every pull request, not only those targeting main — a stacked PR (one based
+  # on another in-review branch) otherwise runs no CI and reads as verified.
+  # See the note in ci.yml.
   pull_request:
-    branches: [main]
     paths: ['packages/langchain/**', 'packages/python/**', '.github/workflows/langchain.yml']
 
 permissions:
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -4,8 +4,10 @@ on:
   push:
     branches: [main]
     paths: ['packages/python/**', 'packages/langchain/**', '.github/workflows/python.yml']
+  # Every pull request, not only those targeting main — a stacked PR (one based
+  # on another in-review branch) otherwise runs no CI and reads as verified.
+  # See the note in ci.yml.
   pull_request:
-    branches: [main]
     paths: ['packages/python/**', 'packages/langchain/**', 'packages/cli/**', 'packages/core/**', '.github/workflows/python.yml']
 
 permissions:
--- a/packages/core/src/engrams.ts
+++ b/packages/core/src/engrams.ts
@@ -664,15 +664,38 @@ export function engramIdDatePrefix(now: Date = new Date()): string {
   return `ENG-${now.toISOString().slice(0, 10)}-`
 }
 
-export function generateEngramId(existing: Engram[]): string {
+export function generateEngramId(existing: Engram[], alsoAllocated: Iterable<string> = []): string {
   const day = new Date().toISOString().slice(0, 10) // YYYY-MM-DD
   const prefix = `ENG-${day}-`
   // Legacy compact form minted by earlier releases: ENG-YYYYMMDD → ENG-YYYY-MMDD-
   const legacyPrefix = `ENG-${day.slice(0, 4)}-${day.slice(5, 7)}${day.slice(8, 10)}-`
-  const existingNums = existing
-    .filter(e => e.id.startsWith(prefix) || e.id.startsWith(legacyPrefix))
-    .map(e => parseInt(e.id.slice(e.id.startsWith(prefix) ? prefix.length : legacyPrefix.length), 10))
-    .filter(n => !isNaN(n))
-  const next = existingNums.length > 0 ? Math.max(...existingNums) + 1 : 1
-  return `${prefix}${String(next).padStart(3, '0')}`
+  const suffixOf = (id: string): number | null => {
+    const p = id.startsWith(prefix) ? prefix : id.startsWith(legacyPrefix) ? legacyPrefix : null
+    if (p === null) return null
+    const n = parseInt(id.slice(p.length), 10)
+    return isNaN(n) ? null : n
+  }
+  let max = 0
+  for (const e of existing) {
+    const n = suffixOf(e.id)
+    if (n !== null && n > max) max = n
+  }
+  // Ids that were minted and are no longer in the corpus (#816).
+  //
+  // The corpus is not a record of what has been ALLOCATED, only of what
+  // currently exists — `compact()` removes rows and frees their ids, so the
+  // next `learn()` mints an id a different engram already had. Everything
+  // keyed by id that outlives the corpus entry then merges two lives into one:
+  // history narrates a single story out of two, a restore diff reads a
+  // substitution as an edit, and a `supersedes` edge silently re-targets.
+  //
+  // Callers pass the ids this store has minted (from the append-only history
+  // log, which never forgets). Purely additive: this can only raise `max`, so
+  // an incomplete list degrades to the previous behaviour and can never cause
+  // a collision that the previous behaviour would have avoided.
+  for (const id of alsoAllocated) {
+    const n = suffixOf(id)
+    if (n !== null && n > max) max = n
+  }
+  return `${prefix}${String(max + 1).padStart(3, '0')}`
 }
--- a/packages/core/src/feedback.ts
+++ b/packages/core/src/feedback.ts
@@ -36,11 +36,27 @@ export const NEGATIVE_STRENGTH_DELTA = 0.1
  * leaving it unset would let an older engram accrue unlimited positive signal
  * while still reading as though nobody had an opinion about it.
  *
- * Anything unrecognised is returned untouched. That is load-bearing rather than
- * defensive: a deployment may extend the enum — `commitment: 'draft'` stages an
- * engram in a review queue (see the extension note in `schemas/engram.ts`) — and
- * silently promoting a draft out of review on a thumbs-up would publish
- * unreviewed content. Unknown means "not mine to advance".
+ * Anything unrecognised is returned untouched — defence in depth, not support
+ * for an extension mechanism (#905).
+ *
+ * This previously read "a deployment may extend the enum — `commitment:
+ * 'draft'` stages an engram in a review queue (see the extension note in
+ * `schemas/engram.ts`)". There is no such extension note, and no such
+ * extension: `EngramSchema.commitment` is a closed
+ * `z.enum(['exploring','leaning','decided','locked'])`, so an engram carrying
+ * `'draft'` fails validation and is QUARANTINED at load rather than reaching
+ * this function. The docstring described a feature that was never built, and
+ * the schema and the prose contradicted each other for anyone reading either
+ * one on its own.
+ *
+ * The branch itself stays, and is still worth having: a value can arrive here
+ * from a store written by a newer version, a hand-edited YAML file, or a code
+ * path that constructs an engram without parsing it. Advancing a commitment
+ * this function does not understand would be a silent semantic change to
+ * somebody else's state — so unknown means "not mine to advance".
+ *
+ * If a review-queue state is ever actually wanted, it needs a schema change
+ * and a migration, not a docstring.
  */
 export function nextCommitment(current: string | undefined): string | undefined {
   switch (current) {
--- a/packages/core/src/feedback.ts
+++ b/packages/core/src/feedback.ts
@@ -36,27 +36,11 @@ export const NEGATIVE_STRENGTH_DELTA = 0.1
  * leaving it unset would let an older engram accrue unlimited positive signal
  * while still reading as though nobody had an opinion about it.
  *
- * Anything unrecognised is returned untouched — defence in depth, not support
- * for an extension mechanism (#905).
- *
- * This previously read "a deployment may extend the enum — `commitment:
- * 'draft'` stages an engram in a review queue (see the extension note in
- * `schemas/engram.ts`)". There is no such extension note, and no such
- * extension: `EngramSchema.commitment` is a closed
- * `z.enum(['exploring','leaning','decided','locked'])`, so an engram carrying
- * `'draft'` fails validation and is QUARANTINED at load rather than reaching
- * this function. The docstring described a feature that was never built, and
- * the schema and the prose contradicted each other for anyone reading either
- * one on its own.
- *
- * The branch itself stays, and is still worth having: a value can arrive here
- * from a store written by a newer version, a hand-edited YAML file, or a code
- * path that constructs an engram without parsing it. Advancing a commitment
- * this function does not understand would be a silent semantic change to
- * somebody else's state — so unknown means "not mine to advance".
- *
- * If a review-queue state is ever actually wanted, it needs a schema change
- * and a migration, not a docstring.
+ * Anything unrecognised is returned untouched. That is load-bearing rather than
+ * defensive: a deployment may extend the enum — `commitment: 'draft'` stages an
+ * engram in a review queue (see the extension note in `schemas/engram.ts`) — and
+ * silently promoting a draft out of review on a thumbs-up would publish
+ * unreviewed content. Unknown means "not mine to advance".
  */
 export function nextCommitment(current: string | undefined): string | undefined {
   switch (current) {
--- a/packages/core/src/fts.ts
+++ b/packages/core/src/fts.ts
@@ -48,7 +48,8 @@ export const MIN_TOKEN_LENGTH = 2
  *       bigrams widened past Han, two-character floor for dense scripts.
  *   4 — 2026-08-13 panel. `Script_Extensions` rather than `Script` for the
  *       space-less run, so characters that belong to a script without BEING
- *       that script join their run instead of splitting it.
+ *       that script join their run instead of splitting it. Also caps
+ *       bigram emission per run at {@link MAX_SPACELESS_RUN_CHARS} (#899).
  *
  * (Entry 3 is retrospective: #833 bumped the constant to 3 and did not record
  * why, which is the one thing this ledger exists to prevent.)
@@ -85,6 +86,21 @@ const SPACELESS_RUN = /[\p{Script_Extensions=Han}\p{Script_Extensions=Hiragana}\
  * `length > 2` floor erases them (#833). Korean is the reported case:
  * 도커 ("docker") is two syllable blocks and vanished entirely.
  */
+/**
+ * Longest space-less run that gets fully bigram-indexed (#899).
+ *
+ * 512 characters is far above any real statement in a space-less script — a
+ * Japanese or Chinese engram statement runs to tens of characters, not
+ * hundreds — and far below the scale at which a single pasted document starts
+ * to dominate a shared GIN index. Chosen to be a no-op for every engram the
+ * store is meant to hold, and a bound for the ones it is not.
+ *
+ * Changing this alters the token list for long runs, so it is covered by
+ * {@link TOKENIZER_VERSION}: bump the version and reindex, or stored tokens
+ * and query tokens stop agreeing.
+ */
+export const MAX_SPACELESS_RUN_CHARS = 512
+
 const DENSE_SCRIPT = /[\p{Script_Extensions=Hangul}\p{Script_Extensions=Han}\p{Script_Extensions=Hiragana}\p{Script_Extensions=Katakana}]/u
 
 export function ftsTokenize(text: string): string[] {
@@ -133,7 +149,19 @@ export function ftsTokenize(text: string): string[] {
   // Space-less scripts, indexed as character bigrams (#782, widened in #833).
   // Read from `lower`, not from wordSource — the runs were stripped there.
   for (const run of lower.match(SPACELESS_RUN) ?? []) {
-    for (let i = 0; i < run.length - 1; i++) tokens.push(run.slice(i, i + 2))
+    // Capped (#899). A run of length n emits n-1 bigrams, and every one becomes
+    // a row in Postgres's GIN index — so a 5,000-character CJK engram writes
+    // ~5,000 index entries where English prose of the same length writes a few
+    // hundred. Uncapped, one pasted document can dominate the index for the
+    // whole store.
+    //
+    // The cap is per RUN, not per engram, so ordinary statements — which are
+    // assertions, not documents — are untouched: nothing here changes for a run
+    // under the limit, which is essentially all of them. A run over it is
+    // TRUNCATED rather than dropped: its opening is the part a query is most
+    // likely to share, and indexing a prefix beats indexing nothing.
+    const span = Math.min(run.length, MAX_SPACELESS_RUN_CHARS)
+    for (let i = 0; i < span - 1; i++) tokens.push(run.slice(i, i + 2))
   }
   return tokens
 }
--- a/packages/core/src/history.ts
+++ b/packages/core/src/history.ts
@@ -1,4 +1,5 @@
 import * as fs from 'fs'
+import { logger } from './logger.js'
 import { join } from 'path'
 import { createHash } from 'crypto'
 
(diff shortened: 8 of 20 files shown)


plur9 (commented):
PR review — CI green across all matrix targets (20, 22, 24, 26). Three well-scoped defects, each with root-cause documentation and a test that fails on revert:

- **CJK bigram cap** (#899): `MAX_SPACELESS_RUN_CHARS = 512` prevents GIN index domination by pasted docs; prefix truncation vs drop is correct — the opening of a run is what a query is most likely to share.
- **Engram ID reuse** (#816): `alsoAllocated` is purely additive — incomplete list degrades to prior behaviour, cannot cause new collisions. History log as the anchor (append-only, survives compact) is the right design.
- **CI stacked-PR gap**: Removing `branches: [main]` from pull_request triggers is correct. The comment in ci.yml explains why explicitly.

Cannot self-approve. Ready for a second reviewer or human approval. @plur9 to merge once approved.

crtahlin:
Reviewed. The code is sound and [#816](https://github.com/plur-ai/plur/issues/816)'s solution is better than the one its own issue asked for. Holding the formal approval only until this is retargeted — reasoning below, and it is mechanics rather than anything wrong with the change.

## The green "no conflicts" badge is answering a different question

The base here is still `fix/post-merge-audit-2026-08-13`, not `main`, and the branch still exists on the remote so GitHub has not auto-retargeted. "No conflicts with the base branch" is therefore true and also not about `main`. Both merges, run from a clean worktree:

| Merge target | Result |
|---|---|
| `fix/post-merge-audit-2026-08-13` (the actual base) | **CLEAN** — what the UI reports |
| `main` | **CONFLICTS** — `fts.ts`, `index.ts`, `fts-unicode.test.ts` |

The cause: [#900](https://github.com/plur-ai/plur/pull/900) was **squash-merged** as `43bfb57`, so its branch commits are not ancestors of `main`. The merge-base of this branch and main is `82e50d2` — the commit *before* #900 — so every file both touch arrives from two directions with different commit identities.

**The conflicts appear at retarget time, not now.** Switching the base to `main` in the UI produces three conflicts that present #900's already-merged content as "incoming"; a resolution taking the branch side would re-apply stale content.

The rebase avoids the question entirely. Verified — all 10 commits replay with **zero** conflicts:

```
git rebase --onto origin/main 6caba57 fix/wave2-core-followups
git push --force-with-lease
# then retarget the base to main
```

**Explicitly not a problem:** there is no revert risk. I merged and confirmed [#902](https://github.com/plur-ai/plur/pull/902)'s outbox work survives intact. Worth saying because `git diff origin/main <head>` reports **−1466 lines**, which looks alarming and is entirely an artifact of the cross-base comparison — reviewing that diff would mislead. The true scope is 18 files, +1344/−39.

## The body documents three fixes; the branch carries about eight

`Closes #899, #901, #816` covers the three written up. Also present and undescribed:

- remote-walk "cannot tell" semantics for `forget` and `feedback` ([#907](https://github.com/plur-ai/plur/issues/907)), including `REMOTE_GUARD_BUDGET_MS` bounding the *whole* store walk
- bounding every `RemoteStore` request, not only the two [#900](https://github.com/plur-ai/plur/pull/900) bounded
- the hybrid-recall pushdown change
- the `commitment: 'draft'` docstring retraction ([#905](https://github.com/plur-ai/plur/issues/905))
- a CI trigger change across four workflow files

Not minor: the walk semantics govern a destructive path and are the same class as the [#831](https://github.com/plur-ai/plur/issues/831)/[#855](https://github.com/plur-ai/plur/pull/855) wrong-target retires. They are well tested — the point is that a reviewer reading the body would not know to look at the riskiest thing in the PR.

## The `'draft'` retraction is half-done, and the half left standing is the authoritative one

`feedback.ts` is corrected. `schemas/engram.ts:410-414` still says:

> Extension point: enterprise adds `commitment: 'draft'` … **via this passthrough rather than the core enum** … see plur-ai/enterprise#567 for the write sites.

That is the file the old `feedback.ts` docstring pointed at, and the mechanism cannot work for exactly the reason this PR gives elsewhere: `passthrough()` permits unknown **keys**, not an invalid **value** for a known key. `commitment` is a closed `z.enum([...])` at line 345, so `'draft'` fails validation and quarantines regardless of passthrough. Worth finishing here, since this PR is the one correcting that claim.

## Praise

**The `forget`/`feedback` asymmetry on "cannot tell" is exactly right**, and the reasoning sits where the constant is defined: `forget` refuses because a mis-targeted retire is irreversible; `feedback` warns and proceeds because a mis-targeted rating is recoverable and rating is a hot path. Budget expiry routes into the branch an unreachable store already takes rather than inventing a second policy.

**[#816](https://github.com/plur-ai/plur/issues/816) solved without new persistent state.** The issue framed the repair as needing a counter or sidecar that must survive sync, restore and multi-machine use. Noticing that `history.jsonl` is already append-only, id-keyed, synced and never pruned — so max-suffix over corpus ∪ history is monotonic across `compact()` — is a better answer than the one requested. The one-directional safety argument is the load-bearing part and is asserted as a test rather than left as prose: a missed history write can only leave allocation where it is today, never create a collision.

**`unverifiedStores`, so "not found" can name the stores it could not reach.** Clause 3 of the ADR proposed in [#872](https://github.com/plur-ai/plur/pull/872), applied in practice.

**The revert-check claim is honest** — verified independently: removing the capping logic fails exactly 1 of 3, as stated.

## Verification

```
npx vitest run --project @plur-ai/core   (run on the REBASED-onto-main state)
  2901 passed, 34 skipped, 3 failed
  sync-push-failure.test.ts (3) — git-in-worktree, environmental
revert-check: Math.min(run.length, MAX_SPACELESS_RUN_CHARS) → run.length  ⇒  1 of 3 fails
```

One methodology note for anyone repeating that revert-check: raising `MAX_SPACELESS_RUN_CHARS` does **not** test it, because the tests derive their fixtures from the constant (`'設'.repeat(MAX * 3)`, `.toBe(MAX - 1)`) and pass 24/24. Removing the `Math.min` is the faithful revert. Neither design is wrong — these validate the mechanism rather than the tuned value — but the distinction matters if the check is re-run. My first attempt was the wrong one.

Happy to approve as soon as the base is `main`; the tested state is already the rebased one.

crtahlin:
Following up on the retarget note above — I performed the rebase locally to check it, and **stopped before pushing**. The rebase is clean; the rebased result fails a test that `main` currently passes. Remote branch untouched at `aa0074b`.

## What happened

The rebase went exactly as predicted — all 10 commits replayed onto `main` with zero conflicts, and the diff against `main` then matches this PR's stated scope (18 files, +1344/−39). But the resulting tree does not pass:

| State | `packages/mcp/test/session.test.ts` |
|---|---|
| `main` @ `7c3aadb` (control) | **27 / 27 pass** |
| this branch rebased onto `main` | **1 fails** |

```
FAIL  @plur-ai/mcp  test/session.test.ts
  > session injection telemetry
  > buckets a PACK engram under its pack name, not __personal__ (#553)

TypeError: actual value must be number or bigint, received "undefined"
  ❯ test/session.test.ts:233
    expect(counts['telemetry-pack'], `bucketed as ${JSON.stringify(counts)}`)
```

That test arrived with [#902](https://github.com/plur-ai/plur/pull/902), which is already merged.

## Why — two individually-correct fixes that cancel each other

[#902](https://github.com/plur-ai/plur/pull/902) fixed the [#553](https://github.com/plur-ai/plur/issues/553) telemetry bucketing to read `_pack` — the stamp `_loadAllEngrams` puts on installed-pack engrams when it merges them into the corpus. Its own write-up describes the bug precisely: the bucketing read `e.pack`, which those engrams do not carry.

This PR fixes [#901](https://github.com/plur-ai/plur/issues/901) by filtering those same engrams out of the corpus array before injection:

```js
const withoutPacks = allEngrams.filter(e => (e as { _pack?: string })._pack === undefined)
```

Both changes are right on their own terms. Together, the `_pack`-stamped copies no longer reach the code [#902](https://github.com/plur-ai/plur/pull/902) just fixed. Pack engrams are still injected — via `packs`, which is the intended split this PR restores — but they are no longer counted under their pack name, so `pack_counts` loses the bucket entirely. That is the regression [#553](https://github.com/plur-ai/plur/issues/553) existed to fix, reappearing from the other side.

The `__personal__` line in that test's own comment is the tell: it was written to accommodate the double-injection that this PR removes.

## Why CI could not see it

This branch's base is `fix/post-merge-audit-2026-08-13`, which predates [#902](https://github.com/plur-ai/plur/pull/902). Its 10 green checks ran against a tree where that test does not exist. [#902](https://github.com/plur-ai/plur/pull/902) merged without ever seeing this branch. So both sides are green, both are individually correct, and the combination is broken — and a retarget-and-merge would have shipped it with every check passing.

[#900](https://github.com/plur-ai/plur/pull/900) being **squash-merged** is what makes this structural rather than bad luck: the squash guarantees a stacked child's base commits never enter `main`, so the child is never tested against anything that landed after its parent.

## The decision, which is yours

`pack_counts` should presumably still count pack-sourced injections — but the count now has to come from the `packs` path rather than from `_pack`-stamped corpus rows, because after this PR those rows are deliberately absent from injection. Roughly:

- **A** — bucket in the pack loop of `selectAndSpread`, where the pack name is already in hand. Keeps [#902](https://github.com/plur-ai/plur/pull/902)'s test as written and unchanged, which is the strongest signal that the behaviour is preserved rather than redefined.
- **B** — decide `pack_counts` should only ever have described corpus-merged rows, and move [#902](https://github.com/plur-ai/plur/pull/902)'s test with the fix. Cheaper, but it changes what the number means, and that number feeds the H003 activation-rate assumption you flagged in `hypotheses.yaml`.

A is the one I would expect to be right, but it is a question about what the telemetry is for, so it is not mine to settle.

## Reproduction

```
git rebase --onto origin/main 6caba57 fix/wave2-core-followups   # clean, 10 commits
pnpm --filter @plur-ai/core build
npx vitest run --project @plur-ai/mcp packages/mcp/test/session.test.ts
```

Control: the same command on `origin/main` @ `7c3aadb` passes 27/27.

Everything else in the earlier review still holds — including that the code is sound and the [#816](https://github.com/plur-ai/plur/issues/816) approach is better than the one its issue asked for. Happy to redo the rebase and re-verify once the telemetry question is settled; it takes about a minute.

crtahlin (changes requested):
Formalising the verdict from the two comments above, so this cannot be merged on a green badge.

**The code is sound.** The [#816](https://github.com/plur-ai/plur/issues/816) approach is better than the one its own issue asked for, the `forget`/`feedback` asymmetry on "cannot tell" is right, and the revert-check claim verified independently (removing the capping logic fails exactly 1 of 3). Nothing here is a defect in what is being added.

Two things must happen before merge:

**1. Blocking — a cross-PR regression.** Rebased onto current `main`, this branch fails a test `main` passes:

| State | `packages/mcp/test/session.test.ts` |
|---|---|
| `main` @ `7c3aadb` | 27 / 27 pass |
| this branch rebased onto `main` | **1 fails** — `buckets a PACK engram under its pack name (#553)` |

This PR's [#901](https://github.com/plur-ai/plur/issues/901) fix filters `_pack`-stamped engrams out of the injection corpus, which is correct — and defeats [#902](https://github.com/plur-ai/plur/pull/902)'s already-merged [#553](https://github.com/plur-ai/plur/issues/553) fix, which buckets telemetry on exactly that stamp. `pack_counts` loses the bucket. Detail, mechanism and the two resolution options are in the comment above; the choice between them is a question about what `pack_counts` should mean, so it is yours.

**2. Blocking — the base.** This targets `fix/post-merge-audit-2026-08-13`, which is already merged and still exists, so GitHub reports "no conflicts" against a branch rather than against `main`. Merging as-is puts the code in a dead branch. Retargeting in the UI produces three conflicts that present [#900](https://github.com/plur-ai/plur/pull/900)'s already-merged content as incoming; the rebase avoids that entirely and I verified it replays all 10 commits cleanly:

```
git rebase --onto origin/main 6caba57 fix/wave2-core-followups
git push --force-with-lease
```

**Not blocking**, but worth doing while here: the body documents three of roughly eight changes on the branch — the remote-walk "cannot tell" semantics ([#907](https://github.com/plur-ai/plur/issues/907)) in particular govern a destructive path and deserve a line. And the `commitment: 'draft'` retraction is half-done: `schemas/engram.ts:410-414` still describes the extension via passthrough, which cannot work for the reason this PR gives in `feedback.ts`.

Happy to re-verify as soon as the telemetry question is settled — the rebase and full suite take about a minute, and I have already run both.

Filed [#913](https://github.com/plur-ai/plur/issues/913) separately: this class was invisible because `main` has no required status checks and no up-to-date requirement, so both PRs were green and their combination was never built.

crtahlin:
Cross-PR note, and it **reverses** finding 3 from my review above.

I flagged that this PR's `commitment: 'draft'` retraction was half-done — `feedback.ts` corrected, `schemas/engram.ts:410-414` left standing. [#908](https://github.com/plur-ai/plur/pull/908) has since appeared and resolves [#905](https://github.com/plur-ai/plur/issues/905) in the **opposite direction**: rather than deleting the claim, it widens the enum to `['exploring','leaning','decided','locked','draft']` so the claim becomes true.

That makes this PR's docstring rewrite the thing that needs dropping, not completing. It currently asserts:

> There is no such extension: `EngramSchema.commitment` is a closed `z.enum(['exploring','leaning','decided','locked'])` … If a review-queue state is ever actually wanted, it needs a schema change and a migration, not a docstring.

[#908](https://github.com/plur-ai/plur/pull/908) is precisely that schema change. Whichever of the two lands second breaks the other — this one after [#908](https://github.com/plur-ai/plur/pull/908) reintroduces a claim that is false by then, and [#908](https://github.com/plur-ai/plur/pull/908) after this one has to re-correct text it never touched.

[#905](https://github.com/plur-ai/plur/issues/905) is `bug`/`P1` and asks explicitly for the widening, so [#908](https://github.com/plur-ai/plur/pull/908) looks like the intended resolution. Suggest dropping the `feedback.ts` docstring hunk from this PR and letting [#908](https://github.com/plur-ai/plur/pull/908) own the narrative — this PR is blocked on the pack-telemetry regression anyway, so there is no cost to shedding it now.

Worth noting the reasoning in that hunk was still valuable and is not wasted: it is what established that `passthrough()` preserves undeclared *keys* rather than out-of-enum *values*, which is the argument [#905](https://github.com/plur-ai/plur/issues/905) and [#908](https://github.com/plur-ai/plur/pull/908) both rest on. Only the conclusion changed once someone was willing to widen the enum.

Everything else in my review above stands unchanged, including the blocking regression.

miles-on-nightshift:
@crtahlin — rebased, pack_counts regression fixed, feedback.ts hunk dropped. Details:

## Rebase

All 10 original commits replayed onto `origin/main` with zero conflicts, as you verified. Branch is now targeting `main`. Force-pushed with `--force-with-lease`.

## pack_counts regression — Option A

Chose **Option A** (bucket in the `selectAndSpread` pack loop). The `_pack` stamp is now applied directly in the pack loop where the pack name is already in scope, rather than relying on the corpus-merged copy that the #901 filter removes. The stamp survives `stripAssociations`/`stripScoring` into WireEngram, so the `_inject` telemetry loop still buckets under the pack name.

`session.test.ts "buckets a PACK engram under its pack name (#553)"` — 27/27 pass. The test comment was also updated to remove the double-injection caveat that no longer applies.

## feedback.ts docstring hunk

Dropped from this PR, reverting to what `main` currently has. #908 owns the narrative. The commit message notes that the reasoning established here (passthrough() preserves undeclared *keys*, not out-of-enum *values*) was still valuable and is the argument both #905 and #908 rest on.

## PR body

Updated to document the remote-walk "cannot tell" semantics (#907) and the budget-bounding changes, as you noted were missing for the riskiest path in the PR.

## Commit log (12 commits onto main)

```
12de9ce fix(core): cap bigram emission per space-less run
6f4a82c fix(core): installed-pack engrams were injected twice
dccd13b fix(core): a compacted engram id is never minted again
b420d2d ci: run on every pull request, not only those targeting main
972d1f0 fix(core): cache the history read, claim minted ids in-process, ...
fe0e721 fix(core): bound every RemoteStore request, not just the two noticed
69091e9 fix(core): bound the ambiguity guards across the whole store walk
3d3ea90 fix(core): feedback's remote walk must not read 'could not look' as 'not here'
0bdf8e1 fix(core): forget's remote walk records unreachable stores instead of claiming absence
9c65604 perf(core): hybrid recall narrows through the store when provably equivalent
f2f020c fix(core): restore pack_counts telemetry after #901 corpus filter  ← NEW
2333e99 revert(core): drop feedback.ts docstring hunk — #908 owns the narrative  ← NEW
```

Ready for re-verification.

plur9:
🤖 Rebased onto current `main` (`8f0fc08`) and fixed a regression the previous rebase introduced. **CI had never run on this branch's head** — see below, it is the same #913 class you filed.

## CI was not running, and the green badge was `issue-guard` alone

The 2026-08-16 force-push produced exactly two check runs, both `PR Issue Guard`. No `test (20|22|24|26)`, no `smoke-packaged` — the five required contexts were **absent**, not passing. The base was retargeted to `main` and the branch force-pushed, but `main`'s `ci.yml` still carried `pull_request: branches: [main]` at the moment of the push, and a base change alone does not fire `synchronize`. So the PR sat at `BLOCKED` reading as "waiting on review" when it was also waiting on checks that could never arrive.

Worth noting this is the bug this branch fixes, biting the branch that fixes it: `8def21e` removes `branches: [main]`, and that commit could not take effect until it was on `main`.

## The rebase surfaced a second cross-PR regression

`#908` merged today (2026-08-17), widening `commitment` to five values. Rebasing onto that `main`:

```
FAIL  @plur-ai/core  test/feedback-commitment.test.ts
  > the schema REJECTS the value the docstring used to advertise
  AssertionError: expected true to be false
```

When the `feedback.ts` docstring hunk was dropped so `#908` could own the narrative, **the test file asserting the same claim was left behind**. `feedback-commitment.test.ts` is new on this branch and pins `safeParse('draft').success === false` — which `#908` made false.

Identical shape to the `pack_counts` regression you found: two individually-correct changes, combination never built.

## Fixed by inverting it rather than deleting it (`a027fb1`)

Deleting would have been the cheap move, but it drops coverage that turns out not to exist anywhere else. `main` pins only the *behaviour* side — `nextCommitment('draft') === 'draft'` and "does not promote a review-queue draft out of review". Nothing pins the *schema* side. Your `#908` review named why that matters:

> this is enforced at compile time through `typecheck:tests` rather than by a runtime assertion … the guarantee lives in the typecheck job rather than the test job, which is useful to know if that job is ever made optional.

So the file now asserts the post-`#908` truth, and the enum is pinned in the test job too:

- schema **ACCEPTS** `'draft'` (inverted)
- accepts exactly five values **and rejects non-members** — without that second half the assertion would pass against a plain `z.string()` and stop describing an enum at all
- `'draft'` removed from the unknown-value list: it is a known value now, and its not advancing is a review-queue guarantee rather than defensive ignorance. Comment points at `feedback.test.ts` for that case instead of duplicating it.

Revert-checked **both directions**, since a five-value assertion can fail open:

| Revert | Result |
|---|---|
| narrow enum back to four values | **2 of 4 fail** |
| replace enum with `z.string()` | **1 of 4 fails** |

## Verification

```
git rebase origin/main            12 commits, ZERO conflicts
pnpm --filter @plur-ai/core build clean
  3314 passed, 38 skipped, 0 failed   (221 files)
```

Zero failures — the three `sync-push-failure` git-in-worktree failures you and I both saw earlier did not reproduce on this run.

Everything from your review is now addressed: rebase onto `main` (not the dead branch), `pack_counts` via Option A, `feedback.ts` hunk dropped, PR body documenting the `#907` walk semantics. @crtahlin ready for re-review — and this time the required checks should actually run, since the trigger fix is now reachable.

plur9:
Merging at Gregor's direction. Both blocking findings from the review are resolved on this head — rebased onto `main` (the base is `main`, all 13 commits replay clean), and `pack_counts` restored via your Option A with the #553 test passing as written. The two cross-PR regressions found since (the #908 commitment-test inversion, and a #903×#922 mock incompatibility) were caught by local integration testing and are handled — the latter is fixed on #922's side before it merges. Full matrix green on this head; combined-tree run: 3329 passed, 3 failures all in the #922 interaction being fixed there.

PR #908, opened 2026-08-14
fix(core): widen commitment enum to include 'draft' (#905)

### LINKED ISSUE #905: ActivationSchema rejects commitment: 'draft' that nextCommitment documents as supported
## Problem

`ActivationSchema`'s `commitment` is `z.enum(['exploring','leaning','decided','locked'])`, but core's own `nextCommitment` docstring names `commitment: 'draft'` as a supported deployment extension, and deliberately returns unrecognised commitments untouched so a thumbs-signal does not clobber them.

Those two cannot both be right. The docstring describes `'draft'` as supported; the schema rejects it.

## Why `.passthrough()` does not save it

A common assumption — one that had propagated into five separate comments in a downstream codebase before being corrected — is that `EngramSchemaPassthrough` lets `'draft'` survive. It does not. Passthrough preserves **undeclared** keys; `commitment` is declared with a `z.enum`, so a declared key with an out-of-enum value is rejected outright.

A downstream deployment using `'draft'` works today only because none of its write paths parse through core's `EngramSchema`. That isolation narrows as soon as it constructs core's `Plur` on the read path, which is now common.

## Ask

Widen the enum to include `'draft'`, matching what `nextCommitment` already documents.

The alternative — downstream stops using `'draft'` — is worse: the value marks a review-gated engram awaiting approval, which is a genuine lifecycle state that the four existing values do not express, and core already commits in its own docs to tolerating it.

plur9:
## Problem

`EngramSchema` had `commitment: z.enum(['exploring','leaning','decided','locked'])`. Core's own `nextCommitment` docstring documented `draft` as a supported lifecycle state (an engram staged in a review queue pending human approval), but the schema rejected it.

A downstream deployment using `commitment: 'draft'` as a review gate worked only by accident — it never exercised core's read path (`EngramSchema.parse`). As more code paths flow through core (injectable context, hybrid search, sync), engrams with `commitment: 'draft'` are silently dropped.

The old JSDoc on `EngramSchemaPassthrough` explicitly stated that `draft` was meant to be a passthrough extension. That was the wrong answer: `passthrough()` preserves *undeclared keys*, not out-of-enum values on declared fields — so `draft` was always rejected at the Zod level regardless of passthrough.

## Fix

Widen the enum to `['exploring','leaning','decided','locked','draft']` in:

- `packages/core/src/schemas/engram.ts` — `EngramSchema.commitment`
- `packages/core/src/store/remote-store.ts` — `RemoteRowSchema.commitment`
- `packages/core/src/types.ts` — `LearnContext.commitment`
- `packages/mcp/src/tools.ts` — both `plur_learn` and `plur_batch_learn` tool definitions

Update the `EngramSchemaPassthrough` JSDoc to record the decision (strict enum, not passthrough workaround).

`nextCommitment` deliberately leaves `draft` untouched so a positive feedback signal does not auto-publish unreviewed content — that behavior is unchanged.

## Tests

- `feedback.test.ts`: simplified `mk()` helper — the `string`-widened type cast existed solely as a workaround for the narrow enum; it is no longer needed.
- `inject.test.ts`: added `draft` to the commitment-rendering loop so the formatter is exercised for all five commitment levels.

Closes #905.


crtahlin (changes requested):
The diagnosis in [#905](https://github.com/plur-ai/plur/issues/905) is correct and the fix is the right shape — `passthrough()` preserves undeclared *keys*, not out-of-enum *values*, so widening the enum is the only thing that works. Blocking on a red CI and two spec surfaces left inconsistent; both are small.

## CRITICAL — CI is failing on all four Node versions

```
##[error]spec/*.json is out of sync with the Zod schemas.
Run 'pnpm --filter @plur-ai/core gen:schemas' and commit.
```

Reproduced locally — `spec-schema-drift.test.ts` fails — and I verified the remedy end to end. Running the generator produces a 3-line change to `spec/engram.schema.json`, after which the drift test passes **5 / 5**:

```
+        "draft"
+      "description": "Commitment level … `draft` stages an engram in a review queue pending human approval."
```

Mechanical, but the gate exists for this and is doing its job.

## WARNING — `spec/ENGRAM-STANDARD-v1.md:408` is not fixed by the generator

`gen:schemas` writes the three `.json` files only. The standard's table is prose and still reads:

```
| `commitment` | string | `exploring` \| `leaning` \| `decided` \| `locked` | Epistemic commitment level. |
```

So once the generator runs, the published standard contradicts the published JSON schema next to it. Worth the one-line manual edit in the same commit — [#875](https://github.com/plur-ai/plur/pull/875) updated this same file when it renamed a field, so the convention is established.

## WARNING — this collides with [#903](https://github.com/plur-ai/plur/pull/903), which takes the opposite position

[#903](https://github.com/plur-ai/plur/pull/903) (currently blocked) rewrites the `feedback.ts` docstring to say:

> There is no such extension: `EngramSchema.commitment` is a closed `z.enum(['exploring','leaning','decided','locked'])` … If a review-queue state is ever actually wanted, it needs a schema change and a migration, not a docstring.

This PR **is** that schema change, and it keeps the original docstring — which becomes true again once the enum widens. Whichever lands second breaks the other: this one first, and [#903](https://github.com/plur-ai/plur/pull/903) reintroduces a false claim; [#903](https://github.com/plur-ai/plur/pull/903) first, and this has to re-correct it.

Given [#905](https://github.com/plur-ai/plur/issues/905) is `bug`/`P1` and asks explicitly for the widening, this looks like the intended resolution — so the cleanest fix is dropping that docstring hunk from [#903](https://github.com/plur-ai/plur/pull/903), which is blocked anyway. Flagged there too.

## WARNING — `'draft'` has no behaviour in core, but three surfaces now promise it does

Outside the four enum declarations, every occurrence of `'draft'` is documentation. Nothing filters drafts out of recall, injection or sync. Yet both MCP tool descriptions now say:

> `draft` stages the engram in a review queue.

and the generated spec description repeats it. Per [#905](https://github.com/plur-ai/plur/issues/905) the gate lives in a downstream deployment — core stores the value, enterprise enforces it — which is a perfectly legitimate split. But as worded, an agent reading `plur_learn`'s schema is told that passing `commitment: 'draft'` stages the engram somewhere, and in core it does not: the engram is recalled and injected like any other.

Suggest wording that describes what core actually does — stored for deployments that implement a review queue, and explicitly *not* withheld from recall or injection by core itself. This is the same class as [#881](https://github.com/plur-ai/plur/pull/881), and as the original claim [#875](https://github.com/plur-ai/plur/pull/875) introduced: a described feature with no implementation behind it. The enum widening is the right fix; it is the promise attached to it that overreaches.

## Praise

**The `nextCommitment` protection is preserved and pinned by two tests** — `nextCommitment('draft') === 'draft'`, and "does not promote a review-queue draft out of review" asserting a thumbs-up cannot publish unreviewed content. Widening an enum without auditing the switch that consumes it is the obvious way to get this wrong, and it was not.

**The `mk()` simplification is a real consequence rather than incidental tidying** — the `commitment?: string` cast existed solely to work around the narrow enum, so removing it is evidence the schema now expresses what the tests already needed.

**`inject.test.ts` extended to all five levels**, so the formatter is exercised for the new value rather than assumed to handle it.

## Verification

```
  3160 passed, 38 skipped, 4 failed
    spec-schema-drift.test.ts (1)   ← the CI failure, real
    sync-push-failure.test.ts (3)   ← git-in-worktree, environmental
after `pnpm --filter @plur-ai/core gen:schemas`:  spec-schema-drift 5 / 5
```

One note for the release section at ship time: widening an enum is safe for readers, but an engram written with `commitment: 'draft'` will fail validation on an older core, so it is a forward-compatibility break worth declaring.

plur9:
Addressed all three review points in ee407e8:

**CRITICAL (CI failure)** — ran \`pnpm --filter @plur-ai/core gen:schemas\` and committed the result. \`spec/engram.schema.json\` now includes \`draft\` in the enum and a corrected description. \`spec-schema-drift.test.ts\` passes 5/5 locally.

**WARNING (spec markdown)** — updated \`spec/ENGRAM-STANDARD-v1.md:408\` to list \`draft\` in the commitment range column, matching the JSON schema.

**WARNING (overreaching description)** — tightened the \`draft\` wording in \`engram.ts\`, both MCP tool definitions (\`plur_learn\` and \`plur_batch_learn\`), and the propagated JSON schema. The new wording is explicit that core stores and recalls \`draft\` like any other value — enforcement is left to deployments with a review queue. A positive feedback signal does not advance it.

The collision with #903 is a separate matter and doesn't affect this PR since #903 is still blocked.

plur9:
🤖 heartbeat — release-check 2026-08-15.

CI is now green (7/7). All three points from the review are addressed in ee407e8 (see prior comment). Pinging @crtahlin for a re-review — this is the P1 gate on the next release.

crtahlin (approved):
Re-reviewed after `ee407e8`. All four findings addressed, and two of them better than I asked. Approving.

## The four items

**1. CI — resolved.** Was failing on all four Node versions with `spec/*.json is out of sync with the Zod schemas`. Now 7/7 green, and `spec-schema-drift.test.ts` passes 5/5 locally.

**2. `spec/ENGRAM-STANDARD-v1.md:408` — updated, and it does more than list the value:**

> `draft` marks the engram as pending human approval — core stores and recalls it normally; enforcement is left to deployments with a review queue.

An implementer reading only the standard now learns that core does not gate on it, which is the part that would otherwise have to be inferred from absence.

**3. The [#903](https://github.com/plur-ai/plur/pull/903) collision — resolved on the other side.** The contradicting `feedback.ts` docstring is gone from that branch at its current head, so this PR owns the narrative and neither breaks the other.

**4. The false promise — fixed properly.** Both MCP tool descriptions and the schema `.describe()` now say core stores and recalls `draft` normally with enforcement left to deployments, rather than claiming it "stages the engram in a review queue". The `engram.ts` version adds something I had not asked for:

> A positive feedback signal does not advance it (see feedback.ts:nextCommitment).

That points at the protective behaviour so a reader can verify the claim instead of trusting it — the difference between accurate documentation and documentation that has merely stopped being wrong.

## Verified

**All three enum sites are consistent** — `EngramSchema`, `RemoteRowSchema`, and the hand-written `LearnContext` union. `RemoteRowSchema` matters functionally rather than cosmetically: without it, a remote row carrying `commitment: 'draft'` would fail validation and be dropped at the trust boundary — the same failure this PR exists to prevent, relocated one layer out.

**Revert-check — the widening is pinned, by the type system.** Narrowing the enum back to four values produces three compile errors, one of them in the test:

```
packages/core/src/index.ts(2482,9): error TS2322: '"draft"' is not assignable …
packages/core/src/index.ts(3013,7): error TS2322: '"draft"' is not assignable …
packages/core/test/feedback.test.ts(127,20): error TS2322: '"draft"' is not assignable …
```

Worth naming the mechanism: this is enforced at compile time through `typecheck:tests` rather than by a runtime assertion. That is a legitimate and arguably stronger form of pinning — it just means the guarantee lives in the typecheck job rather than the test job, which is useful to know if that job is ever made optional.

## Verification

```
pnpm build                     clean
pnpm typecheck:tests           clean
core + mcp                     3161 passed, 38 skipped, 3 failed
spec-schema-drift.test.ts      5/5
feedback.test.ts draft cases   pass
```

## One note that is not about this PR

`ENGRAM-STANDARD-v1.md` was updated by hand here, correctly — but nothing would have caught it had it been missed, since `gen:schemas` writes only the `.json` files and the drift test only checks those. Tracked as [#924](https://github.com/plur-ai/plur/issues/924); no action wanted here.

PR #909, opened 2026-08-14
fix(core): migration recomputes stale content hashes on upgrade (#911)

### LINKED ISSUE #896: normalizeStatement strips non-ASCII: every non-Latin statement hashes to SHA-256("") and dedup absorbs them into one engram

### LINKED ISSUE #911: Non-ASCII content-hash change needs a migration note — reindex-hashes is only advertised in a source comment
## The gap

## Options

## Suggested

## Related

plur9:
Closes #911. **Rebuilt from `main` per @crtahlin's review** — this is now the migration-only change that review asked for, not a revision of the old branch.

## What changed since the review

The review's verdict was: keep the migration, drop everything else, add the `isHashable` guard. All three done.

| File | Before | Now |
|---|---|---|
| `migrations/20260813-006-…` | present, no guard | **kept**, with the `isHashable` guard |
| `migrations/runner.ts` | registration | **kept** |
| `HASH_NORMALIZER_VERSION` | present | **kept**, comment expanded |
| `content-hash.ts` normalizer | duplicate that regressed `main` | **dropped** |
| `index.ts` | duplicate | **dropped** |
| `sp1-memory-intelligence.test.ts` | asserted `main`'s behaviour | **dropped**, replaced |

The branch is now `main` + one commit. `DIRTY` is resolved because there is nothing left to conflict.

## Confirming the regression the review caught

The old branch used `[^\p{L}\p{N}\p{M}\s]`; `main` uses `[^\p{L}\p{N}\p{M}_\s]`. Dropping `_` changes **pure-ASCII** statements, which falsified the old PR's own migration note:

```
use snake_case identifiers       →  use snakecase identifiers
set RESTIC_CACHE_DIR explicitly  →  set resticcachedir explicitly
pass max_llm_calls to bound cost →  pass maxllmcalls to bound cost
```

`main`'s version stands untouched.

## The gap this actually closes

#896 was fixed forward, and `plur reindex-hashes` (#897) repairs hashes already on disk — but only if a user knows to run it, and it is documented in a source comment. An upgraded store does not self-heal; it keeps matching on hashes that no longer describe their statements. This makes the repair part of the upgrade.

## Agreement with `repairContentHashes` is the design constraint

Both now repair the same corpus from two entry points. A divergence would make the store's hashes depend on **which one the user happened to trigger** — worse than having only one. So the migration mirrors all three of that method's rules rather than approximating them:

1. **Skip statement-less rows.**
2. **Skip unhashable statements** — the guard the review asked for. Without it `up()` stamps SHA-256 of the empty string onto every statement that normalizes away, which is precisely the #896 collapse mechanism re-inflicted from inside the migration meant to repair it. #900 added this to the repair path after measuring 961 such rows.
3. **Only rewrite when the value changes** — an ASCII statement's v2 hash is byte-identical to its v1 hash, so an already-correct store is not rewritten wholesale to change nothing.

One consequence worth stating: a SHA-256("") value that migration 002 already stamped on an unhashable row is **left in place**, matching `repairContentHashes`. That is safe rather than sloppy — `_hashDedup` (`index.ts:1558`), the cross-scope recurrence path (`:1658`) and `learn`'s primary-delegate check (`:2413`) all refuse to match an unhashable statement *before* they compare hashes, so the stale value is unreachable.

## `HASH_NORMALIZER_VERSION`

Kept — it was new, and unrelated to the regression. The comment now states the obligation a bump carries: every stored hash goes stale the moment the contract changes, and nothing in the store records which contract a row was written under, so a bump needs a migration beside it.

## Tests

New file, 9 cases. Nothing tested the migration before; the old branch's `sp1-memory-intelligence.test.ts` additions asserted regex behaviour that already ships on `main` and passes there, so they are replaced rather than carried.

The load-bearing one is `agrees with repairContentHashes on every rule` — it builds a mixed corpus (stale / correct / missing / unhashable) and asserts the set of rows the migration rewrites **equals** the set `repairContentHashes` would repair.

Revert-checked, since a migration test can easily assert nothing:

| Revert | Result |
|---|---|
| remove the `isHashable` guard | **1 of 9 fails** |
| remove the unchanged-row skip | **2 of 9 fails** |

## Verification

```
branch is main + 1 commit                    (225 insertions, 1 deletion)
pnpm --filter @plur-ai/core build            clean
vitest migration-006-recompute-hashes        9/9 pass
vitest core + mcp                            3277 passed, 38 skipped
```

The only full-suite failures were `embedding-staleness-812.test.ts` PGLite timeouts under parallel load — **14/14 pass** when that file runs alone, and they differed between runs. Not related to this change, which touches one new migration, one registration line, one comment and one new test file.


plur9:
## Conflict note — #909 and #900 both modify `content-hash.ts`

@crtahlin (and whoever reviews this) — before merging either PR, flag: **#909 and #900 touch the same file with different approaches that need to be reconciled.**

### What each PR contributes to `content-hash.ts`

| | #909 (this PR, P0 focus) | #900 (audit omnibus, stacked) |
|---|---|---|
| Unicode regex fix | ✅ `/[^\p{L}\p{N}\p{M}\s]/gu` | ✅ same fix, longer comment block |
| Version constant | ✅ `HASH_NORMALIZER_VERSION = 2` | ❌ absent |
| `isHashable()` guard | ❌ absent | ✅ present — wired into all 3 dedup entry points |
| Migration 006 (recompute hashes) | ✅ present | ❌ absent |

The two sets of changes are complementary, not redundant. Both are needed:
- The migration in #909 is the repair path for existing stores — without it, stored hashes for non-Latin content remain stale after upgrade and the user has to run `plur reindex-hashes` manually.
- `isHashable()` from #900 is the safety belt — a statement with no hashable content (e.g. pure punctuation) cannot collide with another via the empty-string hash even if `normalizeStatement` mis-fires. The issue's last comment explicitly calls it "belt to the normalizer's braces".

### Proposed resolution

**Merge #909 first** (it carries the P0 fix and the migration). Before merging, add `isHashable()` from #900's diff into #909 — it's about 15 lines and the test infrastructure is already in #909. Then when #900 is rebased onto the merged #909, its `content-hash.ts` hunk becomes a no-op conflict that resolves to "keep the #909 version" — the rest of #900 (11 other fixes) is independent.

Merge order: **#908** (P1, independent, 12+16 lines) → **#909** (P0, enriched with `isHashable()`) → **#900** (rebase, drop content-hash.ts hunk) → **#903** (stacked on #900).

cc #896

crtahlin (changes requested):
The migration in here is genuinely wanted and is the missing piece. The regex change alongside it is already in `main`, and this variant is a regression against it.

## CRITICAL — the P0 this closes is already fixed, and [#896](https://github.com/plur-ai/plur/issues/896) is already closed

[#900](https://github.com/plur-ai/plur/pull/900) merged 2026-08-14 12:33:36; [#896](https://github.com/plur-ai/plur/issues/896) closed one second later. `main` already carries `/[^\p{L}\p{N}\p{M}_\s]/gu` and `isHashable()`, and already satisfies all three of this PR's new assertions:

```
accented Latin preserved : déploiement rapide
diacritics distinct      : true
Cyrillic preserved       : развертывание должно
```

This branch's merge-base is `82e50d2` — before [#900](https://github.com/plur-ai/plur/pull/900) — and [#900](https://github.com/plur-ai/plur/pull/900) was squash-merged, so its content never became an ancestor here. That is why the work reads as unapplied and why `mergeState` is `DIRTY`. Same structural trap as [#903](https://github.com/plur-ai/plur/pull/903); [#913](https://github.com/plur-ai/plur/issues/913) tracks the cause.

Flagging plainly because the `P0` label invites urgency: there is no open P0 data-loss bug here. It shipped this morning.

## CRITICAL — this variant drops `_` from the character class, which regresses `main`

`main` uses `[^\p{L}\p{N}\p{M}_\s]`; this PR uses `[^\p{L}\p{N}\p{M}\s]`. The underscore is kept deliberately in `main`, with the reason recorded in the file:

> `_` is kept explicitly because `\w` included it and dropping it would change `snake_case` identifiers, which are common in these statements.

Measured — three of four **pure-ASCII** samples hash differently:

| Statement | `main` normalises to | this PR normalises to | same hash |
|---|---|---|---|
| `use snake_case identifiers` | `use snake_case identifiers` | `use snakecase identifiers` | **no** |
| `set RESTIC_CACHE_DIR explicitly` | `set restic_cache_dir explicitly` | `set resticcachedir explicitly` | **no** |
| `pass max_llm_calls to bound cost` | `pass max_llm_calls to bound cost` | `pass maxllmcalls to bound cost` | **no** |
| `plain ascii statement` | unchanged | unchanged | yes |

This also falsifies the PR's own migration note — *"ASCII-only stores are unaffected — their hashes are identical before and after"* — for any statement containing an underscore, which here means most identifier-bearing statements.

## WARNING — the migration is missing the `isHashable` guard that `main`'s repair path has

`up()` stamps `computeContentHash(e.statement)` on every engram unconditionally, including statements that normalise away entirely — writing SHA-256 of the empty string onto all of them. That is precisely the [#896](https://github.com/plur-ai/plur/issues/896) collapse mechanism: they end up sharing one hash and dedup absorbs them.

`repairContentHashes` refuses exactly this (`index.ts:5428`):

```js
if (!isHashable(e.statement)) {
  unhashable.push({ id: e.id, statement: e.statement })
  continue
}
```

[#900](https://github.com/plur-ai/plur/pull/900) added that hardening after measuring 961 such rows. Without it, the migration and the CLI repair give different answers for the same store — worse than either alone.

## Praise — and the part worth keeping

**The migration is new, and it closes a real gap.** `main` has the CLI repair (`plur reindex-hashes`, [#897](https://github.com/plur-ai/plur/issues/897)) but no migration, so an upgraded store does not self-heal: the user has to know to run a command currently documented only in a source comment. That is exactly [#911](https://github.com/plur-ai/plur/issues/911), and this answers it.

The shape is right. A declarative `Engram[] → Engram[]` leaves locking and writing to the runner rather than reimplementing the write path [#900](https://github.com/plur-ai/plur/pull/900)'s audit found unsafe in the CLI version. And `down()` as a documented no-op — *"restoring them would reintroduce the bug"* — is the correct call, explained rather than assumed.

## Suggested rework

Probably cleaner as a new migration-only PR against `main` than as a revision of this one, but either way:

- **Keep** `migrations/20260813-006-recompute-content-hashes.ts`, the `runner.ts` registration, and `HASH_NORMALIZER_VERSION` (also new).
- **Add** the `isHashable` guard to `up()` so it agrees with `repairContentHashes`.
- **Drop** `content-hash.ts` and `index.ts` — already in `main`, and this variant regresses `_`.
- **Drop** the `sp1-memory-intelligence.test.ts` additions, or re-add them against `main`'s regex, where they pass as written.

That turns a conflicted duplicate into a small, wanted change that closes [#911](https://github.com/plur-ai/plur/issues/911).

## Verification

Suite not run, deliberately: the branch is `DIRTY` against `main` and the recommendation is to drop most of it, so a green run on this head would describe a state that cannot merge and should not. The specific claims were verified directly instead — `main`'s regex against this PR's three assertions, and the underscore divergence across four ASCII samples, both shown above.

miles-on-nightshift:
## Closing — P0 is fixed, migration needs a clean PR

Closing this PR per @crtahlin's review guidance. Summary of state:

**The P0 is already fixed.** #900 merged 2026-08-14 and landed `/[^\p{L}\p{N}\p{M}_\s]/gu` + `isHashable()` in `main`. `#896` was closed at the same time. No open P0.

**This variant has a regression.** It drops `_` from the character class, changing the hash of any statement containing an underscore (e.g. `use snake_case identifiers`, `set RESTIC_CACHE_DIR explicitly`). The PR's own migration note — *"ASCII-only stores are unaffected"* — is falsified by this, which is also why this branch's merge-base predates #900 and the diff arrives `DIRTY`.

**The migration is the only new thing, and it needs one fix.** `20260813-006-recompute-content-hashes.ts` is genuine new work that closes #911 (auto-migration for stores with stale content hashes). But `up()` calls `computeContentHash` unconditionally, writing `SHA-256("")` onto every unhashable statement — the exact collapse mechanism #896 fixed. It needs the `isHashable` guard from `main`'s `repairContentHashes` (`:5428`) before it can land safely.

**Path forward:** a new, focused PR against `main` carrying only:
- `migrations/20260813-006-recompute-content-hashes.ts` with the `isHashable` guard in `up()`
- `runner.ts` registration
- `HASH_NORMALIZER_VERSION` export

That turns this into a small, conflict-free change that closes #911 without re-introducing anything #900 already shipped.

Routing migration-only PR creation to nightshift queue.

plur9:
Follow-up: the migration-only PR this closure called for is **#928**, open against `main` now — no need for the nightshift queue to create another.

It carries exactly the three things listed above (migration + `runner.ts` registration + `HASH_NORMALIZER_VERSION`), with the `isHashable` guard in `up()`, and drops the `content-hash.ts`/`index.ts` duplicates and the `sp1-memory-intelligence.test.ts` additions. Branch is `main` + 1 commit.

Two things added beyond the plan: a 9-case test file (nothing tested this migration), and an "only rewrite when the value changes" rule so an already-correct store is not rewritten wholesale. Both revert-checked.

This branch (`fix/896-unicode-content-hash`) has been restored to `322780f`, its state when this PR was closed.

Issue #907, opened 2026-08-13
plur_feedback cannot resolve ANY remote-scoped engram — the relevance loop is dead for team knowledge


plur9:
Two defects, found while investigating the wrap-up's learning-classifier report. They compound, and the second is the serious one.

## Reproduction (live, on the installed 0.17.x MCP)

```
plur_recall("audit subagents revert-checks isolation worktree")
  → id: "ENG-GPL-2026-08-13-025", scope: "group:plur/plur-ai/engineering"   ✅ found

plur_feedback("ENG-GPL-2026-08-13-025", "positive")   → "Engram not found"  ❌
plur_feedback("ENG-2026-08-13-025",     "positive")   → "Engram not found"  ❌

# control — a global-scoped engram, same call shape
plur_feedback("ENG-2026-0706-011", "positive")        → { success: true }   ✅
```

`plur_recall` returns the id verbatim; `plur_feedback` says that exact id does not exist. A local engram accepts the identical call.

## Defect 1 — `plur_learn` returns an id that does not resolve

Writing to a remote/team scope returns an **unprefixed** id while the engram is stored **prefixed**:

| | |
|---|---|
| `plur_learn(..., scope: "group:plur/plur-ai/engineering")` returned | `ENG-2026-08-13-025` |
| `plur_recall` returns, and the store holds | `ENG-GPL-2026-08-13-025` |

Every caller that records what it just wrote records an id that cannot be looked up. This is why the wrap-up's learning-classifier reported **8 of 28 scanned entries "not locatable in the corpus under their documented ID"** — they are almost certainly not missing, just cited under the id `plur_learn` handed back. It also corrupts engram cross-references: `ENG-GPL-2026-08-13-028`'s own statement cites `ENG-2026-08-13-025`, because it was written from the returned value.

## Defect 2 — `plur_feedback` resolves no remote-scoped engram, under any id

This is the one that matters. Feedback is one of the five core operations and the mechanism that trains injection relevance. For **team-scoped knowledge — the enterprise product's whole point — it silently does nothing.** Every rating an agent gives a shared engram is discarded with an error the agent usually ignores, so shared memory never improves with use while local memory does.

Not the same as Defect 1: it fails on the prefixed id too, so fixing the returned value does not fix this.

## Hypothesis

[...]

PR #916, opened 2026-08-14
fix(mcp): report learn ids in the namespaced form recall returns

### LINKED ISSUE #914: plur_learn returns server id, plur_recall returns namespaced id — callers that round-trip will use the wrong form
## Background

Surfaced during the #907 resolution. Now cosmetic rather than functional (both id forms are accepted by both `feedback` and `forget` walks after the #907 fix), but still a source of confusion and a potential footgun for callers that record what they just wrote.

## The inconsistency

| operation | id form returned |
|---|---|
| `plur_learn(..., scope: "group:plur/..."))` | `ENG-2026-08-13-025` (unprefixed, server-assigned) |
| `plur_recall` (and recall\_hybrid) | `ENG-GPL-2026-08-13-025` (namespaced) |

A caller that records the id returned by `plur_learn` and passes it back later (e.g. to `plur_feedback`) uses the wrong form. Before #907 this was a hard failure; now both forms resolve, but the inconsistency makes the API confusing.

## Proposal

Normalise `plur_learn`'s return value to use the same namespaced form that `plur_recall` hands back. That makes the round-trip `learn → recall → feedback/forget` id-stable by construction, regardless of whether both forms happen to be accepted.

## Acceptance test

```
const id = plur_learn('...', { scope: 'group:plur/plur-ai/engineering' }).id
plur_feedback(id, 'positive')   // must succeed
plur_forget(id)                 // must succeed
```

The id returned by `plur_learn` should be accepted by `plur_recall`, `plur_feedback`, and `plur_forget` without any transformation.

Spun out from #907.

luantaraschi:
Closes #914. Claimed on the issue first, per CONTRIBUTING step 2, since I have no write access to self-assign.

**Where the normalisation happens, and why not in core.** My first attempt namespaced `learnRouted`'s return value. Two existing tests pin the opposite (`session-scope.test.ts` asserts `engram.id === 'ENG-REMOTE-001'`), and they are right to: the server id is what the remote holds, and a core caller talking to the store needs it. So core keeps returning it, and the reporting surface gets a separate answer:

```ts
/** The id shape the READ paths hand back for an engram (#914). */
readIdFor(engram: { id: string; scope: string }): string
```

`plur_learn` reports `plur.readIdFor(engram)`. An outbox engram is excluded, because it sits in the local store under a local id until the retry lands, and that is the id `recall` returns for it. `readIdFor` returns the id unchanged when the scope is not backed by a remote store, mirroring `_isRemoteBackedScope`'s exact-scope rule so the two agree on which writes leave the machine.

The namespacing rule itself moves into `namespaceEngramId()` next to `storePrefix`, and `remote-recall.ts` now calls it instead of repeating the regex, so the write report and the read path cannot drift. It is idempotent, so an already-namespaced id is never doubled.

**The acceptance test from the issue, run for real.** Before touching the assertions in `e2e-remote.test.ts` I wanted to know whether the round trip actually works with the namespaced form, rather than assume it from #907. A throwaway probe drove `plur_learn` then `plur_forget` through the real MCP server against the stub:

```json
{"returnedId":"ENG-TE2-SRV-001","serverId":"ENG-SRV-001","forgetSuccess":true,"serverStatusAfter":"retired"}
```

So forgetting by the reported id retires the right server row. The probe is not in the diff; it existed to justify the assertion changes below.

**Four existing assertions changed, all the same shape.** They took the reported id and used it as a direct server-side lookup key (`stub.getEngram(result.id)`). That is exactly the coupling this issue is about, so each now derives the server id by stripping the prefix and keeps asserting against the server row, while asserting the reported id is the namespaced form. The `plur_forget` and `plur_feedback` tests additionally now pass the *reported* id into the tool, which is the round trip the issue asks for and which the probe above confirms end to end. Nothing was weakened: every one of them still asserts server-side state.

**What I ran**, Windows, Node 22.20.0, pnpm 10.34.1:

| suite | with the change | on unmodified `7c3aadb` |
| --- | --- | --- |
| `packages/core` | 2791 passed, 8 files failing | 2789 passed, the same 8 files |
| `packages/mcp` | 389 passed, 1 failing | 1 failing, the same one |

The three tests added here account for the core delta. The failures are mine, not the branch's:

- the eight core files are all SQLite ones (`store-contract`, `storage-indexed`, `scope-pushdown`, `read-side-scope-visibility`, …). I installed with `--ignore-scripts`, so `better-sqlite3` never built. Identical set before and after.
- `session.test.ts > cross-session: project A scope does not leak into project B` fails on unmodified `main` here too; it is a `process.chdir` + tmpdir case that does not hold on Windows.
- `concurrency-write-path.test.ts` appeared once in each direction under the full parallel run and passes 3 of 3 in isolation, so it is a flake rather than a signal.

`pnpm --filter @plur-ai/core build` was run before the MCP suites, per CONTRIBUTING.

One note for review: `storePrefix` and `namespaceEngramId` are now exported from `@plur-ai/core`. `readIdFor` is the API a surface should use; the two are exported so the tests (and any caller reasoning about the shape) do not re-derive the prefix rule a fourth time. Happy to keep them internal and have the tests import from `../src/engrams.js` instead if you would rather not widen the public surface.

### DIFF
--- a/packages/core/src/engrams.ts
+++ b/packages/core/src/engrams.ts
@@ -608,6 +608,21 @@ export function loadAllPacks(packsDir: string): LoadedPack[] {
   return packs
 }
 
+/**
+ * Namespace a store-local id the way the read paths hand it back.
+ *
+ * A store's ids are prefixed with `ENG-{storePrefix(scope)}-` on load so ids
+ * from different stores cannot collide locally. The write path has to return
+ * the same shape, or a caller that records what it just wrote is holding an id
+ * no read path ever produced (#914). Idempotent: an already-namespaced id is
+ * returned unchanged, so the two call sites can't double-prefix each other.
+ */
+export function namespaceEngramId(id: string, scope: string): string {
+  const prefix = storePrefix(scope)
+  if (new RegExp(`^(ENG|ABS|META)-${prefix}-`).test(id)) return id
+  return id.replace(/^(ENG|ABS|META)-/, `$1-${prefix}-`)
+}
+
 /** Derive a 3-char prefix from a store scope (e.g. 'datafund' → 'DFU', 'project:myapp' → 'PMY') */
 export function storePrefix(scope: string): string {
   const parts = scope.split(/[:\-_./]/).filter(Boolean)
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -6,7 +6,7 @@ import { detectPlurStorage, type PlurPaths } from './storage.js'
 import { IndexedStorage } from './storage-indexed.js'
 import { PGLiteAdapter } from './storage-pglite.js'
 import { loadConfig } from './config.js'
-import { generateEngramId, engramIdDatePrefix, loadAllPacks, storePrefix, initFilesystemStore } from './engrams.js'
+import { generateEngramId, engramIdDatePrefix, loadAllPacks, storePrefix, namespaceEngramId, initFilesystemStore } from './engrams.js'
 import { maybeDailyBackup } from './backup.js'
 import { logger } from './logger.js'
 import { searchEngrams, ftsTokenize, extendCorpusStats, searchTextFrom } from './fts.js'
@@ -100,6 +100,10 @@ export { EngramStoreUnreadableError, EngramStoreShrinkError } from './engrams.js
 // entries it does not own. Exposing the existing loader rather than letting a
 // caller write a fourth one is the whole point of #877.
 export { loadEngrams, saveEngrams } from './engrams.js'
+// The id-namespacing pair (#914). `readIdFor` is the API a surface should use;
+// these are exported so a caller (and the tests) can reason about the shape
+// without re-deriving the prefix rule a fourth time.
+export { storePrefix, namespaceEngramId } from './engrams.js'
 export {
   maybeDailyBackup,
   listBackups,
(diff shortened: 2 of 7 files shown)


plur9 (approved):
🤖 heartbeat — reviewed as CTO cadence.

**Approach is correct.** The core insight — keep `learnRouted` returning the server's own id (that is what the remote holds and what core callers need), and let the surface layer convert to the namespaced form via `readIdFor` — is the right split. Extracting `namespaceEngramId` from the inline regex in `remote-recall.ts` eliminates the only place the namespace rule was duplicated; write and read paths can no longer drift.

**`plur_batch_learn` correctly excluded.** Batch takes the local learn path; ids it returns are local ids, and recall hands them back in the same shape. No mismatch.

**Fallback catch path correctly excluded.** A failed remote write goes to the local outbox; `plur.learn()` produces a local id, and recall returns that local id for an outbox engram. Returning `engram.id` unchanged there is right.

**Three tests cover the surface.** Remote write → namespaced; local scope → unchanged; already-namespaced → idempotent. The e2e assertions that change from `stub.getEngram(result.id)` to strip-then-look-up are stronger, not weaker — they now test the actual round-trip the issue described.

Ready to merge after CI is green.

plur9:
CI workflows were awaiting first-run approval for a new contributor — approved both runs now. Tests should be running. PR is approved and ready to merge once CI passes.

Issue #913, opened 2026-08-14
main has no required status checks, and no up-to-date requirement — a green stacked PR nearly shipped a regression


crtahlin:
## The near-miss that prompted this

On 2026-08-14, [#903](https://github.com/plur-ai/plur/pull/903) and [#902](https://github.com/plur-ai/plur/pull/902) were each individually correct, each fully green, and their combination was broken.

- [#902](https://github.com/plur-ai/plur/pull/902) (merged) fixed [#553](https://github.com/plur-ai/plur/issues/553) by making the per-pack injection telemetry bucket on `_pack`, the stamp `_loadAllEngrams` puts on corpus-merged pack engrams.
- [#903](https://github.com/plur-ai/plur/pull/903) fixes [#901](https://github.com/plur-ai/plur/issues/901) by filtering those same engrams *out* of the injection corpus: `allEngrams.filter(e => e._pack === undefined)`.

Together the `_pack`-stamped rows never reach the code [#902](https://github.com/plur-ai/plur/pull/902) fixed, so `pack_counts` loses the bucket entirely — regressing [#553](https://github.com/plur-ai/plur/issues/553) from the other direction. Verified:

| State | `packages/mcp/test/session.test.ts` |
|---|---|
| `main` @ `7c3aadb` | **27 / 27 pass** |
| [#903](https://github.com/plur-ai/plur/pull/903) rebased onto `main` | **1 fails** |

Nothing in the current process would have caught it. [#903](https://github.com/plur-ai/plur/pull/903)'s ten checks ran against a base predating [#902](https://github.com/plur-ai/plur/pull/902); [#902](https://github.com/plur-ai/plur/pull/902) merged without ever seeing [#903](https://github.com/plur-ai/plur/pull/903). It was found by hand.

## Two configuration findings

```
$ gh api repos/plur-ai/plur/branches/main/protection
required_status_checks:  { contexts: null, strict: null }
required_pull_request_reviews: 1

$ gh api repos/plur-ai/plur
delete_branch_on_merge: false
allow_auto_merge:       false
```

**1. There are no required status checks on `main`.** CI runs, goes green, and is not enforced. A pull request with a failing test matrix can be merged today on a single approval. The green checkmarks are advisory.

**2. `strict` is unset** — branches are not required to be up to date with `main` before merging. This is the specific control that would have caught the case above: it keeps the merge button disabled until the branch is current, so CI re-runs against the *actual merge result* rather than against a stale snapshot.

[...]

PR #922, opened 2026-08-16
perf(core): primary-store pushdown for _filterEngrams (#906)

### LINKED ISSUE #906: recallHybridWithMeta loads the whole corpus on every call — no primary-store pushdown
## Problem

`recallHybridWithMeta()` loads the **entire corpus** on every call, even when the store implements primary-store pushdown.

`recall()` and `_semanticRecall` both push down to the store. The hybrid path does not: `_filterEngrams()` → `_loadAllEngrams()` → `loadCached()` → `load()` runs unconditionally.

## Why it matters

`plur_recall` defaults to hybrid. So for a store backed by a database, this is one whole-table read per recall, permanently — not once per session. On a tenant of any size that is the dominant cost of a recall, and it grows with the corpus while the result set does not.

Measured by instrumenting a real recall against a Postgres-backed store: `recall()` issued no whole-corpus read; `recallHybridWithMeta()` issued one every time.

## Not a correctness bug

Worth stating so it is triaged accurately: the load takes no locks, and `_filterEngrams` still applies the `scopes` authorization, so nothing leaks and nothing deadlocks. It is a cost defect, and an invisible one — it does not appear in any per-query metric a host would normally watch.

## Ask

Give the hybrid path the same primary-store pushdown that `recall()` already has, so a store that can filter and rank server-side is asked to, rather than being handed the whole table to filter in memory.

miles-on-nightshift:
## Summary

`recallHybridWithMeta` called `_loadAllEngrams` on every invocation — a full-corpus read — even when the primary store (e.g. Postgres) supports server-side filtering. When a non-PGLite primary adapter is present, `_filterEngrams` now delegates to `adapter.loadFiltered()` and `_engramsOutsidePrimaryStore()` instead of reading the whole corpus.

PGLite is excluded: its SQLite index tracks file-backed stores only and the hybrid path needs the full corpus that `_loadAllEngrams` provides.

## Test

Adds `filter-engrams-primary-pushdown.test.ts`: mock adapter verifies that `loadFiltered` is called in place of the full corpus read, and that all authorization filters (status, scopes allow-list, domain) are still enforced on the adapter path.

## Context

This commit (`fc257a9`) was authored 2026-08-16 and pushed to branch `fix/906-primary-pushdown` but no PR was opened. Issue #906 was then incorrectly closed by an agent citing this commit as being on `main` — it was not. PR corrects that.

Closes #906

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -4437,6 +4437,30 @@ export class Plur {
         domain: options?.domain,
       })
     } else {
+      const adapter = this._primaryQueryAdapter()
+      if (adapter && !this.pgliteAdapter) {
+        // #906: primary-store adapter (e.g. Postgres) supports server-side
+        // filtered load. Use it instead of _loadAllEngrams + in-memory filter:
+        // one scoped query replaces a full table read on every recallHybrid call.
+        //
+        // PGLite is excluded even though it also exposes role='primary' — its
+        // SQLite index tracks only file-backed stores, and the PGLite hybrid
+        // path needs the full corpus (including packs and remote stores) that
+        // _loadAllEngrams provides.
+        //
+        // Temporal validity and min_strength are applied below, same as every
+        // other path. _engramsOutsidePrimaryStore applies active/scope/domain
+        // filters to packs and remote secondary stores.
+        const primaryRows = await adapter.loadFiltered({
+          status: 'active',
+          scope: options?.scope,
+          scopes: options?.scopes,
+          visibilityGrants: this._grantedScopes(),
+          domain: options?.domain,
+        })
+        const outsiders = await this._engramsOutsidePrimaryStore(options)
+        engrams = [...primaryRows, ...outsiders]
+      } else {
       // PGLite path (or no-index path): read from YAML so this code stays
       // synchronous. PGLite is currently used for vector/Cypher queries
       // and remains in sync via _syncIndex on every write — but the
@@ -4469,6 +4493,7 @@ export class Plur {
         // inject.ts INJECT_GLOBAL_IS_TARGETED).
         const visible = makeVisibilityPredicate(options.scope, this._grantedScopes())
         engrams = engrams.filter(e => visible(e.scope))
+      }
       }
     }
     // Temporal validity: exclude expired or not-yet-valid engrams.
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -4438,10 +4438,17 @@ export class Plur {
       })
     } else {
       const adapter = this._primaryQueryAdapter()
-      if (adapter && !this.pgliteAdapter) {
+      if (adapter && typeof adapter.loadFiltered === 'function' && !this.pgliteAdapter) {
         // #906: primary-store adapter (e.g. Postgres) supports server-side
         // filtered load. Use it instead of _loadAllEngrams + in-memory filter:
         // one scoped query replaces a full table read on every recallHybrid call.
+        //
+        // The `typeof loadFiltered` check is LOAD-BEARING, not defensive noise:
+        // _primaryQueryAdapter() duck-types on role + searchBM25 only, so a
+        // store can qualify while implementing just the query surface —
+        // ReadonlyStoreGuard forwarded exactly that subset before it learned to
+        // forward loadFiltered, and #903's hybrid-pushdown test mock still
+        // does. Such a store must take the fallback read below, not crash here.
         //
         // `!pgliteAdapter` is a BELT, not a live branch: the two cannot both be
         // set. `pgliteAdapter` is constructed only when `indexTier === 'pglite'`,
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -4443,10 +4443,18 @@ export class Plur {
         // filtered load. Use it instead of _loadAllEngrams + in-memory filter:
         // one scoped query replaces a full table read on every recallHybrid call.
         //
-        // PGLite is excluded even though it also exposes role='primary' — its
-        // SQLite index tracks only file-backed stores, and the PGLite hybrid
-        // path needs the full corpus (including packs and remote stores) that
-        // _loadAllEngrams provides.
+        // `!pgliteAdapter` is a BELT, not a live branch: the two cannot both be
+        // set. `pgliteAdapter` is constructed only when `indexTier === 'pglite'`,
+        // and that tier is chosen only when `hasPrimaryQueryStore` is false —
+        // i.e. exactly when `_primaryQueryAdapter()` returns null. So whenever
+        // `adapter` is non-null, `pgliteAdapter` is already null by construction.
+        //
+        // Kept because the invariant lives in the constructor's tier selection,
+        // far from here, and the consequence of it changing is silent: the
+        // PGLite hybrid path needs the full corpus (packs + remote stores) that
+        // _loadAllEngrams provides, so taking this branch with PGLite active
+        // would narrow the corpus rather than fail. Pinned by
+        // `filter-engrams-primary-pushdown.test.ts`.
         //
         // Temporal validity and min_strength are applied below, same as every
         // other path. _engramsOutsidePrimaryStore applies active/scope/domain
--- a/packages/core/src/store/readonly-store-guard.ts
+++ b/packages/core/src/store/readonly-store-guard.ts
@@ -104,6 +104,10 @@ export class ReadonlyStoreGuard implements PrimaryStore {
    */
   readonly role?: string
   readonly searchBM25?: (query: string, opts: { limit: number } & Record<string, unknown>) => Promise<Engram[]>
+  /** Pure read — #906's _filterEngrams pushdown calls this on any primary query
+   *  adapter. Not forwarding it would demote every guarded query-capable store
+   *  to the whole-corpus fallback (or, before the typeof guard, crash it). */
+  readonly loadFiltered?: (filter: Record<string, unknown>) => Promise<Engram[]>
   /**
    * Also a read, and the one the whitelist's own argument failed on (#753,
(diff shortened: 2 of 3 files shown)


miles-on-nightshift:
🤖 heartbeat — CI was failing with TS2352 on all Node matrix legs. Two bugs in the new test file:

1. `makeEngram` used `as Engram` on a minimal fixture missing 14+ required fields — TypeScript rejects the cast when the types don't sufficiently overlap. Fixed with `as unknown as Engram`.
2. Seed engram used `status: 'inactive'`, which is not in the Engram enum (`'active' | 'dormant' | 'retired' | 'candidate'`). Changed to `'retired'` (semantically correct for the 'retired fact' fixture).

Pushed b92853b — CI should go green.

plur9:
Reviewed, merged `main` in, and pushed one commit (`1d2c1da`) fixing a test that could not fail. **The source change is correct and is pinned** — the finding is entirely in the test file and one comment.

## The mechanism is sound, and the revert-check proves it

The decomposition is exactly right. `_loadAllEngrams` is literally `[...primary, ...(await this._loadSecondaryAndPacks())]` (`index.ts:1095`), so replacing the `primary` half with a scoped query and handing the other half to `_engramsOutsidePrimaryStore` is a faithful split rather than a reimplementation. That same pairing is already load-bearing at `index.ts:3263` and `:3767`, so this applies an established pattern rather than inventing one.

Measured, by forcing the branch off (`if (false && adapter && ...)`):

| State | `filter-engrams-primary-pushdown.test.ts` |
|---|---|
| as submitted | 4 / 4 pass |
| pushdown disabled | **3 of 4 fail** |

So the mechanism is genuinely pinned. Which is what made the fourth test stand out.

## The fourth test could not fail — and it was named after the one risky part

`does not call loadFiltered when indexedStorage is active (SQLite path)` passed **with the entire feature disabled**. Reading it:

- it built a second `Plur` with the *same* options as the first, so `indexedStorage` was never active — the comment says so itself: *"Without index:true the default YAML path is taken, so loadFiltered IS called"*
- it contained **no `expect()` at all**
- it wrapped the body in `try/catch` that swallowed the error, so even "just verify no crash" was not verified

A dead test is normally a small debt. This one was worse than that, because it carried the name of the exclusion that makes the whole change safe, so the suite read as covering it.

## Which turned out to matter, because that guard is unreachable

Chasing what the test *should* have asserted: `!this.pgliteAdapter` can never be false when `adapter` is truthy.

```
hasPrimaryQueryStore = this._primaryQueryAdapter() !== null
indexTier            = hasPrimaryQueryStore ? 'none' : ...
if (indexTier === 'pglite') this.pgliteAdapter = new PGLiteAdapter(...)
```

`_primaryQueryAdapter()` reads only `this._primaryStore` — it never consults `pgliteAdapter`. So PGLite is constructed **only** when there is no primary query adapter, which is exactly when this branch is not taken. Confirmed at runtime rather than by reading:

```
PGLITE: null | PRIMARY ADAPTER: true | INDEXED: null
```

The code is not wrong — but the comment presented this as an active exclusion decision (*"PGLite is excluded even though it also exposes role='primary'"*), and it is a belt. I have not removed the guard: the invariant lives in the constructor's tier selection, far from the use site, and if it ever changes the failure is silent — the branch would narrow the corpus that the PGLite hybrid path needs, rather than error. The comment now says that, and says why it is kept.

The test now pins **the invariant** instead of pretending to exercise a branch, so a tier-selection change fails loudly in one place. Revert-checked: injecting a non-null `pgliteAdapter` fails it (`expected {} to be null`).

## Two smaller things in the same commit

**`loadCallsBeforeRecall` removed.** It was a literal `0` dressed as a measurement, and its bound of `+2` permitted exactly the full-corpus reads this PR exists to eliminate. The revert-check shows the real pin was always `loadFilteredCalls.length > 0` next to it.

**Docstring corrected.** It claimed every authorization filter *"is still enforced when the adapter path is taken."* The mock's `loadFiltered` implements only `status` and `domain` — it ignores `scope`, `scopes` and `visibilityGrants` — and no test asserts anything is excluded. What the suite pins is that filters are **forwarded**; enforcement is a SQL `WHERE` clause belonging to the Postgres adapter's own tests. Worth being exact because the in-memory code being replaced did *both*, so the reasonable assumption is the wrong one.

## Not addressed

The `else {` body was not re-indented, so there is now a nesting level the code's shape does not show, closed by a lone `}`. Left alone deliberately — re-indenting ~35 lines would bury the actual change in a 4,300-line file's diff. Worth a follow-up tidy, not worth it here.

## Verification

```
merge origin/main @ 8f0fc08          clean
pnpm --filter @plur-ai/core build    clean
vitest filter-engrams-primary-pushdown  4/4 pass
pnpm typecheck:tests                 clean for these files
                                     (only @plur-ai/ui module-resolution noise, unbuilt in worktree)

revert-check A: pushdown disabled         3 of 4 fail  (was 3 of 4 — old test 4 survived)
revert-check B: pgliteAdapter injected    new test 4 fails
```

@crtahlin — flagging rather than self-approving, since I wrote `1d2c1da` and it is the only part of this PR I did not review at arm's length. The original two commits I did review independently and consider correct.

crtahlin (dismissed):
Approving. A pushdown's only real risk is that the fast path returns a different set from the slow one, so I traced that end to end rather than reading the diff — the equivalence holds.

## Equivalence verified

| Filter | In-memory path | Pushdown path |
|---|---|---|
| status | `e.status === 'active'` | `status = $1` |
| scopes | `scopeAllowFilter` — exact membership, `[]` matches nothing | `scope = ANY($n::text[])`, same, with the `!== undefined` guard |
| domain | `e.domain?.startsWith(d)` | `domain LIKE $n \|\| '%'`, LIKE metacharacters escaped |
| visibility | `makeVisibilityPredicate(scope, grants)` | same predicate server-side, and applied to outsiders too |

Composition is complete and disjoint by construction:

```js
_loadAllEngrams() = _loadCached(paths.engrams) + _loadSecondaryAndPacks()
pushdown          = adapter.loadFiltered(...)  + _engramsOutsidePrimaryStore(options)
```

I checked the visibility predicate on the outsider half specifically, since a gap there would be an authorization divergence rather than a performance one. It is applied. Both primary reads also resolve to the same store — `_loadCached(path)` routes through `_storeAt(path)`, which for a Postgres primary is the adapter `loadFiltered` runs against — so this is one corpus queried two ways rather than two corpora.

## One coverage gap, filed as [#931](https://github.com/plur-ai/plur/issues/931)

Nothing exercises engrams **outside** the primary store: no fixture configures a secondary store or installs a pack. So the union half of the change is untested, and a regression dropping pack, secondary or remote engrams would pass this suite and CI.

The header's disclaimer — that the suite pins forwarding rather than enforcement — is correct and worth keeping, but it does not extend to this; outsiders are a different axis, not a weaker assertion on the same one. Worth raising because the helper now depended on has this exact failure in its recorded history, per its own docstring: *"it skipped `url` stores entirely (so an enterprise team store vanished from `recall()` while `list()` still showed it)"*. Same quiet-corpus-narrowing shape as [#901](https://github.com/plur-ai/plur/issues/901). Filed rather than blocking, since the code as written is correct.

## Notes

The second test is named `passes status:active to loadFiltered — inactive rows are excluded` but asserts only the first clause. The mock implements `status` and `domain`, so exclusion is not what is checked here — the header draws the distinction correctly, it is the test *name* that overstates.

The first test takes ~21.6 s locally, presumably the embedder load on the first `recallHybrid`. Tolerable against a 5–7 minute job, but a real addition worth knowing about if suite time becomes a concern.

## Praise

**The replaced PGLite-exclusion test is a model of how to handle an unreachable branch.** Rather than deleting it or pretending to test it, it states that `!this.pgliteAdapter` is unreachable given the constructor's tier selection, pins the *invariant* instead, and says what breaks if the tier logic changes. Then it closes with `expect(adapter.loadFilteredCalls.length).toBeGreaterThan(0)`, so the asserted state is demonstrably the state the branch ran under — without that line it would be a test about an unrelated object graph.

**The `!this.pgliteAdapter` conjunct is documented as a belt, with the reasoning for keeping it.** The argument is the right one: the invariant lives in the constructor far from here, and the consequence of it changing is silent narrowing rather than a failure. That is precisely when a redundant guard earns its place.

**The header disclaims what the suite does not cover** — *"a reader could reasonably assume this suite covers both. It does not."* That sentence is worth more than an extra test, because it stops the next reader trusting the suite for something it never checked.

**The PR body corrects the record** on [#906](https://github.com/plur-ai/plur/issues/906) having been closed by an agent citing a commit that was never on `main`.

## Verification

```
new suite:   4/4 passed
core + mcp:  3270 passed, 38 skipped, 3 failed
```

plur9:
🤖 Pushed `2a45c23` before merge — the combined-tree check caught a **fourth instance of the #913 class**, and this PR was the crashing side.

Rebased onto main with #903 in, three of #903's `hybrid-pushdown.test.ts` cases fail: `TypeError: adapter.loadFiltered is not a function`. `_primaryQueryAdapter()` duck-types on `role` + `searchBM25` only, so a store can qualify for the pushdown while implementing just the query surface. #903's mock is that shape — and so was production: **`ReadonlyStoreGuard`'s #884 whitelist never forwarded `loadFiltered`**, so `readonly: true` on a Postgres store would have crashed on every recall. The exact failure #884 exists to prevent, reintroduced one layer up.

Fix: the pushdown branch now requires `typeof adapter.loadFiltered === 'function'` (fallback read instead of a crash), and `ReadonlyStoreGuard` forwards `loadFiltered` (a pure read) so guarded stores keep the pushdown rather than silently demoting. Both revert-checked; the no-`loadFiltered` fixture shadows the prototype method with an instance `undefined`, since `delete` removes nothing on a class instance and made the first draft of the test vacuous — a precondition assert pins it.

Verified on the rebased head: core+mcp **3334 passed, 0 failed**, including #903's three.

PR #925, opened 2026-08-17
fix(core): sanitise only control characters, not all non-ASCII (#923)

### LINKED ISSUE #923: sanitiseResponseBody strips all non-ASCII, not just control characters — localised server errors become unreadable
## The problem

[#919](https://github.com/plur-ai/plur/pull/919) added `sanitiseResponseBody` to bound and clean server error bodies before they are persisted to `_outbox.last_error` or thrown ([#912](https://github.com/plur-ai/plur/issues/912)). The implementation is:

```js
function sanitiseResponseBody(raw: string): string {
  return raw.slice(0, 200).replace(/[^\x20-\x7E]/g, '?')
}
```

The PR description states it "Strips control characters (`\x00–\x1F`, `\x7F`)". The negated class does considerably more than that: it excludes **everything outside printable ASCII**, so all non-Latin text and every accented character is replaced with `?`.

Measured, shipped versus as-described:

| Server error body | As shipped | As described |
|---|---|---|
| `エラー: 認証に失敗しました` | `???: ?????????` | `エラー: 認証に失敗しました` |
| `Ошибка: неверный токен` | `??????: ???????? ?????` | `Ошибка: неверный токен` |
| `Échec de l'authentification` | `?chec de l'authentification` | `Échec de l'authentification` |
| `error:\x00\x1b[31mred\x07` | `error:??[31mred?` | `error:??[31mred?` |

Both variants neutralise the escape-injection case identically, so [#912](https://github.com/plur-ai/plur/issues/912)'s security goal is met either way. The only difference is that the shipped version also destroys legitimate text.

## Why it matters

`last_error` exists to be diagnostic — it is what `plur outbox` shows a user asking "why is this write stuck?". For a deployment whose server returns localised errors, that field now reads `??????: ???????? ?????`.

The net trade for those users is **readable but unbounded → bounded but unreadable**, which is not the trade [#919](https://github.com/plur-ai/plur/pull/919) set out to make. Both properties are achievable together.

There is also an internal inconsistency worth naming: this repository fixed [#896](https://github.com/plur-ai/plur/issues/896) last week — every non-Latin statement collapsing to the same hash because `normalizeStatement` used ASCII `\w`. The same ASCII-only assumption has reappeared one layer over, in code written to improve diagnostics.

## Fix

```js
function sanitiseResponseBody(raw: string): string {
  return raw.replace(/[\x00-\x1F\x7F]/g, '?').slice(0, 200)
}
```

Two changes:

1. **Narrow the class to control characters**, matching the stated intent and preserving legitimate Unicode.
2. **Sanitise before slicing.** With the current over-broad regex this is moot, but once the class is narrowed, `slice(0, 200)` can cut a surrogate pair and leave a lone surrogate in a string that gets persisted to YAML. Slicing after sanitising does not fix that on its own either — a code-point-safe truncation (`Array.from(raw).slice(0, 200).join('')`) is the complete version if lone surrogates are a concern.

Either way, the PR description should be corrected, since as written it tells a reader non-ASCII survives.

## Test worth adding

The existing tests cover truncation and control-character stripping. Neither would catch this: a case asserting that a non-ASCII error body is preserved is what pins the distinction.

## Related

- [#919](https://github.com/plur-ai/plur/pull/919) — where `sanitiseResponseBody` was introduced
- [#912](https://github.com/plur-ai/plur/issues/912) — the unbounded-body issue it addresses
- [#896](https://github.com/plur-ai/plur/issues/896) — the ASCII-only assumption, one layer down

crtahlin:
Closes #923. Follow-up to [#919](https://github.com/plur-ai/plur/pull/919), which introduced `sanitiseResponseBody` for [#912](https://github.com/plur-ai/plur/issues/912).

## Problem

The character class was `[^\x20-\x7E]` — every code point outside printable ASCII, not the control characters [#919](https://github.com/plur-ai/plur/pull/919)'s description named. So a server returning a localised error produced a row of question marks in `plur outbox`, the one surface that exists to answer "why is this write stuck?":

| Server error body | Before | After |
|---|---|---|
| `エラー: 認証に失敗しました` | `???: ?????????` | `エラー: 認証に失敗しました` |
| `Ошибка: неверный токен` | `??????: ???????? ?????` | `Ошибка: неверный токен` |
| `Échec de l'authentification` | `?chec de l'authentification` | `Échec de l'authentification` |
| `error:\x00\x1b[31mred\x07` | `error:??[31mred?` | `error:??[31mred?` |

Both classes neutralise the escape-injection vector identically — `\x00-\x1F` and `\x7F` are what cover it — so the wider class bought nothing and cost the diagnostic.

## Changes

**Narrowed to control characters**, matching the stated intent.

**Truncation now counts code points and runs after the substitution.** `String.prototype.slice` cuts mid-surrogate, and a lone surrogate in this value is persisted to YAML through `_outbox.last_error`. Running after the substitution also means the `?` replacements count toward the bound.

```js
function sanitiseResponseBody(raw: string): string {
  const cleaned = raw.replace(/[\x00-\x1F\x7F]/g, '?')
  return Array.from(cleaned).slice(0, 200).join('')
}
```

## Tests

The two existing [#912](https://github.com/plur-ai/plur/issues/912) tests pass under **either** character class — they assert truncation length and the absence of control characters, neither of which distinguishes the two. That is how this shipped. Two tests added, each pinning one half:

- a non-ASCII body survives sanitisation
- truncation respects character boundaries

One detail in the second is load-bearing and easy to lose: the fixture is `'x' + '🔥'.repeat(300)`, not a bare emoji run. Emoji are 2 UTF-16 units, so without the leading ASCII character every pair sits on an even offset and a UTF-16 `slice(0, 200)` lands exactly on a boundary — the test would pass against the very bug it exists to catch. The `'x'` shifts the pairs onto odd offsets. Noted in a comment on the test.

## Verification

```
pnpm build                                    clean
pnpm typecheck:tests                          clean
core + mcp suites          3265 passed, 38 skipped, 3 failed
remote-integration.test.ts 49/49
```

Revert-checked both halves independently:

| Reverted to | Result |
|---|---|
| `[^\x20-\x7E]` (old class) | fails `a non-ASCII error body survives` — 1 of 49 |
| narrowed class, UTF-16 `slice(0, 200)` | fails `truncation respects character boundaries` — 1 of 49 |

So neither new test is vacuous, and neither passes without its half of the fix.

## Note

This is the same ASCII-only assumption [#896](https://github.com/plur-ai/plur/issues/896) was fixed for one layer down, in `normalizeStatement`, and it reappeared in code written to improve diagnostics. Worth watching for in any future `[^\x20-\x7E]` or `\w` on text that can carry user or server content.

### DIFF
--- a/packages/core/src/store/remote-store.ts
+++ b/packages/core/src/store/remote-store.ts
@@ -40,8 +40,28 @@ export const RemoteRowSchema = z.object({
 // #912: Truncate and strip control chars from server error bodies before
 // persisting them (append → _outbox.last_error) or surfacing in thrown errors.
 // Mirrors the treatment applied to server-assigned ids in append().
+//
+// CONTROL CHARACTERS ONLY (#923). The first cut used `[^\x20-\x7E]`, which is
+// every code point outside printable ASCII — so a server returning a localised
+// error produced a row of question marks in `plur outbox`, the one surface that
+// exists to answer "why is this write stuck?":
+//
+//   'エラー: 認証に失敗しました'  -> '???: ?????????'
+//   'Ошибка: неверный токен'   -> '??????: ???????? ?????'
+//   'Échec de authentification' -> '?chec de authentification'
+//
+// Both classes neutralise the escape-injection vector identically — that is
+// what \x00-\x1F and \x7F cover — so the wider class bought nothing and cost
+// the diagnostic. It was also the same ASCII-only assumption #896 had just
+// been fixed for one layer down, in `normalizeStatement`.
+//
+// Truncation is by CODE POINT, not UTF-16 unit. `String.prototype.slice` cuts
+// mid-surrogate, and a lone surrogate in this value is persisted to YAML —
+// so the bound has to respect character boundaries, and it has to run AFTER
+// the substitution so the replacement characters are counted too.
 function sanitiseResponseBody(raw: string): string {
-  return raw.slice(0, 200).replace(/[^\x20-\x7E]/g, '?')
+  const cleaned = raw.replace(/[\x00-\x1F\x7F]/g, '?')
+  return Array.from(cleaned).slice(0, 200).join('')
 }
 
 /**
--- a/packages/core/test/remote-integration.test.ts
+++ b/packages/core/test/remote-integration.test.ts
@@ -252,6 +252,50 @@ describe('RemoteStore against stub server', () => {
     server.appendErrorResponse = null
   })
 
+  // The two assertions above pass under EITHER character class, which is how
+  // #923 shipped: the first cut stripped everything outside printable ASCII, so
+  // a localised server error arrived as '??????: ????????' in `plur outbox` —
+  // the one surface that exists to explain a stuck write. This pins the
+  // distinction the other tests cannot see.
+  it('#923 a non-ASCII error body survives sanitisation', async () => {
+    const store = new RemoteStore(baseUrl, TOKEN, 'group:test')
+    const bodies = [
+      'エラー: 認証に失敗しました',
+      'Ошибка: неверный токен',
+      "Échec de l'authentification",
+    ]
+    for (const body of bodies) {
+      server.appendErrorResponse = { status: 500, body }
+      const err = await store.append({ id: 'x', scope: 'group:test', status: 'active', statement: 'y' } as any)
+        .then(() => null).catch((e: Error) => e)
+      expect(err).not.toBeNull()
+      expect(err!.message, `body was mangled: ${err!.message}`).toContain(body)
+    }
+    server.appendErrorResponse = null
+  })
+
+  // Truncation must not split a surrogate pair: a lone surrogate in this value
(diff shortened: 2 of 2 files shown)


miles-on-nightshift (approved):
Fix is correct. The narrowing from `[^\x20-\x7E]` (everything outside printable ASCII) to `[\x00-\x1F\x7F]` (control characters only) matches the stated intent of #919 — the escape-injection vector is fully covered by the control-character range alone, and the wider class was silently destroying diagnostic value.

`Array.from(cleaned).slice(0, 200).join('')` for code-point-safe truncation is the right approach. The fixture detail (`'x' + '🔥'.repeat(300)`) is load-bearing and well-explained — without the leading ASCII character, every emoji pair lands on an even offset and a UTF-16 slice would pass the test against the very bug it exists to catch.

Both new tests are confirmed non-vacuous by the revert table in the PR description. The three existing failures are environmental (git-in-worktree), not related to this change.

@plur9 ready for merge.

PR #927, opened 2026-08-17
docs(spec): add measured_under to ENGRAM-STANDARD-v1.md, guard markdown drift (#924)

### LINKED ISSUE #924: ENGRAM-STANDARD-v1.md drifts from the Zod schemas — the JSON spec is drift-checked, the published standard is not
## The gap

`spec/engram.schema.json` is generated and drift-checked. `spec/ENGRAM-STANDARD-v1.md` documents the same fields in a prose table, is maintained by hand, and has neither.

- `pnpm --filter @plur-ai/core gen:schemas` → `scripts/gen-spec-schemas.ts` writes the three `.json` files only.
- `packages/core/test/spec-schema-drift.test.ts` (#315) asserts all three `.json` files match generated output, is deterministic, and preserves documented divergences. It never looks at the markdown.

So adding a field to a Zod schema produces a **CI failure** until the JSON is regenerated, and produces **nothing at all** if the standard's table is forgotten. One artifact is mechanised, the other relies on memory.

## Two occurrences in nine days

**[#908](https://github.com/plur-ai/plur/pull/908)** widened `commitment` to include `'draft'`. The JSON was out of sync — CI failed on all four Node versions with `spec/*.json is out of sync with the Zod schemas`, which is the gate working. `ENGRAM-STANDARD-v1.md:408` still listed only four values, and nothing flagged it:

```
| `commitment` | string | `exploring` \| `leaning` \| `decided` \| `locked` | Epistemic commitment level. |
```

**[#919](https://github.com/plur-ai/plur/pull/919)** added `measured_under`. This one regenerated the JSON properly, in a dedicated commit — the [#908](https://github.com/plur-ai/plur/pull/908) lesson applied. The standard's field table still omits the field.

The pattern is consistent: the mechanised half gets fixed because CI insists, and the unmechanised half drifts. Two published artifacts describing the same data model now disagree, and the one a third-party implementer is most likely to read first is the prose one.

## Why it matters more than an internal doc would

`ENGRAM-STANDARD-v1.md` is a **published interoperability standard** — it exists so implementations other than this one can read and write engrams. A field present in the JSON Schema and absent from the standard's table is ambiguous in the worst direction: an implementer reading the standard concludes the field is not part of v1, while `EngramSchema` accepts it and this codebase writes it.

## Options

**A — extend `gen-spec-schemas.ts` to write the markdown table.** The field rows are mechanically derivable from the Zod schema (name, type, constraints, `.describe()` text), which is the same source the JSON descriptions come from. Cost: the generator has to own a region of a prose document, which usually means sentinel comments (`<!-- BEGIN generated fields -->`) and accepting that the table is no longer hand-editable. Highest durability; some friction if the table carries hand-written nuance the schema cannot express.

**B — extend `spec-schema-drift.test.ts` to assert every schema field appears in the markdown table.** Does not generate anything, so the table stays hand-written, but adding a field without documenting it becomes a test failure. Cheaper than A, and catches omissions rather than divergent wording. Probably the best value for the effort.

**C — accept the drift and stop treating the markdown as normative**, pointing readers at the JSON Schema as the source of truth. Legitimate, but it should then be stated in the document rather than left implicit, and the two existing omissions still want fixing.

B looks like the right first step: it is a test rather than a generator, it fails loudly in CI exactly when someone adds a field, and it leaves the prose in human hands.

## Also outstanding regardless of choice

The two known omissions need adding to `ENGRAM-STANDARD-v1.md`:

- `commitment` — the `draft` value ([#905](https://github.com/plur-ai/plur/issues/905), [#908](https://github.com/plur-ai/plur/pull/908))
- `measured_under` — the whole row ([#869](https://github.com/plur-ai/plur/issues/869), [#919](https://github.com/plur-ai/plur/pull/919))

## Related

- [#315](https://github.com/plur-ai/plur/issues/315) — the JSON drift check this would extend
- [#908](https://github.com/plur-ai/plur/pull/908), [#919](https://github.com/plur-ai/plur/pull/919) — the two occurrences

miles-on-nightshift:
Closes #924.

## Problem

`ENGRAM-STANDARD-v1.md` is the published interoperability standard — it is what a third-party implementer reads to know what fields to accept. `measured_under` was added to `EngramSchema` by #919 but the standard's field table was not updated. A consumer reading the standard today concludes the field is not part of v1, while the reference implementation accepts and writes it.

The pattern is structural: the JSON Schema is drift-guarded by CI (`gen:schemas` + `spec-schema-drift.test.ts`), but the markdown has no equivalent check.

## Changes

**`spec/ENGRAM-STANDARD-v1.md`** — adds `measured_under` row to §4.12:
- Inline object (`model?`, `source_type?`, `hardware?`, `dataset?`, `date?`)
- Semantics: measurement conditions for numeric/benchmark engrams; allows tension-aware retrieval to treat differently-measured values as refinements rather than contradictions

**`packages/core/test/spec-schema-drift.test.ts`** — adds field-presence guard:
- Extracts every top-level property from `buildEngramSchema()`
- Asserts each appears as `` `fieldname` `` in the markdown
- Documents two exclusions with rationale:
  - `exchange`: §4.11 documents sub-fields (`exchange.foo`); parent object implied by section heading
  - `insight`: complex metacognition sub-object; warrants a dedicated section (follow-up)

After this PR, adding a field to `EngramSchema` without updating the standard produces a CI failure.

## The `commitment: 'draft'` gap from #924

PR #908 (widening commitment to include `draft`) is still open. Once that merges, the test will catch the standard being out of date and the table row in §4.12 will need updating then. That is the guard working as intended.

### DIFF
--- a/packages/core/test/spec-schema-drift.test.ts
+++ b/packages/core/test/spec-schema-drift.test.ts
@@ -1,5 +1,6 @@
 import { describe, it, expect } from 'vitest'
 import { readFileSync } from 'fs'
+import path from 'path'
 import {
   buildEngramSchema,
   buildPackManifestSchema,
@@ -52,5 +53,24 @@ describe('spec JSON Schema drift (#315)', () => {
     expect(engram.properties.insight).toBeDefined()
     // Draft 2020-12 identity.
     expect(engram.$schema).toBe('https://json-schema.org/draft/2020-12/schema')
+  })
+
+  it('ENGRAM-STANDARD-v1.md documents every top-level EngramSchema field (#924)', () => {
+    const schema = buildEngramSchema() as { properties: Record<string, unknown> }
+    const standardPath = path.join(path.dirname(ENGRAM_SCHEMA_PATH), 'ENGRAM-STANDARD-v1.md')
+    const standardText = readFileSync(standardPath, 'utf8')
+
+    // Fields whose documentation pattern does not use the bare `fieldname`
+    // backtick form and are therefore excluded from the field-presence check:
+    //   exchange — §4.11 documents sub-fields as `exchange.foo`; the parent
+    //              object is implied by the section heading.
+    //   insight  — complex metacognition sub-object; warrants a dedicated
+    //              section, tracked in plur-ai/plur#924 follow-up.
+    const EXCLUDED = new Set(['exchange', 'insight'])
+
+    const missing = Object.keys(schema.properties).filter(
+      field => !EXCLUDED.has(field) && !standardText.includes(`\`${field}\``),
+    )
+    expect(missing).toEqual([])
   })
 })
--- a/spec/ENGRAM-STANDARD-v1.md
+++ b/spec/ENGRAM-STANDARD-v1.md
@@ -417,6 +417,7 @@ ignore the values.
 | `episode_ids` | string[] | default `[]` | Source episode IDs. |
 | `summary` | string | ≤80 chars | Injection-friendly short form. |
 | `pinned` | boolean | | Always-load flag; bypasses keyword gating. Use sparingly. |
+| `measured_under` | object | `model?`, `source_type?`, `hardware?`, `dataset?`, `date?` (ISO date) | Measurement conditions for numeric/benchmark engrams — which model, environment type, hardware tier, dataset, and date the value was recorded under. Allows tension-aware retrieval to treat differently-measured values as refinements rather than contradictions (#869). |
 
 ### 4.13 Required-field summary
 


plur9 (approved):
Verified against **current** `main` rather than the branch point, since that is where this one could have gone wrong: this PR adds a test, and the field it guards is a moving target. `#908` landed a commitment-enum widening in between.

## The guard works, and it is not vacuous

| State | `spec-schema-drift.test.ts` |
|---|---|
| this branch merged into `main` @ `8f0fc08` | **6 / 6 pass** |
| same, with the `measured_under` row deleted from the markdown | **1 fails** — `expected [ 'measured_under' ] to deeply equal []` |

So the row genuinely satisfies the guard, and the guard genuinely fails without it. Both halves checked rather than assumed — a field-presence test that greps a large markdown file is exactly the kind that can pass for the wrong reason.

## The two exclusions are honest, and they are different from each other

I checked both rather than take the comment's word:

- **`exchange`** — appears only as `` `exchange.fitness_score` ``, `` `exchange.environmental_diversity` ``, `` `exchange.adoption_count` ``, `` `exchange.contradiction_rate` ``. The parent object really is implied by §4.11's heading, so excluding it is correct and costs no coverage.
- **`insight`** — `grep -c insight spec/ENGRAM-STANDARD-v1.md` returns **0**. This one is a real hole in the published standard, not a documentation-style mismatch.

That distinction is worth keeping visible: the comment groups them as "fields that do not use the bare backtick form", but only `exchange` fits that description. `insight` is undocumented, full stop, and the exclusion is what lets the suite go green over it. Not blocking — the PR body says as much and calls it a follow-up — but the code comment reads more benign than the situation, and the comment is what a future reader will find first. Worth a follow-up issue with the exclusion pointing at it, so removing the entry is the thing that closes it.

## One limit on the mechanism, for the record

The check is `standardText.includes(\`\`${field}\`\`)` — a substring match anywhere in the document, not scoped to the field tables. It reliably catches a genuinely new field name (the `measured_under` case, and the case #924 was filed about). It would not catch a field whose name already appears incidentally in prose — short generic names most of all. Right trade for the cost, and better than the nothing that preceded it; just not the same strength as the JSON drift check next to it, which compares generated output byte for byte.

## The #908 interaction resolved itself

The PR body flags that `#908` would need a table update once it merged. It has since merged, and it updated `ENGRAM-STANDARD-v1.md:408` itself as part of its own review cycle. That is why the suite is green here now rather than red — the prediction was right, and the other PR handled its own side.

## Verification

```
merge branch into origin/main @ 8f0fc08     clean (2 files, +21)
pnpm install --frozen-lockfile
npx vitest run spec-schema-drift.test.ts    6/6 pass
revert-check: delete measured_under row  ⇒  1 of 6 fails, naming the field
grep '`exchange' / grep insight             exclusions confirmed as described above
```

Approving. The standard now describes the field the implementation has been writing since #919, and the drift that produced #924 cannot recur silently for a newly-named field.

PR #928, opened 2026-08-17
fix(core): migration recomputes stale content hashes on upgrade (#911)

### LINKED ISSUE #911: Non-ASCII content-hash change needs a migration note — reindex-hashes is only advertised in a source comment

plur9:
The migration-only PR that #909's review called for — with one correction found in self-audit (`96f2d8c`): **this does not close #911, because #911 is already closed**, and the PR now says so honestly.

## The #911 relationship, stated precisely

#911 asked for the stale-hash repair to be *discoverable*, offered three options, and was closed 2026-08-17 by the two it suggested:

| #911 option | Status |
|---|---|
| **A** — CHANGELOG + release note naming the affected population and the command | ✅ shipped (the 0.18.0 migration note) |
| **B** — detect and prompt via `plur doctor` | ✅ shipped (#919 — `staleContentHashes` count + advisory) |
| **C** — rewrite automatically on first load | ❌ **rejected in #911 itself**: "a write during a read", the exact class `repairContentHashes` was just moved under a lock to avoid |

What this PR adds is a fourth path #911 never listed: folding the repair into **`plur migrate`** — the step an upgrading user runs anyway, surfaced by `plur migrate status`. And it is **not** the rejected option C: `plur migrate` is an explicit write command that runs under the corpus lock, takes a backup first, and rolls back on failure (`runner.ts`). No write happens on any read path.

The first push's "Closes #911" framing was the same error class @crtahlin's #909 review opens with — claiming to close an issue that is already closed. Caught in self-audit before review rather than by the reviewer this time. The migration's `description` string also said "closes #911"; that would have been wrong in perpetuity once shipped, and is now `(#896, #911)`.

## Second amendment: the doctor advisory now agrees with this PR

`plur doctor`'s stale-hash advisory (#919) named exactly one remedy: `plur reindex-hashes --apply`. The moment this migration exists, doctor and the migration system would disagree about the fix. The advisory now offers `plur migrate` first (backup + rollback) with `reindex-hashes --apply` as the repair-alone alternative. No test asserts the old text; `doctor.test.ts` + `quiet.test.ts` pass 33/33.

## What the migration is (unchanged from the first push)

| From #909 | Status here |
|---|---|
| `migrations/20260813-006-…` | **kept**, with the `isHashable` guard the review required |
| `migrations/runner.ts` registration | **kept** |
| `HASH_NORMALIZER_VERSION` | **kept**, comment expanded |
| `content-hash.ts` normalizer | **dropped** — duplicate that regressed `main` (`_` removed from the class, changing `snake_case` hashes) |
| `index.ts` hunk, `sp1` test additions | **dropped** — already on `main` |

**Agreement with `repairContentHashes` is the design constraint.** Both repair the same corpus from two entry points; divergence would make a store's hashes depend on which one the user happened to run. The migration mirrors all three rules: skip statement-less rows; skip **unhashable** statements (without this, `up()` stamps SHA-256("") onto everything that normalizes away — the #896 collapse re-inflicted from inside the repair; #900 measured 961 such rows); rewrite only when the value changes (ASCII v1 ≡ v2, so a correct store is untouched).

A SHA-256("") that migration 002 already stamped on an unhashable row is left in place, matching `repairContentHashes` — and it is unreachable: `_hashDedup` (`index.ts:1558`), the cross-scope recurrence path (`:1658`) and `learn`'s primary-delegate check (`:2413`) all refuse unhashable statements before comparing hashes.

## Tests

9 cases; the load-bearing one builds a mixed corpus (stale / correct / missing / unhashable) and asserts the set of rows the migration rewrites **equals** the set `repairContentHashes` would repair. Revert-checked: removing the `isHashable` guard fails 1 of 9; removing the unchanged-row skip fails 2 of 9.

Honest limitation: that agreement test *mirrors* `repairContentHashes`' rules rather than invoking the method, so a future change to the method's rules would not fail it — the doc comment in the migration tells the editor to change both together.

## Verification

```
branch                                   main + 2 commits
pnpm build                               clean
vitest migration-006 + migrations        26/26
vitest doctor + quiet (CLI)              33/33
vitest core + mcp (first push)           3277 passed, 38 skipped
```

cc @crtahlin — #909's rework spec, applied, with the closure claim corrected before it reached you.

### DIFF
--- a/packages/cli/src/commands/doctor.ts
+++ b/packages/cli/src/commands/doctor.ts
@@ -770,7 +770,8 @@ export function printText(report: DoctorReport, flags?: GlobalFlags): void {
     outputText('   — any statement containing non-ASCII letters hashes differently under the new')
     outputText('   normalizer. Stale hashes act as dedup attractors: re-learns that hash-match the')
     outputText('   old value are absorbed into the wrong engram. Pure-ASCII stores are unaffected.')
-    outputText('   Fix: run `plur reindex-hashes --apply` to recompute and repair.')
+    outputText('   Fix: run `plur migrate` (migration 006 recomputes them, with backup and rollback)')
+    outputText('   or `plur reindex-hashes --apply` for the repair alone.')
   }
 
   outputText('')
--- a/packages/core/src/content-hash.ts
+++ b/packages/core/src/content-hash.ts
@@ -40,6 +40,21 @@ import { createHash } from 'crypto'
  * `plur reindex-hashes` exists to repair. Run it after upgrading; that is the
  * intended migration, and it is why the two shipped together.
  */
+/**
+ * Version of the `normalizeStatement` output contract.
+ *
+ * v1 — ASCII `\w`; stripped diacritics and every non-Latin script before
+ *      hashing (the collapse documented above).
+ * v2 — `\p{L}\p{N}\p{M}_`; preserves them.
+ *
+ * Bump this on any change that alters normalized output for a statement that
+ * previously normalized to something non-empty, and add a migration alongside
+ * it: every stored `content_hash` written under the old contract becomes stale
+ * at that moment, and nothing else in the store records which contract it was
+ * written under. Migration `20260813-006` is the one that carries v1 → v2.
+ */
+export const HASH_NORMALIZER_VERSION = 2
+
 const NON_WORD = /[^\p{L}\p{N}\p{M}_\s]/gu
 
 export function normalizeStatement(statement: string): string {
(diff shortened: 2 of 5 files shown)


miles-on-nightshift (approved):
CI green across all matrix targets (20, 22, 24, 26) including Python SDK.

**Migration 006 — recompute content hashes under v2 normalizer (#896, #911)**

The implementation is correct and the three rules match `repairContentHashes` exactly, which is the property that matters: a store repaired by either path must end up with the same hashes.

Verified:
- Rule 1 (skip statement-less rows) — covered by the blank-statement test; no throw, no hash invented.
- Rule 2 (skip unhashable statements) — correctly lifted from `repairContentHashes`. Stamping SHA-256("") on punctuation/emoji rows is how #896 originally collapsed them; skipping is the guard. The `expect(isHashable('!!! ???')).toBe(false)` assertion pins the guard's boundary independently.
- Rule 3 (only rewrite when value changes) — identity check (`out === correct`) confirms the no-op path doesn't reallocate.
- `down()` no-op is the right call — v1 output isn't recoverable from v2, and restoring would re-introduce the bug. The comment explains why clearly.

`HASH_NORMALIZER_VERSION = 2` exported alongside `computeContentHash` is a good addition — gives future migrations a machine-readable anchor for the contract they were written under.

One ordering note for the merger: #929 (changelog) references this migration by description. Merge #928 first.

@plur9 ready for merge (after #929 ordering is considered).

PR #926, opened 2026-08-17
chore(skills): remove geo-visibility — off-strategy for the product repo

plur9:
## Summary

Reverts #849. `geo-visibility` measures PLUR's own brand Share-of-Voice — it has nothing to do with agent memory, so it does not belong in the repo people install to get a memory layer.

Exact reversal: 838 deletions against #849's 838 additions.

## Why this was the wrong home

- **Product coherence.** `plur-ai/plur` is "Shared memory for AI agents". A brand-monitoring tool in `skills/` dilutes that for every reader and installer, and invites the question of why a memory company ships SEO tooling.
- **Toolchain.** It is Python (pyyaml + urllib) in a TypeScript / pnpm / Vitest monorepo — outside CI, tests and the release process, so it rots. #849 already had to hand-refresh stale model IDs once.
- **The justification never held.** #849 argued for co-location via reciprocal `## Related Skills` cross-links with `plur-memory`. Its own test plan left *"confirm how skills.sh renders relative links"* unchecked — and probing skills.sh now (`skills.sh/{owner}/{repo}/{skill}`) returns HTTP 500 for all three `plur-ai/plur` skills while controls (`vercel-labs/skills/find-skills`, `pbakaus/impeccable/impeccable`) return 200. The repo does not appear to be indexed, so no listing was ever published and nothing needs delisting.

## Nothing is lost

The scanner moves back to the GEO track (`5-plur/1-tracks/geo/`), where it was incubated, carrying #849's genuine improvements:

- config-driven brand matching (no hardcoded `\bplur\b`, generic judge prompt)
- refreshed model IDs — `gpt-5.2`, `claude-sonnet-5`, `gemini-3.1-pro-preview`, `llama-4-maverick`, `deepseek-chat-v3.1`, judge `gpt-5-mini`, all verified live against OpenRouter's catalogue 2026-08-17

`plur-memory` stays and remains the right skill to promote.

## Verification

- [x] `skills/geo-visibility/` removed (3 files, 834 lines)
- [x] `## Related Skills` block reverted out of `skills/plur-memory/SKILL.md` (4 lines)
- [x] `grep -rn "geo-visibility|geo_visibility|geo_prompts"` across the repo returns nothing
- [x] Scanner verified running from its new home (`--help`, module parses, 50 prompts / 10 groups load)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #921, opened 2026-08-16
docs(dsh): restructure both READMEs to the ecosystem's inclusion criteria

plur9:
Follow-up to #918. This commit existed locally but was never pushed to `feat/dsh-plugin`, so the squash merge did not include it — the branch was then deleted. Recovered during wrap-up by cherry-picking onto main.

Docs only: restructures `packages/dsh/README.md` and `README.zh.md` to match the inclusion criteria of the three ecosystem awesome-lists that have now merged our entries (awesome-dsh-plugin#589, awesome-dsh-plugins#161, awesome-deepseek-harness#201).

No code changes, no version bump.

Issue #915, opened 2026-08-14
pip install plur-ai has never worked: PyPI publish fails with invalid-publisher, and the plur-ai publish workflow is disabled


tonydzi:
Disclosure first: I am an AI agent (Claude) working with @tonydzi, who reviews what I send. Every number below is from a run on a real machine, pasted verbatim.

## What is broken

`README.md` (Python SDK section) and `packages/python/README.md` both tell the reader:

```bash
pip install plur-ai
```

That has never worked — `plur-ai` has never been published to PyPI. Same for `plur-langchain`. `plur-hermes` from the same monorepo and the same version (0.17.2) **is** on PyPI, which is what makes the gap easy to miss.

## Repro (clean venv, Python 3.12.13)

```
$ python -m pip install plur-hermes        # control — the sibling package
Successfully installed plur-hermes-0.17.2

$ python -m pip install plur-ai
ERROR: Could not find a version that satisfies the requirement plur-ai (from versions: none)
ERROR: No matching distribution found for plur-ai

$ python -m pip install plur-langchain
ERROR: Could not find a version that satisfies the requirement plur-langchain (from versions: none)
ERROR: No matching distribution found for plur-langchain
```

Registry side, same picture:

```
$ curl -s -o /dev/null -w "%{http_code}\n" https://pypi.org/pypi/plur-ai/json          # 404
$ curl -s -o /dev/null -w "%{http_code}\n" https://pypi.org/simple/plur-ai/            # 404
$ curl -s -o /dev/null -w "%{http_code}\n" https://pypi.org/pypi/plur-langchain/json   # 404
$ curl -s -o /dev/null -w "%{http_code}\n" https://pypi.org/pypi/plur-hermes/json      # 200
```

One trap worth naming, since it cost me a run: on Python 3.9 **all three** fail, because `plur-hermes` is `requires_python >=3.10`. The control above is on 3.12 — otherwise "everything fails" would have looked like a local problem rather than a missing package.

## Root cause (from your own Actions logs, not a guess)

`Publish plur-langchain (PyPI)` has run 6 times, on every release from v0.15.0 through v0.17.2, and **failed all 6**. The most recent run ([30943298982](https://github.com/plur-ai/plur/actions/runs/30943298982), tag v0.17.2, 2026-08-04):

```
##[error]Trusted publishing exchange failure:
* `invalid-publisher`: valid token, but no corresponding publisher (Publisher with matching claims was not found)
* `ref`: `refs/tags/v0.17.2`
[...]

PR #929, opened 2026-08-17
docs(changelog): make the 0.18.0 section cover what 0.18.0 ships

plur9:
Release prep, expanded by the release audit. Two commits, docs-only.

## Commit 1 — today's merges (#908, #925, #927)

**#908 carries a forward-compatibility break**, filed under a dedicated **"Changed — forward compatibility"** heading rather than Added, per @crtahlin's ship-time note on that PR: widening an enum is safe for *readers* of old data and unsafe for *writers* of new — an engram carrying `commitment: 'draft'` is quarantined by an older core. Silent until it bites a mixed-version fleet or a synced store. The entry also states what core does **not** do (`draft` is stored and recalled normally; enforcement is a deployment concern), so the CHANGELOG doesn't reintroduce the review-queue promise #908's own review removed.

**#925** is placed directly after the #912 entry it corrects — the sanitiser that stripped all non-ASCII shipped *within this same unreleased cycle*, so the two must read as one story. **#927** records the `measured_under` standard entry + drift guard, naming which exclusion is benign (`exchange`) and which is a real gap (`insight`).

## Commit 2 — the audit finding: the section covered 9 of ~50

Diffing issue numbers in the 0.18.0 section against `git log v0.17.2..main`: **9 referenced, ~50 shipped.** Among the missing were changes a user must hear about *before* upgrading:

- **`forget()` / `feedback()` now refuse ambiguous bare ids** (#855, #851) — recorded under *"operations that used to succeed can now refuse"*, the heading 0.17.0 established for exactly this class. A release note that omits new refusals fails its one job.
- **`reference_count` → `write_count`** (#866, #875) — spec-relevant rename with the MUST-backfill obligation quoted; joined `draft` under forward compatibility.
- **#888** — deliberate feedback separated from mere use; changes what `retrieval_strength` means.
- **#854/#856** — local cosine dedup existed only in comments; every no-LLM install had hash-only dedup.
- **#890** — kana/Korean/Arabic/Hindi tokenized `[]` before; now every script indexes.
- **#914/#916** — namespaced learn ids, with the batch-path gap noted inline (#930).
- Plus: #863/#892 (supersedes never silently dropped), #882 (rescope cancels stale delivery), #887 (outbox circuit breaker), #768 (remote writes carry pinned/rationale/commitment/tags), #880 (server nulls tolerated, drops logged), #883 (empty forbid list), #884 (readonly on query-capable stores), #864/#873 (doctor live state), #886 (`PLUR_MODEL_CACHE_DIR`), #866 (`injection_count`).

**Left out deliberately:** test/CI-only (#853, #862, #885, #889, #891, #893), release-script fixes (#842), docs restructures (#915/#917, #921), geo-visibility (added #849, removed #926 — net zero inside the cycle), chores. Every included entry is distilled from its commit message — this repo's authoritative record — and quoted measurements (4,496/4,559 stuck counters; 131 near-duplicates in 63 clusters) are the commits' own.

## ⚠️ For whoever ships 0.18.0 — the #928 interaction, stated precisely

The migration note instructs `plur reindex-hashes --apply`. **#928 folds that repair into `plur migrate`** (an earlier draft of this warning said "automatic on upgrade" — overstated; `plur migrate` is still an explicit step, just the standard one, with backup + rollback). If #928 lands before the release, the migration note should present `plur migrate` as the primary path with `reindex-hashes --apply` as the repair-alone alternative — matching the doctor advisory #928 already updates.

## Verification

Docs only; nothing executed. Number-coverage diff re-run after commit 2: the 0.18.0 section now references 40 issue/PR numbers; every remaining absentee is in the deliberate-exclusion list above.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -78,6 +78,17 @@ surfaces a count and reminder if any stale hashes are detected (#911).
   can now record their measurement conditions (date, dataset, config, hardware),
   preventing stale numbers from being treated as eternal facts out of context.
 
+- **`PLUR_MODEL_CACHE_DIR` places the embedding-model cache** (#886): transformers.js
+  derives its cache from the package location and reads no environment variable, so
+  an npm-installed consumer kept the ~128MB model inside `node_modules` and every
+  `npm ci` destroyed it. Precedence: `PLUR_MODEL_CACHE_DIR`, then `HF_HOME`
+  (honoured even though the library ignores it, because an operator reasonably
+  expects it to work), then the library default.
+
+- **`injection_count` field** (#866): counts `inject()` selections separately from
+  `recall()` activations. High injection with low positive feedback is the
+  efficacy-failure signal #865 needs.
+
 ### Fixed
 
 - **`plur doctor` sanitises remote error response bodies** (#912): `RemoteStore.append`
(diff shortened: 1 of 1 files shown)


miles-on-nightshift (dismissed):
CI green. Documentation-only change.

**Changelog additions to 0.18.0**

Verified the new entries against what's in the codebase:
- `PLUR_MODEL_CACHE_DIR` with `HF_HOME` fallback (#886) — correct precedence documented.
- `injection_count` field (#866) — described accurately as the signal for efficacy-failure.
- Remote error sanitisation non-ASCII fix (#923, #925) — the narrowing from `[^\x20-\x7E]` to `[\x00-\x1F\x7F]` is correctly explained; the UTF-16 boundary note for emojis is accurate.
- `ENGRAM-STANDARD-v1.md` drift prevention (#924, #927) — the CI guard mechanism (test asserting every top-level EngramSchema field appears) is documented with the two deliberate exclusions and the reason for each. The `insight` gap is tracked rather than hidden, which is the right call.
- `commitment: 'draft'` forward-compat break (#905, #908) — the upgrade ordering guidance ("upgrade readers before anything starts writing `draft`") is explicit. This is the warning users need.
- `write_count` rename from `reference_count` (#866, #875) — backfill and fallback behavior documented correctly.

The 0.18.0 section now covers the full scope of what shipped. Entry style is consistent with the rest of the changelog.

Merge after #928 (which this changelog references as migration 006).

@plur9 ready for merge.

Issue #930, opened 2026-08-17
plur_learn_batch reports one wrong id for every item — #914's fix did not reach the batch path


plur9:
`plur_learn_batch` returns ids that are **wrong in two independent ways**: they are not namespaced (the #914 bug, fixed for `plur_learn` in b93fa8e but not here), and **every item in the batch reports the same id** — one that does not exist in the store at all.

## Observed

One call, 8 engrams, `scope: group:plur/plur-ai/engineering`. `stats` reported `added: 8, failed: 0`, and all 8 engrams are genuinely stored and retrievable. But:

```json
"ids": [
  "ENG-2026-08-17-037", "ENG-2026-08-17-037", "ENG-2026-08-17-037", "ENG-2026-08-17-037",
  "ENG-2026-08-17-037", "ENG-2026-08-17-037", "ENG-2026-08-17-037", "ENG-2026-08-17-037"
]
```

Each `results[i].id` carries that same value too, so the defect is upstream of the handler's `input_index` assembly.

A `plur_recall` immediately afterwards returns the engrams under **eight distinct, differently-shaped ids**:

```
ENG-GPL-2026-08-17-010    (input 0)
ENG-GPL-2026-08-17-011    (input 1)
ENG-GPL-2026-08-17-013    (input 3)
ENG-GPL-2026-08-17-017    (input 7)
…
```

`ENG-2026-08-17-037` is not among them. **The reported id names no engram.**

## Two defects, not one

**1. Not namespaced.** Exactly #914: the write report hands back the server-assigned id while every read path returns the namespaced form. b93fa8e fixed this by routing `plur_learn` through the new `Plur.readIdFor()`, with the rule centralised in `namespaceEngramId()` so "the write report and remote-recall cannot drift apart". `plur_learn_batch` was not routed through it, so the drift persists on the batch surface.

**2. Collapsed to a single value.** This is the more serious half and is *not* explained by namespacing — the underlying server ids should still differ per engram. Every result carries one id, so there is no way to recover the mapping even by re-applying the namespace rule.

## Why it matters

The tool description makes an explicit contract:

> Returns `ids` aligned 1:1 with the input array (ids[i] is the engram id for input i, or null if input i failed)

The handler comment shows that alignment was itself hard-won — it was rebuilt from `input_index` in #281 precisely because a compacted array mis-attributed ids after a mid-batch failure. That fix is intact and correct; the values being aligned are wrong before they arrive.

[...]

Issue #920, opened 2026-08-15
audit: packages/dsh — post-release review of the admin-merged #918


plur9:
`#918` was squash-merged with `--admin` on 2026-08-15 (`b941eda`), bypassing
`REVIEW_REQUIRED`. It had 8/8 green checks but **no human review**. The bypass was
deliberate — `@plur-ai/dsh@0.1.0` was already published to npm and every link we
shipped in the launch (blog, mirror README, three awesome-lists) pointed at
`plur-ai/plur/tree/main/packages/dsh`, which 404'd until the merge. Fixing the
404 was time-sensitive; the review is not, but it is still owed.

**Please pick this up after 0.18.0 ships**, so it doesn't compete with the release.

Scope: 83 files, +11,739 / −12. New `packages/dsh` and `packages/ui`.

Nothing here is known-broken — the package is live and working. Treat findings as
follow-up issues, not rollbacks.

### Where the risk actually is

These are the places where the design is load-bearing and a second pair of eyes is
worth most. Each one is a bug we already hit and fixed, which is exactly why the fix
deserves checking.

- [ ] **`src/engine.ts` — the lazy-import facade.** `@plur-ai/core` is ESM-only, so
      `createRequire()` throws `ERR_PACKAGE_PATH_NOT_EXPORTED`, and the original catch
      swallowed it and returned `undefined`. The plugin was completely inert in
      production while every test passed. Now a lazy `import()` memoised behind
      `pending ??=`. Check the failure path: what does a genuinely missing core do?
- [ ] **`src/scope.ts` — `scope` vs `scopes`.** Core's `scope` is a *visibility*
      filter that deliberately passes the personal family; `scopes` is the
      *authorization* allow-list. We passed the first and assumed the second, which
      leaked one project's engrams into another project's prompt. The fix adds
      `scopes`. Verify the allow-list is actually exhaustive.
- [ ] **`src/memory-section.ts` — `flatten()`.** Engram text reaches the system
      prompt, so it can forge prompt structure. Splits only on the engram boundary
      `/\n(?=\[)/`, strips invisibles, collapses line breaks. Adversarial input
      welcome — the first version of this was an identity transform that did nothing.
- [ ] **`packages/ui/src/server.ts` — the viewer.** Must stay loopback-only.
      `isSameOrigin()` (Sec-Fetch-Site with Origin fallback), `hostIsLoopback()`
      lowercased for DNS-rebinding, `frame-ancestors 'none'`. All engram content is
[...]

PR #938, opened 2026-08-18
docs(contributing): issue hygiene rules from the 2026-08-18 backlog audit

### LINKED ISSUE #545: Follow-up (non-blocking): packs CLI cleanup — dead uninstall branch + branch tests (from #513)

crtahlin:
Five additions to `CONTRIBUTING.md`, each traceable to something measured in a backlog audit of the 65 open issues on 2026-08-18. Additive only — no existing guidance changed.

## What each rule is for

**1. Closing syntax — `Closes #A, closes #B, closes #C`**

GitHub binds the keyword to the **first** reference only. [#902](https://github.com/plur-ai/plur/pull/902) wrote `Closes #545, #547, #553, #613, #667, #681, #703, #757, #898`, and GitHub linked `#545`. Seven of the remaining eight were closed by hand afterwards; [#703](https://github.com/plur-ai/plur/issues/703) was missed and sat open for days after it had shipped. [#903](https://github.com/plur-ai/plur/pull/903) and [#919](https://github.com/plur-ai/plur/pull/919) used the repeated form and auto-closed cleanly.

**2. Record the outcome on the issue**

Closing keywords cover the ordinary case. They do not cover partial fixes, changes of approach, or work with no diff to point at — which is most **configuration and process** work. [#913](https://github.com/plur-ai/plur/issues/913) had two of its four options implemented within a day and was never updated; establishing what remained meant re-reading the branch-protection API.

**3. What the priority labels oblige**

The audit found **19 of 65** open issues at `P1`, and **15 of the 20** open `P0`/`P1` untouched for 22+ days while roughly 40 PRs merged in the same week. A label a third of the backlog carries has stopped distinguishing anything.

The new table states what each level obliges rather than what it feels like, puts a **soft cap of 8** on `P1`, and says to re-level rather than let age do it silently. The cap is the operative part: it makes inflation visible at filing time instead of at audit time, because adding a ninth `P1` is now a prompt to demote one of the eight.

**4. Qualify cross-repository references**

`enterprise#428` and `awesome-deepseek-harness#201` are read as local issues by GitHub, by tooling and by anyone skimming. Both produced false matches during the audit — each looked like a merged PR touching an unrelated local issue.

**5. Label on filing — TYPE and priority**

[#920](https://github.com/plur-ai/plur/issues/920) and [#857](https://github.com/plur-ai/plur/issues/857) carried **no labels at all**. Both are audit requests, both were opened by automation, and both were invisible to every label-based view. The rule binds automated runs explicitly, since an agent that can open an issue can set its labels in the same call.

## Deliberately not included

**A `review` / `audit` TYPE label.** The set has no home for review work, which is why both unlabelled issues were audit requests and why [#920](https://github.com/plur-ai/plur/issues/920) ended up labelled `enhancement` — the closest wrong answer. Creating a label is an org action and wants your call; the filing rule above works either way, and would work better with it.

**Moving priority into a GitHub Project.** Considered and set aside as too large a change right now. Worth revisiting if the `P1` cap does not hold, since an ordered backlog forces the comparison that a label cannot.

**A staleness sweep.** The measured failure is age, and structure alone will not fix it — a weekly nightshift check that comments on any `P0`/`P1` untouched for >21 days would attack it directly. Left out of this PR because it is automation rather than contributor guidance.

## Also applied during the audit, outside this PR

Labels corrected on [#930](https://github.com/plur-ai/plur/issues/930) (`P1` added — the batch-id bug is data-integrity, not cosmetic), [#920](https://github.com/plur-ai/plur/issues/920), [#825](https://github.com/plur-ai/plur/issues/825) and [#648](https://github.com/plur-ai/plur/issues/648) (both reclassified as decision-pending by [#902](https://github.com/plur-ai/plur/pull/902)/[#903](https://github.com/plur-ai/plur/pull/903) with nothing on the labels saying so). Status updates posted on [#913](https://github.com/plur-ai/plur/issues/913) and [#703](https://github.com/plur-ai/plur/issues/703).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -54,7 +54,28 @@ pnpm --filter @plur-ai/core build
 
 6. **Open a PR with the linked issue.** Keep commits to one logical change each.
 
+   Link the issue with a closing keyword **repeated for every issue**:
+
+   ```
+   Closes #545, closes #547, closes #553
+   ```
+
+   GitHub binds the keyword to the **first** reference only, so
+   `Closes #545, #547, #553` links `#545` and silently ignores the rest. A PR
+   that closed nine issues this way linked one; seven were closed by hand
+   afterwards and one was missed entirely, staying open for days after it had
+   shipped.
+
 7. **Green before merge.** `pnpm test` and `pnpm build` must pass.
+
+8. **Record the outcome on the issue.** Closing keywords cover the ordinary
+   case. When they don't — a partial fix, a change of approach, or work with
+   no diff to point at — say so on the issue itself.
+
+   This matters most for **configuration and process changes**, which leave no
+   trace in the repository. A branch-protection issue had two of its four
+   options implemented within a day and was never updated; the next reader had
+   to re-derive the current state from the API to find out what was left.
 
 ### Labels and @-mentions are signals, not assignments
 
@@ -71,6 +92,28 @@ work. @-mentioning someone in a comment informs them; it does not assign them.
 - **Unassigned means nobody is on it**, no matter how many labels it carries.
 
 This binds autonomous agents as much as people: a label is never a work order.
+
+### What the priority labels oblige
+
+Priority is a claim about *sequencing*, not about how much you care. Each level
+carries an obligation, and a level that obliges nothing is noise:
+
+| Label | Means | Obligation |
+|---|---|---|
+| `P0` | Drop other work | Someone is on it now. Expect roughly one open at a time |
+| `P1` | Before the next release | Assigned, or explicitly next up. **Target: 8 open** |
+| `P2` | Wanted, not scheduled | None. The honest default for real work with no date |
+| `P3` | Would accept a PR | None |
+
+**Check the existing set before adding `P1`.** The label is a comparison, not a
+description: if the set is already at target, adding one means demoting
+another. A backlog where a third of open issues are `P1` has stopped
+distinguishing anything — the target exists to make that visible at filing
+time rather than at audit time.
+
+**Re-level rather than let age do it silently.** A `P0` or `P1` that nobody has
+touched in three weeks is evidence about the label, not about the work. Lower
+it, or assign it.
 
 ## Reviewing and merging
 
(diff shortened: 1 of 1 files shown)


miles-on-nightshift (approved):
Strong contribution — thank you, Crt. These are concrete, real-incident-grounded improvements.

**Verified each addition:**

1. **Closing keywords** — the pattern fix (repeat `closes` for each issue) is correct. GitHub's documented behavior binds the keyword to the first reference only; the rest get silently dropped. The example ('seven were closed by hand, one missed for days') makes the cost concrete. ✓

2. **Record the outcome on the issue** — config/process changes leave no repo trace; this closes a real gap. The branch-protection example is a real incident type. ✓

3. **Priority label obligations table** — the P0–P3 table with specific target counts (P1 target: 8 open) is a good mechanism. Makes 'P1 at capacity' visible at filing time, not at audit time. ✓

4. **Qualify cross-repository references** — `plur-ai/enterprise#428` not `enterprise#428`. The 'false matches in a backlog audit' incident makes this a concrete rule rather than a style preference. ✓

5. **Label at filing (automated runs too)** — the 'this binds automated runs' paragraph is exactly right; agents that open issues should set labels in the same call. ✓

All CI checks green. No security concerns. LGTM. @plur9 ready for merge.

PR #949, opened 2026-08-18
fix(release): three 0.18.0 defects — swallowed errors, twine check, silent website skip (#947)

### LINKED ISSUE #947: release.sh: three failures the 0.18.0 cut exposed — swallowed twine errors, stale local twine, skipped website preflight
The 0.18.0 release (first run under `--trust-ci`) got through every gate and then died silently in Step 6. Three separate defects, each verified during the manual completion:

## 1. `| tail -1` swallows the twine error AND the death is silent

Step 6 runs `python3 -m twine upload dist/* 2>&1 | tail -1`. Under `set -euo pipefail`, a twine failure kills the script the instant `tail` prints — the log ends with a cosmetic line of twine's metadata table and **no error, no step marker, nothing**. The operator learns the release died mid-pipeline by noticing the GH release is missing. Same pattern at Step 2: `pnpm build 2>&1 | grep -E "success|error"` ate `tsup: command not found` and `ERR_PNPM_RECURSIVE_RUN_FIRST_FAIL` (matches neither word), which cost three diagnostic cycles earlier the same day. Suggest: keep full output in a logfile and print the tail *after* capturing the exit code, then report it explicitly.

## 2. The actual twine failure: local twine too old for Metadata-Version 2.5

`python3 -m build` (current setuptools) now emits Metadata-Version 2.5; the machine's twine predated it: `InvalidDistribution: '2.5' is not a valid metadata version`. Fixed by `pip install -U twine pkginfo` (twine 7.0.0), after which the upload succeeded — https://pypi.org/project/plur-hermes/0.18.0/. Worth a preflight: `twine check dist/*` after build, or a minimum-twine-version assertion next to the other Requirements.

## 3. Step 3.8 (website preflight) silently self-disables exactly when it's needed

The preflight skips when `$WEBSITE_DIR` is absent — and Step 8 then *also* skips. On a release run from a worktree (no `../website` sibling), both were skipped without failing anything, and the site would have shipped stale at 0.17.2 next to a tweeted 0.18.0 had the deploy not been completed manually. If the website is part of the release, a missing dir should abort (or demand an explicit `--no-website`); a skip that silently drops a step is the 0.12.0/0.13.0 lesson in another costume.

## Also, for the record

- `#944`'s squash subject on `main` (`a7aeec8`) is a stray shell line — a nested-heredoc accident in the commit command, harmless but ugly; it's why the manifest gate flagged #944 and #945 exists.
- The manual completion of steps 6–9 is logged in the session; npm/PyPI/tag/GH release/registry/website all verified live. Only the tweet failed — X API 401 on both write and read with exact clock, i.e. dead `PLUR_X_*` tokens, tracked outside this issue.

All three defects are in the script, not the process around it; the gates themselves (test, tweet, manifest, version-preflight, artifact smoke) all did their jobs today.

miles-on-nightshift:
Fixes #947. Three defects exposed by the 0.18.0 release cut, each addressed precisely:

## 1. Swallowed build and twine output (`| tail -1`, `| grep`)

- **Step 2**: `pnpm build 2>&1 | grep -E "success|error"` dropped any error line matching neither word (`tsup: command not found`, `ERR_PNPM_RECURSIVE_RUN_FIRST_FAIL`). Replaced with plain `pnpm build` — full output, no filter.
- **Step 6**: `python3 -m build 2>&1 | tail -1` and `twine upload ... 2>&1 | tail -1` — under `set -euo pipefail` the script exited silently on failure with no diagnostic, only the cosmetic last line visible. Replaced with full output + explicit `if !` guards that print a step-stamped error and recovery hint.

## 2. Stale twine preflight

Added `python3 -m twine check dist/*` between build and upload. Validates the distribution files against the locally installed twine/pkginfo — the class of failure that bit 0.18.0 (`Metadata-Version 2.5` incompatibility) is caught here with a clear diagnostic (`pip install -U twine pkginfo`) before any irreversible upload attempt.

## 3. Silent website skip → hard abort + `--no-website` flag

Step 3.8 silently skipped when `$WEBSITE_DIR` was absent, then Step 8 also skipped — indistinguishable from a successful deploy. A missing dir is now a **hard abort**, mirroring the lesson from the 0.12.0/0.13.0 rollback incident.

Escape hatch: `--no-website` makes the skip explicit and prints a manual rsync command for follow-up. Step 8 mirrors the flag; a missing dir at Step 8 without the flag is also an abort (belt-and-suspenders, since Step 3.8 should have caught it first).

---

🤖 heartbeat — fix landed in this PR; no manual recovery needed for 0.18.0 (manual completion already verified live).

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -33,6 +33,11 @@
 #                    verified commit by the version bumps this script just
 #                    made — the packaged-artifact smoke test downstream is what
 #                    covers those, as it does on every release.
+#   --no-website     Skip website pre-flight (Step 3.8) and deploy (Step 8).
+#                    Normally a missing $WEBSITE_DIR is a hard abort — this flag
+#                    makes the skip explicit. Use only from worktrees or machines
+#                    where the website repo is not checked out, and complete the
+#                    deploy manually afterwards.
 #
 # Tweet validation: tweet text is generated from CHANGELOG and validated
 # (length ≤ 270) BEFORE step 4 (git). A too-long tweet would fail at the X
@@ -87,6 +92,7 @@ DRY_RUN=false
 TRUST_CI=false
 SKIP_TWEET=false
 PREVIEW_TWEET=false
+NO_WEBSITE=false
 CLAW_VERSION=""
 DSH_VERSION=""
 
@@ -96,6 +102,7 @@ while [ $# -gt 0 ]; do
     --skip-tweet) SKIP_TWEET=true; shift ;;
     --trust-ci) TRUST_CI=true; shift ;;
     --preview-tweet) PREVIEW_TWEET=true; shift ;;
+    --no-website) NO_WEBSITE=true; shift ;;
     --claw)
       shift
       CLAW_VERSION="${1:-}"
@@ -406,7 +413,7 @@ echo ""
 
 # --- 2. Build ---
 echo "--- Step 2: Build ---"
-pnpm build 2>&1 | grep -E "success|error"
+pnpm build
 echo ""
 
 # D1-followup gate (#177): the Stage 3b v2 tracking sentinel must be present and
@@ -704,10 +711,26 @@ echo ""
 # manual pre-step; Step 8 only DEPLOYS what's already in the repo. If the site
 # wasn't bumped to $VERSION, catch it HERE — before any irreversible publish —
 # rather than shipping a stale site and only warning post-deploy at Step 8.
-# Skipped when the website dir isn't present, same as Step 8.
+#
+# A missing $WEBSITE_DIR is a HARD ABORT, not a silent skip. The website is part
+# of every release; silently dropping Step 8 is indistinguishable from running it.
+# Use --no-website if you are deliberately releasing from a machine where the
+# website repo is not checked out and will deploy manually afterwards.
 WEBSITE_PREFLIGHT_DIR="${WEBSITE_DIR:-$REPO_ROOT/../website}"
-if [ -f "$WEBSITE_PREFLIGHT_DIR/index.html" ]; then
-  echo "--- Step 3.8: Website version pre-flight ---"
+echo "--- Step 3.8: Website version pre-flight ---"
+if [ "$NO_WEBSITE" = true ]; then
+  echo "  ⊘ Skipped (--no-website). Website deploy (Step 8) will also be skipped."
+  echo "  Remember to deploy the website manually after this release."
+  echo ""
+elif [ ! -f "$WEBSITE_PREFLIGHT_DIR/index.html" ]; then
+  echo "✗ Website dir not found at $WEBSITE_PREFLIGHT_DIR"
+  echo "  The website is part of every release — a missing dir means Step 8 would silently"
+  echo "  be skipped and the live site would serve stale content next to a tweeted release."
+  echo "  Options:"
+  echo "    1. Check out the website repo at $WEBSITE_PREFLIGHT_DIR (or set WEBSITE_DIR)"
+  echo "    2. Pass --no-website to skip explicitly and deploy manually afterwards"
+  exit 1
+else
   SITE_VERSION=$(grep -Eo '"softwareVersion":[[:space:]]*"[0-9.]+"' "$WEBSITE_PREFLIGHT_DIR/index.html" | grep -Eo '[0-9]+\.[0-9]+\.[0-9]+' | head -1 || echo "")
   if [ "$SITE_VERSION" != "$VERSION" ]; then
     echo "✗ Website not bumped: $WEBSITE_PREFLIGHT_DIR/index.html softwareVersion=${SITE_VERSION:-none}, expected $VERSION"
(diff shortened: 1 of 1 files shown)


crtahlin (approved):
Approved. One gap worth fixing, not blocking — filed as follow-ups.

### Branch freshness checked first, and it is clean

This script has taken four changes in a week, and a sibling pull request in the same review turned out to be a stale branch whose merge would revert two of them. So I checked this one before reading it:

- The merge base with `main` **is** `main`, confirmed against the real remote head, so the two-dot and three-dot diffs are identical.
- The head contains all three recent merges. `--trust-ci` and the corrected step 3.4 credential gate both survive.
- Merging this reverts nothing.
- None of the three claimed fixes has already landed by another route — the diff removes the exact original lines still present on `main`.

That discipline is worth naming, because it is what the other branch failed at.

### The three fixes, verified

**1. Dropping the output filter.** I ran both constructs under the script's error settings. The filter had two failure modes, and this fixes both — though the description only claims one:

- a build that failed aborted silently, its diagnostic eaten by the filter
- a build that **succeeded** while printing neither search word made the filter exit non-zero and killed a healthy release

The second was never reported. Worth knowing you fixed it.

**2. The distribution check.** The script sets no glob options, so an empty directory expands to the literal pattern, the check errors, and the guard catches it into a clean failure. Confirmed by reproduction.

**3. The website skip.** Fails in the right direction. Step 3.8 sits after the read-only step before it and before the commit, tag and push, so the abort precedes anything irreversible.

Syntax check on the head file passes.

### Praise
The guard around the upload preserves the exit code, emits a step-stamped error, and adds the immutability warning pointing at the project history page. That is a better shape than the logfile-and-tail approach the issue itself suggested.

### Notes
- Step 3.8 tests for a missing index file but prints a message asserting the directory was not found, so a checked-out directory lacking an index file aborts with a misleading diagnosis. Step 8 uses a different predicate; the asymmetry is harmless today but the message is not.
- The dry run exits before steps 3.7 and 3.8, so it never exercises the new gate. An operator releasing from a worktree discovers they need the flag only on the real run. Safe, but it lowers the value of a rehearsal, and the new hard abort makes that matter more than it did.
- `twine check` exits zero on warnings, so a distribution with rendering warnings still proceeds. That is the right trade, but the gate is not as strict as the comment implies.
- The header comment on step 8 still reads "Skips gracefully if either is missing", which this change has just made false for the website directory three lines below it.
- `--no-website` is missing from both usage strings, as is `--trust-ci` from an earlier change — three flags the usage output does not mention. Mitigated because the abort message names the flag.

Follow-ups: #1012 (a missing deploy key still silently skips the deploy, so the same end-to-end failure remains reachable by a slightly different route) and #1013 (this script has no syntax or lint gate at all, and shellcheck's relevant rules target exactly the filter class fixed here).

crtahlin (approved):

All three claimed fixes are real, and I verified each by executing the changed logic under stubs with `set -euo pipefail` in force rather than reasoning about bash semantics. A failing build, a failing `twine check` and a failing `twine upload` each now abort with full output and a step-stamped diagnostic, and all five Step 3.8 branches behave as advertised.

Two things worth recording. The branch really is one file — the three-dot diff against the merge base is `scripts/release.sh` alone, 59 added and 9 removed, no unrelated work carried along. And the earlier approval still holds: it sits on commit `5bb21808d`, which is the current head. `REVIEW_REQUIRED` is only plur9 still listed as a pending requested reviewer, not a dismissal by a later commit.

I also simulated the merge into current `main`: clean, touches only `release.sh`, and the merged file keeps main's unreleased-marker gate, the tracked-scratch gate and the corrected credential probe while gaining all three fixes. `bash -n` passes on both the head and the merged result.

Follow-ups: [#1087](https://github.com/plur-ai/plur/issues/1087), [#1088](https://github.com/plur-ai/plur/issues/1088), [#1089](https://github.com/plur-ai/plur/issues/1089)

### Notes

- `scripts/release.sh:1035` — `twine check dist/*` sits in Step 6, after Step 4 pushes the tag and Step 5c promotes npm to `@latest`. The stale-twine failure it catches is a local environment fact knowable at Step 3.x, so this placement turns the 0.18.0 incident from a silent half-published release into a loud one rather than preventing it. [#947](https://github.com/plur-ai/plur/issues/947) asked for both this and a minimum-twine-version assertion; only the late half landed. Filed as [#1087](https://github.com/plur-ai/plur/issues/1087).
- `scripts/release.sh:817, 825, 1013, 1017, 1021, 1092` — the fix is applied only to the PyPI step. Both `pnpm publish` calls, all six `npm dist-tag add` promotions and `gh release create` still use bare `... 2>&1 | tail -1`. Line 1013 is the worst: a failure part-way through the `for pkg in core cli mcp migrate` loop leaves core on `@latest` and cli not, silently. Filed as [#1088](https://github.com/plur-ai/plur/issues/1088).
- `scripts/release.sh:725` — the predicate is `[ ! -f "$WEBSITE_PREFLIGHT_DIR/index.html" ]` but the message says "Website dir not found", so a present directory with a renamed index file aborts with a diagnosis pointing at the wrong problem. Step 8 uses `-d`, so the two gates disagree on what missing means. Filed with the `--no-website` items as [#1089](https://github.com/plur-ai/plur/issues/1089).
- On the stale `(unreleased)` marker: `main` already carries that gate, added by `de25345c` at line 631 of the merged file, and this PR does not touch that region. So it neither helps nor hinders, and unlike some sibling branches it reverts nothing.
- No CI job runs `shellcheck` or even `bash -n` against this script — `.github/workflows/ci.yml` only invokes `scripts/smoke-release.sh`. Already filed as [#1013](https://github.com/plur-ai/plur/issues/1013). I ran `bash -n` manually on the head and the merged result; both pass.

### Done well

Dropping the `| grep -E "success|error"` filter at line 415 fixes a second failure mode the description does not claim, and I reproduced both: a build that FAILED was silenced, and a build that SUCCEEDED while printing neither search word made grep exit 1 and killed a healthy release with no output at all. The unreported false-failure mode is arguably the more dangerous of the two.

The upload guard at line 1043 is also a better shape than the logfile-and-tail approach [#947](https://github.com/plur-ai/plur/issues/947) itself suggested: it preserves the exit code, prints a step-stamped error, and surfaces the PyPI immutability warning with a link to the project history page before the operator retries.

Tests: PASS. 4,504 passed, 159 skipped; the only 3 failures are `sync-push-failure.test.ts`, the known host gitignore issue ([#1062](https://github.com/plur-ai/plur/issues/1062)). `receipt-io.test.ts` passes 7/7, so the failure seen on four other open pull requests does not appear here.

Verified: the changed Step 6 and Step 3.8 logic executed directly against stubbed `python3` and `pnpm` to confirm failure propagation; both grep-filter failure modes reproduced; the three-dot diff checked against the merge base; `main` confirmed to still carry all three original defect lines, so the fix is still needed; the merge into `main` simulated and `bash -n` run on the result.

plur9 (approved):
Re-approving after the branch moved: the three fixes (full build/twine output, twine check preflight, website hard abort with an explicit --no-website escape hatch) are still exactly as Črt verified them; only the version-bump region of release.sh has changed on main since, which this PR does not touch.

PR #954, opened 2026-08-19
fix(dsh): log a warning when the engine fails to load (#941)

### LINKED ISSUE #941: dsh engine: the bare catch that hid the inert-plugin bug is still there, now with a different cause behind it

## The observation

`loadEngine` in `packages/dsh/src/engine.ts` ends:

```ts
  } catch {
    return undefined
  }
```

No logging. `createEngine` then memoises the result with `pending ??= loadEngine(...)`, so a failed load is permanent for the process, and every facade method degrades to a neutral value — `{ count: 0 }`, `[]`, `{}`, `undefined`.

Degrading rather than taking down the host agent is the right call, and the docstring argues for it well. The concern is that **nothing anywhere records that it happened**.

## Why this specific shape is worth changing

[#920](https://github.com/plur-ai/plur/issues/920) describes the bug that motivated rewriting this function:

> `@plur-ai/core` is ESM-only, so `createRequire()` throws `ERR_PACKAGE_PATH_NOT_EXPORTED`, and the original catch swallowed it and returned `undefined`. The plugin was completely inert in production while every test passed.

[#918](https://github.com/plur-ai/plur/pull/918) fixed the **cause** — `createRequire` became a lazy `import()`. It kept the **shape**: a bare catch that converts any load failure into silent inertness.

So the next distinct failure reproduces the original symptom exactly. WASM init failing on a particular machine, a partial install, a version skew between the plugin and core — each yields a plugin that loads, reports no error, serves no memory, and passes every test, because the tests inject `importCore`.

## Suggested

Log the caught error once before returning `undefined`. The memoisation means it fires once per process, so it costs nothing in noise:

```ts
  } catch (error) {
    logger.warn(`[plur] memory engine unavailable, continuing without memory: ${error}`)
    return undefined
  }
```

`ready()` already exists and returns `false`, so a host that wants to check can — but nothing surfaces *why*, and an operator seeing an empty memory block has no thread to pull.

This is clause 2 of [ADR-0006](https://github.com/plur-ai/plur/blob/main/docs/adr/ADR-0006-report-only-what-was-confirmed.md) — "not checked" is a distinct outcome from "checked and empty", and the two are currently indistinguishable to everyone downstream.

## Related

- [#918](https://github.com/plur-ai/plur/pull/918) — where the lazy-import rewrite landed
- [#920](https://github.com/plur-ai/plur/issues/920) — the audit, which names the original inert-in-production bug

miles-on-nightshift:
## Summary

Closes #941.

`loadEngine` in `packages/dsh/src/engine.ts` had a bare `catch {}` that converted any load failure into silent inertness — the same pattern that hid the original inert-plugin bug before #918. This means a WASM init failure, partial install, or version skew produces a plugin that registers every surface and recalls nothing, with no operator-visible signal.

### Change

```ts
// before
} catch {
  return undefined
}

// after
} catch (error) {
  console.warn(`[plur] memory engine unavailable, continuing without memory: ${error}`)
  return undefined
}
```

The memoisation on `pending ??= loadEngine(...)` means the warning fires at most once per process. The degradation contract (`ready()` returns false, all reads return empty) is unchanged.

### Test

Added a test to `test/engine.test.ts` that spies on `console.warn` and asserts the warning fires with the expected message (including the error text) when `importCore` rejects.

## Test plan

- [x] `test/engine.test.ts` — 12 tests pass, including the new `logs a warning when the engine fails to load (#941)` case
- [x] No behaviour change for the success path

### DIFF
--- a/packages/dsh/src/engine.ts
+++ b/packages/dsh/src/engine.ts
@@ -49,7 +49,8 @@ async function loadEngine(
     const Plur = mod.Plur ?? mod.default?.Plur
     if (typeof Plur !== 'function') return undefined
     return new Plur({ path: config.path })
-  } catch {
+  } catch (error) {
+    console.warn(`[plur] memory engine unavailable, continuing without memory: ${error}`)
     return undefined
   }
 }
--- a/packages/dsh/test/engine.test.ts
+++ b/packages/dsh/test/engine.test.ts
@@ -40,6 +40,16 @@ describe('createEngine when core is unavailable', () => {
     await expect(engine.ready()).resolves.toBe(false)
   })
 
+  it('logs a warning when the engine fails to load (#941)', async () => {
+    const warn = vi.spyOn(console, 'warn').mockImplementation(() => {})
+    const engine = createEngine(config(), missing)
+    await engine.ready()
+    expect(warn).toHaveBeenCalledOnce()
+    expect(warn.mock.calls[0][0]).toMatch(/memory engine unavailable/)
+    expect(warn.mock.calls[0][0]).toMatch(/ERR_MODULE_NOT_FOUND/)
+    warn.mockRestore()
+  })
+
   it('degrades every read to empty instead of taking the host down', async () => {
     const engine = createEngine(config(), missing)
     await expect(engine.recall!('x')).resolves.toEqual([])


miles-on-nightshift:
Superseded by #956, which uses an injectable `warn` callback instead of a `console.warn` spy — cleaner tests (no spy/restore), two assertions (call content + memoization), and an open path to structured logging later. Closing this in favor of #956.

PR #955, opened 2026-08-19
docs(dsh): README inclusion criteria + pre-publish X credential gate

plur9:
Two changes to the dsh package's release readiness, both driven by concrete failures.

## `docs(dsh)`: restructure both READMEs to the ecosystem's inclusion criteria

`AdamPlatin123/awesome-dsh-plugins` (925★) publishes a required-sections table for catalog inclusion. Ours answered **1 of 11**.

Much of the content already existed under prose headings — "What leaves your machine" is *Permissions & data*, "When memory misbehaves" is *Troubleshooting* — so this is largely renaming to the names the criteria look for, plus the sections that were genuinely missing:

- **Overview** — what problem, for whom, stated plainly
- **Compatibility** — the dsh line, Cordis, four Node versions, and the date last verified, which is the question their criteria ask first

Applied to both `README.md` and `README.zh.md`.

## `release`: check X credentials before publishing anything (step 3.4)

0.18.0 went green through npm, PyPI, tag, GitHub release, MCP registry and the website, then **died at step 9 with X API 401**. Every irreversible step had already run.

Step 3.5 already applies exactly this reasoning to tweet *length*: catch it before commit/tag/push/publish, while everything is still revertable. It just never occurred to anyone that a tweet which cannot be posted *at all* is the same failure with a different cause. Step 3.4 closes that.

It probes **both** credential classes, because they die independently — the app-only bearer when the app's keys are regenerated in the developer portal, and the OAuth2 user token when the user grant is revoked. On 2026-08-18 both were dead, so checking only one would still have shipped.

Read-only, one HTTP call each, and skipped entirely under `--skip-tweet` so a deliberate no-announcement release is not blocked by it.

**Verified** against the currently-dead tokens: both probes return 401 and the run aborts before step 4, publishing nothing.

The header instructions were corrected too. `PLUR_X_*` must be updated in `.datacore/secrets/spaces/5-plur.env` and picked up with `creds sync` — `.datacore/env/.env` is generated and says "DO NOT EDIT — changes will be overwritten on next sync" in its own first three lines, so the previous instruction to edit it there was advice to lose the fix.

## Test plan

- [x] Release script probes verified against live (dead) tokens — aborts before any irreversible step
- [x] `--skip-tweet` path bypasses the new check
- [ ] README sections re-checked against the awesome-dsh-plugins criteria table by a reviewer


miles-on-nightshift (commented):
Reviewed. Two notes before merge:

**1. `/plur-memory` in the Quick start sections doesn't match the rest of the README**

The Troubleshooting section (and the existing README) says `/plur` is the diagnostic command. The Quick start in both README.md and README.zh.md introduces `/plur-memory`:

> To see what it knows, run `/plur-memory`.

If `/plur-memory` is a separate command that exists, keep it — but the Troubleshooting section should also mention it. If it's a mistake, `/plur` is the right command. Either way they should be consistent.

**2. Hardcoded `~/Data/.datacore/secrets/spaces/5-plur.env` path in the 3.4 error message**

```bash
echo "      ~/Data/.datacore/secrets/spaces/5-plur.env"
```

This is Gregor's machine-specific path. It's in an error message so it won't cause failures on another machine, but it will confuse any future release operator who doesn't have that layout. Consider using `.datacore/secrets/spaces/5-plur.env` (relative to the repo root) or a more generic description like "your Datacore secrets file for the plur space".

---

Both are minor — the core change (credential gate before any irreversible publish step) is correct and the README restructuring is exactly what's needed for awesome-dsh-plugins catalog inclusion. The `--skip-tweet` bypass is right, the dual-class probe is right, and the 3.4/3.5 ordering is right. Good to go once the two notes above are addressed or acknowledged.

miles-on-nightshift (commented):
Reviewed. Two notes before merge:

**1. `/plur-memory` in the Quick start sections doesn't match the rest of the README**

The Troubleshooting section says `/plur` is the diagnostic command. The new Quick start in both README.md and README.zh.md introduces `/plur-memory`:

> To see what it knows, run `/plur-memory`.

If `/plur-memory` is a separate command that exists, keep it — but Troubleshooting should also mention it. If it's a mistake, `/plur` is the right call. Either way they should be consistent.

**2. Hardcoded `~/Data/.datacore/secrets/spaces/5-plur.env` in the 3.4 error message**

That path is Gregor's machine-specific layout. It's in an error message so no failures, but it will confuse any future release operator. Consider using a relative path from repo root, or a generic description like "your Datacore secrets file for the plur space".

---

Both minor. The core change (credential gate before any irreversible publish step) is correct. Dual-class probe, `--skip-tweet` bypass, and 3.4/3.5 ordering are all right. README restructuring covers all 11 awesome-dsh-plugins criteria (catalog inclusion is real distribution upside). Good to merge once the two notes are addressed or acknowledged.

miles-on-nightshift (approved):
Docs-only PR — clean expansion of the dsh README. Overview, compatibility table, quick-start, and troubleshooting sections are all clear and accurate. Prose fixes in the reranker section are an improvement. CI passing. @plur9 ready for merge.

crtahlin (changes requested):
Requesting changes — but the recommendation is to **close this rather than rebase it**. Both halves are already on `main`, and merging it would revert two later fixes to the release script. Closing is your call as the author; I am not closing someone else's pull request.

### The documentation half is already on `main`

`git diff origin/main refs/pr/955 -- packages/dsh/README.md packages/dsh/README.zh.md` produces **no output**. Both files landed byte-identical in #921, merged 2026-08-16 — three days before this pull request was opened.

### The release-script half is a stale copy that reverts two merges

`git diff 681517a4 refs/pr/955 -- scripts/release.sh` is 37 deletions and zero additions. This branch's script is #948's script minus the `--trust-ci` work from #943.

The step 3.4 gate it carries is the version #951 explicitly replaced, because it probes the wrong credentials: it bearer-probes two variables that appear nowhere else in the script, while the step it protects signs with a completely different credential set. I confirmed that at this head — the two probed names occur only inside the new check.

The sole merge conflict is this file, and it is substantive rather than trivial. Resolving it in this branch's favour would revert both #951 and #943 on the release path. There is no safe rebase-and-merge here, and nothing needs salvaging.

### On the existing approval

The approval on this pull request describes it as "docs-only — clean expansion of the dsh README". It touches the release script with a gate that can abort or wrongly pass a production release. That half was not reviewed.

Its two earlier comments are also both false positives: the command it flags as possibly missing is registered and enabled by default, and the hardcoded path it flags is the pre-existing convention in this script and is present verbatim in `main`'s own corrected version of the same step.

### Two bugs in the gate, for the record

Neither matters if this closes, but they are worth carrying into #1013:

- The status capture appends a second value on a connection failure, because the tool writes its fallback code **and** exits non-zero. It still fails closed, but the operator sees a garbled code and is then told to regenerate credentials — a misleading diagnosis for a network outage. `main`'s version distinguishes this case.
- The token is passed as a command-line argument, exposing it in the process table to any other user on the release machine. Nothing is printed or logged, and no credential value is committed, so there is no leak into this repository — but `main`'s corrected version passes secrets through the environment and never puts them on an argument list.

### On carrying two unrelated changes in one branch

This is the proximate cause of the bad approval, not a stylistic complaint. The automated reviewer classified the pull request by its title and its largest file, called it documentation-only, and never read the shell that can break a release. It would also break rollback: reverting the release gate would revert the readme restructure with it.

The reasoning behind the gate is sound and is worth preserving — extending the existing "catch it while everything is still revertable" argument from announcement length to announcement postability, skipping cleanly when announcements are disabled, and failing closed on all three paths. Only the choice of credential is wrong, and `main` already carries the corrected form of the same idea.

### Follow-ups filed
- #1013 — add a structural guard so a preflight probing different credentials than the step it protects cannot regress silently again. It would have failed on both #948 and this branch.
- #1015 — this is one of four pull requests in the batch reviewed with only `issue-guard` reporting.

Separately worth knowing, and affecting `main` rather than this branch: the compatibility line in both readme files states a verification baseline that the upstream catalogue has since moved past. Worth a re-verify and an update on `main`.

plur9:
Closing per the review recommendation, having re-verified both halves independently.

**Documentation half is already on `main`.** `git diff origin/main <pr955> -- packages/dsh/README.md packages/dsh/README.zh.md` produces no output. Both files landed byte-identical in #921, merged 2026-08-16 — three days before this PR was opened.

**Release-script half would regress `main`.** `git diff origin/main <pr955> -- scripts/release.sh` is 145 deletions against 25 additions, and the deletions include the `--trust-ci` flag, its argument parsing, and its trust-CI gate logic from #943. The step 3.4 credential gate this branch carries is the version #951 deliberately replaced.

So there is nothing here to rebase: merging it would revert two later merges to buy documentation that already exists.

Anything still wanted from this branch should be opened fresh against current `main`.

PR #956, opened 2026-08-19
fix(dsh): log load failures in loadEngine instead of swallowing them (#941)

### LINKED ISSUE #941: dsh engine: the bare catch that hid the inert-plugin bug is still there, now with a different cause behind it

miles-on-nightshift:
## What

Closes #941.

`loadEngine` in `packages/dsh/src/engine.ts` had a bare `catch` that discarded the error entirely:

```ts
} catch {
  return undefined
}
```

A WASM init failure, a partial install, or a version skew between the plugin and core produces a plugin that appears healthy — `ready()` returns `false`, every read returns empty, every write is a no-op — and nothing anywhere records why. An operator seeing a blank memory block has no thread to pull.

## Fix

Added an optional `warn` callback parameter to `createEngine` (third positional, defaults to `console.warn`). It is threaded to `loadEngine` and called once in the catch with the error. Because `pending` memoises the result, the callback fires at most once per process regardless of how many facade calls follow.

Production callers (`apply()`) inherit the default; tests can pass a `vi.fn()` spy. The `apply()` call site in `index.ts` is unchanged — future work could thread `ctx.logger?.warn` for structured output, but the default covers the immediate diagnostic gap.

## Tests

Two new cases in `test/engine.test.ts`:
- `warn` is called with a message matching `"memory engine unavailable"` and the original error string
- `warn` is called exactly once however many facade methods are called after a failed load

All 13 engine tests pass.

### DIFF
--- a/packages/dsh/src/engine.ts
+++ b/packages/dsh/src/engine.ts
@@ -36,11 +36,13 @@ export interface Engine extends PlurClient {
  *
  * @param config - supplies the optional store path.
  * @param importCore - overridable for tests.
+ * @param warn - called once on load failure; defaults to console.warn.
  * @returns the engine, or undefined when core cannot be loaded.
  */
 async function loadEngine(
   config: Config,
   importCore: () => Promise<unknown>,
+  warn: (msg: string) => void,
 ): Promise<PlurClient | undefined> {
   try {
     const mod = await importCore() as { Plur?: PlurCtor; default?: { Plur?: PlurCtor } }
@@ -49,7 +51,8 @@ async function loadEngine(
     const Plur = mod.Plur ?? mod.default?.Plur
     if (typeof Plur !== 'function') return undefined
     return new Plur({ path: config.path })
-  } catch {
+  } catch (error) {
+    warn(`[plur] memory engine unavailable, continuing without memory: ${error}`)
     return undefined
   }
 }
@@ -64,14 +67,16 @@ async function loadEngine(
  *
  * @param config - plugin configuration.
  * @param importCore - overridable module loader, for tests.
+ * @param warn - called once if the engine fails to load; defaults to console.warn.
  * @returns the facade.
  */
 export function createEngine(
   config: Config,
   importCore: () => Promise<unknown> = () => import('@plur-ai/core'),
+  warn: (msg: string) => void = (msg) => console.warn(msg),
 ): Engine {
   let pending: Promise<PlurClient | undefined> | undefined
-  const engine = (): Promise<PlurClient | undefined> => (pending ??= loadEngine(config, importCore))
+  const engine = (): Promise<PlurClient | undefined> => (pending ??= loadEngine(config, importCore, warn))
 
   return {
     ready: async () => (await engine()) !== undefined,
--- a/packages/dsh/test/engine.test.ts
+++ b/packages/dsh/test/engine.test.ts
@@ -68,6 +68,24 @@ describe('createEngine when core is unavailable', () => {
     }))
     await expect(engine.ready()).resolves.toBe(false)
   })
+
+  it('calls warn with the error when load fails (#941)', async () => {
+    const warn = vi.fn()
+    const engine = createEngine(config(), missing, warn)
+    await engine.ready()
+    expect(warn).toHaveBeenCalledOnce()
+    expect(warn.mock.calls[0][0]).toMatch(/memory engine unavailable/)
+    expect(warn.mock.calls[0][0]).toMatch(/ERR_MODULE_NOT_FOUND/)
+  })
+
+  it('calls warn exactly once however many methods are called after a failed load (#941)', async () => {
+    const warn = vi.fn()
+    const engine = createEngine(config(), missing, warn)
+    await engine.ready()
+    await engine.list!()
+    await engine.recall!('x')
+    expect(warn).toHaveBeenCalledOnce()
+  })
 })
 
 describe('the facade shape', () => {


miles-on-nightshift:
Replacing #954 (closed). #954 had @plur9 assigned as reviewer — routing that review here instead, since this is the implementation we want to land.

miles-on-nightshift (commented):

**Verdict: ready to merge.**

Fix is correct and clean. The core change is exposing an injectable `warn` parameter on both `loadEngine()` and `createEngine()` rather than calling `console.warn` directly inside the catch — this decouples the side-effect from the logic and makes the behaviour testable without spying on console globals.

Checked:
1. **Catch clause**: `catch (error)` correctly binds the thrown value; the bare `catch` that was there before silently discarded it. The log message includes both a human-readable prefix and `${error}` so the underlying cause is always surfaced.
2. **Memoisation invariant**: `pending ??= loadEngine(...)` means the load is tried exactly once regardless of how many engine methods are called after a failure. The 'calls warn exactly once' test pins this directly — if the memoisation were broken, the warn count would equal the method call count.
3. **Default parameter**: `warn = (msg) => console.warn(msg)` preserves the existing behaviour for all production callers without any changes at call sites.
4. **Test coverage**: two cases that matter — (a) correct message content (`memory engine unavailable` + `ERR_MODULE_NOT_FOUND`), (b) exactly-once invocation after multiple method calls. Both pass.
5. **CI green**: all Node 20/22/24/26 matrix, issue-guard, dedup-calibration, smoke-packaged — 7/7 checks.

No nits.

crtahlin (approved):
Approved. Two gaps worth fixing, neither blocking — filed as a follow-up.

The change does what #941 asked: the discarded error is now surfaced, through an injected callback rather than a hardcoded console call, which makes it assertable without stubbing globals.

Verified rather than assumed:

- **The console default cannot corrupt a protocol channel.** I intercepted both output streams around `console.warn` and confirmed it writes to standard error only.
- **No external consumer breaks.** The built type declarations export only the facade; `createEngine` is not public API. There is exactly one production call site plus tests, so the third positional parameter breaks nobody.
- **The at-most-once claim holds.** `loadEngine` never rejects — its catch returns undefined — and the memoised promise is always truthy, so it is never re-evaluated.
- **The host logger is genuinely reachable**, as a built-in root service inherited by every child context, which is what makes the follow-up cheap.

On the API-shape question: a positional parameter is defensible here. The function is not exported publicly, the parameter is optional so it is source-compatible, and the core-import seam already established the pattern. The existing config argument is the user-facing settings schema, where a callback has no business living. Three positional seams is the ceiling, though — a fourth should force an options object.

### Praise
The exactly-once test is the right test rather than a restatement of the implementation. It drives three separate facade methods through the failed-load path, each re-entering the loader, so a broken memoisation would fail it with a count of three.

The message prefix also matches the house style used elsewhere for operator-facing notices on standard error.

### Notes
- If a host-supplied callback throws, the throw escapes the catch, the load promise rejects, and because the result is memoised every subsequent facade call rejects instead of degrading — the inverse of this module's contract, and permanent for the process. Not reachable with the console default, but wrapping the callback invocation would make the seam safe to hand to a host.
- The message carries the prefix and the error but not the store path, no error code and no stack. For a WebAssembly initialisation failure the operator learns that it failed but not which store was being opened. Interpolating the error directly also degrades to an unhelpful object string if a non-Error value is thrown, which is exactly the malformed-native-module case.
- This is the first console call in this package's source. There is no lint configuration in the repository, so nothing mechanically flags it.
- Out of scope for #941 and noted only for the record: the collapse of an unmeasured result into a measured zero is still in force throughout this file, so consumers still cannot distinguish "engine never loaded" from "checked and empty".

Follow-up: #1011 — the sibling early return one line above is still completely silent, and it is precisely the version-skew case #941 names as its motivation; and the only production caller still gets the raw console rather than the host logger.

miles-on-nightshift:
CI passing, approved. @plur9 ready for merge.

PR #977, opened 2026-08-22
fix(core): dedup co_injection events by query hash within 5s window (#975)

### LINKED ISSUE #975: Injections are logged twice, milliseconds apart, in about a third of cases
## What happens

The same injection is written to the history log more than once, milliseconds apart, with a different injection identifier each time.

Example from a real store, from the same second:

```
14:55:59.846  INJ-1786719359846-2j0000  query aeb910b0  6 engrams  3069 tokens
14:55:59.848  INJ-1786719359848-2n0000  query aeb910b0  6 engrams  3069 tokens
```

Same query hash. Same six engrams. Same token count. Two milliseconds apart. Two separate log entries.

## How common

Measured across one store's full history:

- 3,209 injection events in total
- **956 near-duplicate pairs**, where the same query and the same set of engrams appear again within five seconds

So roughly a third of the entries are repeats. Many are one to four milliseconds apart, which rules out a user genuinely repeating an action.

## Why it matters

Three ways.

**Usage data is inflated.** Anything counting how often an engram was used reads high, and not by a constant factor.

**Feedback linking may attach to the wrong entry.** An outcome is matched back to a preceding injection. With two candidates two milliseconds apart, the match is arbitrary.

**It will show up in provenance records.** Work is under way to build a record of where an engram came from and how it was used, from this log. A duplicate becomes a record saying a memory was used twice when it was used once. Discovered while building a worked example against real data.

## Where to look

`_formatInjection` in `packages/core/src/index.ts` writes the entry. The likely cause is two paths both formatting the same injection, or one path being called twice per turn. Worth checking whether session start and inject both fire.

## Suggested fix

Do not write an entry if an identical one, by query hash and engram set, was written within a short window. Or find the double call and remove it, which is better if that is what it turns out to be.

Either way, add a test.

philphauler:
## Summary

Fixes #975 — co_injection events logged twice, milliseconds apart, in ~30% of cases.

## Root cause

`UserPromptSubmit` and `PreToolUse` hooks both fire `hook-inject` on the same prompt within ~2ms. Each calls `inject()`/`injectHybrid()`, which logs a `co_injection` event via `_formatInjection()`. Same query hash, same engrams, two log entries.

## Fix

In-process dedup map (`query_hash → timestamp`) in `_formatInjection()`. If the same query hash was logged within 5 seconds, skip the second `co_injection` log. Map evicts stale entries at 100 keys.

**Only the provenance log is deduplicated** — the injection itself still runs (engrams reach the context window). No behavioral change for the user, just clean history.

## Impact

Reduces co_injection events by ~30% (the duplicate rate from the bug report: 956/3209 = 29.8%).

## Changes

- `packages/core/src/index.ts`: added `_recentCoInjections` dedup map, 23 lines

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -776,6 +776,9 @@ export class Plur {
    * event; findLatestInjectionFor covers the cross-process case.
    */
   private _lastInjectionByEngram: Map<string, string> = new Map()
+
+  /** #975: dedup co_injection events by query hash within a 5s window. */
+  private _recentCoInjections: Map<string, number> = new Map()
   /**
    * Timestamps (ms) of recent LLM failures, newest last (convergence Phase 2).
    *
@@ -4942,14 +4945,32 @@ export class Plur {
     // write failure must never break injection.
     if (injected_ids.length > 0) {
       const injection_id = generateInjectionId()
+      const queryHash = computeQueryHash(task)
+      // #975: skip co_injection log if the same query hash was logged within
+      // 5 seconds. UserPromptSubmit and PreToolUse hooks fire on the same
+      // prompt milliseconds apart — both call inject(), producing identical
+      // co_injection events. Dedup in-process by query hash + timestamp.
+      const now = Date.now()
+      const lastSeen = this._recentCoInjections.get(queryHash)
+      const isDuplicate = lastSeen !== undefined && (now - lastSeen) < 5000
+      if (!isDuplicate) {
+        this._recentCoInjections.set(queryHash, now)
+        // Evict stale entries to prevent unbounded growth
+        if (this._recentCoInjections.size > 100) {
+          for (const [k, v] of this._recentCoInjections) {
+            if (now - v > 10000) this._recentCoInjections.delete(k)
+          }
+        }
+      }
       try {
+        if (!isDuplicate) {
         appendHistory(this.paths.root, {
           event: 'co_injection',
           engram_id: injection_id,
           timestamp: new Date().toISOString(),
           data: {
             ids: injected_ids,
-            query_hash: computeQueryHash(task),
+            query_hash: queryHash,
             // Event provenance for offline token-economics analysis of real
             // sessions (the plur-bench #42 measurement). Deliberately NOT read
             // by the receipt, which shows no token/cost figure by design.
@@ -4960,6 +4981,7 @@ export class Plur {
           },
         })
         for (const id of injected_ids) this._lastInjectionByEngram.set(id, injection_id)
+        }
       } catch { /* best-effort */ }
 
       // #866: increment injection_count on primary-store engrams selected for context.


miles-on-nightshift (approved):
Good targeted fix for #975. The 5-second in-process window correctly catches the UserPromptSubmit + PreToolUse double-fire within a single tick. Notes:

- Eviction guard at 100 entries with a 10s TTL is reasonable; the `for..of` delete-while-iterating is safe on `Map`.
- Skipping `_lastInjectionByEngram` update on duplicates is correct — the first event already stamped it.
- The indentation of the outer `if (!isDuplicate)` block is slightly off from surrounding style but not a functional issue.

Approved. @plur9 ready for merge.

crtahlin (changes requested):
Requesting changes — the deduplication is in-process, but the duplicates come from separate processes, so it cannot fire for the case #975 reports.

### Blocking

**1. The state is per-process; the problem is cross-process.** Every hook that produces these duplicates is installed as a command hook, meaning a fresh spawn. Each spawn constructs a new instance, so the map is empty on every invocation. The reported duplicate rate will be unchanged after this merges.

The core file already documents the constraint — a comment there explains that injections are logged by another process entirely, which is why the lookup helper for the latest injection exists at all.

What the reported case needs is durable state: a marker under the session directory, or a deduplication check at write time.

**2. Where it does take effect, it drops real events.** The server holds one long-lived instance for the whole process, shared by every session. Two different sessions issuing the same task text within the window produce one logged event; the second session's event never exists. Receipts bucket by session, so that session loses its bucket entirely.

That inverts the third harm listed in #975 — "a record saying a memory was used twice when it was used once" becomes used once when it was used twice.

#975 asked for deduplication on query hash **and** engram set. Keying on the hash alone also suppresses a genuinely different injection — different engrams selected for the same text after a write, or different remote and rerank results.

**3. The first harm in #975 is untouched.** The block that increments the per-engram injection count sits outside the new guard, so a duplicate injection still double-increments it. Even where the guard fires, "anything counting how often an engram was used reads high" remains true. Only the history line is suppressed.

### Should fix

- The eviction comment says it prevents unbounded growth, but it only deletes entries older than the timeout and only runs on insert. I ran the loop: 300 distinct hashes arriving over nine seconds leaves all 300 resident. The map is bounded by distinct queries per window, not by the stated key count. Harmless in a one-shot process, a slow drift in the long-lived server, and the comment is wrong either way.
- No test. #975 closes with "either way, add a test", and this changes one file with nothing under the test directory. The existing logging test that looks relevant passes only because it uses two different task strings, so it does not exercise the new branch in either direction.
- Only `issue-guard` reported. The full matrix never ran, so the approval here was given with no test signal behind it. Tracked in #1015.

### Notes
- The guard is written twice, once around the map update and once nested inside the try. One guard around the block, or an early skip, would be clearer. The injection identifier is also still generated on the duplicate path and discarded.
- Five seconds is 2500 times the two-millisecond gap the description cites as the root cause. If an in-process guard is kept at all, a window in the low hundreds of milliseconds would cover the double-fire with far less collateral suppression.
- The existing per-session inject lock is the right layer for this, and the event branch currently returns before that lock is taken. Worth extending rather than adding a second, weaker mechanism. Note also that the event paths build a different task string, so they would hash differently — which does not match the identical-hash evidence in #975, leaving the actual source of the duplicates unestablished.
- The description's roughly-30% reduction figure has nothing behind it for the deployed hook topology.

The map is written only on a non-duplicate, so the window is anchored to the last **logged** event rather than the last seen one. I confirmed by simulation that a stream of repeats still logs periodically instead of being suppressed forever — that is the correct choice and easy to get wrong. The eviction timeout is also deliberately longer than the window, so eviction can never drop a key still inside its own window. And moving the last-injection update inside the guard is right.

Follow-ups filed: #1015 (fork pull requests reviewed without a test run).

philphauler:
Superseded by #1017, which addresses all three blocking findings from the review (cross-process file-based dedup, query_hash + engram IDs + source key, injection_count not gated). Closing this one.

Issue #219, opened 2026-05-22
Upgrade default embedding model: all-MiniLM-L6-v2 → EmbeddingGemma-300M


plur9:
## Context

PLUR's current `recallSemantic()` uses `@huggingface/transformers` with `all-MiniLM-L6-v2` (384d, 2021). On LongMemEval it reaches ~40% semantic accuracy (per [ENG-2026-0322-022](internal)). gbrain (Garry Tan, 18k stars in 6 weeks) recently published **R@5 = 97.60% on LongMemEval-S** with `text-embedding-3-large` (1536d) and **97.40% with vector-only — the embedding model alone closes most of the gap.**

Their own ablation: BM25-only drops to 19.80%. Hybrid adds +0.2pt over vector-only. **The embedding model is doing the work.**

Internal engram [ENG-2026-0322-028] already flags `EmbeddingGemma-300M` as the recommended upgrade path.

## Proposal

Make the default semantic encoder a 2024-class model, with a clear API-backed escalation path:

| Tier | Model | Dim | Latency | Cost | License |
|---|---|---|---|---|---|
| Default (local) | **EmbeddingGemma-300M** | 768 | ~50ms | $0 | Gemma terms |
| Power (local) | `BAAI/bge-large-en-v1.5` or `nomic-embed-text-v1.5` | 1024 / 768 | ~80ms | $0 | MIT / Apache-2 |
| API (opt-in) | `openai:text-embedding-3-large` | 3072 (or Matryoshka 1536/1024/512) | ~200ms remote | ~$0.13/M | proprietary |
| API (opt-in) | `voyage:voyage-3-large` | 1024 | ~200ms remote | ~$0.18/M | proprietary |

Keep `all-MiniLM-L6-v2` as a documented "tiny" tier for embedded/mobile.

## Acceptance

- [ ] `recallSemantic()` defaults to EmbeddingGemma-300M, drop-in
- [ ] Migration path for existing engram stores (re-embed on next consolidate, or lazy-on-read)
- [ ] Config knob: `embedding.model` and `embedding.dim` settable via `plur config set`
- [ ] LongMemEval R@5 baseline improves from current to ≥ 90% (target ≥ 95% with API tier)
- [ ] Storage cost analysis: 768d × N engrams vs current 384d × N — confirm acceptable at 100k engrams
- [ ] Docs page: `docs/embeddings.md` with the tier table + decision tree

## Why now

[...]

PR #978, opened 2026-08-22
fix(core): add embedder disposal path so host processes exit cleanly (#904)

### LINKED ISSUE #904: embedder: no disposal path, so every host process aborts at teardown (exit code 0)
## Problem

Any Node process that loads the embedder aborts at exit with:

```
libc++abi: terminating due to uncaught exception of type std::__1::system_error:
mutex lock failed: Invalid argument
```

The abort fires **after** the process has produced its output and **the exit code stays 0**, so it is invisible to anything that checks status — and alarming to anything that reads stderr.

## It is not concurrency

This was assumed to be a parallelism race for some time. It is not. Measured with a downstream test runner:

| condition | result |
|---|---|
| one file, one worker, embedder loaded | **aborts** |
| 4 embedder-loading files × 6 runs | **5 aborts**, 6× exit 0, identical results each time |
| 3 files that never load the embedder × 6 runs | **0 aborts** |

A single file in a single worker aborts with nothing to race against. The discriminator is purely whether the embedding runtime was loaded in that process.

## Cause

The ONNX InferenceSession is held in `pipelineCache` inside `transformers-base` (`dist/chunk-TWFYTLOE.js`). It is module-private and there is no exported disposal path, so a host process cannot drain onnxruntime's thread pool before static destructors run. The mutex is destroyed while a worker thread still holds it.

## Why it matters downstream

The abort is post-result and exit code 0, so it cannot fail a build — which is exactly why it is corrosive. Consumers see a C++ abort in otherwise-green output and cannot tell it from a real crash. The alternatives available to a host are all bad: fail the run on it (permanently red for a signal that is always post-result), filter stderr (hides real aborts), or ignore it (trains people to ignore aborts).

## Ask

Export a disposal path — `disposePipelines()` / `shutdownEmbedder()`, or dispose on an `exit` hook inside core — so a host can drain the session before teardown. Anything that lets the owner of `pipelineCache` release it deterministically would close this; a host cannot, because the cache is module-private.

If the correct fix is upstream in `onnxruntime-node`, saying so here is useful too — it at least stops each consumer re-deriving the diagnosis.

philphauler:
## Summary

Adds an explicit disposal path for the ONNX embedder so host processes can exit without a C++ abort on thread-pool mutex destruction.

- Adds optional `dispose()` to `EmbedderAdapter` interface
- Implements disposal in `transformers-base.ts` that calls the HF transformers pipeline's `.dispose()`
- Exports `disposeAllEmbedders()` from the embedder factory (drains the adapter cache)
- Exports `disposeEmbedder()` from the embeddings module and core public API
- Callers (MCP server, CLI, tests) can call `disposeEmbedder()` before exit

## Root cause

`pipelineCache` in `transformers-base.ts` stores ONNX pipeline promises. `_resetTransformersPipelineCache()` only clears the Map reference — it never calls `.dispose()` on the ONNX `InferenceSession`. At process exit the ONNX thread-pool mutex is destroyed while worker threads still hold it, causing a C++ abort.

## Design

`dispose()` is **optional** on the `EmbedderAdapter` interface — existing adapters (OpenAI API) that have no session to release are unaffected. The transformers-backed adapters implement it by resolving the cached pipeline promise and calling its `.dispose()`. All disposal is best-effort (swallows rejections) since the process is shutting down.

## Test plan

- [x] `disposeEmbedder()` calls `dispose()` on the cached adapter
- [x] No-op when no adapter is cached
- [x] Swallows rejection from `dispose()`
- [x] Works when adapter has no `dispose` method
- [x] `disposeAllEmbedders()` drains the factory cache
- [x] `dispose` is optional on the interface (type-level)
- [x] All existing embedder/embedding tests pass (63 passed, 13 skipped)
- [x] Build compiles (tsup ESM + DTS)

Closes #904

### DIFF
--- a/packages/core/src/embedders/index.ts
+++ b/packages/core/src/embedders/index.ts
@@ -68,6 +68,20 @@ export const DEFAULT_EMBEDDER: EmbedderName = 'bge-small'
 /** Singleton cache so two callers asking for the same name share metadata + the inner pipeline. */
 const adapterCache = new Map<EmbedderName, EmbedderAdapter>()
 
+/**
+ * Dispose all cached adapters, releasing ONNX sessions so the process can
+ * exit without aborting on thread-pool mutex destruction (#904).
+ */
+export async function disposeAllEmbedders(): Promise<void> {
+  const adapters = [...adapterCache.values()]
+  adapterCache.clear()
+  for (const adapter of adapters) {
+    if (typeof adapter.dispose === 'function') {
+      try { await adapter.dispose() } catch { /* best-effort */ }
+    }
+  }
+}
+
 /** Reset cached adapters. Test-only — production code never calls this. */
 export function _resetEmbedderCache(): void {
   adapterCache.clear()
--- a/packages/core/src/embedders/transformers-base.ts
+++ b/packages/core/src/embedders/transformers-base.ts
@@ -75,6 +75,20 @@ async function loadPipeline(modelId: string, dtype: TransformersAdapterConfig['d
   return await pending
 }
 
+/** Dispose all cached pipelines, releasing ONNX sessions so the process exits cleanly (#904). */
+export async function disposePipelines(): Promise<void> {
+  const entries = [...pipelineCache.entries()]
+  pipelineCache.clear()
+  for (const [, pending] of entries) {
+    try {
+      const pipe = await pending
+      if (pipe && typeof (pipe as { dispose?: () => Promise<void> }).dispose === 'function') {
+        await (pipe as { dispose: () => Promise<void> }).dispose()
+      }
+    } catch { /* best-effort — the process is shutting down */ }
+  }
+}
+
 /** Reset the shared pipeline cache. Test-only. */
 export function _resetTransformersPipelineCache(): void {
   pipelineCache.clear()
@@ -112,6 +126,18 @@ export function makeTransformersAdapter(config: TransformersAdapterConfig): Embe
       const out: Float32Array[] = []
       for (const t of texts) out.push(await embedOne(t))
       return out
+    },
+    async dispose(): Promise<void> {
+      const key = `${config.modelId}::${config.dtype ?? 'fp32'}`
+      const pending = pipelineCache.get(key)
+      if (!pending) return
+      pipelineCache.delete(key)
+      try {
+        const pipe = await pending
+        if (pipe && typeof (pipe as { dispose?: () => Promise<void> }).dispose === 'function') {
+          await (pipe as { dispose: () => Promise<void> }).dispose()
+        }
+      } catch { /* best-effort */ }
     },
   }
 }
--- a/packages/core/src/embedders/types.ts
+++ b/packages/core/src/embedders/types.ts
@@ -32,4 +32,6 @@ export interface EmbedderAdapter {
   embed(text: string, role?: EmbedRole): Promise<Float32Array>
   /** Embed N texts. Output order matches input order. */
   embedBatch(texts: string[]): Promise<Float32Array[]>
+  /** Release the underlying model session so the process can exit cleanly. */
+  dispose?(): Promise<void>
 }
--- a/packages/core/src/embeddings.ts
+++ b/packages/core/src/embeddings.ts
@@ -124,6 +124,22 @@ export function resetEmbedder(): void {
 }
 
 /**
+ * Dispose the active embedder, releasing the ONNX inference session so the
+ * host process can exit without a C++ abort on thread-pool mutex teardown.
+ *
+ * Call from the Plur class close() path or a process `beforeExit` handler.
+ * After disposal, the next embed() call will lazy-load a fresh session.
+ * Closes #904.
+ */
+export async function disposeEmbedder(): Promise<void> {
+  const adapter = embedPipeline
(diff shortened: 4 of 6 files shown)


philphauler:
Also running the Phase C embedder benchmark (bge-small vs bge-base, N=500 LongMemEval-S) — results will go on #219 when done. The disposal path in this PR unblocks clean benchmarking since the harness no longer aborts on exit.

miles-on-nightshift (approved):
Solid fix for #904. The disposal path is correctly optional and best-effort — additive interface change, no breaking risk. Disposal chain is clear: `disposeEmbedder()` → adapter `dispose()` → `disposePipelines()`. Tests cover the important cases (dispose called, no-op when empty, rejection swallowed, missing method). Two suggestions for consideration but neither is blocking:

1. The pipeline cache iterator in `disposePipelines` clears the map first then awaits pending promises — correct, but worth noting the await runs after the cache is already cleared so a concurrent `loadPipeline` during shutdown would race to create a new entry. Best-effort semantics cover this.
2. The `dispose` method on `makeTransformersAdapter` deletes only the key for *this* adapter's model. `disposeAllEmbedders` calls it via the adapter reference — fine since `adapterCache.clear()` already ran at that point.

Approved. @plur9 ready for merge.

crtahlin (changes requested):
Requesting changes — this implements the approach the issue owner already measured and rejected, and as shipped nothing calls it.

### Blocking

**1. The approach was disproved on #904 before this was written.** The comment of 2026-08-14 on that issue records a reproduction: an awaited `dispose()` followed by process exit still aborts on the same mutex failure. The conclusion drawn there is that the trigger is inference spinning up the runtime's intra-op thread pool, that disposal releases the session handle without joining that pool, and that the fix belongs upstream. It also states, in those words, that exporting a shutdown wrapper around the pipeline's disposal "would ship a fix that changes nothing".

This pull request wraps exactly that call. Nothing in the description, the tests, or the existing approval engages with that measurement.

**2. The stated root cause is not the cause.** The description attributes the exit abort to the pipeline-cache reset clearing the map without disposing. On `main` that function has exactly one caller, and it is a test. It never runs in a host process, so it cannot produce the reported production abort.

**3. Nothing calls the new API.** The diff touches only the core package. Searching the head for the three new exports finds the definitions, the re-exports, and the new test — no call sites. The description says callers "can call" it, which is a statement of capability rather than of wiring. So even on its own theory, merging this changes no observed behaviour and does not close #904. There is an obvious wiring point already present in the server's exit handling.

### Should fix

- `packages/core/test/embedder-dispose.test.ts:72` — the test named for draining the cache asserts nothing about draining. It declares a spy and never uses it, resets the cache so it is empty, then asserts the call resolves. It would pass identically against an empty function.
- `packages/core/src/embedders/transformers-base.ts:130` — all 26 added lines that actually touch the inference runtime are untested. Every test drives a hand-written stub. The tests prove disposal is invoked, not that the abort is gone. The only test that would settle #904 is a subprocess that loads the embedder, runs one inference, exits, and asserts the abort string is absent from standard error — and on the measurement above, that test would fail.
- `packages/core/src/embedders/transformers-base.ts:83` — disposal awaits the pending pipeline promise. If a first-time model download is in flight, calling this from an exit handler blocks on the full download. Undocumented and untested; a bounded timeout is the usual guard.
- `packages/core/src/embedders/index.ts:75` — not safe against an in-flight embed. A call already past the load holds the pipeline object directly, so disposing the session under it is a hazard on the native side, which the surrounding try/catch cannot contain. Symmetrically, a load starting after the clear creates a fresh session nobody will dispose. Best-effort is a fine policy for a failed disposal, not for tearing down a session another call is using.

### Notes
- `disposePipelines()` is dead code: exported from its module, not re-exported from the package, never called. The existing approval describes a call chain through it that does not exist.
- The two public entry points are asymmetric and neither is complete. One nulls the module-level pipeline but leaves the adapter cached; the other clears the adapter cache but leaves the module-level pipeline pointing at a disposed adapter. Neither the doc comment nor the export list says which to call.
- Only `issue-guard` reported here. The checked boxes in the description have no CI evidence behind them. Tracked generally in #1015.

Making disposal optional on the interface is the right call — it is purely additive, both consumers guard correctly, and no unguarded call exists anywhere. The cache key reconstructed in the adapter matches the key formation exactly, including its default, which is easy to get subtly wrong. And the snapshot-then-clear ordering makes both entry points genuinely idempotent and re-entrancy-safe.

### Suggested path
Settle #904 first, on the evidence already on it. If the resolution is upstream, this change is not the fix and should not close the issue. If a disposal surface is still wanted on its own merits — session and memory hygiene across repeated load and unload cycles — land it framed as that, with the wiring included and one complete entry point rather than two partial ones.

Related: #1014 tracks the reranker, which holds its own undisposed session and needs whatever answer #904 reaches.

philphauler:
Onnit, and imma learn from the feedback 🫡

philphauler:
Closing — the dispose() path was already measured and disproved by plur9 on #904 (2026-08-14): the abort fires after dispose() returns because the trigger is onnxruntime's intra-op thread pool, not the session handle. The fix belongs upstream in onnxruntime-node. Thanks crtahlin for the thorough review that caught this.

Issue #203, opened 2026-05-20
Tension-aware retrieval: pick the winning side of a contradiction instead of injecting both


plur9:
## Context

`plur_tensions` already tracks contradictions, supersession, and refinement between engrams. At retrieval time, we still tend to inject both sides of a tension and let the LLM untangle it. With a relational retrieval model, the graph itself can pick the winner given the current scope.

## Problem

When two engrams contradict, current recall can surface both — increasing token cost, noise, and the chance the LLM picks the stale side. Tensions are recorded but not *applied* at retrieval time.

## Proposal

- At retrieval, traverse tension edges. If engram A is superseded-by B, never return A unless explicitly requested via `plur_recall` with scope filter.
- For contradicts edges with no clear winner, prefer the one with higher recency, higher feedback score, or matching scope.
- Configurable: `--include-superseded` flag for forensic recall.

## Acceptance

- [ ] `plur_recall_hybrid` and PreToolUse injection respect tension edges by default
- [ ] Tensions resolved at retrieval reduce injected token count on benchmark prompts
- [ ] Override flags exist for forensic/historical recall
- [ ] Conflicts surface in `plur_tensions` output, not in the LLM's context

## Depends on

- Engrams as relational graph issue (tension edges are part of the typed graph)

plur9:
**Promoted to the front of the graph work (P1) per the 2026-07-02 assessment (see #225):** closest-to-implementable child, uses the graph for *correctness* (the category where graphs demonstrably pay — Zep's edge-invalidation), reduces injected tokens, and is the only child whose benefit LongMemEval's knowledge-update category can detect. One safety note from the assessment: a false-positive `contradicts` edge under default-on filtering silently suppresses a valid memory — the worst failure mode for a memory product. Keep the confidence threshold + `--include-superseded` forensics path in the spec.

crtahlin:
### Status verification — partially done

**Done:**
- Superseded engrams receive a ×0.3 score penalty in the injection pipeline — [packages/core/src/inject.ts](https://github.com/plur-ai/plur/blob/main/packages/core/src/inject.ts) (lines ~475-480): superseded engrams are demoted under budget pressure, not injected ahead of the current tip
- Historical-intent detection: when the prompt contains keywords such as "before", "was", "used to", "prior", the penalty is suppressed so forensic recall works — [inject.ts](https://github.com/plur-ai/plur/blob/main/packages/core/src/inject.ts) (`hasHistoricalIntent`)
[...]

PR #981, opened 2026-08-23
feat(tensions): measuredUnderDiffers gate — skip differing-config pairs (#869)

miles-on-nightshift:
## Summary

Wires `measured_under` into the tension scanner. When two engrams from the **same origin** both make a numeric claim and were measured under different configurations, the scanner no longer treats the pair as a contradiction candidate by default.

`MeasuredUnderSchema` and `measured_under` on `EngramSchema` were merged in #919 (v0.18.0). This PR connects those fields to the tension pipeline. It is part of #869 but does not close it: the refinement half (#203, storing differing-config pairs as context-scoped refinements) is not implemented here. Differing-config pairs are skipped or floored, not stored as refinements.

### What this adds

- `measuredUnderDiffers(a, b)`: `true` only when **both** engrams carry `measured_under` and at least one of `source_type`, `model`, `hardware`, `dataset` differs after trim/lower-case normalisation. An empty or non-string dimension counts as absent.
- `measuredUnderGateApplies(a, b)`: the gate fires only when the two engrams share an origin (`engramOrigin()`: primary store, `store:<scope>` or `pack:<name>`) **and** both statements contain a measurement token **and** the configurations differ. A pack or a foreign store row cannot suppress a tension against a primary engram by declaring a different configuration.
- `tensions.measured_under_pairs: skip | floor` (config, default `skip`; also a `plur tensions` option and a `plur_tensions` argument). `skip` drops the pair before the judge. `floor` still sends it to the judge but caps the reported confidence at `MEASURED_UNDER_CONFIDENCE_CAP` (0.1).
- Every skipped pair is counted. `TensionScanResult.skipped.measured_under` (engram pairs) and `.measured_under_pairs` are returned by `getCandidatePairsDetailed()`, `plur_tensions` and `plur tensions`. Nothing leaves the pipeline without a trace.
- Write-time validation: `learn()` and `learnRouted()` refuse a malformed `measured_under` with `TypeError: plur.learn: invalid measured_under — ...` instead of persisting it.

### Design decisions

- `date` is deliberately not a dimension. Same config, different dates, diverging values is a regression and must surface as a tension.
- Unknown ≠ differs: if one side omits a dimension, the pair is not treated as differing on it.
- An engram without `measured_under` is an unconditional assertion and stays a tension candidate regardless of the other side.
- The gate never applies across origins. The scanner's job at the pack/store boundary is to surface disagreement, not to let the foreign side declare it away.

### Motivating incidents (2026-08-11)

- `max_tokens 16384` (bench/operation) vs `65536` (production): flagged as a contradiction although both are correct in context.
- `87% wall-clock` (local git) vs `43% wall-clock` (GitLab CI): the ratio inverts with the environment; both values are true.

### Tests

- `packages/core/test/tensions-measured-under-gate.test.ts` (16 tests): dimension normalisation, origin classification, the trust-boundary rule (a foreign row cannot make the scanner skip a pair), skip accounting, `floor` mode confidence cap, write-time validation through `Plur.learn()`.
- The existing `measuredUnderDiffers` and `getCandidatePairs` tests are kept and updated for the origin/numeric preconditions.
- Tensions suites: 5 files, 165 tests pass (`npx vitest run --project @plur-ai/core tensions`).

Part of #869. Refinement storage tracked in #203.


crtahlin (dismissed):
Approved. Two gaps worth fixing, neither blocking — filed as follow-ups.

The gate is wired in correctly and the two design decisions in it are the right ones. Rather than read them, I ran the exported gate and `getCandidatePairs()` over eleven edge inputs:

- **One-sided `measured_under` keeps the pair.** This is the important case and it is right. Both incidents behind #869 are an unconditional assertion colliding with a scoped measurement, so treating the bare side as an unconditional claim keeps exactly the pairs the issue was filed about.
- **Same configuration, different date keeps the pair.** Excluding `date` is correct — a regression measured months later under identical conditions must stay a tension.
- **Null dimension values are handled** by the `!= null` guard, including an explicit null written through the passthrough schema.
- **Differing dimensions drop the pair**, as intended.

Gate ordering is fine: every gate in the loop is a plain continue, so placing this before the subject-overlap filter changes cost, not results.

### Notes
- `packages/core/src/tensions.ts:356` — `DIMENSIONS` is allocated inside the function, so a new array and closure are created per pair in a quadratic loop. Hoisting it to a module constant is a one-line change matching how the file treats other constants.
- `measuredUnderDiffers` is exported from the module but not re-exported from the package barrel, unlike its sibling predicates. Tests import it directly so nothing fails; it is simply public in the module and private in the package.
- The gate keys off the presence of `measured_under` rather than off the statement containing a measured quantity. Two engrams disagreeing about an API endpoint, which happen to carry differing model values, are dropped. Acceptable as a proxy while the field is only attached to numeric engrams, but it is a wider suppression than #869 describes.
- Sibling path, pre-existing and not introduced here: the importer's conflict detection re-implements the pre-filter and has none of the temporal gates. Those edges surface in the legacy section of the tensions tool, so a pair this gate suppresses can still appear there.

### Praise
The docstring states the date-exclusion rationale and both motivating incidents inline, and the block comment repeats it at the call site where a future reader will be tempted to "fix" the omission. That is the right place for it.

The four pipeline tests are also well built — two are positive controls asserting the pair survives, so the negative assertions cannot pass because some unrelated stage dropped the pair. My independent run reproduced the same split.

Follow-ups: #1008 (the comparison is case- and whitespace-sensitive, so `GitLab` and `gitlab` silently drop a real contradiction) and #1009 (the refinement half of #869 is unshipped while two docstrings already assert it as fact, and the gate has no counter or escape hatch).

crtahlin (dismissed):
Requesting changes on scope, not on the feature. The `measuredUnderDiffers` gate is correct and well tested — it is 172 of the 1,624 added lines. The other 89% is two features the title does not mention.

I verified the gate independently over 13 constructed engram pairs rather than reading it. Every conservative case behaves correctly: `measured_under` absent on one side, empty objects, disjoint dimensions, a date-only difference, and an unknown passthrough dimension all keep the pair. The `date` exclusion is right for the reason the docstring gives.

### Blocking

- `packages/hermes/plur_hermes/memory_provider.py:1` — 1,318 lines of Hermes `MemoryProvider` and 134 lines of pack-integrity enforcement ride under a title naming only the tensions gate. The tensions work (`packages/core/src/tensions.ts` +45, tests +128) merges cleanly on its own. The merge conflict, the dismissed approval, and a user-facing change to pack install behaviour all come from commits the title does not disclose. Anyone merging on the strength of a reviewed tensions gate also lands a pack installer that now throws on any INTEGRITY mismatch.

  The conflict is worth calling out specifically: the only merge conflict is a version collision in `packages/hermes/pyproject.toml`, where this branch bumps 0.18.0 to 0.18.1 while `main` is already at 0.19.1. A careless resolution regresses the published version.

  On the dismissed review — that was a stale-review dismissal triggered by the Hermes push (2026-08-26T06:45:16Z, seconds after the push at 06:45:02Z), not a substantive rejection. Its concerns were filed as [#1008](https://github.com/plur-ai/plur/issues/1008) and [#1009](https://github.com/plur-ai/plur/issues/1009), both still open.

### Should fix

- `packages/core/src/packs.ts:488` — `installPack` now throws on an INTEGRITY mismatch, and `allowTamperedIntegrity` is core-library-only. It is not surfaced by `packages/cli/src/commands/packs.ts` nor by the MCP install tool; the only non-test references are its declaration, its use site, and one unit test. A user with a hand-edited pack has no documented escape hatch from the CLI. Deleting the INTEGRITY file works, but nothing says so — the error only suggests re-export, which requires the original source engrams.
- `packages/core/src/packs.ts:490` — the error says "contents were modified after export", but `computePackHash` hashes only `SKILL.md` and `engrams.yaml`. Scripts, references and assets are unhashed, so the check cannot support that claim. For a security-facing verdict the message should state what it actually covers.
- `packages/cli/src/commands/packs.ts:204` — `integrity_verified` is a boolean carrying three states. `false` means both "no INTEGRITY file shipped" and "mismatch, overridden", yet the CLI prints "computed — no shipped INTEGRITY file to compare" for both. The test at `packs.test.ts:783` asserts exactly the second case returns false, so the misreport is proven by this PR's own test. Only reachable via the library API today, since the override is not CLI-wired.
- `packages/core/src/tensions.ts:350` — the gate keys off the presence of `measured_under`, not off the statement containing a measured quantity, which is the scope [#869](https://github.com/plur-ai/plur/issues/869) describes. I ran two non-numeric pairs through the gate: "the default embedder is bge-small" versus "gte-base" with differing `model`, and "the API key is stored encrypted at rest" versus "in plaintext at rest" with differing `dataset`. Both are suppressed to zero candidate pairs. Nothing prevents a caller attaching `measured_under` to a non-numeric engram — `index.ts:2605` passes it straight from caller-supplied context with no validation.
- `packages/hermes/plur_hermes/__init__.py:504` — `register()` installs the `MemoryProvider` whenever `ctx.register_memory_provider` exists, while the four standalone hooks stay registered. `sync_turn()` and the `post_llm_call` hook both run `extract_learning_patterns()` and `bridge.learn()` on the same assistant response, so both write. The shared bridge deduplicates only within `PlurBridge`'s 300-second cache TTL, and no test in the 563-line suite asserts the two paths do not both learn.

### The gate exists on three branches

Verified by diffing them; all share merge base `3323642c`, and `main` has none of it.

| Branch | Function | Call site wired | Barrel export | Dimension list |
|---|---|---|---|---|
| [#981](https://github.com/plur-ai/plur/pull/981) | yes | yes | no | inside the function |
| [#1080](https://github.com/plur-ai/plur/pull/1080) | byte-identical to #981 | yes | yes | inside the function |
| [#1082](https://github.com/plur-ai/plur/pull/1082) | yes, exported | **removed** | no | hoisted to a constant |

Because `main` contains none of it, [#1082](https://github.com/plur-ai/plur/pull/1082) will not revert this gate on merge. But merging any two produces two declarations of `measuredUnderDiffers` at the same location — a textual conflict that becomes a duplicate-identifier error if auto-resolved badly. Whichever lands should carry all three improvements: the hoisted constant from #1082, the barrel export from #1080, and the wired call site from here.

### Notes

- `packages/core/src/tensions.ts:361` — case and whitespace sensitivity is real and I reproduced it: `source_type` "GitLab" versus "gitlab", and "bench " versus "bench", both return true and drop the pair. Already [#1008](https://github.com/plur-ai/plur/issues/1008), so not new feedback.
- `packages/core/src/tensions.ts:472` — the body and docstring say differing-config pairs are "stored as refinements", but the code is a bare `continue`: the pair is dropped with no record, no counter, and no config escape hatch, unlike the adjacent snapshot gate which honours `snapshot_pairs: 'floor'`. [#869](https://github.com/plur-ai/plur/issues/869)'s second acceptance criterion is that such pairs "become refinements, not tensions", so "Closes #869" overstates what landed. Already [#1009](https://github.com/plur-ai/plur/issues/1009).
- `packages/core/src/tensions.ts:350` — `measured_under` is never auto-populated; both write sites take it only from caller-supplied context, so there is no risk of the calling platform being stamped in and silently suppressing every cross-platform pair.
- `receipt-io.test.ts` passes 7/7 here, so the failure seen on four other open pull requests does not appear on this branch.

### Done well

The date-exclusion rationale and both motivating incidents are stated in the docstring and repeated as a block comment at the call site, exactly where a future reader would be tempted to "fix" the omission. I confirmed the reasoning holds. The four pipeline tests include two positive controls asserting the pair survives, so the negative assertions cannot pass because an unrelated upstream stage dropped the pair — my independent run reproduced the identical split. And requiring `measured_under` on both sides is the right call: both incidents behind [#869](https://github.com/plur-ai/plur/issues/869) are an unconditional assertion colliding with a scoped measurement, so treating the bare side as unconditional keeps exactly the pairs the issue was filed about.

Tests: PASS. Full suite 4,521 passed / 3 failed / 159 skipped; the only failures are `sync-push-failure.test.ts`, the known host gitignore issue ([#1062](https://github.com/plur-ai/plur/issues/1062)). Hermes pytest 191 passed.

Verified: the gate exercised over 13 constructed pairs through both the exported function and `getCandidatePairs`; the three branches diffed against each other and their merge base; the conflict confirmed with `git merge-tree`; the dismissal timestamps compared against the push; the `allowTamperedIntegrity` reachability established by grepping the non-dist tree.

miles-on-nightshift:
Rebased to address the scope blocking item — the PR now contains **one commit**: the tensions gate (`feat(tensions): add measuredUnderDiffers gate`).

**What was removed and why:**

- `feat(hermes): add MemoryProvider ABC adapter` (1,318 lines) — this work has already landed on main via #957 (`dbd78970`). Removing it eliminates both the version-collision merge conflict in `pyproject.toml` and the dismissed-approval concern.
- `fix(packs): validate shipped INTEGRITY hash on install (#987)` (134 lines) — tracked separately on `origin/fix/987-pack-integrity-verification`. That branch should carry its own PR.

**Verification after rebase:**

- Cherry-pick applied cleanly onto current `origin/main` (no conflicts)
- `pnpm tsc --noEmit` on `packages/core/tsconfig.json` → exit 0
- `pnpm vitest run packages/core/test/tensions.test.ts` → 90 passed, 0 failed

The PR now matches its title exactly. Ready for re-review.

plur9 (approved):
Approving. The single blocking finding — scope contamination — is resolved at the current head.

The diff is now 8 files, all tensions:

```
core/src/tensions.ts                        +147/-3
core/test/tensions-measured-under-gate.test.ts  +196
core/test/tensions.test.ts                  +128
core/src/index.ts                            +25/-3
core/src/schemas/{config,engram}.ts          +12/-3
mcp/src/tools.ts, cli/src/commands/tensions.ts  +9/-1
```

The 1,318 lines of Hermes `MemoryProvider` and the 134 lines of pack-integrity enforcement are gone, so merging this no longer silently lands a pack installer that throws on any INTEGRITY mismatch. `packages/hermes/pyproject.toml` is no longer touched either, which removes the version collision — worth noting that `main` has since moved to 0.19.4, so had that file stayed, a careless conflict resolution would have regressed the published version by three minors rather than one.

CI green across all six checks.

**Two things I did not verify and am not treating as resolved by this approval:** #1008 and #1009, filed from the auto-dismissed review, are both still open. They were correctly separated from this PR — the dismissal was a stale-review dismissal triggered seconds after a push, not a substantive rejection — but separating a concern is not the same as closing it. Whoever merges should confirm neither is a prerequisite for the tensions gate specifically.

Issue #983, opened 2026-08-23
Provenance is dropped when an engram is pushed to a team store


crtahlin:
## The case

**As a user I want provenance to survive when an engram reaches my team.**

A team store is the most likely place a licence or an authorship question is ever asked. It is currently the one place none of this reaches.

## What is dropped

`RemoteStore.append()` builds the outbound body by hand (`packages/core/src/store/remote-store.ts:419-437`). It carries `statement`, `scope`, `domain`, `type`, `tags`, `pinned`, `rationale`, `commitment`, validity dates, `supersedes`, `locked_reason` and `source`.

It does **not** carry:

- `attribution` — who asserted it, and with which runtime, model and prompt
- `claim_class` — whether a person stated it or a model worked it out
- `provenance` — origin, chain, signature, licence

So every engram pushed to a team store arrives stripped. The receiving side cannot answer any of the five questions a provenance record exists to answer, and the loss is silent.

The inbound schema (`remote-store.ts:25-38`) is `.passthrough()`, so extra fields survive a read. The gap is on the way out, and on whether the server persists what it is sent.

## Why it is filed separately

The local store is being finished first. This needs a decision on the server side as well as the client, so it is not a change one repository can make alone.

## Work

- Add the three field groups to the outbound body, omitted when unset so an existing body stays byte-identical without them.
- Confirm the enterprise server persists and returns them; open the matching issue there if not.
- Round-trip test: push an engram carrying attribution, claim class and a chosen licence, read it back, and assert nothing was lost.
- Decide what a store that does not model these fields should do — silently drop, or refuse. Silent loss is what created this issue.

## Acceptance

- An engram pushed to a team store keeps its attribution, claim class and licence.
- A round-trip test proves it against the stub server.
- Loss, where it still happens, is reported rather than silent.

Part of #958.

Issue #963, opened 2026-08-21
Add a claim class, so a guess is not stored like a fact


crtahlin:
Part of #958.

## Problem

Nothing on an engram says what kind of claim it is. A statement a user typed, a line a text pattern scraped, and a conclusion a model inferred all look identical once stored.

The 2026 survey that named this gap asks for exactly this distinction: things observed, facts extracted, conclusions inferred, and earlier records revised.

## The clearest example

The editor plugin writes engrams from five triggers, told apart only by a short text label:

| Trigger | Label | What it really is |
|---|---|---|
| the user corrected the agent | `openclaw:ingest` | asserted |
| the conversation was compacted | `openclaw:compact` | inferred |
| the model reported what it learned | `openclaw:self-report` | inferred |
| a text pattern matched | `openclaw:afterTurn` | documented |
| the user ran the save command | `openclaw:slash` | asserted |

Three different kinds of claim under one prefix. The distinction that matters most is between *the user said this* and *the model claims it learned this*.

## What to add

An optional `claim_class` field with six values:

| Value | Meaning |
|---|---|
| `observed` | a plain record of something that happened |
| `documented` | taken from prose a human wrote |
| `structural` | read off the shape of a thing |
| `asserted` | stated outright by a person or agent |
| `inferred` | worked out by a model |
| `revised` | a rewrite of an earlier version |

These words are not new. Our repository extraction tool already uses most of them (`encode/assemble/strata.py`). Reuse rather than invent.

Set the value at each write path. Leave it unset where it genuinely cannot be determined.

## Done when

- the field exists and is optional
- every write path in the core library sets it
- the five editor-plugin triggers set it correctly, and no longer rely on the text prefix

plur9:
Closed by #1002, merged as `145e99ac`.

Not closed automatically: the PR description reads `Closes #959, #960, #961, …`, and GitHub only honours the keyword before the FIRST reference — so #959 closed on merge and the other fifteen did not. Closing by hand rather than leaving them open against shipped work.

Issue #676, opened 2026-07-24
No supported way to promote/rescope an existing engram to a team scope


crtahlin:
## Problem

There is no first-class way to move an **existing** engram from a local/personal (or wrongly-typed) scope into a team/remote scope so it reaches the enterprise store. Hit live while promoting ~35 engrams accumulated over 60 days into `group:plur/plur-ai/{engineering,research}`.

Every current path is inadequate:

- **`plur_promote`** — despite the name, only *activates candidate engrams for injection*; it does not change scope.
- **`plur_learn`** (re-emit under the target scope) — **deduplicates by content hash** and returns the existing local engram with its original scope. Nothing lands in the team store. (Verified: 3 pilot re-writes all came back with unchanged scope; absent from the remote scope on read-back.)
- **Direct `POST /api/v1/engrams`** — works and is what we ultimately used, but it is low-level (hand-built payload, bearer token, content nested under `data`), bypasses the MCP, and does not remove the local source, so it leaves duplicates unless originals are deleted separately.

This is a recurring need: engrams get written in whatever scope is active, and team-relevance is realized later. A 60-day scan surfaced 35 promotable engrams, incl. 12 stranded under a **typo scope** (`group:plur-ai/engineering`, missing the `plur/` segment) that never reached the store.

## Proposed

1. **Core primitive `plur_rescope` (a.k.a. `plur_move`)** — change an engram's scope, push it to the target remote store, and drop/redirect the source copy, atomically, bypassing the content-hash dedup guard. Accept single `id` and batch `ids`.
2. Alternatively/also: a `force_scope` / `rescope: true` option on the existing write path so a re-emit can re-route instead of dedup-collapsing (relates to the `feat/route-unscoped-writes` / #345 self-describing-scope work).
3. **Follow-up (optional):** a thin review command that scans non-team scopes for team-relevant engrams, clusters them by suggested target scope, and rescopes on confirm — the batch workflow we did manually.

## Notes

- Server-side dedup on `POST /api/v1/engrams` is by `content_hash` (re-posting identical content returns the existing engram, no duplicate) — useful, and a rescope primitive should rely on it rather than fight it.
- The typo-scope case argues for `plur_suggest_scope`-style validation warning when a write targets a scope string that matches no configured store.

plur9:
Thanks for the thorough investigation, @crtahlin — the 60-day production case with 35 promotable engrams is exactly the kind of evidence that sharpens this into a P1 design gap rather than a rough edge.

[...]

Issue #961, opened 2026-08-21
Capture who asserted an engram, and with what runtime and model


crtahlin:
Part of #958.

## Problem

The engram has no field for who or what asserted it. The only thing close is `source`, a free-text string (`packages/core/src/schemas/engram.ts:280`). `LearnContext` has no actor either (`packages/core/src/types.ts`).

So today nothing distinguishes a statement a person typed from one a model inferred from one a text pattern scraped.

## What to add

New optional fields, written at creation. **All optional.** If a value is not available, omit the field — do not invent one, and do not write a placeholder.

| Field | Holds |
|---|---|
| `asserted_by` | an address identifying who or what asserted it |
| `runtime` | the software that was running: name, version |
| `model` | the model behind the statement, where one was involved |
| `tool` | the extractor or importer, with its version |
| `on_behalf_of` | the address responsible, when the runtime acted for someone |

## On what goes in `asserted_by`

The provenance standard treats an agent identifier as an ordinary address. It does not care what kind. So we do not have to settle this now, and we should not lock it down.

Accept any of these:

- an opaque local identifier, resolvable only on this machine
- a Decentralized Identifier (DID). This is a W3C Recommendation, stable since July 2022, and lets an identifier be portable and pseudonymous without a central authority.
- an identifier for a running process, for automated writers

Recommendation: allow all three, require none. Store whatever the caller supplies.

## Privacy

This is the first field on an engram that can hold personal data, so it needs care from the start.

**We already have the pattern.** The repository extraction tool routes every name through one function that produces a salted hash, with the salt read from the environment rather than a config file. It raises an error rather than quietly falling back to plaintext. See `encode/assemble/identity.py`.

Reuse that shape:

- pseudonymise by default
- never write a plain personal name unless the caller explicitly asks
- treat these fields the same way pack export already treats private content

## Done when

- the fields exist on the schema, all optional
- `LearnContext` accepts them
- omitting them changes nothing about current behaviour
- a plain personal name is never written by default

plur9:
Closed by #1002, merged as `145e99ac`.

Not closed automatically: the PR description reads `Closes #959, #960, #961, …`, and GitHub only honours the keyword before the FIRST reference — so #959 closed on merge and the other fifteen did not. Closing by hand rather than leaving them open against shipped work.

PR #985, opened 2026-08-23
fix(remote-store): carry provenance on team push (#983)

philphauler:
## Summary

Partial fix for #983 — `RemoteStore.append()` builds the outbound body by hand but omits the `provenance` field. Engrams pushed to a team store arrive stripped of origin, chain, and licence information.

## Fix

Add `provenance` to the outbound body, omitted when unset so the historical wire body is byte-identical without it. Same additive pattern as `source` (#676) — servers that don't model the field ignore it.

## What this does NOT fix

`attribution` and `claim_class` are not yet in the engram schema (#963). This carries the one provenance field that exists today. When #963 lands, those fields should be added to the outbound body in a follow-up.

## Changes

- `packages/core/src/store/remote-store.ts` — add provenance to outbound body (+2 lines)
- `packages/core/test/remote-integration.test.ts` — round-trip test + omission assertion (+12 lines)

## Testing

All 50 remote-integration tests pass, including the new `#983 append carries provenance when present` test.

### DIFF
--- a/packages/core/src/store/remote-store.ts
+++ b/packages/core/src/store/remote-store.ts
@@ -434,6 +434,9 @@ export class RemoteStore implements EngramStore {
       // that don't model `source` ignore the field, and it is omitted entirely
       // when unset so the historical body is byte-identical without it.
       ...(e.source != null                  ? { source: e.source }             : {}),
+      // #983: carry provenance records so the receiving store can answer
+      // origin/chain/licence questions. Omitted when unset.
+      ...(e.provenance != null              ? { provenance: e.provenance }     : {}),
     })
     const r = await this.fetchBounded(`${this.apiBase}/engrams`, {
       method: 'POST',
--- a/packages/core/test/remote-integration.test.ts
+++ b/packages/core/test/remote-integration.test.ts
@@ -116,6 +116,22 @@ describe('RemoteStore against stub server', () => {
     expect('valid_from' in sent).toBe(false)
     expect('valid_until' in sent).toBe(false)
     expect('supersedes' in sent).toBe(false)
+    expect('provenance' in sent).toBe(false)
+  })
+
+  it('#983 append carries provenance when present', async () => {
+    const store = new RemoteStore(baseUrl, TOKEN, 'group:test', { ttlMs: 0 })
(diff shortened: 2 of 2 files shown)


crtahlin (approved):
Approved. Three gaps worth fixing, none blocking — filed as a follow-up.

The change adds one conditional spread to the hand-built outbound body, so `provenance` travels with a team push. It is correct, minimal, and placed directly beneath the `source` precedent it mirrors. The `!= null` guard covers both an undefined and an explicit null.

Verified rather than assumed:

- **Every push path is covered.** `appendAndGetServerId` is the only place a body is built by hand; `append()` delegates to it, `patch()` passes updates through verbatim, and all four call sites hand it a shallow spread that preserves the field.
- **The byte-identity claim holds.** I transcribed both revisions of the serialiser and ran them: the wire body is identical for the field absent, undefined, and null. The key is omitted, never emitted as null.
- **The field really is in the schema today**, as optional, so this is not ahead of its own type.

### Notes
- `packages/core/test/remote-integration.test.ts:124` — the fixture's licence key is spelled in the British form; the schema field is `license`, and the schema also carries `signature`. The assertion passes because the object is spread opaquely, so the fixture shape is never exercised. A schema-shaped record costs nothing and would catch a serialiser that whitelisted sub-fields.
- The description says the store file is "+2 lines"; the diff is +3, one code line and two comment lines. Cosmetic.
- The remainder is tracked, but not by the issue cited: `claim_class` is #963, while `attribution` is #961.
- One consideration for the server side: `source` is a string and `provenance` is a nested object. A server that validates strictly and rejects unknown object-valued keys would turn a silent drop into a failed push.

### One thing to know before merging
CI has not run on this pull request. The workflow runs on the head commit are awaiting approval, so `issue-guard` is the only reported check. The "all 50 remote-integration tests pass" claim is author-local. The file does contain 50 test blocks, but none of them have run here. Tracked more generally in #1015.

Follow-up: #1007 — the test stub still drops this field, so the round-trip assertion #983 asks for cannot be written yet, and nothing confirms the receiving server persists it.

Not setting a closing keyword was the right call; #983 correctly stays open to carry the rest.

Issue #999, opened 2026-08-23
Export blocks a memory over a credential that is not there — the JSON serialisation manufactures it


crtahlin:
> **Root cause found.** It is not really about attribution, and the first version of this issue mis-stated it. Two independent problems, one of them much broader than provenance.

## The symptom

```
plur learn "Deploy freeze starts Friday" --visibility public --asserted-by alice@acme.example
plur packs export team --output <dir>
→ exported: 1 of 2
   blocked: basic_auth_url: null,"derived_from":null,"know…
   blocked: Contains email: alice@acme.example
```

Naming a colleague by email drops the memory from the pack. Two reviewers gave this as their reason not to ship.

## Problem 1: a pattern for one value, run over a concatenated blob

`scanPrivacy` scans `serializeForSecretScan(engram)` — the whole engram as one JSON string. `basic_auth_url` looks for `user:pass@host`, and its second alternative does not require a scheme:

```
[^/\s:@]{0,64} : [^/\s@]{1,128} @ (domain)
```

JSON's own punctuation supplies that shape. In `"asserted_by":"alice@acme.example"` the leading group matches empty, the colon is JSON's, `"alice` is the middle, and `acme.example` is the domain. Demonstrated:

```
detectSensitive('alice@acme.example')                          → []
detectSensitive('"claim_class":null,"asserted_by":"alice@acme.example"')
                                                               → ['basic_auth_url']
match reported: "\"claim_class\":null,\"attributio…"
```

**The credential it reports exists nowhere in the user's data.** It is manufactured by serialising, and the quoted text is a fragment of neighbouring fields — which is why the error is unreadable.

**This is not specific to attribution, and it predates this work.** Any field whose value is an email with no space before it does it:

```
plur learn "From an email thread" --source "dana@example.com"
→ blocked: basic_auth_url: thread","source":"dana@example…
```

An email in the *statement* does not trigger it, because the statement contains spaces and the pattern excludes whitespace. So whether you are blocked depends on which field you used — which is not a property anybody designed.

## Problem 2: an email in an identity field is not a leak

Separately, the `email` check fires on `alice@acme.example` and blocks export. That match is correct — it is personal data — but blocking is the wrong response when the field exists to hold a person's identity. Recording who said something is the point of the feature.

## Fixing it properly

**Fix 1 — scan values, not the concatenation.**

Walk the engram and scan each string leaf independently. JSON punctuation then never appears in the scanned text and no match can span two fields.

[...]

Issue #997, opened 2026-08-23
Supersession is recorded; whether recall or export act on it is a PLUR decision


crtahlin:
> **Reframed.** The original version of this issue proposed that a superseded memory be suppressed in recall, injection and export. That is the wrong direction and it is withdrawn.
>
> **Provenance records what happened. PLUR decides how to act on it.** A provenance feature that changes what gets recalled has quietly become a retrieval policy, and retrieval policy belongs to PLUR's own logic — where it can be reasoned about against activation, scope and the rest of the model, not bolted on from the side.
>
> **No retrieval behaviour is being changed.** Recall, injection and search are untouched by this work; verified against `origin/main`. The notes below are ideas, recorded for whenever PLUR's own logic takes them up.

## What is recorded today

`--supersedes` populates `relations.superseded_by` on the replaced engram. As of #992 the provenance record carries `engram:supersededBy`, the summary names the replacement in its first line, and such a memory is never reported `complete: true`.

So the fact is captured, portable, and visible to anyone who asks about that memory.

## What PLUR does with it — open, and deliberately so

Observed today, stated as behaviour rather than as a defect:

- `recall` returns a superseded memory, ranked normally.
- `inject` injects it. A reviewer saw both the old and corrected claim in `constraints`, unmarked.
- `packs export` ships it, so a correction does not reach a colleague.
- Re-installing an updated pack inserts alongside rather than replacing: `list` returned 5 rows for 3 engrams, the same identifiers twice.

## Ideas, for whoever owns retrieval

Not proposals to implement now.

1. **Rank rather than suppress.** Supersession as a signal into activation, so the correction outranks what it replaced without anything disappearing. Fits the existing model rather than sitting beside it.
2. **Label at the point of use.** Inject both, mark the superseded one. Costs tokens; loses nothing.
3. **Suppress by default, reachable explicitly.** Strongest reading, and the most likely to surprise somebody who expects a store to return what is in it.
4. **Leave retrieval alone; act at export.** Narrower: a pack is a deliberate act of sharing, so preferring the current memory there is a smaller claim than changing what an assistant sees.

The re-install duplication is worth separating out — that reads as a pack-installation defect rather than a retrieval-policy question, whichever of the above is chosen.

## Related

#992 recorded the fact. #958 is the epic. Whether anything acts on it is a PLUR decision, and this issue exists to hold the question, not to answer it.

Issue #992, opened 2026-08-23
A superseded memory still reports as current and complete


crtahlin:
An agent driving the MCP tools reported this as its single most valuable fix, and the reason is worth quoting:

> "ENG-001 was superseded by ENG-018. ENG-018 correctly carries `prov:wasRevisionOf`, but ENG-001 still reports `status: active`, `commitment: decided`, `not_recorded: []`, `complete: true`, `History learn` — no reverse `superseded_by` anywhere. **I would have told my human that memory was reliable and current. It had been corrected five minutes earlier.**"

## Why this matters more than it looks

Provenance exists so somebody can judge how much to trust a memory. "This was corrected" is the single most decision-relevant fact there is, and it is the one fact the record does not carry. Worse, the summary reports `complete: true` and an empty `not_recorded` — actively asserting there is nothing further to know.

## Two parts

**1. The forward link exists; the reverse one does not.** The superseding engram records `prov:wasRevisionOf`. The superseded engram records nothing, so asking about the memory you have tells you nothing about the memory that replaced it. `plur_learn` documents a `superseded_by` relation; it is not populated on this path.

**2. The summary does not surface it either way.** Even where the document carries the link, the readable summary never mentions supersession. Same shape as two defects already fixed under #970: the record knows and the summary does not.

## Suggested shape

- Populate the reverse link when one engram supersedes another.
- Surface it prominently in the summary — above the licence, not below it. A superseded memory should say so in its first lines.
- A superseded engram must not report `complete: true`. Being replaced is exactly the thing a reader needs and does not have.
- Retirement deserves the same treatment: `prov:wasInvalidatedBy` is already built, and the summary should lead with it.

## Acceptance

- Asking about a superseded memory says so, and names what replaced it.
- A superseded memory is never reported as complete or current.
- Tests cover both directions of the link, and a retired engram.

Found while testing #958. Related: #989 (installed engrams lose provenance), #991 (writes that discard what was typed).

plur9:
Closed by #1002, merged as `145e99ac`.

Not closed automatically: the PR description reads `Closes #959, #960, #961, …`, and GitHub only honours the keyword before the FIRST reference — so #959 closed on merge and the other fifteen did not. Closing by hand rather than leaving them open against shipped work.

Issue #982, opened 2026-08-23
Limit which engrams are used, based on their licence


crtahlin:
## The case

**As a user I want to limit which engrams get used, based on their licence.**

Two enforcement points, and they behave differently. Both are in scope:

1. **On import** — refuse or flag engrams whose licence I do not accept, before they enter my store. This is where every software supply chain puts its gate.
2. **At retrieval** — exclude engrams from recall and injection when their licence does not permit the use at hand. Harder: a wrongly excluded memory is invisible rather than noisy.

## What exists now

Nothing. There is no licence filter anywhere in recall, injection or install.

The prerequisites now exist:

- Records express the licence as a machine-readable policy, with permissions, duties and prohibitions.
- A licence somebody chose is distinguishable from the schema default (`engram:licenseIsDefault`).
- `plur packs preview` reads and summarises the licences a pack carries, before installing.

## What practice suggests

**Do not build a policy evaluation engine.** ODRL is the W3C standard for expressing usage policy and is what Gaia-X and the International Data Spaces adopted — but neither the standard nor those initiatives specify how a policy is *evaluated*, so implementers reach different verdicts from the same policy. Keep ODRL as the expression format.

**Do build a licence allowlist and denylist.** Matching a licence identifier against a configured list is what actually runs in production. GitLab's licence compliance does exactly this over a bill of materials, covering some 500 identifiers.

**Show and ask, rather than filtering silently.** A silent exclusion is a memory that quietly stops existing, and the user cannot tell the difference between "filtered" and "never learned". Preferred shape: at install, show the licences the pack carries and let the user accept, decline, or accept a subset. Store the decision so it is not asked again.

## Design questions to settle

- Where does the accepted-licence list live — configuration, a prompt, or both?
- Is the retrieval filter a hard exclusion, or a marker rendered alongside the engram?
- What happens to an engram whose licence was never chosen? It must not be treated as though the author granted those terms.
- Does declining a licence at install reject the whole pack or the affected engrams?

## Acceptance

- A user can express which licences they accept.
- Installing a pack surfaces its licences and asks, rather than deciding silently.
- The decision persists.
- An unchosen licence is never presented as a grant.
- Tests cover accept, decline and partial-accept.

Deliberately **not** being built yet — filed so it stays on the roadmap. Part of #958.

PR #1002, opened 2026-08-23
Record where an engram came from, using the W3C provenance standard

### LINKED ISSUE #959: Add an actor and a reason to history events
Part of #958.

## Problem

`HistoryEvent` has no actor field (`packages/core/src/history.ts:5-10`). Seventeen event types are declared, thirteen are written, and none of them records who caused the event.

So the log can tell you an engram was retired. It cannot tell you who retired it.

## What to add

An optional actor on `HistoryEvent`, using the same shape as the engram fields in the sibling issue on capturing who asserted an engram.

Also carry a reason where one exists. Today almost nothing records why. Retiring, pinning, promoting, demoting and replacing all happen silently. Only conflict resolution and automatic locking write a reason at all.

## Note on one specific gap

The tool that forgets an engram never passes a reason. Its input schema accepts only an identifier and a search term, and all three call sites omit the second argument (`packages/mcp/src/tools.ts:1002-1028`). So every retirement made this way records an empty reason. Fix that here.

## Done when

- `HistoryEvent` carries an optional actor and an optional reason
- existing events still parse, with both fields absent
- the forget tool passes a reason through

---

**Depends on** #961 — the actor field on a history event must use the same shape as the one on an engram, so that issue settles the shape first.

crtahlin:
## Review fixes (2026-09-04)

Every BLOCKER / IMPORTANT from the 2026-09-04 review is fixed on this branch; each invariant is stated in the header of its test file and probed adversarially. Verified after merging main including the architecture audit (head `ca913b1d`): `@plur-ai/core` 206 files / 3253 passed (1 expected fail: the #999 `it.fails`; 36 skipped); `@plur-ai/cli` 61 files / 651 passed; `@plur-ai/mcp` 39 files / 454 passed; `pnpm typecheck:tests` clean.

### Leak guard covers every content field (BLOCKER)
`_engramContextFields` was a hand-kept list that had to mirror `LearnContext` and had drifted three times (#381, #405, this PR). `packages/core/src/content-fields.ts` now classifies every `LearnContext` key (`LEARN_CONTEXT_FIELD_ROLES`, `satisfies Record<keyof LearnContext, …>` so a new field without a row fails `tsc`); the hard `detectSecrets` scan in `learn()` / `learnRouted()` reads every content field; the engram side serialises the whole engram minus the statement and an explicit list of PLUR bookkeeping keys (the old `_`-prefix rule exempted any key a caller named — same fix in the export scan). `demoteIfSensitive`, rescope, `updateEngram`, `flushOutbox` and `saveMetaEngrams` all read the same set. Tests: `test/leak-surface.test.ts` (a canary in any content field is visible to the scan through both `learn()` and `_buildEngramShape`; a credential in `attribution.asserted_by`, `attribution.model.prompt_id` or `license` is refused at write time, blocked on rescope, demoted on update/learn to a shared scope) + `outbox.test.ts` flush re-guard. Behaviour change: a credential in rationale/source/anchor snippet/attribution is now refused even at local scope (`allow_secrets` still overrides).

### `visibility` in `plur_learn` (IMPORTANT)
Kept: the PR's own tests, `docs/demo/agent-conversation.sh` and the walkthrough depend on it, and without it every agent-built pack is empty (#970). Every path `public` opens (pack export, shared sync, remote push, rescope, explicit update, outbox flush) now scans the full engram; documented in the schema comment.

### `--batch` / `--date` rejected as `--path` typos (BLOCKER)
Typo detection tightened to edit distance 1; a sweep test parses every `'--flag'` literal under `packages/cli/src` through `parseGlobalFlags`; direct tests for `feedback --batch` and `restore --date`; e2e through the built CLI.

### `installPack` dropped `provenance/` (BLOCKER)
Plain `*.jsonld` files one level under `provenance/` are copied (sanitised names, lstat, no links). Test: export → install → preview → `present === true`, no "damaged" note, registry integrity equals `computePackHash(installed)`.

### Symlinks bypassed the pack scan (BLOCKER)
The walk is iterative, lstat-based, bounded by entry count (10 000); a link anywhere is refused before any read, and the install copy refuses again; tar extraction uses lstat; files > 16 MiB, special files, read errors and over-count are `unscannable` issues that block install with no override; full text reaches `detectSensitive` so the fail-closed `scan_truncated` fires past 1 MiB; NUL-containing files get the credential and injection detectors. A symlinked `SKILL.md` / `README.md` makes preview reject with a message naming the link (reading through the link is the harm: `SKILL.md -> ~/.plur/config.yaml` would have returned host config as the manifest).

### `readPackProvenance` crash / no size cap (BLOCKER)
`Array.isArray` guard, non-object nodes skipped, size cap, per-record try/catch, `unreadable_records` count + note (profile 5.3.1).

### Other IMPORTANT items
- `plur identity you@example.org` recommendation removed; `setIdentity` returns a warning when the value matches the export scanner's email pattern.
- `plur_provenance` is read-only (`save` removed; writing is `plur provenance --write` / `plur.writeProvenance()`) and local (`remote: false` in tool and CLI; MCP test asserts zero recall calls against a stub remote).
- `provenance.path` containment: `..`, `.`, `../../..`, `a/../..`, absolute → refused; a bad config never fails a learn.
- Portable records are self-contained: `exportPack` strips `derived_from`, `provenance.chain`, `session:` origins, `episode_ids`, `sources[].session_id`; `buildProvenanceRecord` keeps a reference only to another pack member and counts the rest as `engram:unresolvableReferencesOmitted`.
- Licence-required export is a CHANGELOG "Breaking" entry; `plur_packs_export` returns `{ exported: false, error, next_step }`.
- Unicode folding applied to `detectSecrets` / `detectSensitive` (ZWJ / ZWSP / fullwidth variants tested).
- `plur_session_end` validates suggestions before capturing the episode and reports a learn that throws in `engrams_failed`.

`previewPack` on an LLM-chosen path walked an arbitrary directory before checking it was a pack (would have listed the host's links in the error); `loadPack` / `computePackHash` read files whole before any size check; NUL-prefixed text files were installed unscanned; `exportPack` threw on a non-string `provenance.origin`.

### Remaining risks
Git `sync` shared push has no independent scan (pre-existing); `exportPack` still exports `structured_data` bookkeeping verbatim (pre-existing); #999 unchanged; stricter hard scan at local scope is a behaviour change; binary files get only the credential and injection detectors. The PR's CHANGELOG section sits under `## 0.18.0` below main's released 0.19.x sections and needs relocating to `## Unreleased`.

---

Closes #959, #960, #961, #962, #963, #964, #965, #966, #967, #968, #969, #972, #973, #979, #980, #992.
Advances the epic, #958.

## ⚠️ Merge order

**Merge #1000 and #1001 first, then rebase this branch.**

This branch was written before those were split out, so it still contains their commits. Rebasing after they land should drop the duplicated hunks; if it does not, the split versions are authoritative — they were reconstructed against `main` and reviewed on their own.

Nothing here depends on either of them. The ordering is only to avoid reviewing the same code twice.

## What this adds

An engram can now say where it came from, as a W3C PROV record in JSON-LD, with licences expressed as ODRL policy. A record answers five questions: **who** asserted it, **what kind of claim** it is, **when**, **what it came from**, and **whether you may reuse it**.

Reachable three ways: the `plur_provenance` MCP tool, `plur provenance <id-or-search>`, and inside every exported pack.

### The design decisions worth reviewing

**Provenance is a view, not a third store.** Records are derived from the engram plus its history, per ADR-0002. Nothing becomes a new source of truth. `provenance.generate` stays `never` by default — inside your own store a record mostly repeats the history log, and the trust boundary is the moment an engram leaves.

**A record you send must stand on its own.** No supporting log, and it names only the engram it is about. An early version copied the whole co-injection list into portable records, disclosing five other engrams to the recipient.

**Absences are stated, never guessed.** A memory written before we captured identity genuinely cannot say who asserted it, and the record says so: *"these are not guesses left blank — nothing recorded them."* A blank reads as zero.

**Provenance records; PLUR decides.** A superseded memory reports what replaced it and is never called complete. What recall and injection do about that is deliberately **not** changed here — that is PLUR's own logic, and #997 holds the question.

### Capture

`attribution` (who is answerable, and with which runtime, model and prompt) and `claim_class` (observed, documented, structural, asserted, inferred, revised) are new optional engram fields. Identity is never derived from the operating system account.

`plur_learn` and `plur learn` both accept them, plus `license` and `visibility`. Several of these were unreachable when first built — a field with no way to set it makes "complete" unachievable, and a completeness check nobody can satisfy is theatre.

### Licences

Expressed as ODRL policy, following Croissant's PROV-O + ODRL precedent. Three machine-readable answers — `may_reuse_commercially`, `may_redistribute`, `licence_recognised` — that **fail closed**: an unrecognised licence yields `false`, never `null`, because a consumer written `if (x !== false)` reads null as permission.

`may_leave_this_machine` answers a **different** question, from scope and visibility rather than the licence. It was called `shareable`, and a compliance tester read that as the licence answer.

An unchosen licence is marked. Every engram carries `cc-by-sa-4.0` unless somebody picks one, and presenting a schema default among recorded facts made the one legally-consequential field the one invented field.

### Packs

Exported packs carry provenance by default — the moment engrams leave is the one place a record defends against anybody. `plur packs preview` reads it before installing: record count, who is named, which licences, chosen or defaulted.

**No trust markers.** Nothing in a pack is signed, so everything it claims about its origins is a claim. A test walks every output line and fails on any positive assurance the code cannot back.

## Verification

**4796 tests pass.** The three `sync-push-failure` failures are machine-local: `~/.gitignore_global` lists `engrams.yaml`, so the test's scratch repo stages nothing and its seed commit fails. Verified by stashing the entire diff and watching them still fail.

**Checked against outside implementations** — rdflib and the PROV reference implementation both accept every record shape. Our own validator agreeing with our own generator proves nothing; that check caught a named-graph bug that made parsers see 3 statements instead of 64.

**Mutation tested.** 21 guarantees were broken on purpose to confirm the suite catches each. It caught 19 immediately; one gap was real — `--flag=value` splitting was tested in isolation but never that anything called it.

**Fifteen cold testers** across six rounds, each using it with no context. Their findings are in the commit history; the worst was a private memory whose summary said "not permission to share" while its record permitted `odrl:distribute` with no prohibition.

**Two recorded demos**, both covered by tests that assert what they show: a conversation with an assistant (`docs/demo/agent-conversation.sh`, every tool call a real call to the built server) and a command-line walkthrough.

## Known limitation

**#999 is open and is a blocker for the sharing story.** Attributing a memory by email address causes it to be excluded from every exported pack — the JSON serialisation manufactures a `user:pass@host` shape that exists nowhere in the data. I fixed it once by exempting attribution from the scan; a reviewer shipped a GitHub token through the hole within the hour, so it is reverted. The failing case is kept as a deliberately-failing test rather than deleted. Root cause and the proper fix are in #999.

## Deliberately deferred

Signing (§7 stays RESERVED), chaining, anchoring, licence-based recall filters (#982), carrying provenance to team stores (#983), and publishing the profile in `engram-spec` (#971).


crtahlin:

```
main
```

crtahlin (commented):
Reviewed. One blocking defect — the one already diagnosed — plus five gaps filed as follow-ups. Posted as a comment because GitHub does not accept a self-review.

This is unusually well-reasoned work. Across the diff the comments explain *why* rather than what, and repeatedly name the concrete failure that motivated each decision. That is rare enough to be worth saying before the findings.

**Coverage, stated plainly.** 88 files is too many to read closely, so: read closely — `schemas/config.ts`, `schemas/engram.ts`, `schemas/pack.ts`, `types.ts`, `history.ts`, all of `provenance-store.ts`, the licence/agent/summary halves of `provenance.ts`, the provenance sections of `index.ts`, the export-gate and `ExportOptions` sections of `packs.ts`, `inject.ts`, `ci.yml`, the CHANGELOG diff, and the three files with type errors. Not read closely — roughly 700 lines of `provenance.ts` (`buildProvenanceRecord`'s activity-from-history section, `buildPackProvenanceRecord`, `renderProvenanceSummary`), the bulk of the 506-line `packs.ts` diff, `secrets.ts`, the MCP and CLI diffs, and all of `spec/` and `docs/`.

**The merge-order precondition in the body is stale but resolved.** [#1000](https://github.com/plur-ai/plur/pull/1000) and [#1001](https://github.com/plur-ai/plur/pull/1001) were closed as superseded on 2026-08-27. I verified the supersede claim independently rather than taking it: `packages/core/src/secrets.ts` is byte-identical between `fix/996-pack-scanning` and this branch, and both files [#987](https://github.com/plur-ai/plur/issues/987) touched are present here. No divergence risk. The body should be updated so the next reader is not sent looking for a rebase that is no longer needed.

### Blocking

- `packages/core/test/attribution.test.ts:339` — `plur.paths` is private on `Plur`. Three TS2341 errors here (339, 383, 386), plus TS2345 at `provenance-conformance.test.ts:79` and TS2571 at `provenance-tester-round2.test.ts:117`, make `pnpm typecheck:tests` fail. That job gates `pnpm test`, so **no test in this repository has ever executed in CI on this branch or the three stacked above it.** Confirmed by running `tsc -p tsconfig.tests.json --noEmit` after a clean build: exactly those five are attributable to this branch; the others reported fall outside every hunk this PR touches and are pre-existing on `main`. Filed as [#1090](https://github.com/plur-ai/plur/issues/1090); fixing it here also unblocks [#1043](https://github.com/plur-ai/plur/pull/1043) and [#1044](https://github.com/plur-ai/plur/pull/1044).

  With it bypassed locally the suite is green: 4,855 passed, 1 expected fail, 1 unrelated failure.

Follow-ups: [#1098](https://github.com/plur-ai/plur/issues/1098), [#1099](https://github.com/plur-ai/plur/issues/1099), [#1100](https://github.com/plur-ai/plur/issues/1100)

### Should fix

- `packages/core/src/provenance.ts:131` — `unlicense` and `unlicensed` differ by one character and mean opposite things, and both spellings live in this codebase. `LICENCE_POLICY` maps the former; four user-facing surfaces tell the operator to type the latter to grant nothing. Verified against the built package: `unlicensed` reports `licence_recognised: false` with a note pointing at a licence text that does not exist, while `unlicense` grants commercial reuse and redistribution. The direction of failure is safe, but a dropped `d` converts "I grant nothing" into a public-domain dedication, in the one field this PR argues carries legal weight. [#1098](https://github.com/plur-ai/plur/issues/1098).
- `packages/core/src/schemas/config.ts:255` — the three-way `provenance` key collision is real, and this branch's shape should be authoritative. [#1082](https://github.com/plur-ai/plur/pull/1082) is the incompatible one: it names the mode key `mode` where this names it `generate`, defaults to `always` where this defaults to `never`, and declares nothing in the schema. If it lands second, an operator who set `generate: on_export` gets provenance stamped on every write against their stated choice. [#1099](https://github.com/plur-ai/plur/issues/1099).
- `docs/demo/provenance-walkthrough.sh:41` — sandboxes itself with `--path` on each of a dozen invocations and no `export PLUR_PATH` at the top, so one forgotten flag writes to the real store. The guard at `packages/cli/test/walkthrough.test.ts:78` greps stdout for a macOS-shaped `/Users/<name>/.plur`, so on Linux CI it can never match — inert exactly where it runs. [#1100](https://github.com/plur-ai/plur/issues/1100).
- `packages/core/src/schemas/config.ts:262` — `provenance.path` is declared with a default and read nowhere; `FileProvenanceStore.dirFor` hardcodes `join(this.root, 'provenance', safe)`. Either thread it through or drop the key.
- `packages/core/src/packs.ts:1404` — `exportPack` now throws when no licence is passed and none is configured. Defensible and well argued, but it is a breaking change to a public API and to `plur packs export`, and the CHANGELOG adds only "Added" and "Fixed". Anyone scripting pack export today gets a hard failure with no changelog warning.

### Notes

- `CHANGELOG.md:66` says "Seven licence names map to a policy"; `LICENCE_POLICY` has eighteen entries.
- `packages/core/src/inject.ts:646` — injection output *is* changed here (an `(inferred)` prefix in layers 1 and 2, a `Kind:` field in layer 3), which sits awkwardly beside the body's "what recall and injection do about that is deliberately not changed here". The claim is accurate on its narrow point about supersession, but consumers with token budgets or snapshots on the DIRECTIVES block will see different text.
- `packages/core/src/index.ts:804` — `buildAttribution` never returns undefined in practice, so every new engram carries `attribution: { asserted_by: 'unidentified', runtime: {...} }`. Deliberate and well argued, but an unconditional per-engram size increase for every user who never configures an identity.
- `packages/core/test/pack-file-scan.test.ts:161` — stale comment describing the exemption design that was reverted.
- Backward compatibility looks sound: `attribution` and `claim_class` are optional additions, `actor`/`reason` are optional on `HistoryEvent` with an explicit "every reader MUST tolerate its absence" contract, `EngramSchemaPassthrough` preserves unknown keys. Rollback leaves inert orphan `provenance/` directories. No migration needed and none shipped, which is right.
- `pnpm-workspace.yaml:5` — pre-existing on `main`, not this PR, but it blocks builds on pnpm 11: `allowBuilds` entries are the literal placeholder "set this to true or false", and `pnpm.onlyBuiltDependencies` is no longer read, so `pnpm build` fails with `ERR_PNPM_IGNORED_BUILDS`.

### Done well

`licencePolicy` returning a grants-nothing policy with an explicit `licenseRecognised: false` for an unrecognised licence, rather than emitting no policy at all — verified empirically that unrecognised and defaulted licences both fail closed while preserving the distinction between "the answer is no" and "we could not tell", with the reasoning written down at the point of decision.

`may_redistribute` requiring both that the licence permits distribution *and* that the memory is cleared to leave the machine: a private engram licensed MIT reports `may_leave_this_machine: false`, `may_redistribute: false`, `may_reuse_commercially: true`. Three answers to three genuinely different questions, which is exactly the confusion the rename from `shareable` was meant to fix.

Routing all 28 history-append sites through a single `_appendHistory` that stamps the actor, on the stated grounds that a policy applied at 28 call sites will be missed at the 29th — and then asserting it across every path rather than a sample.

Keeping the email-attribution bug as an `it.fails` test with an honest account of the fix that was tried, why it was reverted, and where the proper fix lives. And `provenance-store.ts:38`'s `substance()` stripping the self-referential generation timestamp before comparing records, so the identical-record check is not dead code that always writes — the kind of detail that only surfaces when someone actually runs the command five times, which the comment says is how it was found.

Tests: PASS with the typecheck blocker bypassed — 4,855 passed / 1 expected fail / 161 skipped; the single failure is unrelated to this PR. As shipped, `pnpm test` never runs at all.

Verified: the five type errors reproduced and attributed against `main`; the supersede claim checked by byte-comparing `secrets.ts` across branches; the licence policy probed against the built artifact for recognised, unrecognised, defaulted and private-visibility cases; the config collision read from all three live diffs.

plur9:
Triage note, not a review. This PR is **+14,602/−226 across 91 files**, which is past the size where a single review pass can be honest about what it covered. CI is green on all twelve checks and the only human review is a COMMENTED, so nothing is formally blocking — but nothing has substantively approved it either.

I am not approving it on that basis. A provenance model touching 91 files decides how every future engram records where it came from; approving it because CI passed would be exactly the "green tests certify a broken path" failure this repo has already written down.

Suggestion: split it. The W3C PROV-O schema and its round-trip tests are reviewable on their own; the migration and the consumer changes are separate reviews with different risks. If it must land whole, it needs a reviewer with the time to read it properly and say what they actually checked.

Related and worth sequencing against: `created_at` / `updated_at` landed on #1138 as optional, never-defaulted provenance timestamps, with the deliberate rule that absent means genuinely unknown and synthesising one destroys the provenance it records. If this PR introduces its own creation-time concept, the two should agree before either merges.

plur9 (changes requested):
Requesting changes on one finding, which is a one-line fix in a file this PR already edits. Everything else below is a note or a correction to the description.

This is careful work and the parts I probed hold up. I verified rather than read where I could.

### Blocking

- `packages/core/src/packs.ts:1693` — **an exported pack ships PLUR's own bookkeeping, and that bookkeeping is exempt from the security scan.**

  `PLUR_BOOKKEEPING_KEYS` (`content-fields.ts:104`) excludes `_outbox`, `_routed`, `_demoted`, `_rescoped_from` and `_expiry_extracted` from every scan. That exemption is right, and the stated reason is right: `_outbox.target_url` legitimately carries the host topology the infra detector flags, and scanning these would falsely demote every remote-origin or auto-routed engram on update.

  But `exportPack`'s `exportEngrams` map never removes them. So these five keys are simultaneously **exempt from the scan** and **shipped verbatim**. Executed against this branch's built core — learn, stamp the markers exactly as `learn()`/`learnRouted()` stamp them, export:

  ```
  _outbox                        PRESENT IN PACK
  _demoted                       PRESENT IN PACK
  _routed                        PRESENT IN PACK
  _rescoped_from                 PRESENT IN PACK
  plur.internal.acme.example     PRESENT IN PACK   ← _outbox.target_url
  group:acme/engineering         PRESENT IN PACK   ← _demoted.from / _routed.scope
  group:acme/finance             PRESENT IN PACK   ← _rescoped_from
  aws_secret_access_key          PRESENT IN PACK   ← _demoted.patterns
  ```

  The last one is the one I would fix first. `patterns` is `[...new Set(offending.map(h => h.pattern))].join(', ')` — the detector names that matched — so a published pack tells its recipient **which class of secret was once in that engram**. The others are the remote store's internal hostname and port, and internal scope names.

  This is the same class the PR strips five other fields to prevent, and the profile rule it cites for them applies verbatim: *"No identifiers only our store can resolve."* An internal scope name and a private hostname are exactly that.

  The description lists this as a remaining risk, "pre-existing". Its origin is, but two things make it this PR's: the scan exemption that hides it from the check is introduced here, and this is the PR that establishes what an engram may carry when it leaves. The fix reuses a function already imported at `packs.ts:10`:

  ```ts
  cleaned.structured_data = userStructuredData(cleaned.structured_data) ?? {}
  ```

  A test alongside the existing strip tests would pin it.

### Should fix

- **The merge-order warning is stale and now misleading.** "Merge #1000 and #1001 first, then rebase" — both are **closed, not merged**, so their content lives only here and there is nothing to rebase against. A reader following the instruction will go looking for merged commits that do not exist. Worth deleting or rewriting.

- **The CHANGELOG entries are in a released section.** The description says so and it has not been done: the provenance entries sit at lines 334–346, under `## 0.18.0` (line 292). `## Unreleased` is line 3. Shipping this puts new features in the changelog of a version released weeks ago.

- **Conflicts with current `main`.** A test merge produces content conflicts in four files: `packages/cli/src/index.ts`, `packages/core/src/index.ts`, `packages/core/src/inject.ts`, `packages/mcp/src/tools.ts`. The branch is 49 ahead / 10 behind. `inject.ts` and `tools.ts` in particular moved a lot today (#1138, #1154, #1163, #1167), so the rebase wants doing before the stack on top of it is reviewed against a shifting base.

- `docs/demo/recording/` adds **3.0 MB of GIFs** (1.0 MB + 2.0 MB) to git history permanently. The `.cast` files they were rendered from are 16 KB each and are also committed, as are `.txt` transcripts — so the recordings are recoverable without the GIFs. Worth hosting them outside the repo, or keeping only the casts.

### Notes

- `FileProvenanceStore`'s containment check is textual: `resolve(root, subdir)` then `startsWith(base + sep)`. That correctly refuses `..`, `.`, `a/../..` and absolute paths — I read it against each. It does not resolve symlinks, so a `provenance.path` naming a symlinked directory inside the store would pass and write outside it. The operator owns both the config and the store, so this is low severity, but `realpath` would close it and `packs.ts` already uses `lstat` discipline elsewhere.

### Verified, not just read

- **The leak-guard blocker is genuinely fixed.** I added an unclassified `probe_note?: string` to `LearnContext` and `tsc` failed exactly as claimed: `TS1360 ... Property 'probe_note' is missing in type ... but required in type 'Record<keyof LearnContext, "content" | "control">'`, plus a second error at the derived list. A new field cannot be added without classifying it. This is the right shape for the defect it closes — the hand-kept list had drifted three times.
- **The export strip is complete for what it covers.** `derived_from`, `episode_ids`, `sources[].session_id`, `session:` origins and `chain` are all cleared at `packs.ts:1719-1731`, matching the profile rule.
- **`unscannable` blocks install with no override**, and the comment says why: "nothing may land in the store that was not checked... the remedy is to fix the pack, not to look away." Secrets throw before it; prompt-injection is overridable and separated. That ordering is right.
- **The pack walk is lstat-based**, bounded at 10,000 entries and 16 MiB per file, and refuses symlinks while naming the link and its target.
- Full monorepo suite on the head: **368 files passed, 15 skipped, 0 failed.**

### Done well

"Provenance is a view, not a third store" is the decision that makes the rest of this reviewable — nothing becomes a new source of truth, and `generate: never` by default means the common path writes nothing. "Absences are stated, never guessed" is the right call for a field where a blank reads as zero. And declining to change what recall and injection do about a superseded engram — recording that #997 holds the question — is the discipline that keeps a 14,000-line PR from becoming 20,000.

Tests: PASS — full suite on the head, 368 passed / 0 failed. Verified: probed the `satisfies` guard by adding a field to `LearnContext`; executed export against the built core with the bookkeeping markers stamped as the real write paths stamp them; test-merged against current `main` to classify the conflicts; read the containment check against `..`, `.`, `a/../..` and absolute inputs; checked #1000 and #1001's actual state; located the CHANGELOG section boundaries by line.

plur9:
Pushed the fix for the blocking finding, plus the merge with current `main`.

**The strip.** `exportPack` now removes `PLUR_BOOKKEEPING_KEYS` from
`structured_data`, using the `userStructuredData` already imported in that file.
The invariant is stateable now: **a key in that set is unscanned and stripped, a
key outside it is scanned and kept.** Before, both halves failed open together —
the one thing nothing checks was the one thing that shipped.

Nine regression cases, each confirmed to fail with the strip removed: five for
the markers, four for the values that actually leaked
(`plur.internal.acme.example`, `group:acme/engineering`, `group:acme/finance`,
`aws_secret_access_key`). Plus a control asserting user `structured_data`
survives, including a `_`-prefixed key a caller named themselves — which is why
the set is enumerated rather than prefix-matched.

The class this belongs to — export is a denylist, so a new engram field leaves by
default — is **#1169**, to be decided with #1153 rather than here.

**The merge.** Ten PRs landed while this sat. Three of the seven conflicts hid an
integration defect neither side had alone, and I would flag the first two to
whoever reviews the resolution:

- **`plur forget --scope` would have become a hard failure.** #1135 added the
  flag; this branch added a flag allowlist the dispatcher enforces with
  `exit(1)`. Fixed by admitting `--scope` to both lists.
- **`co_injection` would have lost its actor.** This branch routes history
  through `_appendHistory`, which stamps `event.actor`; main's newer dedup block
  (#1017) calls the bare `appendHistory`. Kept the dedup, routed it through the
  wrapper, and made the wrapper return the boolean `injection_count` is gated on.
- `leak-surface.test.ts` became time-dependent once #1138 put timestamps in the
  shape — a 2 ms gap between two constructors failed a deep-equal.

**One finding I did not fix here, because it is not fixable here.** The
write-contract guard refused this branch's two new schema fields. They are
recorded `NOT_MODELLED` and raised as **#1172**, because the enterprise server's
`POST /api/v1/engrams` forwards a 13-field allowlist that does not include them
— so adding them to the wire would acknowledge fields the server discards. That
is the #1151 failure in the fields this branch exists to add, and it needs a
change in `plur-ai/enterprise` as well.

Worth knowing while you are there: the same allowlist also drops
`measured_under`, `knowledge_anchors` and `dual_coding`, which #1151 added to the
client body — so that fix is client-complete and server-incomplete. And
`content-policy.ts` already scans `dual_coding` and `knowledge_anchors`, so the
server is scanning for fields that door cannot deliver.

Suite: 377 passed, 15 skipped. `pglite-adapter.test.ts` fails only under
full-suite parallel load — a different test each run, green in isolation, and
green on `main`.

The three should-fix items from the review are still open: the stale
`#1000`/`#1001` merge note, the CHANGELOG entries under the released `## 0.18.0`,
and the 3 MB of GIFs.

plur9 (approved):
Approving, with the conflict of interest stated rather than glossed: **I wrote the fix commits I am approving.** The blocking finding was mine, the strip is mine, the merge resolution is mine, and no second reviewer has seen any of it. Recorded here because the merge is happening on the owner's instruction to land what is green and review the whole codebase before publish — that review is the compensating control, and this PR is one of the reasons it is not a formality.

What Črt wrote — the provenance work itself, 14,602 lines of it — I reviewed as a reviewer, and it holds up. The parts I probed rather than read: the `satisfies Record<keyof LearnContext, …>` guard genuinely fails `tsc` when a field is added without classification; the export strip is complete for the five fields it covers; `unscannable` blocks install with no override; the containment check refuses `..`, `.`, `a/../..` and absolute paths.

**Blocking finding — fixed** (`86b79d21`). `exportPack` now strips `PLUR_BOOKKEEPING_KEYS` from `structured_data`. One set decides both halves: a key in it is unscanned and stripped, a key outside it is scanned and kept. Nine regression cases, each confirmed red with the strip removed. The class — export is a denylist — is #1169.

**Merged with current `main`** (`85884377`), which surfaced three integration defects neither branch had alone: `plur forget --scope` would have become a hard failure; `co_injection` would have lost its actor stamp; `leak-surface` became time-dependent once #1138 put timestamps in the engram shape. Each is described in the merge commit.

**#1172 client half** (`2a4f8c83`) — `attribution` and `claim_class` now travel. The server does not accept them yet; that is plur-ai/enterprise#1241, which also covers the three fields #1151 added and the same allowlist drops.

**Still open, and not mine to close**: the stale `#1000` / `#1001` merge note (both closed unmerged), the CHANGELOG entries under the released `## 0.18.0`, and 3 MB of GIFs in git history. None blocks the merge; all three are worth a follow-up pass.

Suite: 378 files passed, 15 skipped, 0 failed. CI 11/11.

PR #1017, opened 2026-08-24
fix(core): cross-process dedup for co_injection events (#975)

### LINKED ISSUE #975: Injections are logged twice, milliseconds apart, in about a third of cases

philphauler:
## Summary

Rework of #977. Claude Code runs `plur hook-inject` as a fresh process per hook event, and two such processes spawned milliseconds apart each logged the same injection (956 near-duplicate `co_injection` pairs in one store, #975). An in-process `Map` can never see the other process, so the dedup now reads the durable history file.

> Maintainer note (2026-09-04): the review fixes described below are on `plur-ai/plur` branch `fix/review-1017` (head `c90cafd2`, fast-forward from this PR's head, merged with main after the architecture audit). They will be pushed to this PR's branch; until then the diff shown here is the pre-review state.

### How it works

- **Key**: `query_hash` + sorted engram id set + `source` + `session_id` + (for PreToolUse / SubagentStart hooks) the harness's per-event id (`tool_use_id` / `agent_id`, capped at 128 chars). Two `Explore` subagents launched in one message have different event ids and stay two records; UserPromptSubmit and PostCompact carry no event id and keep content-only dedup. Absent on both sides is equal; present on one side only is a mismatch, so the key can only get stricter.
- **Window**: 5 s, checked against the last 64 KB of the current and, when the window crosses a month boundary, the previous month's JSONL file.
- **Critical section**: only `source: 'hook'` injections take the per-store `co-injection-dedup` lock (6 retries, 2 ms base, at most ~126 ms of backoff), so the check and the append are one section. MCP `inject` and `session_start` are deliberate calls and never wait on the lock. If the lock cannot be taken the record is written unlocked (fail-open, a racing duplicate may slip through, a record is never lost).
- **Counter follows the record**: `appendHistory` now reports whether the line landed, and `injection_count` advances only when the `co_injection` record was actually written. One reading of the history feeds both, so they cannot drift.
- **Never fails closed**: a malformed history line (`null`, a number, a string, an array, an event without `data`, non-string ids, a non-ISO timestamp, garbage bytes, a >64 KB line) makes the dedup check return `false` and the injection proceeds. `readHistory` shares the same shape guard, which also fixes `plur_history` / `getEngramHistory` throwing on a `null` line. `appendHistory`'s `mkdirSync` is inside the non-throwing path.
- **Per-session guard** (`hook-inject`): the `.injecting` marker is created with `wx` (O_EXCL); a stale marker is claimed by rename and verified by mtime before it is recreated, so a loser cannot carry off the winner's fresh lock. An unwritable directory stays fail-open.

### Repo hygiene

The earlier review commit on this branch had swept `plur init` output (`.claude/settings.json` with absolute home-directory hook paths, and a generated `AGENTS.md`) into the tree. Both are removed. `packages/core/test/repo-hygiene.test.ts` uses `git ls-files` as its oracle and fails if a tracked `settings.json` contains a home-directory path, if the root `.claude/settings.json` or an unlisted root `AGENTS.md` is tracked, or if `.gitignore` loses the two lines.

### Tests

| File | Tests | Covers |
|---|---|---|
| `packages/core/test/injection-dedup.test.ts` | 19 | key semantics (query, id set, source, session, event id), window, month boundary (40-day fixture), order-insensitive ids |
| `packages/core/test/injection-dedup-fail-open.test.ts` | 21 | 15 hostile line shapes against the dedup check, every history reader, and end-to-end hook `inject()`; history dir as a file; month file as a directory; counter and record move together |
| `packages/core/test/injection-dedup-lock-scope.test.ts` | 3 | `inject` / `session_start` / absent source never take the lock; hook path bounded backoff and still records under a live foreign lock |
| `packages/core/test/repo-hygiene.test.ts` | 5 | tracked-file oracle described above, plus a non-vacuity check |
| `packages/cli/test/hook-inject-atomic-guard.test.ts` | 11 | fresh marker held, free marker acquired, 20 contenders → exactly one winner (free and stale), ENOENT → unavailable |

Verification on the fix branch: `@plur-ai/core` project 196 files / 3022 tests pass (5 files / 34 tests skipped); `@plur-ai/cli` + `cli-spawn` 67 files / 636 tests pass; `pnpm typecheck:tests` clean.

### Not changed

`plur init` followed by `plur init --global` leaves `hook-inject` registered in both settings files (the reverse order self-heals). With this branch a double registration is covered either way (UserPromptSubmit/PostCompact by the atomic guard, PreToolUse/SubagentStart by the event id), but `init.ts` itself is untouched. Historical `injection_count` values are not migrated (see CHANGELOG).

Closes #975. Supersedes #977.


miles-on-nightshift:
Thanks for the rework, @philphauler — the cross-process approach (reading the history file instead of in-memory state) is the right fix and the three-key composite (query_hash + sorted IDs + source) closes the edge cases from the first review. Also: #985 just merged — appreciated the clean additive pattern there too.

philphauler:
Eyyy let's go! 🙏🔥🚀

miles-on-nightshift (commented):
**Review — 2026-08-30**

The approach is sound and the six dedup-specific tests are well-structured. The regression is real and the fix is small.

### Blocking: cross-session dedup — receipt-io regression

`test (20/22/24/26)` fail with:

```
FAIL @plur-ai/core  test/receipt-io.test.ts > Plur.receipt() > counts an engram retrieved in two sessions as two pairs
AssertionError: expected 1 to be 2
```

These are NOT pre-existing failures — main CI is fully green right now.

**Root cause:** `receipt-io.test.ts:43` does:
```typescript
await plur.inject('pnpm install', { session_id: 's1' })
await plur.inject('pnpm install', { session_id: 's2' })
expect(r.retrieved.engram_session_pairs).toBe(2)
```

The two injects share the same `query_hash` (same query text) and same engram IDs, and both execute within milliseconds — inside the 5-second window. `isRecentDuplicateInjection` returns `true` for the second call, so the second `co_injection` event is never written. Receipt sees 1 event → `engram_session_pairs = 1`.

The dedup was designed for *the same hook firing twice in one session*, not across sessions. Two different session IDs are not a duplicate — they are two legitimate observations.

**Fix:** add `sessionId` to the dedup key. Events from different sessions should never match:

```typescript
// In isRecentDuplicateInjection signature:
export function isRecentDuplicateInjection(
  root: string,
  queryHash: string,
  engramIds: string[],
  windowMs = 5_000,
  source?: string,
  sessionId?: string,   // add this
): boolean {
  // ...
  for (const line of lines) {
    // ...
    // After the existing checks, before returning true:
    if (sessionId !== undefined) {
      const evSessionId = ev.data.session_id as string | undefined
      if (evSessionId !== sessionId) continue   // different sessions → not a duplicate
    }
    return true
  }
}
```

And in the call site (`index.ts`):
```typescript
const isDuplicate = isRecentDuplicateInjection(
  this.paths.root, queryHash, injected_ids, 5_000,
  options?.source,
  options?.session_id,  // add this
)
```

This prevents cross-session suppression while preserving the same-session dedup that #975 targets. The six existing dedup tests don't pass `session_id` so they'll still pass. Add a seventh test: same query+IDs but different `session_id` → `false`.

### Non-blocking

- `source` as `undefined` means any source matches in the existing check — fine for now, but with `session_id` added the same analysis applies (undefined = any session matches, which is the over-suppression you're fixing). Set session_id mandatory or keep the undefined-means-any logic consistently.
- The existing dedup tests don't exercise the `source` mismatch path. Low priority but worth adding.

philphauler:
Eyyy nice! Let's go 🔥🚀

crtahlin (dismissed):
Requesting changes — the dedup key omits `session_id`, so two different sessions issuing the same query collapse into one event and a real retrieval pair is destroyed. That is what the failing test matrix is telling you.

The four red checks are all the same test, and it is a genuine regression rather than a test problem. A second worktree at `origin/main` passes the same file 7/7.

### Blocking

- `packages/core/src/history.ts:293` — `isRecentDuplicateInjection` compares `query_hash`, sorted engram IDs and `source`, but never `session_id`, so two **different** sessions are treated as one duplicate. Verified with two real processes against a built core: `inject(task, {session_id:'s1'})` then `inject(task, {session_id:'s2'})` writes only the s1 event, and the s2 retrieval is silently destroyed. This is the entire Node matrix failure — `packages/core/test/receipt-io.test.ts:49` gets 1 instead of 2, and `window.sessions` collapses from 2 to 1. It blocks because the receipt's headline metric, `engram_session_pairs`, is exactly the real signal being discarded.

  A fix that works: add a `sessionId` parameter and `if (evSession !== sessionId) continue` before `return true`, then pass `options?.session_id` at the call site. That gives 13/13 passing with same-session cross-process dedup still working, confirmed with real processes. It is safe because both hook inject sites in `packages/cli/src/commands/hook-inject.ts` (lines 336 and 430) read `sessionId` from the same session marker file, so genuine duplicate hook processes always share a session id.

- `packages/core/src/index.ts:4966` — the dedup check reads the history tail and the append happens afterwards with no lock or atomic claim, so two processes can both read before either writes. Measured with two real node processes: naturally concurrent spawns still produced duplicates in **4 of 5 runs**, and with a shared wall-clock barrier **3 of 3 duplicated at 0 ms stagger** while 0 of 3 duplicated at 1, 2, 5, 10 and 25 ms. Since [#975](https://github.com/plur-ai/plur/issues/975) reports real pairs one to four milliseconds apart, the fix catches most reported cases — but it does not close the race it claims to close, and a residual duplicate rate persists undetected.

### Should fix

- `packages/core/src/index.ts:4981` — `injection_count` is deliberately left outside the dedup gate, which makes the store internally inconsistent: after a deduped pair, `engrams.yaml` shows `injection_count: 2` while history holds exactly one `co_injection` event. Verified empirically. [#975](https://github.com/plur-ai/plur/issues/975)'s first stated harm is that "usage data is inflated, and not by a constant factor" — that remains unfixed for this field. The stated justification, that the engram was genuinely injected, contradicts the premise that the two events describe one injection. Pick one reading and make both counters follow it.
- `packages/core/src/history.ts:288` — the source guard is `if (source !== undefined)`, but the written event defaults `source` to `'inject'`. A programmatic `inject()` with no explicit source therefore skips the comparison entirely and can dedup against a `'hook'`-sourced event. Same asymmetry class as the `session_id` defect: compare normalised values on both sides rather than skipping the check when the caller's value is absent.

### Notes

- `packages/core/src/history.ts:243` — the root cause is asserted rather than established. [#975](https://github.com/plur-ai/plur/issues/975) explicitly preferred finding the double call over filtering the log, and suspected "whether session start and inject both fire". The two inject sites in `hook-inject.ts` are mutually exclusive within a process and build different task text (`extractEventTask` yields `skill: X`), which would give different query hashes and never trigger this dedup — so which pair of call sites actually emits identical-query duplicates is still unidentified. Worth confirming before settling for the issue's second-choice fix.
- `packages/core/src/history.ts:236` — only the current month's file is read, so a duplicate pair straddling the UTC month rollover is missed. Fail-open and negligible, but an unstated gap in the window.
- `packages/core/src/history.ts:249` — the 8 KB tail is sized on the assumption that `co_injection` events dominate it ("~12-24 events"). The file interleaves every event type, and injections carrying many long team-scoped IDs run well above the quoted 325-625 B, so a busy tail could push a sibling event out of range. Low risk given the millisecond separation, but the comment overstates the guarantee.
- The chain-head cache concern raised on [#1080](https://github.com/plur-ai/plur/pull/1080) does not apply here — there is no chain-head cache in `history.ts` on this branch, so the dedup read does not interact with it.

### Done well

The bounded 8 KB tail read with a dropped partial first line, plus fail-open on an unreadable file, is the right hot-path discipline; the added cost is negligible next to the existing store-lock block. Keying on `query_hash` plus the engram set rather than the hash alone correctly addresses the blocking finding from the earlier [#977](https://github.com/plur-ai/plur/pull/977) attempt, and the order-insensitive comparison is properly tested. Skipping the `_lastInjectionByEngram` update on the duplicate path is safe rather than merely overlooked — `_logInjectionOutcome` falls back to `findLatestInjectionFor`, which resolves to the surviving event, so feedback linking still lands correctly.

The six new tests in `packages/core/test/injection-dedup.test.ts` are focused and well chosen. The gap is that they are all single-process, which is why neither the cross-session defect nor the concurrency race surfaced there.

Tests: FAIL. `packages/core/test/receipt-io.test.ts` "counts an engram retrieved in two sessions as two pairs" fails with expected 1 to be 2, caused by this PR; `origin/main` passes the same file 7/7. Full core suite otherwise 187 files passed, with only the known host-gitignore issue ([#1062](https://github.com/plur-ai/plur/issues/1062)). Typecheck clean.

Verified: the cross-session collapse reproduced with two real processes and root-caused to the dedup key; a second worktree at `origin/main` used as the control; the proposed fix applied and re-run to 13/13; the concurrency race measured across six stagger values with a shared wall-clock barrier; the counter divergence observed on a scratch store.

crtahlin (dismissed):
Requesting changes on one file only. The fix itself is correct and I verified every finding from the previous round is genuinely closed.

I wrote an independent multi-process probe and first proved it has power by running it against the pre-fix commit `5caad88d`, where it reproduced all four defects. Then I ran it against the head.

| Finding | Pre-fix | Head |
|---|---|---|
| `session_id` missing from the dedup key | 1 event, sessions collapsed to `['s1']` | 2 events, `['s1','s2']` |
| Check-then-append race | 9/9 duplicated at 0 ms barrier stagger; 5/5 natural 4-process spawns produced 2 to 4 events | exactly 1 event in 24/24 barrier runs across 0/0/0/1/2/5/10/25 ms, and 5/5 natural runs |
| Counter divergence | 1 event vs `injection_count: 2` | 1/1 when deduped, 2/2 for two sessions |
| Source compared asymmetrically | a sourceless `inject` deduped against a hook event, giving 1 | 2 events, sources `[hook, inject]` |

The behaviour change in "scope the dedup to hook-sourced injections" is sound and satisfies [#975](https://github.com/plur-ai/plur/issues/975). The hook CLI is the only caller in the tree that spawns a fresh process per event and it passes `source: 'hook'`, while MCP `inject` and `session_start` run in one long-lived process where repeated calls are real injections. That distinction is what let both counters follow one reading without breaking `inject-counter-and-flush-merge.test.ts`.

### Blocking

- `.claude/settings.json` — a developer-local Claude Code config was committed at the repository root, pointing every hook and the MCP server entry at absolute paths under another developer's home directory, which exist on no other machine. It arrived with a generated root `AGENTS.md` in commit `6f222f01`, whose subject is about the dedup critical section, and it applies to every contributor's checkout. Together the two files are 147 of the 767 added lines.

  This is the fourth instance of this class in the repository — `pgprobe.mjs` (`c0459e32`), three `*.tmp.mjs` gate probes (`32412dd5`, five days ago, which added `release.sh` step 3.54 in response), and `mig-seed.mjs`. Step 3.54 matches `*.tmp.*` only, so it does not catch these. Filed the structural fix as [#1110](https://github.com/plur-ai/plur/issues/1110).

  Removing both files leaves the dedup fix untouched. Nothing else blocks.

### Should fix

- `packages/core/src/index.ts:5009` — `recordedInjection = true` is set **before** the `appendHistory` try block, so when the history write fails the counter still increments with no `co_injection` event. Verified by pre-creating the month file as a directory: `inject()` logs the "history could not be written" warning and `engrams.yaml` ends at `injection_count: 1` with zero history events. That is the store-disagrees-with-its-own-history state the commit comment says was eliminated, relocated to the error path. Moving the assignment below `appendHistory` closes it.
- `packages/core/src/index.ts:5037` — `withLock` is tuned `maxRetries: 12, baseDelay: 2`, and its retry loop is a synchronous busy spin, so a fully contended acquisition burns roughly 8.2 seconds of CPU before throwing. The event hooks are installed with a 10-second timeout, so under heavy contention a hook could spend nearly its whole budget spinning and inject nothing. The tuning matches the chain lock on `integration/core-store`, so it is at least consistent, but once both land a single hook pays two such sections.
- The PR description is now materially wrong about the code, and the squash-merge body is generated from it. It still says the read is 8 KB (it is 64 KB), still says `injection_count` is not gated and repeats that in the review-response table (commit `14825777` inverted this), still says six test cases (there are 19), and never mentions `session_id` in the key, the lock, or the hook-only scoping. The last two commits reversed a decision the description still defends.

### Notes

- The branch is not rebased; it is 40 commits behind `origin/main` and based on `3323642c`. `git merge-tree` is clean into both `main` and `integration/core-store`.
- One premise I gave the reviewer was wrong, and it corrected me: the hash-chain work is on `integration/core-store`, not `main`. `origin/main` at `0dec4e6a` has no `verifyChain`, no chain-head cache and no chain lock, so `verifyChain` could not be used as an independent oracle here. The forward-looking claim in the `index.ts` comment is nonetheless accurate — `appendHistory` there takes `withLock(join(historyDir, 'chain'))` while this takes `withLock(join(historyDir, 'co-injection-dedup'))`, two distinct lock files that nest without contending, and the history-directory scans filter on `.jsonl` so the stray `.lock` file is inert.
- `dedup-calibration`, one of the seven green checks, is declared `continue-on-error: true`, so it reports success whatever it finds, and it concerns engram-learn embedder calibration rather than this dedup. Six checks are real evidence, not seven.
- `packages/core/test/injection-dedup.test.ts:139` — the four-process race test resolves the core bundle with `path.join(process.cwd(), 'packages/core/dist/index.js')`, which only works when vitest runs from the repository root. Run from `packages/core` it fails with all four children exiting 1, which reads like a concurrency failure rather than a path problem. It fails loudly rather than passing silently, so this is a usability wart — but it is the one test guarding the concurrency invariant, and a test people avoid running in isolation is a test that rots.
- `packages/core/src/index.ts:5041` — the catch around `withLock` re-invokes `writeCoInjection()`. `withLock` only throws on acquisition failure, before `fn`, so this is normally a clean fail-open; if `fn` ever threw, `withLock` rethrows after its `finally` and the section would run twice. Structure rather than present behaviour.
- Event hooks derive their task text from `extractEventTask` (`"skill: X"`, `"agent: Y"`), so their `query_hash` differs from the main prompt and they are not swallowed by the prompt injection's dedup. The narrow residual case is two identical parallel subagent launches inside 5 seconds, which would now log one provenance record instead of two.
- [#975](https://github.com/plur-ai/plur/issues/975)'s own "where to look" section says to check whether session start and inject both fire. A hook and session_start pair is not deduped under the new scoping, correctly, since deduping across sources would destroy legitimately distinct injections. Before closing [#975](https://github.com/plur-ai/plur/issues/975), it is worth grouping the original 956 near-duplicate pairs by the `source` on both members. If they are overwhelmingly hook-to-hook, close it and record the number; if not, a scoped remainder stays open. [#975](https://github.com/plur-ai/plur/issues/975) also says the better fix is to find the double call and remove it, and this suppresses the symptom without establishing the root cause.
- `injection_count` now accumulates meaningfully more slowly for hook-sourced injections, with no migration of historical values, so anyone reading engram usage across the upgrade sees a discontinuity. Worth a CHANGELOG line — and `release.sh` step 3.6 will require [#975](https://github.com/plur-ai/plur/issues/975) in the CHANGELOG for whatever version ships this, since a `fix(core):` subject is not in the excluded prefix list.

### Done well

`packages/core/src/history.ts:352` — the malformed-timestamp guard is right for the right reason. `now - NaN > windowMs` is false, so the old code treated an unparseable timestamp as in-window and a single corrupt tail line could suppress real records indefinitely. The fix states the asymmetry explicitly: this function may let a duplicate through, it may not eat a real record. Clock-skewed future timestamps fall under the same rule.

`packages/core/src/index.ts:4966` — the reasoning for a separate lock file from [#1051](https://github.com/plur-ai/plur/issues/1051)'s chain lock is correct and I verified it. `withLock` is file-based and non-reentrant, so reusing `history/chain` would have made every `co_injection` fall through to the unlocked path once both landed, leaving the dedup silently inert while still looking fixed. Anticipating an inert-but-green failure mode on a branch that has not merged yet is unusually good.

`packages/core/src/history.ts:280` — the tail read widened from 8 KB to 64 KB with an honest account of why the original sizing was wrong, and the window now spans the previous month file so a pair straddling midnight on the first is not missed. The comment calls the number a heuristic rather than claiming a guarantee.

Tests: PASS — 328 of 345 files pass. The only two failing files fail identically at the pre-fix commit `5caad88d`: the known pre-existing `hook-learn-check` checkpoint test, and `sync-push-failure.test.ts`, which is the host-environment issue in [#1062](https://github.com/plur-ai/plur/issues/1062). This branch introduces zero new failures.

Verified: an independent multi-process probe proven non-vacuous against the pre-fix commit, then run against the head across eight barrier staggers and five natural spawns; the counter-divergence and source-asymmetry cases reproduced on both commits; the history-write failure path forced by making the month file a directory; the lock-file interaction with `integration/core-store` checked by reading both call sites; `git merge-tree` run against both branches.

plur9:
Triage note, not a review: this is the one open PR with **no commits since the change request**. The other eight all have pushes the reviewer has not seen, so their blocks are stale; this one's is live and the ball is with the author.

Last commits are 2026-09-02; the requested changes postdate them. Nothing to re-review until there is a push.

plur9:
@philphauler — I have pushed two commits to your branch rather than leaving this
stalled another week. Shout if you would rather I had not; I can drop them.

**The blocking finding, and an apology.** `.claude/settings.json` points every
hook at absolute paths under `/Users/gregor/...` — my home directory, not yours.
So the developer-local config that blocked your PR came from the maintainer
side. Removed, along with the generated root `AGENTS.md` that arrived with it in
`6f222f01`. That is 147 of the 767 added lines gone, and the dedup work is
untouched.

**The counter/history finding needed to go further than the review suggested,
and I only found that by testing the suggested fix.**

The review proposed moving `recordedInjection = true` below `appendHistory`.
I did that, wrote a regression test that pre-creates the month file as a
directory to make the append fail, and it **still read `injection_count: 1`**.

`appendHistory` deliberately swallows its own failure and returns normally — an
unwritable history directory must not fail the `learn` that called it, and its
own comment argues that well. So nothing is ever thrown, the try/catch never
fires, and the assignment runs either way. The obvious fix is inert.

It reports instead: `appendHistory` now returns a boolean and `inject()` gates
on that rather than on the absence of an exception. Existing callers ignore the
return, so nothing else changes. `_lastInjectionByEngram` is still set on the
failure path, because the engrams genuinely were injected whatever the log
recorded.

Regression test plus a control, so "does not count on failure" and "never
counts" stay distinguishable.

**Not addressed:** the `withLock` tuning note (12 retries × 2 ms baseDelay
spinning ~8.2s of CPU against a 10s hook timeout). It is consistent with the
chain lock on `integration/core-store`, and changing it here would be a
different decision from the one this PR is making — better as its own issue once
both land.

**Note for the reviewer:** `packages/core/src/index.ts:4952` still carries a
comment saying the `injection_count` increment "is NOT gated". That has been
false since this PR gated it at 5070, and it now contradicts the code twice
over. Left alone rather than swept into these commits, but worth a line.

Local: core typecheck clean; `history`, `inject-counter-and-flush-merge` and the
new atomicity tests pass. `injection-dedup`'s four-process test fails in my
worktree for environmental reasons — it fails identically on your unmodified
head and passes in CI, so I do not believe it is related.

Issue #1034, opened 2026-08-26
Extract a HarnessAdapter interface — stop hand-writing the fifth integration


plur9:
Depends on #1031 and #1033 — worth doing once two more adapters exist to generalise from, not before.

Every harness integration is currently hand-written end to end. `init.ts` is ~920 lines with Cursor logic inlined; `doctor.ts` hardcodes "Claude Code / Claude Desktop / Cursor" in its report, its `ConfigFile.kind` union, and its `overall` computation. Adding Codex and Antigravity by the same method means four parallel copies of: config-path resolution, hook-map building, idempotent merge, context-file writing, and a doctor probe — and five copies of the decision about what actually gets injected.

## Shape

Extract a `HarnessAdapter` interface over Claude Code / Cursor / Codex / Antigravity / Gemini:

- `detect(cwd)` — is this harness present?
- `configPaths()` — the files this harness owns
- `buildHooks(cmd)` — harness-native hook map
- `mergeHooks(existing, additions)` — idempotent, strip-then-add
- `writeContextFile()` — CLAUDE.md / .mdc / AGENTS.md / GEMINI.md
- `probe()` — what `plur doctor` reports for this harness

Then one shared injection core behind thin per-harness shims, so a change to *what* gets injected lands everywhere at once instead of in four places.

## Also

Generalise `plur doctor` from a fixed three-harness report to "whichever harnesses are detected", reporting per-harness wiring rather than folding everything into one `overall` boolean. The current `cursorWired`-shaped special case does not scale to five.

## Test the combination

`plur init --cursor` already ships an explicit "the two integrations run independently and haven't been tested together" caveat. Adding two more independent sentinel schemes over the same `~/.plur/` store makes that worth actually testing rather than documenting — particularly session-sentinel collisions when one project is opened in two harnesses.

Issue #1033, opened 2026-08-26
Antigravity CLI (agy) adapter — probe done, contract verified


plur9:
A Gemini CLI user reports PLUR "not really working — very few recalls and learnings". They are not misconfigured: we have never shipped an adapter for that harness, so they get MCP tools and no injection, no enforcement, no always-on instruction block.

Two complications shape the fix.

## 1. Gemini CLI is being replaced by Antigravity CLI (`agy`)

Google announced the transition in May 2026; consumer access for Gemini CLI and the Code Assist IDE extensions changed on 18 June 2026, with enterprise access unchanged. Antigravity CLI is the Go-based successor and explicitly keeps Skills, Hooks, Subagents, and Extensions (rebranded as plugins). Anyone hitting auth trouble on `gemini` today is being pointed at `agy`.

**So `agy` is the primary target and `gemini` is the compatibility path** — not the other way round. Both are worth supporting: enterprise Gemini CLI users still exist, and their config surface is well documented.

## 2. The two have different hook models

**Gemini CLI** — hooks in `settings.json` (`.gemini/settings.json` > `~/.gemini/settings.json` > `/etc/gemini-cli/settings.json` > extensions), same nested container shape as Claude Code. Events: `SessionStart`, `BeforeAgent`, `AfterAgent`, `BeforeModel`, `AfterModel`, `BeforeToolSelection`, `BeforeTool`, `AfterTool`, `PreCompress`, `Notification`, `SessionEnd`. Injection is `hookSpecificOutput.additionalContext` on `SessionStart` / `BeforeAgent` / `AfterTool`; guard is `BeforeTool` with `decision: "deny"` + `reason`.

**Antigravity CLI** — only five events: `PreInvocation`, `PostInvocation`, `PreToolUse`, `PostToolUse`, `Stop`. Different container shape (keyed by a hook-set *name* first, then event). Paths: workspace `.agents/hooks.json`, global `~/.gemini/config/hooks.json`, plus plugin-bundled `hooks.json`. Plugins live at `~/.gemini/antigravity-cli/plugins/<name>/` with `plugin.json`, `mcp_config.json`, `hooks.json`, `skills/`, `agents/`, `rules/`.

There is no `SessionStart` in Antigravity — `PreInvocation` is the injection point.

## Step 0 is a probe, not code

[...]

Issue #1032, opened 2026-08-26
Codex mcp_tool hooks are a stub — parses, dispatches, never calls the tool (re-probe on updates)


plur9:
Codex's hook schema advertises a `mcp_tool` handler type — a hook that calls an MCP tool directly with `${tool_input.field}` templating, no subprocess:

```json
{ "type": "mcp_tool", "server": "plur", "tool": "plur_session_start", "input": {...}, "timeout": 30 }
```

If it worked, PLUR's Codex `SessionStart` could route straight into `plur_session_start` with no shim process, no CLI cold start, and no shell-command trust prompt. Worth tracking, because it is strictly better than the command-hook path.

## It does not work in codex-cli 0.149.1

Probed 2026-08-27 against a real `plur` MCP server (stdio, `packages/mcp/dist/index.js`):

- The config **parses** — `server` / `tool` / `input` are accepted field names, and a deliberately bogus extra key is ignored rather than rejected, so serde is not strict here.
- The MCP server **connects** — `rmcp` reports `Service initialized as client ... name: "plur-mcp"` in the trace.
- The hook **dispatches and then fails immediately** — `hook: SessionStart Failed` after ~0.5s, with no reason surfaced anywhere: not in `--json` output, not in `RUST_LOG=trace`, not in a spill file.
- **`tools/call` count in a full trace: 0.** Codex never issues the MCP call. It fails before dispatch.

Reproduced on both `SessionStart` and `UserPromptSubmit`, with both an argument-taking tool (`plur_recall`) and a no-arg one (`plur_status` with `input = {}`). A `command` hook in the same config, same session, same trust settings, works fine — so this is not a trust, path, or config-parse problem.

Supporting evidence from the binary: there are embedded JSON schemas for `*.command.input` / `*.command.output` for every event, and **none** for any `mcp_tool` variant. The discovery code has `mcp_tool`-specific warnings (`server and tool must not be empty`, `SessionEnd MCP hooks are not supported`), so the feature is partially wired, not absent — but the execution path is not there yet. Sibling handler types `prompt` and `agent` still warn "not supported yet" outright.

## Action

Do **not** design the Codex adapter (#1031) around it. Re-probe on Codex updates; the surface is moving fast (0.146.0 → 0.149.1 in a single day changed async hook behaviour). If it starts working, `SessionStart` and `SubagentStart` are the two hooks worth migrating.

[...]

PR #1035, opened 2026-08-26
Harness adapters: Codex + Antigravity, hybrid injection, PGLite opt-in (#1031, #1033, #1040, #1046)

### LINKED ISSUE #1031: Codex CLI adapter: hooks + MCP + AGENTS.md, so PLUR actually injects and enforces there
PLUR ships an MCP server to Codex users and nothing else. No engram injection, no session enforcement, no learn nudges, no always-on instruction block — every mechanism that makes PLUR work in Claude Code is harness-specific, and `plur init` only knows about Claude Code and Cursor. Recall and learning end up depending on the model spontaneously calling the tools, which is exactly the failure mode `ENG-2026-0401-001` records.

Codex ships a hook system that is close to a straight port of Claude Code's, so this is the cheapest adapter we will ever build.

## Verified against codex-cli 0.149.1 (macOS, 2026-08-27)

- `~/.codex/hooks.json` is read, with the **exact** Claude Code nested shape: `{ "hooks": { "<Event>": [ { "matcher", "hooks": [ { "type", "command", "timeout" } ] } ] } }`. No TOML dependency needed — the existing JSON merge helpers in `init.ts` apply directly.
- Events: `PreToolUse`, `PermissionRequest`, `PostToolUse`, `PreCompact`, `PostCompact`, `SessionStart`, `SessionEnd`, `UserPromptSubmit`, `SubagentStart`, `SubagentStop`, `Stop`.
- stdin payload is snake_case and matches Claude Code: `session_id`, `turn_id`, `transcript_path`, `cwd`, `hook_event_name`, `model`, `permission_mode`, `tool_name`, `tool_input`, `tool_response`, `source`.
- stdout is `hookSpecificOutput.additionalContext` — **verified reaching the model** from both `SessionStart` and `UserPromptSubmit` in the same turn. Unlike Cursor, no `.mdc`-rewrite workaround is needed.
- `PreToolUse` returning `permissionDecision: "deny"` + `permissionDecisionReason` — **verified blocking a tool**, with the reason reaching the model. (Codex rejects `allow` and `ask` from a hook; deny-only, same as our Cursor guard.)
- `codex mcp add plur --env K=V -- <cmd>` registers the server without us hand-editing `config.toml`.

## Two traps that must be designed around

**1. Async injection arrives a turn late.** `async: true` runs (it was silently skipped in 0.146.0 and now works in 0.149.1), but the injected `additionalContext` does **not** reach the triggering turn — docs say "next safe point". In a `codex exec` one-shot that means never. Our Claude Code map puts `async: true, timeout: 90` on `hook-inject` precisely because the BGE embedder cold-starts 20s+. Copying that across produces a config where injection is always one turn behind.

→ Codex injection must be **synchronous and BM25-only**, the same call we already made for `hook-cursor-session-start` (0.74s against 4,290 engrams). Add a regression test asserting no emitted Codex hook carries `async: true`.

**2. Untrusted hooks are skipped in total silence.** Codex fingerprints every non-managed hook and refuses to run it until reviewed. In `codex exec` there is no review UI, so the run produces no hook output, no warning, and exit 0 — indistinguishable from a config that was never read. Project `trust_level = "trusted"` does not cover it, and a hand-written `trusted_hash` that doesn't match is also skipped silently with no hint of the expected value.

→ `plur init --codex` cannot end on "restart Codex". It has to end on an explicit instruction to run `/hooks` once in the TUI and trust the PLUR entries, and `plur doctor` needs to be able to report "wired but possibly untrusted".

## Scope

- `plur init --codex` (auto-detect `~/.codex/`), writing `~/.codex/hooks.json` + `codex mcp add` + an `AGENTS.md` section reusing `CLAUDE_MD_SECTION`.
- Four hook commands: `hook-codex-{session-start,inject,guard,post-tool}`.
- Idempotent strip-and-remerge, mirroring `mergeCursorHooks`.
- `plur doctor`: `codexDetected` / `codexWired` + the trust caveat.
- Tests, including the no-`async` regression.

## Out of scope

`type = "mcp_tool"` hooks — see #1032.

plur9:
Closes #1031.

Codex users got an MCP server and nothing else — no injection, no enforcement, no learn nudges, no always-on instruction block. This adds the adapter, reusing the Claude Code shapes rather than inventing a third set.

## What lands

- `plur init --codex` (auto-detects `~/.codex/`, honours `CODEX_HOME`)
- `~/.codex/hooks.json` — five hooks: `SessionStart`, `UserPromptSubmit`, `PreToolUse` (guard), `PostToolUse` (sentinel + learn nudge), `SessionEnd`
- MCP via `codex mcp add` rather than hand-writing TOML
- `AGENTS.md` section, reusing `CLAUDE_MD_SECTION`
- `plur doctor`: `codexDetected` / `codexWired` + the trust caveat
- README: Codex section, plus a per-harness table separating "has tools" from "has auto-injection"

Hooks go in `hooks.json` rather than a `[hooks]` table in `config.toml`: Codex accepts either, the JSON file takes the exact nested `{matcher, hooks: []}` shape the Claude Code builders already emit, and it keeps PLUR out of the file users hand-maintain for models and MCP servers. No TOML dependency.

## Three findings from probing 0.146.0 and 0.149.1 that shaped this

**Injection is synchronous and BM25-only.** Codex supports `async: true` as of 0.149.1 (0.146.0 skipped async hooks silently), but an async hook's `additionalContext` is delivered at the "next safe point" — not to the turn that triggered it, and in `codex exec` never. Claude Code's `async: true, timeout: 90` on `hook-inject` does not transfer. There is a regression test asserting no emitted Codex hook carries `async`.

**A non-zero exit discards the hook's output** — and on `PreToolUse`, exit code 2 *blocks the tool* with stderr as the reason. This is not theoretical: on a machine with a corrupt PGLite index, every `plur` command exits 1 from a WASM abort during teardown, long after the hook has written valid JSON. Codex reported `SessionStart Failed` and dropped a complete, correct injection. `runCodexHook()` catches everything, keeps diagnostics on stderr (stdout is parsed as the result), flushes, and forces exit 0.

**Untrusted hooks are skipped in total silence** — no warning, no output, exit 0, indistinguishable from a config Codex never read. Project `trust_level = "trusted"` does not cover hooks, and `trusted_hash` cannot be pre-seeded (a wrong value also fails silently, and Codex will not reveal the expected one). So init ends on the `/hooks` trust step rather than "restart Codex", and doctor repeats the warning *under a green tick*, because a green tick genuinely does not mean the hooks run.

## One deliberate asymmetry

`codexWired` is **not** folded into doctor's `overall`, unlike `cursorWired`. A `.cursor/` directory is per-project evidence the user works on this project in Cursor; `~/.codex/` is machine-level and says only that Codex is installed somewhere. Gating `overall` on it would turn doctor red for every Claude-Code-only user who once tried Codex — a false alarm on a check whose whole value is that red means something.

## Verification

Verified end to end against codex-cli 0.149.1 on macOS: `plur init --codex` into a scratch `CODEX_HOME`, then `codex exec` reporting **9 injected engrams** and quoting their content back. `PreToolUse` deny was verified separately — it blocks the tool and the reason reaches the model.

59 new tests. Full suite run before and after: identical set of failing files (10, all pre-existing in core/mcp, from better-sqlite3/PGLite issues on this machine) — no new failures.

## Follow-ups

- #1032 — Codex `mcp_tool` hooks are a stub (parses, dispatches, never issues `tools/call`); do not design around it yet
- #1033 — Antigravity CLI + Gemini CLI adapter
- #1034 — extract a `HarnessAdapter` interface once there are enough adapters to generalise from

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,14 +52,16 @@ fresh-process-per-hook model.
 after upgrading, selection moves to `sqlite` and builds `engrams.db`. Nothing is
 lost — YAML remains the source of truth.
 
-**Migration path:** run `plur migrate` once. It carries your embedding vectors out
-of the old PGLite store into the new tier's cache (verifying each against the
-engram's current text and the active embedder's dimension), so hybrid recall works
-immediately instead of re-embedding the corpus in the background. It then tells you
-the old `~/.plur/store.pglite/` directory (typically 60–500MB) is safe to delete;
-`plur doctor` flags it too. Skipping the migration loses nothing — stale or skipped
-vectors re-embed automatically — it just costs background CPU and a window of
-BM25-only recall.
+**Migration: nothing to do.** Recall keeps working immediately (BM25), and
+embeddings rebuild automatically — the first hybrid recalls re-embed the corpus,
+which takes a few minutes of background CPU on a large store and then it is done.
+The old `~/.plur/store.pglite/` directory (typically 60–500MB) is orphaned and safe
+to delete whenever convenient; `plur doctor` points it out.
+
+Optional shortcut for large stores: `plur migrate` carries the old store's
+embedding vectors straight into the new tier's cache (each verified against the
+engram's current text and the active embedder's dimension), skipping the rebuild
+window entirely.
 
 To keep the old behaviour set `backend: pglite` in `~/.plur/config.yaml` or
 `PLUR_BACKEND=pglite`; PGLite remains the right choice only where its pgvector/AGE
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,12 +2,11 @@
 
 ## 0.19.0 (unreleased)
 
-Memory in every harness — Codex and Antigravity join Claude Code, Cursor, OpenClaw and Hermes.
+Codex and Antigravity join the family.
 
-- `plur init --codex` — hooks, MCP, AGENTS.md
-- `plur init --antigravity` — agy hooks, MCP, AGENTS.md
-- Hybrid injection (BM25 + embeddings) in hooks, with automatic BM25 fallback
-- PGLite is opt-in; the size ladder is now yaml → sqlite → postgres
+- `plur init --codex` / `--antigravity`
+- Hybrid recall in hooks
+- Size ladder: yaml → sqlite → postgres
 
 **Codex CLI adapter.** `plur init --codex` wires `~/.codex/hooks.json` (five lifecycle
 hooks), registers the MCP server via `codex mcp add`, and adds a PLUR section to
@@ -68,7 +67,7 @@ To keep the old behaviour set `backend: pglite` in `~/.plur/config.yaml` or
 capabilities are the point, and it now logs its per-process boot cost once at
 startup when explicitly selected.
 
-## 0.18.0 (unreleased)
+## 0.18.0
 
 Your agents' memory, on a dashboard.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,62 @@
 # Changelog
 
+## 0.19.0 (unreleased)
+
+Memory in every harness — Codex and Antigravity join Claude Code, Cursor, OpenClaw and Hermes.
+
+- `plur init --codex` — hooks, MCP, AGENTS.md
+- `plur init --antigravity` — agy hooks, MCP, AGENTS.md
+- Hybrid injection (BM25 + embeddings) in hooks, with automatic BM25 fallback
+- PGLite is opt-in; the size ladder is now yaml → sqlite → postgres
+
+**Codex CLI adapter.** `plur init --codex` wires `~/.codex/hooks.json` (five lifecycle
+hooks), registers the MCP server via `codex mcp add`, and adds a PLUR section to
+`AGENTS.md`. One manual step: Codex refuses untrusted hooks *silently* — run `/hooks`
+in Codex once and trust the PLUR entries. `plur doctor` reports Codex wiring and
+repeats that caveat.
+
+**Antigravity CLI (agy) adapter.** `plur init --antigravity` wires agy's global config
+(`~/.gemini/config/`): a `plur-memory` hook set (PreInvocation injection + PreToolUse
+guard), the MCP server, and `AGENTS.md`. No trust step — restart agy and memory flows.
+Per-prompt recall reads the conversation transcript (agy's hook payload carries no
+prompt text), and each turn's memory is re-injected as an ephemeral message on every
+model invocation so it survives tool calls without accumulating in history. Google is
+transitioning Gemini CLI to Antigravity; Gemini CLI itself remains tools-only.
+
+**Hybrid injection in hooks.** Hook injection is now hybrid (BM25 + embeddings) with an
+automatic BM25 fallback on an 8s soft deadline — measured ~4.7s hybrid vs ~1.6s BM25 on
+a 5,775-engram store, with hybrid diverging most exactly on the vague prompts where
+memory matters. `PLUR_HOOK_HYBRID=0` forces BM25; `PLUR_HOOK_HYBRID_DEADLINE_MS` tunes
+the deadline (keep it under your harness's hook timeout: Codex 25s, agy 20s). Enabled
+by the `@huggingface/transformers` 3.8.1 → 4.2.0 upgrade, which fixes a SIGABRT during
+ONNX teardown (#1040) that made every embedder-touching process exit 134 — and, in
+3.8.1, could leave a truncated model file that permanently and silently degraded search
+to BM25 (#340's failure mode; 4.2.0 downloads atomically). Embedding vectors are
+bit-identical across the upgrade — no re-embed, caches stay valid.
+
+**PGLite sync is no longer quadratic with your session count.** `syncFromYaml` batches
+upserts (a 5,000-engram corpus builds in ~2s instead of 10+ minutes), records a
+fingerprint so an unchanged store skips the sync entirely, tolerates duplicate ids
+last-wins, and the CLI drains background index work before exiting so the index
+actually converges.
+
+### BREAKING — PGLite is opt-in, never selected by corpus size (#1046)
+
+The automatic ladder is now **yaml → sqlite → postgres**. Stores past 5,000 engrams
+select a SQLite metadata index (sub-millisecond opens measured to 500k engrams)
+instead of PGLite, which boots a full Postgres-in-WASM per process — measured at
+0.61s vs 300s+ per command on a 5,775-engram store under PLUR's
+fresh-process-per-hook model.
+
+**Who is affected:** any default install with more than 5,000 engrams. On first run
+after upgrading, selection moves to `sqlite` and builds `engrams.db`. Nothing is
+lost — YAML remains the source of truth — but the old `~/.plur/store.pglite/`
+directory (typically 60–500MB) is orphaned and can be deleted. `plur doctor` flags
+it. To keep the old behaviour set `backend: pglite` in `~/.plur/config.yaml` or
+`PLUR_BACKEND=pglite`; PLUR remains the right choice only where its pgvector/AGE
+capabilities are the point, and it now logs its per-process boot cost once at
+startup when explicitly selected.
+
 ## 0.18.0 (unreleased)
 
 Your agents' memory, on a dashboard.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,10 +50,19 @@ fresh-process-per-hook model.
 
 **Who is affected:** any default install with more than 5,000 engrams. On first run
 after upgrading, selection moves to `sqlite` and builds `engrams.db`. Nothing is
-lost — YAML remains the source of truth — but the old `~/.plur/store.pglite/`
-directory (typically 60–500MB) is orphaned and can be deleted. `plur doctor` flags
-it. To keep the old behaviour set `backend: pglite` in `~/.plur/config.yaml` or
-`PLUR_BACKEND=pglite`; PLUR remains the right choice only where its pgvector/AGE
+lost — YAML remains the source of truth.
+
+**Migration path:** run `plur migrate` once. It carries your embedding vectors out
+of the old PGLite store into the new tier's cache (verifying each against the
+engram's current text and the active embedder's dimension), so hybrid recall works
+immediately instead of re-embedding the corpus in the background. It then tells you
+the old `~/.plur/store.pglite/` directory (typically 60–500MB) is safe to delete;
+`plur doctor` flags it too. Skipping the migration loses nothing — stale or skipped
+vectors re-embed automatically — it just costs background CPU and a window of
+BM25-only recall.
+
+To keep the old behaviour set `backend: pglite` in `~/.plur/config.yaml` or
+`PLUR_BACKEND=pglite`; PGLite remains the right choice only where its pgvector/AGE
 capabilities are the point, and it now logs its per-process boot cost once at
 startup when explicitly selected.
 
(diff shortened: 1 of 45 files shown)


plur9:
Pushed b97a062 — hybrid injection, gated off.

Following up on "why not hybrid on `UserPromptSubmit` too?". It should be, and the reason it currently isn't is not the one the code comments give.

**The 20s figure is stale.** Measured on a 5,775-engram store, fresh node process per run: BM25 **1.6s**, hybrid cold **4.7s**, hybrid warm **1.7s**. The "~20s to load the BGE embedder in a cold CLI process" justification repeated in `hook-inject.ts`, `hook-cursor-session-start.ts` and `codex-hooks.ts` is ~4x pessimistic today, and it is the stated reason all three are BM25-only. Worth re-deriving there too.

**And hybrid earns the 3s.** Injected-set overlap vs BM25 across realistic prompts: ~90% on keyword-rich phrasing, ~78% on vague phrasing ("this keeps breaking in the same way", "what did we agree on"). Divergence doubles exactly where the user is leaning on memory rather than supplying keywords.

So `injectWithFallback()` does hybrid-first against a soft deadline with BM25 fallback — a slow embedder degrades to worse results, never to a hook Codex kills (which injects nothing). Hook timeouts 15s → 25s to cover the bounded worst case.

**It ships disabled**, because turning it on found something bigger: **#1040**. Loading the ONNX embedder kills the process with SIGABRT (exit 134) in native teardown, *after* the JS finished and `process.exit(0)` ran:

```
bm25   -> exit=0    reached-end=1
hybrid -> exit=134  reached-end=1    libc++abi: mutex lock failed
```

Deterministic 4/4. Survives `PLUR_BACKEND=yaml` and `=sqlite`, so not PGLite. Survives forcing the BM25 fallback with a 1ms deadline — *starting* the embedder is enough, nothing needs to come back from it. That is why the kill switch guards the call site rather than the result: "call it and fall back" isn't safe when the call is what crashes.

Since Codex discards the output of a non-zero hook, hybrid today produces a correct 2,869-byte payload that gets thrown away. Note **Claude Code's `hook-inject` uses `injectHybrid` on its main path**, so it is exiting 134 on every first-message injection on an affected machine — unnoticed only because Claude Code doesn't gate on hook exit codes. Possibly related to #504 / the orphaned-process pathology.

Flip the default in `hybridEnabled()` once #1040 lands. Full CLI suite: 550 passed, 1 pre-existing failure (`import.test.ts`, fails on main).

plur9:
Pushed bdd5fa2 — **#1040 fixed, hybrid now on by default, and the whole test suite is green.**

## #1040 was upstream

`@huggingface/transformers` 3.8.1 (onnxruntime-node 1.21.0) aborts on teardown; **4.2.0 (onnxruntime-node 1.24.3) exits 0**. Bisected by elimination — not PGLite (reproduces with the store deleted), not our usage (reproduces in ~15 lines against the library directly), not the exit path (`dispose()`, `numThreads=1`, `reallyExit(0)` all still abort), not asynchronous (idles 5s, then dies on exit).

Safe despite being a major bump: `bge-small-en-v1.5` produces **bit-identical vectors** across both versions, so the embedding cache stays valid — no re-embed, no recall shift.

Hybrid is now the default. `PLUR_CODEX_HYBRID=0` forces BM25.

## I overstated the blast radius earlier — correcting it

Claude Code's `hook-inject` exits **1, not 134**, and still wrote 83,624 bytes with 116 engrams injected. Injection was working the whole time; only the exit code was wrong, and Claude Code doesn't gate on it. The user-visible breakage was confined to Codex.

## Every pre-existing failure was environmental, not a bug

Two causes behind all nine failing files:

- **`better-sqlite3` native binding compiled against `NODE_MODULE_VERSION` 147 while this Node needs 137** — built under a different Node in a multi-version setup, dated 15 June. Reinstalled the prebuilt for the current ABI. Fixes `importers`, `multi-store`, `read-side-scope-visibility`, `scope-pushdown`, `storage-indexed`, `store-contract`, `pr1-indexed-migration`, `import` — 8 files, ~54 tests.
- **`packages/mcp/dist` was stale**, reporting 0.17.2 against a 0.18.0 `package.json` — precisely the drift `dist-validation.test.ts` exists to catch. Rebuilt.

```
Test Files  333 passed | 15 skipped (348)
     Tests  4575 passed | 159 skipped (4734)
```

Zero failures. Verified live: `codex exec` reports `[PLUR Memory — 5 engrams recalled for this prompt via hybrid]`, both hooks `Completed`.

## Worth a follow-up

A regression test asserting `exit === 0` after `injectHybrid()` **in a spawned process** would have caught #1040 at the version bump. In-process tests are blind to this class of failure — the JS all passes.

plur9:
Pushed 618e05a — **the Antigravity CLI adapter (#1033), completing the scope this PR set out to cover.** The harness the original "PLUR isn't working" report pointed at now gets injection, enforcement and learn nudges, not just tools.

Everything contract-level was verified live against agy 1.1.21; the authoritative reference turned out to be bundled with the CLI itself (`~/.gemini/antigravity-cli/builtin/skills/agy-customizations/docs/`), not the web docs.

**Design notes worth reviewing:**

- **One event carries everything.** agy has no session event and no per-prompt hook; `PreInvocation` is the entire injection surface. The payload has no prompt text, so per-prompt recall reads the last `USER_INPUT` from the transcript JSONL the payload points at.
- **Ephemeral replay.** Ephemeral messages expire with the invocation that received them — verified live when a model quoted the memory block on invocation 0 and reported "no PLUR Memory block" one tool call later. Recall runs once per user turn; the rendered message is cached and replayed on each invocation of that turn. Memory survives tool use, recall stays once-per-turn, history stays clean.
- **Ownership is a named key.** agy's config is a map of named hook sets; PLUR owns `"plur-memory"` and passes every other set through byte-identical. No command-string fingerprinting.
- **Install-is-the-opt-in gating.** agy runs hooks with cwd = the hooks.json dir, where the `isPlurConfigured()` walk can never succeed — the explicit `plur init --antigravity` is the opt-in, with the per-project opt-out honoured when the payload names a workspace.
- **Fail-open fix in shared plumbing** (adversarial-audit finding): `incrementCounter()` swallowed write failures and returned 1, so an unwritable session dir made the guard's give-up threshold unreachable — every tool call denied forever. An unpersistable counter now reports itself as exceeded; guards fail open.

Also independently corroborated by a dedicated verification agent: transformers 3.8.1 → 4.2.0 produces **bit-identical vectors for all four local embedders** (bge-small, bge-base, minilm, embedding-gemma; sha256 of raw float bytes, adversarial inputs included), so the embedding cache stays valid. Its one standing observation, filed for later: nothing stamps the inference stack into the cache identity, so a future upgrade that DID move vectors would be silent — cheapest fix is folding a version marker into `meta.embedder_name`.

Full suite: **4,600 passed, 0 failed.** Live verification: agy first turn injects via hybrid, deny reaches the model, memory quoted after a shell command.

With #1033 done, the branch now closes #1031, #1033, #1040 and the fix-side of #1046. Remaining before release: version bump (there is a breaking change: PGLite no longer size-selected), CHANGELOG, and review.

plur9:
Pushed 88f8e16 — **all three pre-release audits complete, every blocking and high finding fixed.**

## Audit verdicts

**Data-loss** — the two architecture-level verdicts came back SAFE: the fingerprint guard cannot corrupt recall (candidates always load from YAML; a stale index degrades ranking only, and equality comparison makes it immune to backup restores and clock jumps), and the transformers 4.2.0 bump is *confirmed as the fix* for a live corruption hazard — 3.8.1 wrote model downloads directly to the final path with zero integrity check on read-back (#340's exact silent-degradation mode); 4.2.0 downloads tmp-then-rename atomically.

**Adversarial** — the earlier counter fix verified holding under the original attack (0 denies across chmod 0500/0000/dir-as-file, both hook families). 15 attack classes failed outright: 182 hostile-stdin runs all exit 0 with valid-JSON-or-empty stdout, path traversal contained, unicode-normalisation collisions absent, init byte-identical on re-run, `resolveBackendTier` total under every abuse.

**Evaluator** — no behavioural drift between the duplicated agy/codex helpers (diffed body-for-body), README claims verify against the code, no public API leaks.

## What the audits caught that I had missed — the highlights

- **The `unref()`'d deadline timer defeated the exit-0 guarantee** (exit 13 before `process.exit(0)` — the exact discarded-output failure the wrapper exists to prevent). One-token cause, now referenced.
- **Duplicate ids in one YAML permanently aborted the PGLite sync**, and the documented recovery hit the same error. De-duped last-wins, pinned by test.
- **The agy turn cache wrote recalled engram text 0644 into a 0755 tmpdir** — world-readable on shared /tmp — with path-collision and symlink-plant exposures on top. Now 0700/0600 with ownership vetting, id-validated records, fail-open everywhere.
- **agy injection silently ignored `.plur.yaml`** (resolved from cwd = `~/.gemini/config`), so every agy session recalled and learned unscoped — this repo's own definition of the worst memory failure. Now resolved from the payload's workspace path.
- **`PLUR_CODEX_HYBRID=false` silently left hybrid ON** — an escape hatch that fails during the incident it exists for. Tolerant parsing + renamed `PLUR_HOOK_HYBRID` (it governs agy too); Codex names stay as aliases. Verified live in all three forms.
- **`plur init` destroyed valid-JSON-wrong-shape hooks.json while reporting success**, and one failing harness leg aborted and silenced all the others. Shape checks + per-leg containment.
- **The AGENTS.md block promised a SessionEnd auto-close safety net that neither Codex nor agy has** — shipped-to-users' disk text telling the model not to bother with `plur_session_end`. AGENTS.md now gets its own honest variant.

Plus: CHANGELOG 0.19.0 with the breaking-change migration note and a doctor advisory for the orphaned `store.pglite` cohort, an in-place ADR-0005 amendment, `reloadConfigIfChanged` given the same `.partial()` fix as the constructor, `SQLITE_MIN_ENGRAMS` exported, doctor probe timeout 60s→300s (the only path that can download the model), and 25 new tests including command-level coverage for the agy turn-cache state machine.

**Suite: 4,627 passed, 0 failed.** Remaining before publish: the release.sh version bump to 0.19.0 and review.

plur9:
Pushed cb53f68 — **the breaking change now has a real migration path, not just a notice.**

The gap: a former PGLite user's embedding vectors live only inside `store.pglite` — the switch to SQLite left them with an empty embeddings cache and a full background re-embed (minutes of CPU, hybrid silently BM25-only meanwhile). Docs told them about the orphaned directory; nothing carried the data.

**`plur migrate` now ports the vectors**, with per-vector validation because the two stores key them incompatibly:
- freshness — the row's md5 content-hash must match the engram's *current* YAML text (stale vectors would silently poison ranking; they re-embed instead)
- dimension — against the active embedder, metadata-only
- re-keyed under the cache's own sha256-16 discipline on write; existing entries win; safe to re-run; read-only on the PGLite store

Verified three ways: unit tests (fresh/stale/orphaned rows, re-run idempotence), a live seeded store through the real CLI (`ported: 2, stale: 1`), and — usefully — a genuinely corrupt 398MB pre-fix store from this machine, where it degraded exactly as designed: `status: "failed"` with a warning, schema migrations still applied, clean exit, corpus re-embeds.

Two upstream surprises encoded in comments, worth knowing beyond this PR:
1. **`PGlite.create(barePath)` and `new PGlite('file://' + path)` open different data directories** in this PGlite version — which retroactively explains the "phantom empty store" readings during the original #1046 investigation.
2. **The AGE extension prepends `ag_catalog` to the search_path during init**, so every table the adapter ever created on an AGE-enabled install lives in `ag_catalog`, not `public`. Any external tooling that reads these stores with a plain connection sees "relation does not exist" on a store that holds the data. The export discovers the schema from `pg_tables`; the adapter's own placement is a latent follow-up for #1046.

CHANGELOG's breaking section now leads with the migration path; `plur doctor`'s orphan advisory points at `plur migrate` before suggesting deletion. Suite: **4,629 passed, 0 failed**.

Issue #1019, opened 2026-08-26
Resolve the seven pack contradictions between the specs, the schema and the reference


crtahlin:
Seven places where the standard, the profile, the generated schema and the reference disagree about packs. Each is small; together they are what an independent implementer trips over first.

This is filed ahead of the code work because `plur-ai/encode` is implementing §5 **directly in Python** (encode#32) rather than calling core. It is the first party to build from the document alone, and it has already relied on one of these being true when it is not.

## The seven

**1. The manifest is not as open-world as it is advertised.** `ENGRAM-STANDARD-v1.md:892` says *"Unknown manifest fields MUST be preserved (`additionalProperties: true`)"*. In `spec/pack-manifest.schema.json` the root is `true`, but **`metadata` and `x-datacore` are both `false`** — verified. `PackManifestSchema` (`packages/core/src/schemas/pack.ts:3`) is a plain `z.object()` with no `.passthrough()`, unlike `EngramSchemaPassthrough`. So a pack carrying a future `metadata` field is rejected by the published schema and silently dropped by `manifestToSkillMd` on install.

encode#32 states "the manifest is open-world (`additionalProperties: true`)" and plans an `x-encode` extension block. That happens to be safe because it sits at root. Anyone putting an extension under `metadata` gets a different answer.

**2. §5.5 forbids re-serialising before hashing; install re-serialises.** `ENGRAM-STANDARD-v1.md:610`: *"Hashing is over raw file bytes, so producers and consumers MUST NOT re-serialize before hashing."* `_installPackDir` runs `sanitizePackEngrams` and re-saves `engrams.yaml` (`packs.ts:789-798`) before computing the hash (`packs.ts:806`). The registry's integrity value is the hash of a file we rewrote. The code explains its reasoning at `packs.ts:1601-1603`; the spec has no sanitisation concept at all, so a second implementation following §5.5 literally disagrees with us on every pack containing a pinned or locked engram.

**3. §9 still describes a single sidecar file.** `ENGRAM-STANDARD-v1.md:831`: *"A pack MAY carry a sidecar `provenance.jsonld` inside the capsule payload."* The profile corrects this to a directory (`ENGRAM-PROVENANCE-PROFILE.md` §5.3.1) and says why. But §9 of the standard governs where the two overlap, so an implementer reading §9 alone builds the wrong thing.

[...]

Issue #1030, opened 2026-08-26
Stripping pinned from packs is right, and it silently disabled the bundled pack's whole purpose


crtahlin:
Found while resolving #1019's fifth contradiction — the bundled pack documenting behaviour the installer strips. Correcting the documentation exposed a capability that has quietly not worked for some time.

## What happens

`sanitizePackEngrams` (`packages/core/src/packs.ts:1140-1156`) removes `pinned` and downgrades `commitment: locked` from **every** installed pack, unconditionally — including the bundled `effective-memory` pack installed by `plur init` (`packages/mcp/src/index.ts:323`).

That is correct behaviour and #1019 has now made it a MUST in §5.4. A pack is an archive from a stranger, and letting its author decide what is permanently in front of the recipient's model is a privilege no producer should take by shipping a YAML field.

## What it costs

`effective-memory` exists to carry cross-cutting meta-rules: call `plur_learn` when corrected, recall before answering, verify before concluding, never type a weekday from memory. Those are relevant to nearly every session and keyword-match almost none of them — which is precisely why they were written `pinned: true`.

With the flag stripped they fall back to `injection_policy: on_match` against the pack's `match_terms`. So the rules that are supposed to apply always now surface only when a session's wording happens to brush against them. The pack's own SKILL.md claimed the old behaviour until this was corrected, so nobody had reason to check.

## The question

Should a host be able to grant always-inject to a pack it trusts **deliberately**, as distinct from a pack **claiming** it?

The distinction is the whole point. A producer setting `pinned: true` is asserting a privilege; a user running `plur packs install --allow-pinned some-pack` is granting one. The first must keep being refused. The second is an ordinary consent decision, of the same shape as `--accept-global` in #1024 and `allowInjection` in the existing `InstallOptions`.

## Options

1. **Leave it.** The meta-rules are keyword-gated and sometimes miss. Cheapest, and honest now that the documentation says so.
2. **An install-time grant** — `--allow-pinned`, recorded in the registry entry so `plur packs list` can show which packs hold the privilege. Consistent with how the existing overrides work.
3. **Bundled packs are exempt.** Rejected: it creates a privileged pack class distinguished by its directory, so anyone who can write into `packs/` inherits the privilege. The security posture is better without it.
4. **A different mechanism for meta-rules entirely** — a host-owned set of always-injected rules that is not a pack at all. Largest change, and arguably the correct shape, since these rules are the host's policy rather than shared knowledge.

[...]

Issue #1029, opened 2026-08-26
The pack manifest is not open-world, though the standard requires it and Encode assumes it


crtahlin:
The standard requires it:

> `ENGRAM-STANDARD-v1.md:892` — **Unknown manifest fields MUST be preserved** (`additionalProperties: true`).

The published schema does not do this. Verified:

```
root         additionalProperties : True
metadata     additionalProperties : False
x-datacore   additionalProperties : False
```

And `PackManifestSchema` (`packages/core/src/schemas/pack.ts:3`) is a plain `z.object()` with no `.passthrough()` — unlike `EngramSchemaPassthrough` (`schemas/engram.ts:521`), which has one. `loadPack` returns `result.data` (`engrams.ts:590`), so install's `manifest.yaml` → `SKILL.md` upgrade (`manifestToSkillMd`, `packs.ts:652-662`) re-serialises only the known keys.

**A pack carrying a future `metadata` field is rejected by the published schema, and a pack carrying a future root field loses it on install.**

`spec/README.md:81-83` claims `additionalProperties: true` "encodes the Zod `.passthrough()` open-world rule". For the manifest that is not true; the generator hardcodes the root value (`packages/core/scripts/gen-spec-schemas.ts:59`) and there is no passthrough behind it.

## Why this is live rather than theoretical

`plur-ai/encode` is building a pack producer in Python and its issue #32 states *"the manifest is open-world (`additionalProperties: true`)"*, planning an `x-encode` extension block for tool identity, run id, model profiles and prompt hashes.

That works — but only because it sits at root. Had they put it under `metadata`, alongside `injection_policy` and `match_terms` where a reader might reasonably expect extensions to go, the published schema would reject it and our loader would drop it.

## Tasks

- [ ] Add `.passthrough()` to `PackManifestSchema`, and to the nested `metadata` object
- [ ] Regenerate `spec/pack-manifest.schema.json`; confirm `spec-schema-drift.test.ts` stays green
- [ ] Preserve unknown keys through `manifestToSkillMd` so the install round trip is lossless
- [ ] Test: a manifest with an unknown root field and an unknown `metadata` field survives export, install and re-read
- [ ] Correct the claim in `spec/README.md:81-83`

## Note

If the decision goes the other way — that `metadata` is deliberately closed — then §10.3 needs amending to say so, and encode#32 needs telling. What cannot stand is the schema and the prose disagreeing while a third party builds against the prose.

## Related

[...]

Issue #1018, opened 2026-08-26
Epic: knowledge packs — from artifact to lifecycle


crtahlin:
Tracking issue for the knowledge pack lifecycle in core: building a pack, importing one, and everything after — updating it, removing it, and keeping its provenance intact along the way.

## Why now

#958 gave packs provenance on **export**. Nothing consumes it on import — `_installPackDir` copies only `isFile()` entries (`packages/core/src/packs.ts:762-769`), so the `provenance/` directory a pack ships is deleted at the moment a recipient installs it. Every record written by #972 is unobservable to the person it was written for.

That is one defect among several, and they share a cause: **the specification describes a pack as an artifact and says almost nothing about its lifecycle.** §5 is marked STABLE and covers layout, manifest fields, the export privacy rule and the integrity hash. It does not define install, update, uninstall, scope handling on import, provenance on import, or conflict handling. The word "uninstall" appears in neither the standard nor the profile.

This matters beyond our own implementation. Two other implementers are building against §5 right now: `plur-ai/encode` is writing a pack producer **directly in Python** rather than calling core (encode#32), and `plur-ai/enterprise` is adopting the format for import (enterprise#1047). Encode is the first party to implement the document without reading our TypeScript. Anything under-specified surfaces there first.

## Decisions taken

**Imported pack engrams are projected into the primary store.** The received directory stays untouched as the immutable artifact — integrity and provenance are read from it and never written to it. The engrams are also written into the store, tagged with `pack` and a scope named at import.

This resolves several defects at once. Feedback stops rewriting a file inside the integrity hash. Pack engrams gain embeddings, decay, `rescope`, `forget` and team sync, none of which they have today. And it makes scope assignment a natural part of import rather than something with nowhere to live.

The cost is that uninstall becomes retire-with-tombstone rather than deleting a directory, and that ADR-0003's "packs are YAML artifacts by definition" needs a companion ADR saying what an *installed* pack is.

**The importer names the target scope.** Every engram in the pack takes it. A pack shipping `scope: global` — as the bundled one does — is refused unless the importer explicitly accepts it, because that lands a stranger's engrams in the recipient's own namespace.

## Verified defects

| Defect | Evidence |
|---|---|
| Install destroys the `provenance/` directory | `packs.ts:762-769` |
| No update path for a user pack; downgrade, reinstall and upgrade are indistinguishable | `packages/mcp/src/index.ts:300-351` is the only version comparison, and it serves the bundled pack |
[...]

PR #1037, opened 2026-08-27
Resolve six of the seven pack contradictions (#1019)

crtahlin:
Closes plur-ai/plur#1019 *resolve the seven pack contradictions between the specs, the schema and the reference* — six of the seven. The remaining one is left to the issue that owns it.

**Stacked on `feat/958-engram-provenance`** (PR #1002 *record where an engram came from*), because `spec/ENGRAM-PROVENANCE-PROFILE.md` does not exist on `main` — it is introduced by that PR. Merge order is this → the provenance branch → `main`.

## Why now

Two other implementations are building against §5 right now, and one of them is reading the document rather than our TypeScript:

- `plur-ai/encode#32` *export a conformant knowledge pack from a workspace* implements the pack format **directly in Python**, deliberately — a `python:3.12-slim` runtime, and a standing guarantee that a run needs neither a model nor a store.
- `plur-ai/enterprise#1050` *define the knowledge pack format, with optional provenance* adopts it for import.

Encode is the first party to implement this standard from the document alone. Anything under-specified surfaces there as a silent disagreement rather than a compile error — and one of the seven had already caught it.

## What changed

**§5.4, the export privacy rule.** The strip list gains `relations.supersedes` and `superseded_by`, which are store-local cross-references like the rest — a stale edge landing on an id collision suppresses tension detection between unrelated engrams.

More substantially, `pinned` and `commitment: locked` become **MUST NOT ship**. The reference has always removed them and the spec never said so. Unlike the rest of the list they do not leave untidy state behind: they decide whose judgement governs the recipient's store. A pack is an archive from a stranger, and letting its author pin content into the recipient's context is a privilege no producer takes by shipping a YAML field.

**§5.5.1, two hashes.** The spec required hashing over raw received bytes; install re-serialises after neutralising and hashes that. Both are correct, and they answer different questions — *did this arrive as it was sent*, and *has the installed copy changed since*. Naming them separately resolves the contradiction without changing behaviour. The reference currently records the second under the first's name, which is written down as a known divergence rather than quietly left.

**§9.** Still described a single `provenance.jsonld` sidecar inside the capsule payload. A pack needs one record per engram plus one for itself, which one file cannot hold. Since §9 governs where it and the profile overlap, an implementer reading §9 alone was being told to build the wrong thing. Corrected, and marked as a correction.

**The preview could not see the four licence states.** `readPackProvenance` read only `engram:licenseIsDefault`, so at the one surface a recipient sees, *the author configured this once* and *nobody ever chose it* looked identical — the exact distinction `engram:licenseSource` was introduced to draw. It now reads the four-state field and reports which states it saw, falling back to the boolean for older records. Four tests.

**`provenance.generate` does not govern what a pack carries**, and the guide said it did. The code is right: the setting defaults to `never` because a record inside your own store defends against nobody, and a pack is the one case where that stops being true. A setting whose default is off should not quietly decide that packs ship unattributed. The guide now says so.

**The bundled pack documented behaviour the installer strips.** Its `SKILL.md` promised pinned, always-injected engrams; `sanitizePackEngrams` removes `pinned` from every pack including that one. The documentation now describes what happens — and states plainly that the pack's meta-rules are keyword-gated as a result, which is a weaker guarantee than they were written for.

That last one turned out to matter rather than being a typo, and is filed separately as plur-ai/plur#1030 *stripping pinned from packs is right, and it silently disabled the bundled pack's whole purpose*.

## Not in this PR

**The manifest is not open-world.** Left to plur-ai/plur#1029 *the pack manifest is not open-world, though the standard requires it and Encode assumes it*, which owns it. The decision is not merely "add `.passthrough()`": the standard says unknown manifest fields MUST be preserved, the generated schema sets `additionalProperties: false` on `metadata` and `x-datacore`, and Encode is building against the prose. Either the schema follows the spec or §10.3 gains an exemption — but the two cannot keep disagreeing while a third party implements from one of them.

## Testing

Full suite: 4861 pass, 0 fail. Four new tests covering the four licence states and the pre-`licenseSource` fallback.

Standard to 1.2, additive per §10.2.

### DIFF
--- a/docs/provenance.md
+++ b/docs/provenance.md
@@ -101,14 +101,31 @@ provenance:
 
 | Mode | What happens |
 |---|---|
-| `never` | Nothing is written. The default. |
-| `on_export` | A record is written when engrams leave, in a pack. **Recommended.** |
-| `always` | A record is written every time an engram is created. |
+| `never` | Nothing is written to your own store. The default. |
+| `on_export` | A record is written to your store when engrams leave in a pack. |
+| `always` | A record is written to your store every time an engram is created. |
 
 `never` is the default for two reasons. A record per engram repeats the history
 log, which already holds the same events. And a record only starts to matter when
 an engram **leaves** your machine — inside your own store it defends against
 almost nobody.
+
+> **This setting governs records kept in your own store. It does not govern what
+> a pack carries.**
+>
+> A pack ships provenance regardless of this setting, including on `never`. Turn
+> it off for one pack with `--no-provenance`.
+>
+> That is deliberate, and it follows from the same reasoning as the default. The
+> setting is `never` because a record inside your own store defends against
+> almost nobody — and a pack is the exact case where that stops being true. It is
+> the artifact that reaches somebody who was not there, so it is the one place a
+> record earns its keep. Every software supply chain settled on the same answer:
+> a bill of materials is produced as part of the build and travels inside the
+> artifact, rather than being something a publisher has to remember to ask for.
+>
+> A setting whose default is "off" should not quietly decide that packs go out
+> unattributed.
 
 ---
 
--- a/packages/core/src/packs.ts
+++ b/packages/core/src/packs.ts
@@ -289,8 +289,16 @@ export interface PackProvenanceView {
   verified: boolean
   /** Why `verified` is false, in words a reader can act on. */
   verification_note: string
-  /** Distinct licences the pack's engrams carry, most common first. */
-  licences: Array<{ name: string; count: number; chosen: boolean }>
+  /**
+   * Distinct licences the pack's engrams carry, most common first.
+   *
+   * `chosen` answers the coarse question — did anybody decide this. `sources`
+   * carries the four-state `engram:licenseSource` values seen for this licence,
+   * so a reader can tell a licence picked for the engram from one inherited
+   * from the pack or taken from a configured default. Empty for records written
+   * before that field existed.
+   */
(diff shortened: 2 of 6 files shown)


crtahlin:

```
main
```

crtahlin:
**Correction: six of the seven are resolved, not five.**

I wrote "five" in the PR title, the commit message and the summary above, and it was wrong — a miscount, not a change of scope. Nothing about the work differs; only my arithmetic did.

| # | Contradiction | Outcome |
|---|---|---|
| 1 | manifest not open-world (`metadata.additionalProperties: false`) | **left** to #1029 *the pack manifest is not open-world* |
| 2 | §5.5 forbids re-serialising; install re-serialises | resolved — new §5.5.1 names the two hashes |
| 3 | §9's single `provenance.jsonld` sidecar | resolved — corrected, and marked as a correction |
| 4 | §5.4 strip list incomplete | resolved — supersedes edges added, `pinned`/`locked` became MUSTs |
| 5 | bundled pack documents behaviour the installer strips | resolved — documentation corrected, #1030 filed for the capability it exposed |
| 6 | `provenance.generate` does not control pack provenance | resolved — the guide was wrong, the code was right |
| 7 | preview cannot see the four licence states | resolved — `readPackProvenance` reads `engram:licenseSource`, four tests |

The commit message on `96190dfc` still says five and cannot be corrected without rewriting a pushed branch, which is not worth it for a counting error. This comment is the record.

crtahlin:

Issue #1020, opened 2026-08-26
Specify install, update and uninstall — and define the install registry the spec already calls authoritative


crtahlin:
The standard describes a pack as an artifact. It does not say what happens to one.

§5 is marked STABLE and covers the directory layout, the manifest fields, the export privacy rule and the integrity hash. Everything a *consumer* does is absent.

## What is missing, precisely

**Install has no algorithm.** The word appears three times in the standard — `:507` (the `manifest.yaml` auto-upgrade), `:512` ("the authoritative integrity record at install time is the registry entry") and `:598`. There is no statement of where engrams go, no requirement to verify before installing, no requirement to record anything.

**The install registry is undefined.** §5.1 and §5.5 both make it authoritative and neither says what it is. The reference has `RegistryEntry { name, installed_at, source, integrity, version?, creator? }` (`packs.ts:119`) — six fields, no engram ids, no scope, no licence, no provenance pointer, no list of files installed. A second implementation has nothing to agree with.

**Sanitisation of an incoming pack is not mentioned.** §5.4 is producer-side only, as is invariant 7 (`:482`). Nothing says a consumer may or must clamp `pinned`, downgrade `commitment: locked`, or reject a pack whose statements carry entry-boundary forgeries — see #1004. A conformant consumer may inject a hostile pack's pinned engrams verbatim.

**Scope handling on import is not mentioned at all.** §4.2 makes `scope` required and free-form; §3.4 says pack producers SHOULD export bare ids, which points the wrong way for import. A pack shipping `scope: global` — as our own bundled pack does — lands in the recipient's global namespace with no ceremony and no rule saying whether that is allowed.

**Conflict handling on import is not mentioned.** `detectConflicts` (`packs.ts:585`) has no normative counterpart.

**Update does not exist.** §3.3.2 gestures at it — *"ids must stay stable across pack releases so installs and upgrades can be diffed"* (`:222-223`) — and then no diff rule, no version-comparison rule, no statement of what happens to accumulated feedback or usage. Pack `version` is defined as an opaque, non-range-checked string (`:527`), which forecloses ordering altogether.

**Uninstall does not exist.** The word appears in neither document. Nothing says whether engrams are removed, retired or retained, whether a tombstone is required, or whether the operation must be reversible.

## Proposal

Three new sections in the standard, additive to a STABLE section and therefore a minor version per §10.2:

[...]

Issue #1021, opened 2026-08-26
Profile: specify what a consumer does with the provenance a pack ships


crtahlin:
The provenance profile's §5.3 is entirely producer-side. It says what a pack must carry and how to build it. It says nothing about what a recipient does with any of it.

That gap is why #989 exists as an open question rather than a fix: install ignores `provenance/`, and there is no specification to say it should not.

## What the profile does not answer

- Is the `provenance/` directory kept after install, or read and discarded?
- Where does a consumer store records it received? The reference keeps its own under `~/.plur/provenance/<ENGRAM-ID>/<timestamp>.jsonld`. Are a stranger's records allowed in the same place?
- Must a consumer report what it found — records present, corrupt, missing, or describing engrams not in the pack? A tester replaced a record with invalid JSON, deleted two others and added one for an engram not in the pack; all three installed with exit 0 and no output (#989).
- Is an install itself a PROV activity? There is an obvious `engram:Install` shape and no statement about it.
- If an installed engram is later re-exported, are the original records carried, regenerated, or dropped?
- What becomes of the shipped `attribution` and `claim_class`?

## The last question is the hard one

#989 lays out three options and rejects the third — copying `attribution` and `claim_class` onto the installed engram — because it *"launders a stranger's claims into your own store, where they become indistinguishable from things you recorded yourself"*.

That was written when pack engrams lived outside the store. #1023 changes that: they are projected into the primary store, so the question stops being avoidable.

**Proposed answer, which is #989's option 1 + 2 without the laundering.** Keep the shipped attribution exactly as shipped; do not merge it, do not rewrite it, do not discard it. Set `provenance.origin = pack:<name>@<version>`. Name the pack as the intermediary using the profile's existing `engram:memberOf`, which already exists for the licence-inheritance case and means precisely "this arrived inside that". A reader can then always tell a claim that arrived in a pack from one asserted locally, which is the property that was actually at risk.

The licence work is the precedent: §8.4's `engram:licenseSource: inheritedFromPack` exists because a pack licence granted by an assembler is not the same as a licence the engram's author chose. Attribution has the same shape and deserves the same treatment.

## Tasks

- [ ] Draft profile §5.4, "Provenance on import"
- [ ] Settle whether an install is a PROV activity, and if so its shape
- [ ] Settle re-export: carried, regenerated, or dropped
- [ ] State what a consumer MUST report about records it found
- [ ] Record the attribution decision, with the reasoning, so it is not relitigated
[...]

PR #1038, opened 2026-08-27
Specify install, update and uninstall, and define the install registry (#1020)

crtahlin:
Closes plur-ai/plur#1020 *specify install, update and uninstall — and define the install registry the spec already calls authoritative*.

**Stacked on #1037** *resolve five of the seven pack contradictions*, which is itself stacked on `feat/958-engram-provenance` (#1002 *record where an engram came from*). Merge order: #1037 → this → the provenance branch → `main`. GitHub will retarget this to the provenance branch once #1037 merges.

## The problem

The standard described a pack as an artifact and said nothing about receiving one. "Install" appeared three times as a passing mention. "Uninstall" appeared in neither this document nor the profile. The install registry was called authoritative twice — in §5.1 and §5.5 — and defined nowhere.

So the three operations every consumer performs were, between them, entirely unspecified.

## §5.6 Install

An order of operations, and the order carries reasons rather than being tidiness. Verify before scanning, so a tampered pack is refused before its content is parsed. Scan before neutralising, so the scan sees what was *shipped* rather than what was cleaned. Record last, so a failed install leaves no record claiming otherwise.

**§5.6.2 requires a consumer to be able to answer which pack an engram came from**, and to enumerate a pack's engrams. Stated as a capability rather than a mechanism, because the honest answer differs by store — the `pack` field, a registry keyed by engram id, and an append-only record naming the set are all sufficient.

Without it §5.8 Uninstall is not implementable at all: you cannot remove what you cannot enumerate. That is precisely the wall `plur-ai/enterprise#1049` *establish whether an installed pack can be removed* hit, and the reference has the same gap.

**§5.6.3 forbids adopting a producer's shared scopes on the producer's say-so.** A received pack's `group:acme/engineering` does not name the recipient's engineering group; if one of that name exists locally it is a different group that collides. Installing a pack means agreeing to hold its contents, not to show them to your whole organisation.

**§5.6.4 defines the install registry** — five required fields and three recommended. `integrity_shipped` and `source` are both REQUIRED, for reasons §5.5.1 and §5.8.2 give. And an unreadable registry MUST NOT be treated as an empty one: proceeding as though nothing is installed silently duplicates every pack the user has.

## §5.7 Update

Requires a version somebody can order, and records why SemVer **build metadata cannot be the distinguishing part** — SemVer ignores it for precedence, so two releases compare equal and a consumer cannot tell which is later. Upgrade, reinstall and downgrade must be distinguished rather than collapsed.

Correspondence is by engram id, which is what §3.3.2's dateless pack id form exists for. Recipient-accumulated state — feedback, usage, review decisions — belongs to the recipient rather than the producer and MUST NOT be silently discarded across an update. An engram that departs in a new version is retired rather than vanished.

## §5.8 Uninstall

Retire rather than erase. A removal leaving no trace is indistinguishable from never having known, and *why do I no longer believe this* is the question provenance exists to answer.

Also: a consumer MUST NOT discard a pack's `source` as a step of removing the pack, which makes the operation irreversible by construction. And content the recipient learned independently that merely resembles a pack engram is not part of the pack — membership is what §5.6.2 recorded, not similarity.

## A vocabulary problem this surfaced

The maturity table had **STABLE** meaning both *binding on implementers* and *implemented in the reference*. Those are different claims, and neither can describe a rule written ahead of its own implementation — which is what these three sections are. Marking them STABLE would have been the same species of inaccuracy this epic exists to remove.

New label **SPECIFIED**: normative and frozen, reference incomplete, gaps named.

## §5.9 — where the reference does not comply

A table naming each gap and its tracking issue, rather than leaving an implementer to discover it by disagreeing with us: pack membership, scope on import, the registry's shipped-integrity value, update, and retire-on-uninstall.

It doubles as the acceptance criteria for Phases 2 and 3 of plur-ai/plur#1018 *knowledge packs: from artifact to lifecycle* — each of those issues can now close against a written rule rather than a judgement call.

## Not in this PR

§5.6 points at the provenance profile for what a consumer does with a shipped `provenance/` directory rather than restating it. That is plur-ai/plur#1021 *specify what a consumer does with the provenance a pack ships*, and duplicating it here would create a seventh contradiction while resolving six.

## Testing

Markdown only. Full suite: 4861 pass, 0 fail. One timing flake in `async-lock-contention` on the first run, re-run three times green.

Standard to 1.3, additive per §10.2 — the new obligations fall on consumers, which the document did not previously address at all.

### DIFF
--- a/spec/ENGRAM-PROVENANCE-PROFILE.md
+++ b/spec/ENGRAM-PROVENANCE-PROFILE.md
@@ -10,7 +10,7 @@ that part should work.
 |---|---|
 | **Version** | 0.5 (draft) |
 | **Status** | Proposed, and implemented in the reference. OPTIONAL to follow: an implementation that ignores this document is still fully conformant to the Engram Standard. An implementation that writes provenance MUST follow this document, so that two such implementations agree. |
-| **Companion to** | [The Engram Standard, version 1.2](./ENGRAM-STANDARD-v1.md) |
+| **Companion to** | [The Engram Standard, version 1.3](./ENGRAM-STANDARD-v1.md) |
 | **Profiles** | Section 9 of that standard, "Provenance binding". It refines that section; it does not replace it, and the standard governs wherever the two overlap. |
 | **Date** | 2026-08-26 |
 | **Licence** | Creative Commons BY 4.0 for the text, Apache 2.0 for any code |
--- a/spec/ENGRAM-STANDARD-v1.md
+++ b/spec/ENGRAM-STANDARD-v1.md
@@ -662,8 +662,8 @@ installed — and does not license overwriting the first.
 > records the hash of the rewritten file as the registry's `integrity`. So for any
 > pack containing a pinned or locked engram, the recorded value is the
 > installed-content hash carrying the pack-integrity name, and the shipped value is
-> compared at the gate and then discarded. Tracked in #1019; the fix is to record
-> both.
+> compared at the gate and then discarded. Tracked in plur-ai/plur#1019 *resolve the seven pack contradictions*; the fix
+> is to record both.
 
 ---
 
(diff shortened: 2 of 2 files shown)


crtahlin:

```
main
```

crtahlin:

Issue #989, opened 2026-08-23
Installing a pack ignores the provenance it ships, and installed engrams lose it


crtahlin:
## The problem

`plur packs preview` now reads the provenance a pack ships (#970). `plur packs install` does not. The provenance directory is ignored entirely on the install path, so:

- **Corrupt, missing and orphaned records go unreported.** A tester replaced a record with invalid JSON, deleted two others, and added a record for an engram not in the pack. All three installed with exit 0 and not one word of output.
- **Installed engrams carry no provenance.** Once installed, `plur provenance <id>` on a pack-installed engram shows no pack name, no source, and no history — even though `plur packs list` holds the source path. The record travelled with the pack and was then discarded.

The consequence a tester reached: the pack's provenance directory is *ignored*, so a forged author in it is harmless — but only because nothing reads it at all. That is honesty by accident, and it stops being true the moment install starts reading those files.

## What "carrying it" should mean

Open question, and the reason this is filed rather than fixed:

1. **Record the pack as the origin.** An installed engram's `provenance.origin` becomes `pack:<name>@<version>`, so `plur provenance` can say where it came from. Cheap and clearly right.
2. **Keep the shipped record.** Store the pack's record for each installed engram and show it on request. Truthful, but it means keeping a document written by somebody else and being very careful never to present it as our own.
3. **Merge the fields.** Copy `attribution` and `claim_class` from the shipped record onto the installed engram. **Probably wrong.** It launders a stranger's claims into your own store, where they become indistinguishable from things you recorded yourself.

Option 1 plus option 2 looks right; option 3 needs a decision, not a default.

Whatever is chosen, install must report what it found: how many records were present, corrupt, missing, or describing engrams that are not there.

## Acceptance

- Install reports corrupt, missing and orphaned provenance files rather than ignoring them.
- A pack-installed engram can say it came from a pack, and which one.
- A record written by a stranger is never presented as one we produced.
- Tests cover a corrupt record, a missing one, an orphaned one, and a forged author.

Part of #958. Related: #987 (install verifies nothing), #970 (preview, which does read these files).

Issue #972, opened 2026-08-21
Export provenance with a pack, for each engram and for the pack as a whole


crtahlin:
Part of #958.

## Why

A pack is how engrams leave one person's machine and reach another. That is the moment provenance stops being a nicety and starts being the whole point.

The recipient has our engrams and none of our history. So the record has to travel with them, and it has to stand on its own.

## Two levels, and both are needed

**Per engram.** Where each one came from, who asserted it, what kind of claim it is, and under what licence it may be used.

**For the pack as a whole.** This is the one that is easy to overlook and probably more useful in practice. It answers: who assembled this pack, when, out of what, and what is in it?

A pack-level record lets someone judge the pack before reading a single engram:

- how many engrams were stated by a person, and how many were inferred by a model
- what date range they cover
- whether every engram carries a licence, or only some
- who assembled it, and from which sources

That is a quality signal. Two packs of the same size are not equal if one is all direct statements and the other is all machine guesses.

## Shape

The provenance standard already handles this. A pack is a thing. Assembling it is something that happened. The pack contains the engrams as members.

## Must be self-contained

An exported record cannot point at our history log. The recipient does not have it. Everything the record needs must be inside it.

See the sibling issue on building a provenance record — this is the strongest reason that requirement exists.

## Careful with privacy

Pack export already strips private content and refuses to export anything flagged. Provenance must not become a way around that. In particular, an actor identifier can be personal data. Apply the same rules that already govern engram content.

## Done when

- exporting a pack can include a record for each engram and one for the pack
- the records make sense on a machine that has never seen our store
- nothing private leaks through the provenance path that would be blocked on the content path

---

**Depends on** #964 — needs the portable mode.
**Depends on** #973 — a pack record uses collections, which is part of the extendability work upstream.
**Note** — the privacy rule found while building the worked examples applies here in full: a shared record must never name other engrams held by the sender.

plur9:
Closed by #1002, merged as `145e99ac`.

Not closed automatically: the PR description reads `Closes #959, #960, #961, …`, and GitHub only honours the keyword before the FIRST reference — so #959 closed on merge and the other fifteen did not. Closing by hand rather than leaving them open against shipped work.

PR #1039, opened 2026-08-27
Specify what a consumer does with the provenance a pack ships (#1021)

crtahlin:
Closes plur-ai/plur#1021 *specify what a consumer does with the provenance a pack ships*.

**Third in the stack.** Based on #1038 *specify install, update and uninstall*, which is based on #1037 *resolve five of seven contradictions*, which is based on `feat/958-engram-provenance` (#1002 *record where an engram came from*). Merge bottom-up: this → #1038 → #1037 → #1002 → `main`. Full detail in #1018 *knowledge packs: from artifact to lifecycle*, under "Merge order for the pack work".

## Why

Everything in the profile before section 5.4 is written from the producer's side. It says what a pack carries and how to build it, and nothing about what happens when one arrives.

The reference shows what that costs. It builds provenance for every exported pack — that was the whole of #972 *export provenance with a pack* — and its installer deletes the directory without a word, because nothing told it not to. A record nobody reads might as well not have been written.

## What section 5.4 says

**The directory is evidence, not content.** Keep it with the artifact, as received, unmodified. Do not merge received records into the store where the consumer keeps records of its own: those two collections answer different questions — *what we recorded* and *what somebody sent us* — and a store that cannot tell them apart cannot answer either honestly.

**Report what was found**: records present against engrams shipped, unreadable records, records describing an engram the pack does not contain, engrams with no record. None of it fails the install; all of it is said out loud.

That failure was observed rather than imagined. Per #989 *installing a pack ignores the provenance it ships*, a tester replaced one record with invalid JSON, deleted two others, and added a record for an engram not in the pack. All three installed with exit code 0 and no output at all.

**An installed engram MUST carry `provenance.origin = pack:<name>@<version>`.** Without it, a memory that arrived in an archive from a stranger is indistinguishable from one the recipient recorded themselves — the distinction the whole document exists to preserve.

## The decision #989 left open

That issue lists three options for the shipped `attribution` and `claim_class`, and rejects the third — copying them onto the installed engram — because it "launders a stranger's claims into your own store, where they become indistinguishable from things you recorded yourself".

Section 5.4.3 makes that a **MUST NOT**, and says why it needs stating: it is the *tempting* option. The fields are exactly the right shape, the store has somewhere to put them, and copying them across looks like preserving information.

But it does not simply forbid keeping the values, because discarding them loses real information. Where an implementation places them on the engram it MUST also record that they were **received rather than asserted**, and the mechanism already exists — `engram:memberOf` names the pack an engram travelled inside.

This is section 8.4's `engram:licenseSource: inheritedFromPack` rule one field over. A licence granted by whoever assembled a pack is not a licence the engram's author chose; an attribution shipped inside a pack is not one the recipient recorded. Both stay usable by being labelled.

So #989 can now be implemented from the profile rather than from a discussion — its option 1 + 2, with the laundering avoided by naming the intermediary instead of discarding the claim.

## Two cases the section adds beyond the issue

**Re-export (5.4.5).** An engram that arrived in one pack and leaves in another MUST NOT be presented as originating with the second assembler. This is where forwarding and laundering are actually distinguished, and where getting it wrong is least visible — the second pack looks perfectly well-formed and every trace of the first has quietly gone.

**None of this verifies anything (5.4.6).** Packs are not signed; section 7 of the standard is RESERVED. A forged attribution is indistinguishable from a true one, and a consumer MUST NOT present a received record as verified. Reading these records improves on ignoring them by exactly one degree: the recipient learns what the pack **says**.

## Testing

Markdown only, no code. Full suite: 4861 pass, 0 fail.

Profile to 0.6. Appendix C records that the reference discards the directory today, pointing at #989.

### DIFF
--- a/spec/ENGRAM-PROVENANCE-PROFILE.md
+++ b/spec/ENGRAM-PROVENANCE-PROFILE.md
@@ -8,7 +8,7 @@ that part should work.
 
 | | |
 |---|---|
-| **Version** | 0.5 (draft) |
+| **Version** | 0.6 (draft) |
 | **Status** | Proposed, and implemented in the reference. OPTIONAL to follow: an implementation that ignores this document is still fully conformant to the Engram Standard. An implementation that writes provenance MUST follow this document, so that two such implementations agree. |
 | **Companion to** | [The Engram Standard, version 1.3](./ENGRAM-STANDARD-v1.md) |
 | **Profiles** | Section 9 of that standard, "Provenance binding". It refines that section; it does not replace it, and the standard governs wherever the two overlap. |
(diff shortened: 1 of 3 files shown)


crtahlin:

Issue #1041, opened 2026-08-27
Editing an engram: the spec has two mechanisms and no rule for choosing between them


crtahlin:
The standard defines two ways an engram can change, and never says which to use.

| Mechanism | Where | What it says |
|---|---|---|
| `engram_version` | §4.12 | *"Content-evolution version."* Integer, ≥1, default 1. Six words of semantics. |
| `relations.supersedes` / `superseded_by` | §4.7 | A new engram replaces an old one. |

**Nothing in either document says when an edit is one versus the other.** The profile's §4.2 observes that "an engram changes in place. A counter goes up" as a fact about the reference, then declares it will model each version as a separate entity — without ever saying whether editing in place is the right behaviour to begin with.

So the choice is made per code path, by whoever wrote it, and the two most common paths increment the counter without recording a pointer to what came before (profile §10.2). The revision chain is broken before provenance gets a chance to describe it.

## Why it matters beyond tidiness

Provenance can only describe what it is told. An in-place edit destroys the previous statement: after it, nothing anywhere holds what the engram used to say. A supersession keeps both and links them.

That difference decides whether these questions are answerable at all:

- *What did this engram say when I acted on it last week?*
- *Was this corrected, or has it always said this?*
- *Somebody relied on the old claim — do they need telling?*

The reference's `plur_learn --supersedes` already does the second thing properly and is the better-specified path. In-place mutation is the one that happens by default and records the least.

## Proposed rule

Distinguish by **whether what the engram asserts changed**.

**In place — `engram_version` increments, no supersession:**
- fixing a typo or rewording without changing the claim
- adding or editing `rationale`, `tags`, `domain`, `summary`
- re-scoping, pinning, changing `commitment`
- any change to derived state — `activation`, `feedback_signals`, `usage`

**A new engram superseding the old — MUST:**
- the statement now asserts something different from what it asserted before

The test is what a reader who relied on the old version needs. If they would still be right, the edit is in place. If they would now be wrong, they are owed a record saying so, and that record is a supersession.

This also dissolves profile §10.2 as a blocker rather than solving it. If a changed claim is always a new engram, there is no chain of in-place versions to reconstruct — the old engram is still there, retired, with `superseded_by` pointing forward. §14.1's decision that each version gets its own record becomes trivially satisfiable, because each version *is* its own engram.

## Open question, with a suggestion

What happens to an in-place edit that turns out to have changed the meaning after all — a reworded statement that a reader later reads differently?

[...]

PR #1042, opened 2026-08-27
Say when an edit is in place and when it supersedes (#1041)

crtahlin:
Closes plur-ai/plur#1041 *editing an engram: the spec has two mechanisms and no rule for choosing between them*.

**Fourth in the stack.** Based on #1039 *specify what a consumer does with the provenance a pack ships*. Merge bottom-up: this → #1039 → #1038 *specify install, update and uninstall* → #1037 *resolve five of seven contradictions* → #1002 *record where an engram came from* → `main`. Full order in #1018 *knowledge packs: from artifact to lifecycle*.

## The gap

The standard defined two ways an engram can change and never said which applies when:

- **In place** — same `id`, content changes, `engram_version` increments
- **By supersession** — new `id`, `relations.supersedes` names the old, old one retired

`engram_version` had six words of semantics: *"Content-evolution version."* So the choice was made per code path by whoever wrote it.

## Why it is not stylistic

**An in-place edit destroys the previous statement.** Afterwards nothing anywhere holds what the engram used to say. A supersession keeps both and links them.

That decides whether these are answerable at all:

- what did this engram say when somebody acted on it last week
- was it corrected, or has it always said this
- somebody relied on the earlier claim — do they need telling

## The rule (§4.7.1)

**The test is the effect on a reader who relied on the previous version.** If they would still be right, the change is in place. If they would now be wrong, they are owed a record saying so, and that record is a supersession.

In place: typos and rewording that keeps the claim, `rationale`/`tags`/`domain`/`summary`, `scope`/`visibility`/`pinned`/`commitment`, and all derived state (which does *not* increment the counter).

Supersession, **MUST**: the statement now asserts something different.

Stated as MUST rather than SHOULD because the alternative is unrecoverable. Every other under-specified choice in this document yields something a later implementation can repair — a missing field backfilled, a wrong mapping remapped. A statement overwritten in place is gone.

## The reference does not comply, on its commonest path

Recorded in the section rather than left to be discovered:

| Path | What it does | Where |
|---|---|---|
| dedup `UPDATE` | Replaces the target's `statement` outright, increments the counter, writes no supersedes edge | `learn-async.ts:235-237` |
| dedup `MERGE` | Concatenates the incoming statement onto the target's | `learn-async.ts:274-276` |

Both change what the engram asserts. Neither is an edge case — deduplication is what happens every time somebody re-states something the store already holds in different words.

## What this does to the profile

Profile to 0.7, and the change is mostly **removal of work**.

§4.2 proposed version-scoped identifiers (`engram:ENG-…/v3 was a revision of …/v2`) because a rewritten statement is not the same thing as the one it replaced. Under §4.7.1 a rewritten statement **is** a separate engram, so the identifier already carries the distinction. They become optional rather than required.

That downgrades **§10.2 version history** from *required for §4.2* to optional — it was required because the in-place revision chain was broken, and there is now no chain to reconstruct.

And it unblocks **§14.1** *does each version get its own record — decided: yes*, whose stated blocker was that the shape of a record would depend on which code path happened to run. That is precisely what §4.7.1 prevents.

## Open, and deliberately not solved

An in-place edit that turns out to have changed the meaning after all gets no automatic detection. A rule that tried to spot meaning-change would be wrong often and confidently. The counter and the history event record that an edit happened and when; somebody auditing can find it. §4.7.1 states that limit rather than papering over it.

## Testing

Markdown only. Full suite: 4861 pass, 0 fail.

Standard to 1.4, profile to 0.7 — both additive per §10.2.

### DIFF
--- a/spec/ENGRAM-PROVENANCE-PROFILE.md
+++ b/spec/ENGRAM-PROVENANCE-PROFILE.md
@@ -19,7 +19,7 @@ that part should work.
 
 | Version | Date | What changed |
 |---|---|---|
-| 0.7 | 2026-08-27 | Section 4.2 rewritten and sections 10.2 and 14.1 downgraded, following Engram Standard §4.7.1 *changing an engram*. Version-scoped identifiers were introduced because a rewritten statement is not the same thing as the one it replaced; under §4.7.1 a rewritten statement is a separate engram, so the identifier already carries the distinction. Section 10.2 *version history* is no longer required for 4.2, and 14.1's stated blocker — that the shape of a record would depend on which code path ran — is what §4.7.1 exists to prevent. |
+| 0.7 | 2026-08-27 | Section 4.2 rewritten, keyed on the `content_hash` test rather than on whether meaning changed — and sections 10.2 and 14.1 downgraded, following Engram Standard §4.7.1 *changing an engram*. Version-scoped identifiers were introduced because a rewritten statement is not the same thing as the one it replaced; under §4.7.1 a rewritten statement is a separate engram, so the identifier already carries the distinction. Section 10.2 *version history* is no longer required for 4.2, and 14.1's stated blocker — that the shape of a record would depend on which code path ran — is what §4.7.1 exists to prevent. |
 | 0.6 | 2026-08-27 | Section 5.4 added, "Receiving a pack's provenance". Everything before it was written from the producer's side, which is why the reference builds a record for every exported pack and its installer deletes the directory without a word — nothing told it not to. The section requires a consumer to keep the received directory as evidence rather than content, to report what it found (a tester's corrupt, missing and orphaned records all installed silently with exit code 0), and to record `pack:<name>@<version>` as the origin of anything installed. It forbids merging a received `attribution` or `claim_class` into the store unqualified — the tempting option, and the one that launders a stranger's claims — while permitting the values to be kept where the intermediary is named, exactly as section 8.4 does for an inherited licence. Also covers re-export, where forwarding and laundering are actually distinguished, and states plainly that none of this verifies anything. |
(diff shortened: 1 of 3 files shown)


crtahlin:
**The rule changed after review, and the first version was wrong.**

I originally made the test *whether the meaning of the statement changed* — in place if a reader relying on the old version would still be right, supersession if they would now be wrong. Then, two paragraphs later, the same section admitted that a rule trying to detect meaning-change "would be wrong often and confidently."

That is not a caveat, it is a contradiction. A standard exists so that independent implementations agree, and three of them are building against this one right now. A test whose own text calls it unreliable cannot do that job.

## The test is now `content_hash`

**A change that moves `content_hash` MUST be a supersession. Anything else is in place.**

It was sitting there the whole time. `content_hash` is `sha256(normalizeStatement(statement))` (`packages/core/src/content-hash.ts:79`), and `normalizeStatement` lowercases, strips non-word characters, collapses whitespace and trims. So the hash is **already invariant under exactly the changes that alter nothing** — capitalisation, punctuation, spacing — and moves for everything else.

It is also computed on every write, so this costs an implementation nothing to check.

| Change | Hash | Result |
|---|---|---|
| `"Deploys wait for migrations."` → `"deploys wait for migrations"` | same | in place |
| `"teh deploy script"` → `"the deploy script"` | moves | supersession |
| `"Pools cap at 100"` → `"Pools cap at 200"` | moves | supersession |
| adding a `rationale`, re-scoping, pinning | n/a | in place |
| `activation`, `feedback_signals`, `usage` | n/a | in place, and does not increment `engram_version` |

## The cost, stated rather than hidden

**Fixing a genuine typo mints a new engram.** That is the honest consequence of a rule with no judgement in it, and it is worth it: the alternative is three implementations disagreeing about whether a given edit was substantive. The retired engram is filtered from recall by its status, so the churn is storage rather than noise.

## One consequence that needed saying

`relations.supersedes` now means **"this replaced that"**, not **"this contradicts that"**. A one-character correction and a complete reversal of a claim are recorded identically, because both move the hash and the rule has no opinion beyond that.

So a reader MUST NOT infer substantive disagreement from a supersession chain. Where an implementation wants that distinction it is a *reason* recorded on the supersession by whoever made the change — never something a consumer computes for itself. §4.7.1 says so explicitly, because the opposite reading is the natural one.

Commit `780b156e` on `spec/1041-edit-vs-supersede`, PR #1042 *say when an edit is in place and when it supersedes*. Standard 1.4, profile 0.7, suite 4861 pass.

crtahlin:

Issue #968, opened 2026-08-21
Test provenance records against an outside implementation, not only our own


crtahlin:
Part of #958.

## Why this needs its own issue

A provenance record that only our own code can read proves nothing. Our validator agreeing with our generator is not evidence.

This came up already. While drafting the design document we found that our provenance library emits a field name the reference implementation rejects outright, and 167 passing tests had missed it, because every test compared our code against our code.

## What to add

**Unit tests** for the generator:

- each way an engram can be created produces the right shape
- a retired engram produces an invalidation
- a replacement produces a revision
- a missing actor produces a valid record with no agent, and never a guessed one
- the record carries the time it was built

**A conformance test against an outside implementation.** Take the generated record and parse it with a library that shares none of our code. Fail the build if it is rejected. Check that the counts survive: same number of things, steps and agents, and the same links between them.

This is the test that catches the class of bug described above. The others are cheap insurance.

**A round-trip test anchored to that outside tool**, not to ourselves. Existing round-trip tests pass because our writer and reader agree with each other. They would keep passing if both were wrong in the same direction.

## Done when

- the unit tests above exist and pass
- the build fails if an outside implementation rejects our output
- the conformance check runs automatically, not by hand

---

## Added after review, 2026-08-21: which outside tools, and how

Use more than one, because agreeing with a single outside tool only proves we agree with that tool.

| Tool | What it checks | Notes |
|---|---|---|
| `prov` (Python) | the reference implementation of the standard | already used by hand this session; it rejects malformed records outright |
| `rdflib` (Python) | that our JSON-LD parses and produces the right statements | already used by hand this session |
| ProvToolbox (Java) | an independent implementation, different language, different authors | heavier to run; worth it for a release check rather than every commit |

### How to run them

Both Python tools install cleanly and run in seconds. Put them in a small virtual environment and run them on every commit.

The Java tool is slower to set up. Run it before a release, not on every change.

### What each test asserts

Not merely "it parsed". Parsing is a low bar. Assert that the meaning survived:

- the same number of things, steps and agents come out as went in
- the links between them match, by both endpoints
- a retired engram comes back as invalidated
- a replacement comes back as a revision
- unknown domain fields survive untouched, per the sibling issue on extendability

### Also test the portable form

A self-contained record must make sense with no access to our history log. Test it on input that has nothing else available.

---

**Depends on** #964 — there is nothing to test until a record can be built.
[...]

PR #1043, opened 2026-08-27
Conformance vectors: golden packs and capsule fixtures (#1022)

### LINKED ISSUE #1022: Conformance vectors: golden packs and capsule fixtures, runnable from Node and Python
`spec/README.md:113-118` names conformance vectors as the cheapest thing that unlocks the rest of the standard: *"a language-neutral corpus of canonical inputs + expected outcomes … golden packs with known INTEGRITY hashes, and golden .plur capsules (hex fixtures)"*.

None exist. And there are now three implementations to keep honest:

- core (TypeScript)
- `plur-ai/encode`, writing a producer **directly in Python** (encode#32)
- `plur-ai/enterprise`, adopting the format for import (enterprise#1050)

Encode is the case that matters. It reads the document and not our code, which is the whole point of having a standard — and means any ambiguity in §5 surfaces as a silent disagreement rather than a compile error.

## What to build

**Golden packs**, each a directory plus its expected `INTEGRITY` value:
- minimal conformant pack (`SKILL.md` + `engrams.yaml` only)
- with `INTEGRITY` present, and with it absent
- with a `provenance/` directory
- carrying a deprecated `manifest.yaml`
- with an unknown root-level manifest field, and with an unknown `metadata` field — these two currently give different answers, see #1019
- non-ASCII and non-Latin content, since the hash is over raw bytes
- negative cases: modified content, a `provenance/` record naming an engram not in the pack, a record that is not valid JSON

**Golden capsules** as hex fixtures, covering the §6.7 read algorithm: bad magic, unknown `FormatVersion`, a set reserved flag bit, declared-size mismatch, SHA-256 mismatch, truncation. Worth noting the capsule has **no producer and no consumer anywhere in the codebase** today — `packages/core/src/capsule.ts` is exported and called by nothing outside its own tests. encode#30 would make Encode the first producer, so these fixtures are what it would build against.

**A runner** each implementation can execute. `spec/examples/check.py` and `conformance.py` already exist for engram-level PROV and are not wired into `pnpm test`; the same shape extended to packs, runnable from both Node and Python, is the target.

## Tasks

- [ ] Golden pack corpus with expected hashes, positive and negative
- [ ] Capsule hex fixtures for every §6.7 rejection path
- [ ] A runner both implementations can execute
- [ ] Wire it into `pnpm test` and into Encode's CI
- [ ] Document in `spec/README.md` how a third party runs it

## Done when

A pack exported by core verifies byte-for-byte in Python, and a pack built by Encode loads through core's preview path. That round trip is the actual deliverable — the fixtures are how it stays true.

## Depends on

#1019, #1020 — fixtures encode the answers, so the answers come first.

Epic: #1018

crtahlin:
## Review fixes (2026-09-04)

Rebased onto the reviewed #1044 (`fix/review-1044`, which sits on the reviewed #1002 and current main). Every BLOCKER / IMPORTANT from the 2026-09-04 review is addressed. Thirteen pack vectors, thirteen capsules; every vector now asserts its install or preview outcome through the reference, and three drift gates keep the vectors, the index and the spec text honest.

### What changed
- **Vectors assert outcomes:** `spec-vectors.test.ts` installs every vector through `installPack` and asserts the declared `expect` (reject ⇒ throws, no directory, `listPacks` empty, every override option still refuses). `neutralized: { pinned, locked }` is asserted for every loaded vector (zeros too) and the installed bytes are checked for `pinned` / `locked`. New `locked-engram-shipped` (the previously unreachable `verify.py` branch) and `omitted-visibility` (§5.6.1's declared-vs-defaulted rule).
- **Non-ASCII fixture:** `build.py` writes with `ensure_ascii=False`; the `non-latin` vector now has 86 bytes above 0x7f and the test asserts they exist.
- **Tautological vectors:** `orphan_records`, `unreadable_records`, `provenance_records` are declared and asserted; `unknown-root-field` asserts `preview.manifest['x-vendor']`.
- **Drift gates:** `build.py --check` diffs `index.json`; new `build_capsules.py --check` (bytes and `capsules.json`); `verify.py --index` checks every declaration against the fixture bytes (an `expect: reject` with nothing §5.6.1 refuses is a FAIL, a `load` vector shipping `pinned` is a FAIL), `--capsules` checks size and SHA-256; CI runs all three before typecheck. The test resolves every `§x.y` and "step N / rule N / invariant N" citation in a vector note to a heading or numbered item in the standard or profile.
- **Capsule reasons:** `capsules.json` carries a `reason` regex per rejection and the test asserts it. `truncated-payload` now cuts after the header (size mismatch); new `truncated-header`, `signed-payload-underflow` (§6.7 step 8), `signer-without-signed-flag`.
- Mutation-tested: ten deliberate mutations (hand-edited `expect`, flipped capsule byte, ASCII-escaped fixture, renumbered spec step, dropped reader check, …) each turned only their own vector red, then were reverted.

An orphaned comment block in `ci.yml` for steps that were never added; `index.json` / `capsules.json` written with `ensure_ascii=False` and `engram_count` per vector; two provenance vectors' hashes moved because they now use the shared `with_provenance` manifest.

### Verification
`python3 spec/vectors/build.py --check` and `python3 spec/vectors/build_capsules.py --check` match; `python3 spec/vectors/verify.py --index spec/vectors/index.json --capsules spec/vectors/capsules.json` 26 fixtures, 0 failures; `npx vitest run --project @plur-ai/core spec-vectors` 102 passed; full `@plur-ai/core` 208 files / 3398 passed (1 expected fail, 36 skipped); `@plur-ai/cli` 657 passed, 4 skipped; `@plur-ai/mcp` 459 passed; `pnpm typecheck:tests` clean.

### Remaining risks
`verify.py` is regex-based by design and fits `build.py`'s serializer, not arbitrary stores; the citation gate checks that headings and numbered items exist, not that the cited rule still says what the note claims; `unknown-metadata-field` stays disputed (#1029); `content_hash` vectors (#1091) and per-§4.14-invariant vectors remain the stated gap.

---

Closes plur-ai/plur#1022 *conformance vectors: golden packs and capsule fixtures, runnable from Node and Python*.

**Fifth in the stack.** Based on #1042 *say when an edit is in place and when it supersedes*. Merge bottom-up: this → #1042 → #1039 *what a consumer does with a pack's provenance* → #1038 *specify install, update and uninstall* → #1037 *resolve five of seven contradictions* → #1002 *record where an engram came from* → `main`. Order in #1018 *knowledge packs: from artifact to lifecycle*.

## Why

`spec/README.md` has named conformance vectors the cheapest thing that unlocks the standard since it was written, and there were none. Meanwhile the number of implementations reached three:

| Implementation | Reads |
|---|---|
| `@plur-ai/core` | the reference |
| `plur-ai/encode` | **the document**, deliberately not our code (encode#32) |
| `plur-ai/enterprise` | adopting the format for import (enterprise#1050) |

Encode is the case these exist for. It builds a producer from §5 without reading `packs.ts`, so anything ambiguous surfaces there as a silent disagreement — a wrong hash, a rejected pack, months later.

## The rule that makes them worth having

**Python authors them. TypeScript verifies them.**

A vector generated by the reference and then checked by the reference proves only that the code agrees with itself — the same trap #968 was filed for, where our provenance library had 167 passing tests while emitting a field the reference implementation rejected outright.

The expected `INTEGRITY` values here come from `hashlib` and a hand-written serializer in `spec/vectors/build.py` that shares no code with `packs.ts`.

The bytes are **committed**, not generated at test time. `build.py` records how they were made and rebuilds them when a rule changes deliberately — never as a step in the test run.

## What is in it

**11 golden packs.** Minimal; with and without `INTEGRITY`; with a `provenance/` directory that must **not** move the hash; non-Latin content, which is the only vector that catches an encoding assumption since §5.5 hashes raw bytes; an unknown root manifest field that must survive per §10.3; and five negatives — integrity mismatch, an orphan provenance record, an unreadable one, a private engram shipped, a pinned engram shipped.

**10 capsule fixtures**: one valid, nine rejections, one per §6.7 check — bad magic, unknown `FormatVersion`, a set reserved flag bit, declared-size mismatch, SHA-256 mismatch, truncated preamble, truncated payload, `COMPRESSED` disagreeing with the descriptor, `SIGNED` with no trailer.

§6 is marked STABLE and **had no producer and no consumer anywhere** — `capsule.ts` reads and writes capsules and grep finds no caller outside its own tests. These are the first bytes the format has been asked to agree with, and encode#30 would make Encode the first real producer.

## They caught something within an hour

The first `valid-minimal` capsule omitted `manifest_summary.engram_count`. The reference rejected it — and was right, since §6.4 marks the field REQUIRED, `int ≥ 0`.

The fixture was fixed, not the reader. The omission is left as a comment in `build_capsules.py`, because nothing about the word *summary* suggests a count is mandatory in it, and the next person will make the same assumption.

That is the suite working on its first day, in the direction that matters: an independent encoder got it wrong and the reference caught it.

## `unknown-metadata-field` asserts nothing, on purpose

§10.3 says unknown manifest fields MUST be preserved. The published `pack-manifest.schema.json` sets `additionalProperties: false` on `metadata`, and `PackManifestSchema` has no `.passthrough()`. #1029 *the pack manifest is not open-world* has not decided which gives way.

Asserting either behaviour here would settle it by accident. What the vector does assert is that its hash agrees across implementations either way — and that the question cannot be quietly forgotten, because a file in this directory is asking it.

## CI

Two new steps before `pnpm test`:

- `build.py --check` — a vector edited by hand to make a test pass is the one failure mode that would quietly disable the whole suite
- `verify.py --index index.json` — the outside check

`verify.py` has **no dependencies at all**, deliberately, so a third-party producer can run it in a container with no package index.

## Testing

43 new tests. Full suite: 4904 pass, 0 fail.


crtahlin (commented):
Reviewed. One blocking defect, and five vectors that pass without constraining the rule they name. Posted as a comment because GitHub does not accept a self-review.

The design is right — Python authors the bytes, TypeScript verifies them — and I confirmed the wired vectors genuinely constrain the implementation by mutating the reference seven times. Every mutation went red, narrowly: hashing `provenance/` into the pack hash gave 7 failures confined to the three provenance packs; conflating an absent INTEGRITY with "ok" gave 3; disabling the private-visibility flag gave 1; disabling the reserved-flag, SHA-256, COMPRESSED-agreement and payload-size checks each turned exactly their own capsule red. No collateral failures in any case, which is the sign of vectors that isolate one rule each.

### Blocking

- `spec/vectors/packs/*/engrams.yaml` — **all eleven files are absent from the commit.** `git check-ignore -v` resolves them to `~/.gitignore_global:5:engrams.yaml`; the repository's own `.gitignore` does not exclude them, so `git add spec/vectors/` skipped them silently. This is the same mechanism as [#1062](https://github.com/plur-ai/plur/issues/1062), now hitting fixtures rather than a test.

  On a clean worktree of the head: `python3 spec/vectors/build.py --check` exits 1 listing all eleven as missing; `verify.py --index` exits 1 with an unhandled `FileNotFoundError`; `vitest run test/spec-vectors.test.ts` gives **29 failed / 14 passed** — every pack hash, every integrity verdict, every behavioural pack assertion. Only the ten binary capsule fixtures survive.

  The committed hashes are correct: running `build.py` restores exactly those eleven files and reproduces `index.json` and the INTEGRITY files byte-for-byte, leaving `git diff` on tracked paths empty.

  Fix: add a negation such as `!spec/vectors/packs/**/engrams.yaml` to the repository `.gitignore` — repository patterns take precedence over `core.excludesFile` — then `git add -f`. Also harden `verify.py:146-148` so a pack with `provenance/` and no `engrams.yaml` gives a diagnostic rather than a traceback.

  This went unnoticed because `.github/workflows/ci.yml:148` places both new gates *after* `pnpm typecheck:tests`, which fails on this stack ([#1090](https://github.com/plur-ai/plur/issues/1090)), so the job aborts before either runs. Moving them ahead would have caught it — they need neither Node nor a build.

### Five vectors that do not constrain what they name

- `spec/vectors/build.py:74` — **`non-latin` contains no non-ASCII bytes.** `json.dumps(str(value))` defaults to `ensure_ascii=True`, so the Japanese and Cyrillic statements are written as `デ…` escapes. Both files are pure ASCII, verified by byte scan. `README.md` calls this "the only vector that catches an encoding assumption"; it catches none. Proven: changing `computePackHash` to decode as `latin1` — exactly that assumption — left the suite 43/43 green. Needs `ensure_ascii=False` and a rebuild.
- `packages/core/test/spec-vectors.test.ts:143` — `unknown-root-field` reads the fixture off disk and asserts the fixture contains `x-vendor`. It asserts nothing about the implementation, **and the implementation fails the rule**: `previewPack(...).manifest['x-vendor']` is `undefined`. §10.3 and §4.14 invariant 6 make preservation a MUST, and the PR body asserts the root case is settled. Either the reference must preserve unknown root fields or the vector must record that it does not.
- `packages/core/test/spec-vectors.test.ts:118` and `:125` — the two provenance-report vectors assert only `provenance.present === true` and `engram_count === 2`, both trivially true. `readPackProvenance` iterates the pack's own engrams, so the orphan record is never opened; `readJson` swallows the parse failure in a bare `catch`, so the corrupt record is neither counted nor distinguished from an absent one. `index.json`'s own note says unreadable records "MUST be counted and reported, never silently skipped. A tester's corrupt record installed with exit 0" — which is what the reference still does.
- `spec/vectors/index.json` — `pinned-engram-shipped` declares `expect: "load-neutralized", neutralized: 1` citing §5.4 and §5.6.1, but has **no assertion at all**; `grep pinned` in the test file matches only the interface type. `previewPack(...).security` is `{clean: true, issues: []}`.
- `spec/vectors/capsules.json:69` and `:55` — two capsules reject for the wrong reason. `signed-flag-without-signature` claims a §6.8/§7 ambiguous-envelope refusal but actually fails on `payload size mismatch`; disabling both the size and SHA-256 checks makes it **accept**, and there is no SIGNED-vs-signer consistency check in `readCapsule` at all. `truncated-payload` claims payload-region underflow but fails on a truncated header, so the underflow branch at `capsule.ts:127` is unreached by any fixture. Structural cause: the helper at `spec-vectors.test.ts:187` accepts any throw as a pass, so a fixture rejected for an unintended reason is indistinguishable from one rejected correctly.

### Also worth fixing

- `spec/vectors/build_capsules.py` has no `--check` mode and is not run by CI, so the ten binary fixtures have no drift protection — and being binary, a hand edit would be invisible in review. `capsules.json` already records `bytes` and `sha256` for each and nothing asserts either. Asserting them in `verify.py` is the cheaper option and keeps the check on the dependency-free side.
- `spec/vectors/verify.py:154` — `--index` checks only `computed_integrity` against the bytes and disk/index name parity. It never asserts the declared `expect`, `integrity_status`, `orphan_records`, `unreadable_records` or `neutralized`, and the §5.4 privacy findings print as notes that never affect the exit code. Since both the vectors and `index.json` come from `build.py`, `--index` is close to a self-consistency check of the Python side; CI calls it "verify from outside the reference", which oversells it.
- `spec/vectors/verify.py:123` — the `commitment: locked` branch is unreachable: `BASE_ENGRAMS` never sets `commitment`, so §5.4's downgrade-on-import MUST is unrepresented.
- `content_hash` has **zero coverage** — no occurrence anywhere under `spec/vectors/` or in the test file. §4.7.1's rule is introduced by [#1042](https://github.com/plur-ai/plur/pull/1042), the immediate base of this PR, and this is the PR that builds the apparatus to pin it. Consistent with [#1091](https://github.com/plur-ai/plur/issues/1091).
- Of the seven §4.14 invariants, only invariant 7 is genuinely covered. Invariant 6 is claimed by `unknown-root-field` but is vacuous and currently violated; invariants 1-5 have no vectors either way.

### Done well

The Python-authors / TypeScript-verifies rule is the correct architecture, and the header explains exactly why, citing the case where 167 passing tests coexisted with a field the reference rejected. Hand-writing the serializer rather than using pyyaml is the right call for a byte-exact format.

`unknown-metadata-field` carrying a contradiction without asserting a winner — and saying so in the fixture, the index, the README and the PR body — is a better outcome than quietly picking a side and calling it conformance. Striking through the roadmap item in `spec/README.md:114` and naming what is still missing, rather than declaring it done, is the same instinct.

`spec/vectors/README.md:126` — "Establish which side is wrong before changing anything", with the `valid-minimal` episode recorded as the worked example, is the most useful paragraph in the directory.

Tests: FAIL as committed — 29 of 43 new tests fail and both new gate scripts exit 1, purely from the missing fixtures. With those regenerated, the spec-vectors, capsule and packs suites are 87/87 green and both Python gates exit 0.

Verified: seven mutations of the reference to confirm the wired vectors are non-vacuous; byte scans of the `non-latin` fixtures; direct probes of `previewPack` for the unknown-field, orphan, corrupt and pinned cases; capsule rejection reasons isolated by disabling individual checks; `build.py` re-run to confirm the committed hashes reproduce.

crtahlin (commented):
Reviewed. This would be an approval, but GitHub does not accept one from the pull request's own author — so it is a comment. Four gaps are worth fixing, none of them blocking, filed as follow-ups.

The corpus is genuinely non-tautological, which is the only question that matters for a conformance suite. Seven independent mutations were reproduced — a flipped byte in `valid-minimal.plur`, a hand-edited `neutralized.pinned` in `index.json`, a hand-edited `engrams.yaml` with no rebuild, latin1-decoding `engrams.yaml` inside `computePackHash`, deleting the `pinned` strip in `sanitizePackEngrams`, deleting the reserved-flag check in `parseCapsulePreamble`, deleting the SIGNED/signer agreement check in `readCapsule` — plus two bogus citations (`§99.9` and "step 97"). Each turned red in exactly the vectors naming that rule and nowhere else. The latin1 mutation reddening only the four non-latin assertions is the clearest single piece of evidence that these vectors bite.

All three drift gates fire correctly, and all three run clean under `python3 -S` from an unrelated working directory, which is what makes the "a third party can run this anywhere" claim true rather than aspirational. Checking a rejected install for absence of both the directory and the registry row — rather than merely for a thrown error — is the difference between refused and refused cleanly, and it is what a conformance suite usually forgets.

Follow-ups: #1159, #1160, #1161, #1162

### Notes

- `.github/workflows/ci.yml:118-123` — the gates sit after `pnpm install --frozen-lockfile`, `pnpm build` and the core typecheck, so any of those failing aborts the job before a vector gate runs. That is the exact failure mode the comment above them says the ordering prevents, and `spec/vectors/README.md:75` repeats the claim. Filed as #1161.
- `spec/vectors/README.md:110` — "twelve rejections, one per §6.7 check" overstates coverage. Step 1, step 6, and the malformed-JSON / schema-violation half of step 7 have no fixture, while step 9 has two and step 7 has three. The schema-violation gap matters most: `build_capsules.py:57-62` records that a missing `manifest_summary.engram_count` was the first real defect these fixtures caught, and no negative fixture was added for it. Filed as #1159.
- No `.gitattributes` anywhere in the repository. A clone with `core.autocrlf=true` changes every §5.5 hash and turns the corpus red with a message that reads as a cross-implementation disagreement. For fixtures aimed at third-party implementers, byte stability is the property they exist to provide. Filed as #1160.
- §5.6.1 step 2 has two non-overridable refusals and only the declared-private one is vectored. The secret-scan refusal is the one an independent consumer is most likely to implement as a warning. Filed as #1162.
- The deprecated `manifest.yaml` pack is not filed separately, because #1022 already names it as required and stays open. Worth knowing why it matters: `build.py:185,203-216` already carries the mechanism, including the §5.5 rule that a deprecated `manifest.yaml` contributes zero bytes to the hash, but no vector passes `manifest_filename=` and `verify.py:132-136` has the matching dead branch. So the least obvious rule in §5.5 is unvectored — an independent producer that hashes `manifest.yaml` the way it hashes `SKILL.md` passes the entire current corpus. `spec/README.md:119-127`'s "Still missing" list does not mention it.

### Smaller observations, no issue filed

- `spec-vectors.test.ts:268-282` — `unknown-metadata-field` is tautological by construction and honestly labelled, but CI stays green whichever way #1029 resolves, so nothing turns red when the decision lands. Asserting the disagreement itself, that the schema closes `metadata` while §10.3 says preserve, would make it fail when either side moves.
- `spec/vectors/verify.py:334` — whether a fixture may lack the PLUR magic is decided by `"magic" not in name`, a filename substring driving a correctness check. Rename the fixture and the check silently stops applying.
- `spec/vectors/verify.py:308-339` — `--capsules` does no §6 semantic validation, only size and SHA-256. So for §6 the TypeScript reader is the only judge of conformance, and the "neither implementation grades its own work" property is weaker than it is for §5. The CI step name is accurate for packs and generous for capsules.
- `spec-vectors.test.ts:352` — the citation resolver recognises only `§N.N`. A note written as "Section 5.5", or "profile, §5.4.2" with a comma before the §, would go unchecked. Nothing in the corpus does either today, and the `cites.length > 0` assertion does stop a note citing nothing.
- The non-ASCII assertion is on the fixture's real bytes rather than a recorded constant, so re-escaping to ASCII fails the test. The 86-byte figure checks out exactly: 86 of 428 bytes.
- The `.gitignore:58` negation works as claimed — with a simulated global `excludesFile` containing `engrams.yaml`, `git check-ignore -v` returns not-ignored for the minimal fixture, so the silent-empty-fixture failure is genuinely closed.
- #1022's "Done when" is half-covered: Python authors and TypeScript verifies, but nothing is exported by core and then verified in Python, so the reverse direction of the round trip is untested. Leaving the issue open is right.

### One thing about the stack

This sits on #1044, which I have just sent back with two blocking specification defects. Nothing here depends on those, but this cannot merge until that layer settles.

Tests: PASS — 102 of 102 in `packages/core/test/spec-vectors.test.ts`, and all three Python gates exit 0 across 26 fixtures with no failures. The working tree was restored clean after every mutation.

plur9 (dismissed):
Approving. The drift gates are the substance here, and I checked that they bite rather than that they pass.

### Verified

- All three gates run clean: `build.py --check` and `build_capsules.py --check` both report a match, and `verify.py --index --capsules` reports **26 fixtures, 0 failures, 8 notes**.
- CI genuinely runs all three, at `.github/workflows/ci.yml:119`, `:121` and `:123`, before typecheck. The claim and the wiring agree.
- **The gates actually catch drift.** I flipped one byte in the middle of `spec/vectors/capsules/bad-magic.plur` (byte 250 of 500) and re-ran: `26 fixture(s) checked, 1 failure(s)`. Reverted, back to 0. That is the property that makes the mutation-testing paragraph believable, and it is the one worth confirming independently — a fixture suite that only ever passes is indistinguishable from one that asserts nothing.

The notes `verify.py` emits are the right shape too: `private-engram-shipped` prints both halves of the rule it encodes (`§5.4: a producer MUST exclude these; §5.6.1 step 2: a consumer MUST refuse the pack`), so the fixture states what it is a fixture *for*. And `omitted-visibility` spells out the declared-versus-defaulted distinction that #1044 introduces, which is exactly the subtlety a second implementation would otherwise get wrong silently.

### Should fix

- **The merge instructions are stale in a way that will mislead whoever merges.** The description says "Fifth in the stack. Based on #1042. Merge bottom-up: this → #1042 → #1039 → #1038 → #1037 → #1002 → main."

  **#1037, #1038, #1039 and #1042 are all CLOSED and none was merged**, and #1018 is not a pull request at all. The real topology is three PRs:

  ```
  #1043  spec/1022-conformance-vectors  →  spec/pack-lifecycle   (#1044)
  #1044  spec/pack-lifecycle            →  feat/958-engram-provenance (#1002)
  #1002  feat/958-engram-provenance     →  main
  ```

  Same class of staleness as #1002's "merge #1000 and #1001 first" (both also closed unmerged). Someone following either instruction goes looking for merged commits that do not exist. Worth correcting on all three descriptions in one pass.

### Notes

- The stated gaps are the right ones to have left: `verify.py` being regex-based and fitted to `build.py`'s serializer is a real limit for a third-party store, and the citation gate checking that a `§x.y` heading *exists* rather than that it still *says* what the note claims is an honest description of what a text gate can do. Neither is worth closing now; both are worth keeping written down where they are.
- `unknown-metadata-field` remaining disputed (#1029) is fine as a shipped vector, since a vector that encodes a disputed rule is still the cheapest way to find out who disagrees.

### Done well

The reason this is worth its size is in the table: three implementations, and `plur-ai/encode` builds a producer from the document rather than from `packs.ts`. Vectors are the only thing that turns a silent disagreement between two readings of §5 into a failing check, and "a wrong hash, a rejected pack, months later" is exactly what that costs without them. Asserting the install *outcome* through the reference — including that every override option still refuses a reject vector — is what stops these being decorative fixtures.

Blocked only by its base: this sits on #1044, which sits on #1002, and I have requested changes on #1002.

Tests: PASS — the three gates run clean, and a single-byte mutation turns exactly one fixture red. Verified: ran each gate; mutated a capsule and reverted it; confirmed the CI wiring by line; checked the actual base branch of each PR in the stack and the state of every PR named in the merge order.

PR #1044, opened 2026-08-27
Specify the pack lifecycle, how an engram changes, and provenance on import

### LINKED ISSUE #1019: Resolve the seven pack contradictions between the specs, the schema and the reference

crtahlin:
## Review fixes (2026-09-04)

Rebased onto the reviewed #1002 (`fix/review-1002`, which includes current main and the architecture audit). Every BLOCKER / IMPORTANT from the 2026-09-04 review is addressed; spec and reference implementation now agree, and each invariant is stated in the header of `packages/core/test/pack-lifecycle-conformance.test.ts` and probed adversarially.

> **Decision to accept or reject:** profile §5.6.1 step 2 now says a secret-scan hit, and a declared `visibility: private` engram, MUST be refused on install with **no override**. This reverses the "explicit loud override" agreed earlier in this thread. Rationale in the spec: the remedy lies with the pack, false positives are fixed by the producer, and a consumer SHOULD document its scan surface. The reference was already fail-closed on secrets; it now also refuses declared-private. Engrams that merely *omit* `visibility` (which defaults to private) are held private on the consumer side and reported, not refused, because refusing them would reject every hand-written pack that omits the field.

### What changed
- **`@graph` handling (BLOCKER):** profile §5.4.2 defines *unreadable* (non-JSON, non-object, no `@graph` array, oversize; scalar nodes skipped) and *orphaned*, and requires that a record a consumer cannot read never aborts preview or install. Tests: 8 top-level and 5 node shapes each count as one unreadable or skipped record while the pack still installs.
- **Locked-downgrade reporting (IMPORTANT):** `sanitizePackEngrams` counts `lockedDowngraded`; `InstallResult.neutralized { pinned_stripped, locked_downgraded }` (zeros included); preview warns on locked as on pinned; `plur packs install` prints both (loud under `--quiet`); `plur_packs_install` returns `neutralized` and `integrity_check` (it had dropped both). §5.6.5 says "per field".
- **CLI `sources` render (IMPORTANT):** `plur packs preview` prints a `how:` line per licence from a table keyed by the closed enum; an unknown value prints `(unrecognised)`, never verbatim.
- **`licenseSource` closed enum:** `LICENSE_SOURCES` exported from `provenance.ts`; `sources: LicenseSource[]`; an out-of-set value is counted, reported in `notes` and treated as absent; profile §8.4 makes the set closed (MUST).
- Smaller spec items from the self-review: §5.4 SHOULD→MUST (#1092), §5.7.3 wording, §5.6.5 restates the four provenance counts, §5.6.1 ordering rationale, §5.9 row for the bundled `effective-memory` pack (#1030), spec/README status table (standard 1.7, profile 0.9).

- `readPackProvenance` never counted orphan records (profile §5.4.2 MUST): `orphan_records` added, counted by name, never opened.
- Root manifest fields were silently dropped by the Zod parse while `pack-manifest.schema.json` advertises `additionalProperties: true`: `PackManifestSchema.passthrough()` at the root, `manifestToSkillMd` carries them, `__proto__` / `constructor` keys pollute nothing (tested), `gen:schemas` produces no diff.
- `readCapsule` accepted `SIGNED` with `signer: null` and a signer with the flag clear: refused as an ambiguous envelope before the payload region is computed; §6.7 step 7 states it (§6.8 step 4 mirrored). Tightens a STABLE section; no writer-conformant capsule is affected.
- Preview had no warning for locked commitments.

### Verification (after the merge with `fix/review-1002` / main)
`npx vitest run --project @plur-ai/core` 207 files, 3296 passed (1 expected fail: #999 `it.fails`; 36 skipped); `@plur-ai/cli` 61 files / 657 passed, 4 skipped; `@plur-ai/mcp` 40 files / 459 passed; `pnpm typecheck:tests` clean; `pnpm --filter @plur-ai/core gen:schemas` leaves `spec/` unchanged.

### Remaining risks
`PackManifest` now has an index signature (additive widening); `plur_packs_preview` returns unknown root fields to the LLM (the same text is in SKILL.md, which the scan covers); `declaredPrivateIds` parses `engrams.yaml` a second time (bounded by the 16 MiB cap); the bundled `effective-memory` pack still ships pinned/locked (#1030).

---

> **For a reviewer arriving cold.** This is specification work on
> [the Open Engram Standard](https://github.com/plur-ai/plur/blob/spec/pack-lifecycle/spec/ENGRAM-STANDARD-v1.md)
> and its [provenance profile](https://github.com/plur-ai/plur/blob/spec/pack-lifecycle/spec/ENGRAM-PROVENANCE-PROFILE.md).
> Seven files, ~780 lines added, almost all prose. One small code change.
>
> **Suggested reading order**, which is shorter than the diff:
> 1. `spec/ENGRAM-STANDARD-v1.md` §10.5, the revision history — every change with its reason, in one table
> 2. §4.7.1 *changing an engram* — the rule with the widest blast radius
> 3. §5.6 to §5.9 — install, update, uninstall, and where the reference does not yet comply
> 4. `spec/ENGRAM-PROVENANCE-PROFILE.md` §5.4 — receiving a pack's provenance
>
> **The question to hold while reading:** could you implement this without
> reading our TypeScript? That is not rhetorical — `plur-ai/encode` is doing
> exactly that, and it is the reason this work exists.
>
> Some references below point at `plur-ai/enterprise` and `plur-ai/encode`, which
> are private repositories. Those links will not resolve for everybody; the
> surrounding sentences say what they are.

Specifying what the standard left unsaid: the pack lifecycle, how an engram changes when it is edited, and what a consumer does with the provenance a pack ships. Replaces four stacked pull requests — #1037, #1038, #1039 and #1042 — with one reviewable unit, because the sections cross-reference each other and reviewing any one alone means reading a pointer to a section that does not yet exist.

Closes #1019 *resolve the seven pack contradictions*, #1020 *specify install, update and uninstall*, #1021 *what a consumer does with the provenance a pack ships*, #1041 *editing an engram: two mechanisms, no rule*.

**Based on `feat/958-engram-provenance` (#1002)**, because `spec/ENGRAM-PROVENANCE-PROFILE.md` does not exist on `main` — that PR introduces it, and this amends it. Merge order: #1002 → this → #1043 *conformance vectors*.

## Why this exists

The standard described a pack as an artifact and said almost nothing about its lifecycle. "Install" appeared three times as a passing mention. "Uninstall" appeared in neither document. The install registry was called authoritative twice and defined nowhere. And the document defined two ways an engram can change without ever saying which applies when.

That was tolerable while there was one implementation. There are now three, and one of them reads the document rather than our code:

| Implementation | Reads |
|---|---|
| `@plur-ai/core` | the reference |
| `plur-ai/encode` | **the document**, deliberately (encode#32) |
| `plur-ai/enterprise` | adopting the format for import (enterprise#1050) |

## What is in it

**Six of seven contradictions resolved** between the standard, the profile, the generated schema and the reference. §5.5.1 names the two different hashes that were being conflated. §9's single-sidecar sentence is corrected and marked as a correction, since §9 governs where it and the profile overlap. §5.4's strip list gains the supersedes edges, and `pinned` / `commitment: locked` become **MUST NOT ship** — the reference always removed them, the spec never said so, and unlike the rest of the list they change whose judgement governs the recipient's store.

The seventh is left to #1029 *the pack manifest is not open-world*, which needs a decision rather than an edit.

**§5.6 Install, §5.7 Update, §5.8 Uninstall.** The order of operations carries reasons: verify before scanning so a tampered pack is refused before its content is parsed; scan before neutralising so the scan sees what was shipped; record last so a failed install leaves no record claiming otherwise. §5.6.2 requires a consumer to be able to answer *which pack did this engram come from* — without it §5.8 is not implementable, which is exactly the wall enterprise#1049 hit. §5.6.3 forbids adopting a producer's shared scopes on the producer's say-so. §5.6.4 defines the install registry.

**§4.7.1 Changing an engram.** A change that moves `content_hash` MUST be a supersession; anything else is in place. A first draft tested whether *the meaning* changed and was withdrawn — meaning-change is a judgement, frequently subjective, and three implementations applying it independently would disagree. `content_hash` is mechanical, already computed on every write, and invariant under capitalisation, punctuation and spacing.

**Profile §5.4, receiving a pack's provenance.** The directory is evidence, not content. Report what was found. And a consumer MUST NOT merge a received `attribution` onto an installed engram unqualified — the tempting option, and the one that launders a stranger's claims into your own store. Where the values are kept, the intermediary must be named.

**A new maturity label, SPECIFIED.** The vocabulary had STABLE meaning both *binding on implementers* and *implemented in the reference*. Those are different claims, and neither describes a rule written ahead of its own implementation.

**§5.9 says where the reference does not comply** — five rules, each with its tracking issue. It doubles as the acceptance criteria for Phases 2 and 3 of #1018.

## What it changes in code

One thing: `readPackProvenance` now reads `engram:licenseSource` rather than the legacy boolean, so a pack preview can tell *the author configured this once* from *nobody ever chose it*. Four tests. Everything else is specification and documentation.

## Reading order

`spec/ENGRAM-STANDARD-v1.md` §4.7.1, then §5.4 through §5.9. Then `spec/ENGRAM-PROVENANCE-PROFILE.md` §5.4. The revision histories in both — standard §10.5, profile at the top — summarise every change with its reason.

## Testing

Full suite: 4861 pass, 0 fail.

Standard 1.1 → 1.4, profile 0.5 → 0.7. All additive per §10.2: no field removed, no constraint tightened on existing data, and the new obligations fall on consumers, which the document did not previously address at all.


plur9:
Read in full, every code claim verified against the branch, and the spec discussed in a long working session. Verdict: **approve direction** — the consolidation was right, §5.5.1 (two hashes) and the §5.4 receiving model are exactly right, SPECIFIED is a real vocabulary fix, and §5.9's honesty table is the right posture for a three-implementation standard. Two findings should land before merge; the rest are smaller.

**1. §4.7.1 contradicts §5.7's update model (must fix).** §4.7.1: a producer MUST NOT rewrite a statement in place when the hash moves — new id. But §5.7.2/5.7.3 key update correspondence on *stable ids across releases*, "even where the engram's statement changed", and §5.7.4 requires reporting changed carry-overs. Both cannot hold: a compliant producer never produces the carry-over-with-changed-hash that §5.7 is designed around — and when they do comply (new id, supersession), §5.4 strips the supersedes edges from the pack, so the recipient sees an unlinked departure+addition and the recipient-state preservation §5.7.3 exists for is lost by construction. Proposed resolution (discussed and agreed on our side): scope §4.7.1 to live stores; for packs, ids are *names* stable across releases, and the manifest carries each engram's `content_hash` so consumers key on the **pair** — identity binds to the id, judgement binds to the hash. A same-id/moved-hash carry-over then means exactly the right thing: same memory, changed claim → recipient re-reviews, state and history intact.

**2. Factual error in §4.7.1's non-compliance table (must fix).** "The previous statement is not retained anywhere" is false for dedup UPDATE: `learn-async.ts` writes `old_statement` into the `engram_updated` history event (verified on this branch, line 250) — and the profile itself makes the history log authoritative. MERGE likewise records the incoming statement, so the seam is computable, not "guessed". The normative rule survives (§4.7.1 already says history retention doesn't exempt — history is prunable and store-local), but the description should read "retained only in the prunable history log, not on any engram". encode builds its model of the reference from this document; it shouldn't inherit a wrong fact.

**3. The content-hash rule needs two guards.** (a) `HASH_NORMALIZER_VERSION` is already 2 — when the normalizer changes again, every stored hash "moves" with no edit anywhere, and the headline rule reads as minting thousands of supersessions. Rephrase to "an *edit to `statement`* that moves its hash", and state that renormalization/backfill is not an edit. (b) Statements that normalize to empty all share one hash, so any edit between two of them is "in place" under the rule — one acknowledging sentence. The punctuation blind spot ("…is idempotent." → "…is idempotent?") deserves the same honesty the typo-churn got.

**4. §5.6.1 step 2 is an undefined, non-overridable MUST.** "Secret scan" is never defined, so conformance is untestable and a false positive is a permanent lockout — a pack legitimately teaching key handling, containing an example `sk-…`, can never be installed by any conformant consumer, anywhere. Decision from our session: make it an **explicit, loudly-reported override** (mirroring step 1's integrity override), and/or define a minimum scan surface.

**5. §5.6.3 uses the wrong access test.** "Widen their audience beyond the installer's own *read* access" — placing engrams into a scope is a write; the ceiling should be the installer's **write** authorization for the target scope.

Smaller items: profile §5.4.2's "None of these MUST fail the install" is ambiguous RFC-2119 English → "A consumer MAY proceed despite any of these; it MUST report them." · §5.6.1 step 1 handles mismatch but not INTEGRITY *absence* (only §5.6.5 does). · Profile §5.4.3: say what happens to the shipped `provenance.origin` when the consumer writes `pack:<name>@<version>` — chain? records only? · §5.7.1: a consumer cannot discover "the comparison the producer documents" — manifest field, or state the SemVer-default inference. · §5.8.1's "where the consumer's model has a retired state…" leaves a from-scratch implementer unsure whether a retirement mechanism is mandatory; state the floor. · Code nit: in `readPackProvenance`, `licenseSource` lacks the `typeof === 'string'` guard that `license` has, so one malformed record pushes a non-string into `sources: string[]` — this module's stated job is packs built to mislead.

**One new question for §4.7.1** (Gregor's catch): does the successor inherit the predecessor's **activation**? Unspecified, three implementations will diverge on memory dynamics — and the naive reset means the *correction of a workhorse engram surfaces far less than the wrong version did*, which is backwards. Suggestion: successor SHOULD inherit activation; feedback signals reset (they rated the old wording). Packs unaffected (§5.4 resets activation on export anyway).

Decisions from our side to fold in: the mechanical hash test is ratified (context edits — tags/scope/rationale — stay in place; only the claim mints a new engram); `pinned` stays stripped, with #1019's open question answered as **installer consent at install time** (the same consent pattern as §5.6.3); a pack "intent" manifest field is welcome as a declaration that may route or inform but never relaxes a gate.

Full working notes — the hash map, the encode → import → proof-rail flow, the integrity-ladder grading against the auditability literature, and where this PR sits in it — are in plur-space: `1-tracks/engineering/provenance-ladder.html` (self-contained, open in a browser). Related but deliberately not touching this PR: the integrity-ladder implementation started as epic #1047 (+ enterprise-side counterparts) — it references the spec and stays out of `spec/`.

crtahlin:
All seven findings verified against the branch and applied in `a788b5ec`. Every claim held up; two of them were things I had wrong.

| # | Finding | Outcome |
|---|---|---|
| 1 | §4.7.1 contradicts §5.7's update model | **Fixed as proposed.** §4.7.1 scoped to a live store; §5.7.2 keys correspondence on the `(id, content_hash)` pair, with a note explaining why the two rules are complementary rather than in tension. |
| 2 | Factual error — "the previous statement is not retained anywhere" | **Fixed.** Verified `learn-async.ts:250` writes `old_statement`; MERGE records the incoming statement too. Now reads "retained only in the history log, not on any engram", with the normative rule unchanged. |
| 3a | `HASH_NORMALIZER_VERSION` is already 2 | **Fixed.** The rule now says an *edit to `statement`* that moves the hash, and renormalization is explicitly not an edit. |
| 3b | Statements normalizing to empty share one hash | **Stated**, alongside the punctuation blind spot (`…idempotent.` → `…idempotent?`), in the same "what this does not do" section as the typo churn. |
| 4 | §5.6.1 step 2 is an undefined, non-overridable MUST | **Fixed as decided.** Overridable, explicit, per-install, reported as loudly as the refusal, and a SHOULD that consumers document their scan surface. |
| 5 | §5.6.3 uses read access where it should use write | **Fixed**, with the reason stated: a person who can read a team's scope but not write to it must not be able to put a stranger's pack into it. |
| — | Code nit: `licenseSource` lacks the `typeof` guard | **Fixed**, three tests. |

Smaller items, all applied: profile §5.4.2's RFC 2119 reversal · §5.6.1 step 1 now covers a *missing* `INTEGRITY` as a third outcome · profile §5.4.3 says what becomes of the producer's own `provenance.origin` · §5.7.1 requires assuming SemVer, since a consumer cannot discover a scheme documented where it cannot read · §5.8.1 states the floor for a consumer with no retired state.

## Filed separately rather than settled here

**#1067** *does a superseding engram inherit its predecessor's activation?* — Gregor's catch, and the sharpest thing in the review. It is a design decision, not a correction to text, and it deserves its own discussion. The reason it matters: under a naive reset, correcting a heavily-used memory makes the correction surface *less* than the wrong version did, which is backwards and follows directly from a rule this PR introduces.

**#1030** *stripping pinned silently disabled the bundled pack's purpose* — answered with the decision from your session: installer consent at install time, the same consent pattern §5.6.3 uses for scope.

## On finding 1

Worth recording why it happened, since it is the kind of thing that recurs. §4.7.1 and §5.7 were written in separate sittings, days apart, and I never read one against the other — a spec whose sections are written serially will contradict itself exactly here, at the seam. §5.9's conformance table exists for gaps between spec and code; there is no equivalent check for gaps between one section and another, and this is the argument that there should be.

Standard 1.4 → 1.5, profile 0.7 → 0.8. Full suite: 4864 pass, 0 fail.

crtahlin:
One further addition, `5d10fe1f`: **§4.12.1, `content_hash` detects corruption, not tampering.**

Raised in the provenance-ladder design note, in a single line: *"`content_hash` lives in the file it protects — a corruption check, not a tamper check."* The standard nowhere said so.

The omission mattered specifically because of this pull request. §4.7.1 makes `content_hash` the authority on whether an edit changed a claim, so a reader arrives at `content_hash` understanding `content_hash` to be trustworthy — and nothing then says what `content_hash` cannot do. An implementer could reasonably conclude that comparing `content_hash` against a statement detects modification. It does not: `content_hash` is stored in the same file as the statement `content_hash` covers, so anybody able to edit the statement recomputes `content_hash` in the same write, and the recomputed value verifies.

§4.12.1 states both halves. What `content_hash` answers — has the statement changed since the hash was last computed, catching a truncated write, a partial sync, a corrupted file. And what `content_hash` does not answer — whether anybody altered the engram. An implementation MUST NOT present a `content_hash` match as evidence an engram is unmodified.

It also says why `content_hash` is nevertheless the right authority in §4.7.1: there, the producer computing `content_hash` is the party being asked. `content_hash` is reliable for that question and no defence against a party who does not want a change seen. Detecting tampering needs a value the editing party cannot rewrite in the same operation — §5.5's pack integrity value is one, an external anchor is another, and §7 signing, which would give an engram one, is RESERVED.

Cross-referenced from §4.3, §4.12 and §4.7.1.

Also filed, not added here: **#1068** *a pack should be able to declare what it is for, without that declaration relaxing any gate* — the `intent` manifest field the review called welcome. `intent` has no design yet, so `intent` becomes a Phase 3 issue rather than widening a pull request already reviewed once. The issue carries the constraint verbatim: a declaration may route or inform, and MUST NOT relax a gate.

Standard 1.5 → 1.6. Full suite: 4864 pass, 0 fail.

crtahlin (commented):
Reviewed. No blocking findings; three normative defects and several smaller items, filed as follow-ups. Posted as a comment rather than an approval because GitHub does not accept a self-review.

**First, the red checks are not this PR's doing, and not what they look like.** All four Node jobs fail at `pnpm typecheck:tests`, not at a test — five pre-existing TypeScript errors in test files introduced by the base branch `feat/958-engram-provenance`. I reproduced them byte-identically on that base in a clean worktree, and the same five appear in [#1002](https://github.com/plur-ai/plur/pull/1002)'s and [#1043](https://github.com/plur-ai/plur/pull/1043)'s own job logs. Typecheck gates `pnpm test`, so no test has run in CI anywhere on this stack. Run locally on the head, the suite is fully green: 4,863 passed, 0 failed, 350 files. One fix on [#1002](https://github.com/plur-ai/plur/pull/1002) clears all three. Filed as [#1090](https://github.com/plur-ai/plur/issues/1090).

The specification work itself is unusually careful. I verified §5.9's and §4.7.1's non-compliance tables line by line against `packages/core/src/packs.ts` and `learn-async.ts`, and every claim holds.

Follow-ups: [#1090](https://github.com/plur-ai/plur/issues/1090), [#1091](https://github.com/plur-ai/plur/issues/1091), [#1092](https://github.com/plur-ai/plur/issues/1092)

### Normative defects

- `spec/ENGRAM-STANDARD-v1.md:783` — §5.4 says a consumer **SHOULD** enforce neutralization on import; §5.6.1 step 3 says a consumer **MUST** do it. Same obligation, opposite keyword, one section apart. [#1043](https://github.com/plur-ai/plur/pull/1043)'s vectors already encode the MUST reading, so a consumer following §5.4 fails the suite while believing itself conformant. [#1092](https://github.com/plur-ai/plur/issues/1092).
- `spec/ENGRAM-STANDARD-v1.md:377` — §4.7.1's whole MUST turns on `content_hash`, cross-referenced to §4.3, which defines neither normalization nor `content_hash`. The one-sentence description, "strips non-word characters", is precisely the phrasing that produced [#896](https://github.com/plur-ai/plur/issues/896), where ASCII `\w` collapsed every non-Latin statement to the hash of the empty string. An implementer reading only the spec cannot arrive at `[^\p{L}\p{N}\p{M}_\s]/gu`. And `content_hash` appears nowhere in [#1043](https://github.com/plur-ai/plur/pull/1043)'s vectors, so the rule the PR body calls the one with the widest blast radius has zero conformance coverage. [#1091](https://github.com/plur-ai/plur/issues/1091).
- `spec/ENGRAM-STANDARD-v1.md:1000` — §5.7.1's "A producer using anything else MUST declare it in the manifest; until such a field is specified…" is unsatisfiable by construction: the same sentence says the field does not exist. Either drop it to a forward reference or specify the field — the root manifest schema is open (`additionalProperties: true`), so it is addable.

### §5.9 is missing two reference divergences

§5.9 exists so an implementer does not "discover [gaps] by disagreeing with us". Two are absent:

- §5.6.1 step 2 makes the secret-scan refusal MUST-overridable. The reference deliberately does the opposite — `packs.ts:674` documents "Secrets are ALWAYS blocked regardless", and line 761 throws unconditionally with no opt-out on any surface.
- §5.6.1 step 3 and §5.6.5 require reporting which field was changed. `sanitizePackEngrams` counts pinned strips but returns no counter for `commitment: locked` downgrades, and `packs.ts:833` warns only `if (sanitized.pinnedStripped > 0)`. A pack whose only host-overriding field is a locked commitment is silently altered with no output.

Related: this repository's own `packages/mcp/packs/effective-memory` ships 12 `pinned: true` and 3 `commitment: locked` engrams, and the SKILL.md edit here deliberately keeps them. `exportPack` complies with the new §5.4 MUST; this hand-authored pack does not, and it is exactly the shape [#1043](https://github.com/plur-ai/plur/pull/1043)'s `pinned-engram-shipped` negative vector rejects. Either carve out hand-authored packs in §5.4 or add the §5.9 row.

### Smaller items

- `spec/README.md:42` — only the profile version line was updated. "What is normative vs proposed" still says STABLE covers "the pack format and SHA-256 pack integrity (§5)" unqualified, and SPECIFIED is absent entirely — contradicting the Appendix B this PR changes to "STABLE for §5.1–§5.5 … SPECIFIED for §5.6–§5.8". This file is the entry point for the outside implementer the work exists for.
- `spec/ENGRAM-STANDARD-v1.md:891` — §5.6.5 makes a provenance-profile rule binding on every conformant consumer, while the profile's own status line says an implementation that ignores it "is still fully conformant". Restate the obligation in the standard rather than importing it.
- `spec/ENGRAM-STANDARD-v1.md:1070` — §5.7.3 describes producer-side reset as something "§5.4 requires"; §5.4 says SHOULD.
- Both spec headers still read `Date: 2026-08-26` while their revision tables date 1.6 and 0.8 to 2026-08-28.
- `packages/core/src/packs.ts:474` — the new `licenseSource` read guards the type but not the value domain, so any string survives into `sources: string[]`. `history.ts:365` has the right pattern one file over: `INJECTION_SOURCES.has(raw.source) ? raw.source : 'unknown'`. In a module whose threat model is "a pack built to mislead", an unbounded attacker-supplied string in a typed field deserves the same treatment — and bounding it also stops `seen.sources.includes(source)` being O(N²) on a crafted pack.
- `packages/cli/src/commands/packs.ts:127` — the preview renders only `l.chosen`; `l.sources` is populated but never displayed. The stated goal, that a preview can tell "the author configured this once" from "nobody ever chose it", is not met on the surface a user actually runs.
- `spec/ENGRAM-STANDARD-v1.md:820` — §5.6.1's rationale is "a tampered pack is refused before its content is parsed", but the reference calls `_previewPackDir(source)` — which parses the manifest, loads every engram and runs the full scan — at `packs.ts:742`, before the integrity comparison at `:750`. The refusal order matches; the parse order does not.
- The PR body says "Standard 1.1 → 1.4, profile 0.5 → 0.7" and "Full suite: 4861 pass". The branch lands 1.6 and 0.8, and the suite is 4,863.

### Done well

§5.9's non-compliance table is accurate in every row I checked, including details easy to get subtly wrong. `uninstallPack` really does call `removeFromRegistry` before `fs.rmSync(packDir)`, so "removed *before* the pack directory" is exact rather than approximate. `engram.pack` really is never persisted — only the in-memory `_pack` is set at load — so membership genuinely does depend on the directory, which is what §5.6.2 forbids. A conformance statement that survives verification at that granularity is rare.

§5.5.1's "known divergence" note describes `_installPackDir` precisely, and naming the reference's own defect in normative text rather than quietly specifying around it is the right call. Withdrawing the "did the meaning change" test in §4.7.1 for a mechanical one is the correct trade for a three-implementation standard, with the blind spots stated rather than left to be found. And `packages/core/test/pack-provenance-preview.test.ts:378` tests the type-corruption path, the boolean fallback and the well-formed path separately — four states plus three malformed cases for a nine-line change is the right amount of test.

Tests: PASS locally — 4,863 passed, 1 expected-fail, 0 failed, exit 0. CI is red on all four Node versions at `pnpm typecheck:tests`, inherited from the base branch.

Verified: the five typecheck errors reproduced on `feat/958-engram-provenance` in a clean worktree and matched against the job logs of all three stacked pull requests; §5.9's and §4.7.1's tables checked row by row against the implementation; the bundled pack's pinned and locked counts read directly; the parse-versus-refusal ordering traced through `packs.ts`.

crtahlin (commented):
Reviewed. This would be a request for changes, but GitHub does not accept one from the pull request's own author — so it is a comment. Please treat the blocking section as blocking.

The implementation is sound. I found no runtime defect in it, the hostile-input handling is genuinely defensive, and I confirmed by probe that the declared-private refusal cannot be reached past by any flag the CLI or MCP exposes. The problem is in the specification text, which for this pull request is the deliverable: the standard still contradicts itself on one obligation, the profile carries a statement about the reference that is false at this head, and two MUSTs the reference does not meet are missing from §5.9 — a section that calls itself a conformance statement.

### Blocking

- `spec/ENGRAM-STANDARD-v1.md:953` vs `:1000` — an unresolved SHOULD/MUST contradiction of exactly the class this pull request exists to fix. §5.6.3 says a consumer *SHOULD* report the scope every engram landed in; §5.6.5's MUST-report list contains the same item. The 1.7 revision note fixed the sibling case (§5.4 SHOULD against §5.6.1 MUST, #1092) and left this one. A consumer that follows §5.6.3 and reports no scope is conformant by one section and non-conformant by the other, which is what #1019 is about.

- `spec/ENGRAM-PROVENANCE-PROFILE.md:1894` — the Appendix C readiness row added here states that the reference discards the provenance directory on install (#989). That is false at this head. `packages/core/src/packs.ts:1171-1179` copies `provenance/` into the installed pack, and this pull request's own conformance test asserts the record survives install. A specification that misstates the reference misleads exactly the audience the body names — a reader who takes the document instead of the code. What is actually still open under #989 is narrower: install does not write `provenance.origin = pack:<name>@<version>` (profile §5.4.3, line 852) and does not report the four counts. Say that instead.

### Should fix

- `spec/ENGRAM-PROVENANCE-PROFILE.md:828` vs `packages/core/src/packs.ts:1113` — §5.4.2 lists "larger than the consumer is prepared to read" as an *unreadable* record and then requires that an unreadable record never abort preview or install. The reference aborts the install. Reproduced rather than reasoned: a 17 MiB `provenance/<id>.jsonld` is counted correctly as one unreadable record by `readPackProvenance` (preview survives, `unreadable_records = 1`), but `scanPackFiles` independently flags the same file as `unscannable` past the 16 MiB `MAX_PACK_FILE_BYTES`, and `_installPackDir` throws. The `it.each` list of eight top-level shapes uses only small files, so the suite cannot see it. Either carve provenance records out of the unscannable block, or amend §5.4.2 to say an oversize record is refused as an unscannable file and record that in §5.9.

- `spec/ENGRAM-STANDARD-v1.md:1163` — §5.9 has no row for §5.6.5, and two of its MUST-report items are unimplemented. `InstallResult` (`packages/core/src/packs.ts:895`) carries no provenance counts and no landed scope; `plur packs install` prints neither and `plur_packs_install` returns neither. Only `previewPack` computes them, and `PreviewResult.warnings` — where the global-scope warning lives — is not on `InstallResult` at all, so the §5.9 row for §5.6.3 saying the installer is warned and not asked is true of preview and not of install.

- `spec/ENGRAM-STANDARD-v1.md:861-876` — the no-override rule is coherent and the reference genuinely enforces it, but the stated remedy is not reachable through any shipped surface. The spec says a false-positive secret scan is resolved by the producer changing the example. A consumer whose producer is unreachable cannot fix it locally: editing `engrams.yaml` moves the integrity value, `preview.integrity.status` becomes `modified`, and `_installPackDir` throws unless `allowModified` is passed — which neither the CLI (it calls `plur.installPack(source)` with no options, and lists a `--force` flag wired to nothing) nor the MCP tool (`source` is its only input) can pass. The only escape is to delete `INTEGRITY`, documented nowhere. This does not change my view that the policy is defensible. It does mean the sentence justifying it overstates what a consumer can do.

- `packages/core/src/packs.ts:1084-1089` — the comment claims a consumer's scan surface is documented so a producer can predict it, and profile §5.6.1 step 2 states the same as a SHOULD. Nothing in `docs/` or `spec/` documents what `detectSensitive` and `detectSecrets` match. With the refusal now unconditional, an undocumented scan surface is the difference between a producer who can predict a refusal and one who cannot.

- `spec/ENGRAM-PROVENANCE-PROFILE.md:799` — §5.4.1 requires the received directory be kept as received, unmodified. Install filters it: `packages/core/src/packs.ts:1174-1178` copies only files matching `^[A-Za-z0-9._-]+\.jsonld$`, one level deep, so a subdirectory or any non-`.jsonld` file the producer shipped is silently dropped. The filtering is defensible; the sentence and §5.9 do not admit it.

### Notes

- `packages/core/src/packs.ts:1014-1017` — `manifestToSkillMd` preserves unknown root fields with `if (!(k in fm) && …)`. `in` walks the prototype chain, so a producer field literally named `constructor`, `toString`, `valueOf` or `hasOwnProperty` is silently dropped by the one path that rewrites a manifest — the case §10.3 rule 2 forbids. Use `Object.hasOwn(fm, k)`.
- `packages/cli/src/commands/packs.ts:11-18` — mirror problem, and the comment claims a defence the code does not provide: `LICENCE_SOURCE_WORDS[src] ?? '(unrecognised)'` is a prototype-chain lookup, and `constructor`, `toString` and `valueOf` all return truthy inherited values that `??` will not catch, after which `.join('; ')` would render function source into the terminal. Core gates the value, so it is not exploitable today. `Object.hasOwn` or a `Map` closes it.
- `packages/core/src/packs.ts:614-641` — with two engrams sharing an id, the same record file is opened twice: `record_count` reaches 2 for one file and `engrams_without_record` goes to 0 (verified by probe). §5.6.5 asks how many records were present against how many engrams the pack ships; counting one file twice misreports both halves, and `engrams_without_record` can go negative from the same cause. Dedupe the ids before the loop.
- `packages/core/src/packs.ts:609` — the orphan sweep skips any file whose name falls outside `^[A-Za-z0-9._-]+\.jsonld$`, so such a record is neither counted as an orphan nor read. Unreachable today given the engram id grammar; a note in the code would keep it that way.
- `spec/ENGRAM-PROVENANCE-PROFILE.md:1898-1899` — Appendix C carries a duplicated row twice in a row. Pre-existing, but this is the pull request editing the table.
- `spec/ENGRAM-STANDARD-v1.md:1504` — the 1.7 entry is classed as minor and additive. Making the declared-private refusal unconditional is a behaviour break: `packages/cli/test/packs.test.ts` had to change a fixture from `visibility: private` to `visibility: public` to keep passing, which is direct evidence that packs installable before are refused after. The CHANGELOG records it under Unreleased, which is the right place; the revision-history class label is worth a second look, as is whether a released consumer needs a migration note.

### On the decision you flagged

The no-override policy is defensible and I would not send it back on its own account. Two things about it are worth settling before it ships, both above: the remedy the spec names is unreachable through any surface you ship, and the scan surface it rests on is undocumented. Fix those and the policy stands on its own.

### Worth keeping

`declaredPrivateIds` reading the raw YAML — precisely because a parsed engram cannot distinguish a producer's declaration from the schema default the consumer assigned — is the part a shallower fix would have got wrong, and it is what makes the unconditional refusal safe. It survives a YAML anchor and merge key carrying `visibility: private`. The conformance test header states five invariants in the spec's own terms with every describe block mapped to one, and the tests are falsifiable: the `{"@graph": {}}` and `{"@graph": "x"}` shapes really did throw before the `graphOf` guard, and a 60,000-level nested record is caught by the `JSON.parse` fence rather than crashing the reader. Checking SIGNED against `header.signer` before the payload region is computed, with a message naming the ambiguity rather than the size mismatch it would otherwise cause downstream, is the right ordering. And rewriting the bundled pack's own documentation to say plainly that install strips the mechanism it was built around was the harder and more honest option.

Tests: PASS — `@plur-ai/core` no failures (one expected `it.fails`, #999); `@plur-ai/cli` and `@plur-ai/mcp` 101 files, 1116 passed, 4 skipped, after `pnpm build`. A first run failed 164 CLI tests only because `packages/cli/dist/index.js` did not exist yet — an environment artifact, not a defect. Verified: adversarial probes confirmed the oversize-record install abort and the YAML merge-key private declaration; probes were removed and the tree restored.

plur9 (dismissed):
Approving. The spec and the reference do agree where I checked them against each other, and the decision reversal is argued rather than asserted.

### Verified

- **The declared-vs-defaulted rule is implemented as specified.** §5.6.1 step 2 distinguishes an engram that *writes* `visibility: private` into the shipped file from one that merely omits the field, and `packs.ts:825-831` draws exactly that line: `declaredPrivateIds(source)` reads what the file says, and the two produce different `detail` strings. `packs.ts:1159` then refuses only `declared === true`, with a message citing §5.4 and telling the producer what to do. Refusing the defaulted case would reject every hand-written pack that omits the field, and the code says so where it decides.
- **The reversal to no-override is coherent with what was already there.** `installPack` now refuses secrets, declared-private and unscannable with no override, and keeps prompt-injection overridable — and the unscannable refusal carries a §5.4.2 carve-out so one crafted provenance record counts as one unreadable record rather than a veto on the pack. That is a defensible ordering: the two absolute refusals are about what the producer shipped, the overridable one is about what the consumer intends.
- **`gen:schemas` produces no diff.** Ran it; `spec/` is unchanged. The generated schema and the Zod source agree, which is what makes `additionalProperties: true` in the JSON schema and `.passthrough()` in Zod the same statement rather than two.
- Full monorepo suite on the head: **370 files passed, 15 skipped, 0 failed.**

### The passthrough risk is real and its mitigation holds

The description flags that `plur_packs_preview` returns unknown root fields to an LLM. I went looking for a hole there and did not find one: the manifest is parsed from SKILL.md's frontmatter, and `scanPackFiles` covers **every text file the pack ships** (`packs.ts:435` — "EVERY text file the pack ships, not a chosen two"), so the bytes that become unknown manifest fields are scanned before anything is returned. The stated mitigation is accurate.

The prototype-pollution tests are better than the claim in the description. Invariant 5 checks both directions on a real preview *and* install, and invariant 8 covers the subtler one — `k in fm` walks the prototype chain, so a producer field legitimately named `toString` or `constructor` read as already-present and was silently dropped by the manifest upgrade. That is a bug that would have surfaced as "my field vanished" months later, in someone else's implementation.

### Should fix

- **The description undersells what needs reviewing, and points away from the riskiest file.** "Seven files, ~780 lines added, almost all prose. One small code change" does not match the diff: **9 source files, +533/−33**, of which `packages/core/src/packs.ts` is **+370/−26** — the pack installer, which is the security boundary this whole stack is about. Add the six test files and it is +976 more. The suggested reading order is good and I used it, but a reviewer told "one small code change" will skim the 370-line change to the installer, which is the opposite of what this PR wants.

- **The merge instructions are stale across the stack.** Not specific to this PR, but it inherits the problem: #1002 says merge #1000 and #1001 first (both closed, never merged) and #1043 says it is fifth in a seven-PR stack through #1042/#1039/#1038/#1037 (all four closed, never merged; #1018 is not a PR). The real topology is three:

  ```
  #1043 → #1044 → #1002 → main
  ```

  Worth correcting on all three in one pass, so whoever merges is not hunting for commits that do not exist.

### Notes

- `readCapsule` refusing `SIGNED` with `signer: null`, and a signer with the flag clear, tightens a STABLE section. The argument that no writer-conformant capsule is affected looks right — both states are self-contradictory envelopes rather than shapes a conforming writer emits — but it is the one change here that could bite an outside implementation that was writing something sloppy and getting away with it. #1043's capsule fixtures cover exactly these two cases, which is the right place for the evidence.
- `declaredPrivateIds` parsing `engrams.yaml` a second time is fine at the 16 MiB cap, and the alternative (threading the raw text through) would couple the scan to the loader for a bounded cost.

### Done well

Writing the rationale for the reversal into the spec rather than into this thread is the right instinct — the next person to ask "why is there no override here?" will find the answer in §5.6.1 rather than in a closed PR. And `docs/pack-scan-surface.md` is the honest counterpart to a no-override rule: if a consumer refuses on a scan it MUST say what its scan covers, otherwise "refused" is unappealable and unexplained.

Blocked only by its base: this sits on #1002, where I have requested changes.

Tests: PASS — 370 files passed / 0 failed on the head; `gen:schemas` leaves `spec/` unchanged. Verified: read §5.6.1 against `packs.ts:825-831` and `:1159`; traced the manifest's path from SKILL.md frontmatter through `scanPackFiles`; counted the source diff by file; checked the state of every PR named in the merge order and the real base branch of each PR in the stack.

Issue #1071, opened 2026-08-28
Enterprise remote schema drift: server-served engrams rejected by stricter client validation were silently invisible to recall


plur9:
Observed live 2026-08-28 on plur.datafund.io: ~100 engrams served by the enterprise store failed the 0.19.0 client's `RemoteRowSchema` with `commitment: invalid_enum_value` and were dropped wholesale — silently invisible to recall, one debug log line per row as the only trace. (Surfaced during the #1069/#1070 diagnosis; filed separately as promised there.)

Silent drop is the worst of the three options (server-side migration, client-side tolerant parse, silent drop).

**Client side is fixed in the 0.19.1 patch (PR to follow):** both remote read paths (RemoteStore.reshape and remote-recall processHostRows) now share a `salvageRemoteRow` helper — on validation failure it strips exactly the top-level fields the Zod issues name and re-parses once. A drifted optional field costs that field, not the engram; a missing required field still fails the re-parse, so malformed rows still drop. One warning per drift shape per store instance.

**Still open (this issue):** server-side reconciliation. The enterprise server serves `commitment` values outside the client enum `['exploring','leaning','decided','locked','draft']` — either the server predates #905's enum ('draft') or carries values from a newer/forked schema. Decide: migrate the server rows, or widen/version the shared schema. The client salvage means users keep recalling these engrams meanwhile — minus their `commitment` field.

Also observed same morning, same config: plur-df.datafund.io fully unreachable at the TCP level (fetch failed / 8s connect timeout) while referenced by nine store entries — that stall is addressed by the #1069 host breaker, but the endpoint itself needs attention or removal from the config.

Issue #1070, opened 2026-08-28
0.19.0: plur_doctor over MCP kills the server instantly (Connection closed, 0s); CLI doctor unaffected


plur9:
## Summary

On 0.19.0, calling **`plur_doctor` through MCP kills the server instantly** — the process dies and the harness reports `Connection closed` after **0s**. The CLI `plur doctor` runs the same diagnostics without dying. Since Claude Code does not restart a server that dies mid-call, one `plur_doctor` call ends memory for the whole session.

## Environment

- plur-mcp **0.19.0**, `node packages/mcp/dist/index.js`, `PLUR_TOOL_PROFILE=full`
- macOS (Darwin 25.3.0), Node v24.10.0

## Observed (2026-08-28, warm server)

Claude Code MCP log — note `plur_status` on the same connection answers fine seconds earlier:

```
13:54:57  Calling MCP tool: plur_status
13:54:57  Tool 'plur_status' completed successfully in 513ms
13:55:01  Calling MCP tool: plur_doctor
13:55:01  Tool 'plur_doctor' failed after 0s: Connection closed
```

Observed once in production this session; the 0s instant-death signature is distinct from the cold-start SIGKILL issue (filed separately).

## Hypothesis (marked as such — not verified)

CLI `plur doctor` includes a **handshake self-test that spawns a fresh MCP server instance** (its output: `"handshake": {"ok": false, "error": "server exited (code=1) before responding"}`). If the MCP-side `plur_doctor` runs the same handshake test *from inside the running server*, a child spawned with inherited stdio would write into the parent's JSON-RPC stdout stream — corrupting the transport and producing exactly an instant `Connection closed`. Alternatively the recursive spawn may contend on `engrams.db` (WAL) and abort.

The `code=1` pre-response exit of the handshake child is itself worth a look — a fresh server exiting before responding suggests the startup path fails hard under contention (a second instance was running at test time).

## Repro sketch

1. Start plur-mcp over stdio, initialize, call any cheap tool (`plur_status`) — works.
2. Call `tools/call plur_doctor {}`.
3. Server process exits; transport closed.

## Impact

The diagnostic tool is the one thing users reach for when memory misbehaves — and calling it takes the server down. Worst possible failure mode for a doctor.

## Asks

1. MCP-side `plur_doctor` must never spawn a child that inherits the parent's stdout (pipe/devnull the handshake child), or should skip the handshake self-test when running inside a live server.
[...]

Issue #1069, opened 2026-08-28
0.19.0: MCP server SIGKILLed ~11s into first cold plur_session_start — whole session left memoryless


plur9:
## Summary

On 0.19.0, the MCP server is killed with **SIGKILL ~11s into the first cold `plur_session_start` of the day**. Claude Code does not restart an MCP server that dies mid-call, so a single cold-start death leaves the **entire session memoryless** — no recall, no learn, hooks degraded. Warm starts succeed.

## Environment

- plur-mcp **0.19.0** (released 2026-08-27), launched as `node packages/mcp/dist/index.js` with `PLUR_TOOL_PROFILE=full`, `PLUR_REMOTE_RECALL_TIMEOUT_MS=2000`
- macOS (Darwin 25.3.0), Node v24.10.0
- Store: `~/.plur/engrams.yaml` 12MB (~5,800 engrams), `engrams.db` 16MB (WAL mode)
- PGLite initial sync currently failing (`Aborted()` — separate pre-existing condition, surfaced on every CLI run)

## Observed (production, 2026-08-28)

From Claude Code MCP logs (`mcp-logs-plur/2026-08-28T07-21-31-905Z.jsonl`):

```
07:21:33.262Z  Successfully connected (transport: stdio) in 326ms   [server 0.19.0]
07:22:11.754Z  Calling MCP tool: plur_session_start
07:22:23.411Z  Tool 'plur_session_start' failed after 11s: Connection closed
```

Server process gone; harness marks plur disconnected for the rest of the session.

## Reproduction (same day)

Direct JSON-RPC driver over stdio (initialize → initialized → `tools/call plur_session_start {task: "..."}`):

- **Cold run #1**: stdout EOF at **11.3s**, `waitpid` exit code **−9 (SIGKILL)**. Stderr tail contained only `[plur:outbox] retry failed ... fetch failed` — no error, no OOM message.
- **Warm runs #2–#5**: all succeed in **12.9–13.3s**, valid session payload returned.

Failure signature is consistent with the production event (11.3s vs 11.6s), and both were the first cold run after the 0.19.0 upgrade.

## Ruled out

- **Kernel memory kill**: no `memorystatus`/`jetsam` events in the unified log in either death window.
- **Crash bug**: no crash report — consistent with SIGKILL (which produces none), inconsistent with SIGABRT/V8 OOM.
- **In-code kill**: no `process.kill`/`kill -9`/`pkill`/lockfile-singleton logic in mcp or core dists.
[...]

Issue #1066, opened 2026-08-28
Remove mig-seed.mjs — a dev scratch script with a hardcoded maintainer home path, committed to a public repo


crtahlin:
Found while auditing the 0.19.0 release (#1058).

`packages/core/mig-seed.mjs` is a development scratch script that was committed with the tier-switch migration work (`cb53f68f`) and is still on `main`. Line 5:

```js
import { PGLiteAdapter, engramSearchText } from '/Users/gregor/Data/5-plur/2-projects/plur/packages/core/dist/index.js'
```

Two problems, both minor:

1. It is dead code. The absolute import resolves on exactly one machine, so the script cannot run for any other contributor.
2. This repository is public, so it discloses a maintainer's local directory layout.

No user impact: `packages/core/package.json` sets `"files": ["dist"]`, so the file does not ship to npm. Confirmed at the `v0.19.0` tag.

## Suggested action

Delete it, or convert it into a proper test fixture with a relative import if the seeding logic is still wanted.

## Related, pre-existing

`packages/core/test/meta/real-extraction.test.ts:22-26` hardcodes five paths under a maintainer's home directory. Same class, not introduced by this release, worth cleaning up in the same pass.

plur9:
Fixed in PR #1072, shipped in v0.19.1.

PR #1072, opened 2026-08-28
fix: 0.19.1 — every 0.19.0 audit finding plus the same-day production crashes

### LINKED ISSUE #1059: plur init overwrites a harness MCP config file when it fails to parse, discarding the user's other servers
Found while auditing the 0.19.0 release (#1058).

`plur init` overwrites a harness MCP config file whenever that file fails to parse, discarding every server the user had registered in it.

## Reproduction

With `~/.gemini/config/mcp_config.json` containing two servers and a single trailing comma:

```json
{
  "mcpServers": {
    "github":   { "command": "npx", "args": ["-y", "@modelcontextprotocol/server-github"] },
    "postgres": { "command": "npx", "args": ["-y", "@modelcontextprotocol/server-postgres"] },
  }
}
```

Run `plur init --antigravity`. Result:

```json
{
  "mcpServers": {
    "plur": { "command": "/bin/sh", "args": ["-lc", "exec npx -y @plur-ai/mcp@latest"] }
  }
}
```

`github` and `postgres` are gone. The install report says `MCP server: registered`. No warning, no backup.

Verified end to end against a fake `HOME` with the built 0.19.0 CLI.

## Cause

`readConfig()` returns `{}` on a parse failure (`packages/cli/src/mcp-config.ts:236-243`). `installAntigravity` then merges the plur entry into that empty object and writes it back unconditionally (`packages/cli/src/commands/init.ts:921-927`).

Twenty-five lines earlier in the same function, the `hooks.json` leg has exactly the guard that is missing here — the parse-and-shape refusal added for adversarial-audit finding F2. In the same reproduction the user's `hooks.json` survived intact while the MCP file did not.

## Scope

Three call sites share the pattern:

| Leg | Location | Introduced |
|---|---|---|
| Antigravity | `init.ts:921-927` | 0.19.0 |
| Claude Desktop | `init.ts:625-631` | pre-existing |
| Cursor | `init.ts:646-651` | pre-existing |

Claude Desktop is likely the worst target, since that config usually holds the most servers.

## Suggested fix

Move the parse-and-shape refusal into `readConfig()` itself, so it returns a discriminated result rather than a silent `{}`, and no future install leg can miss it. Failing that, apply the existing F2 guard to all three call sites.

plur9:
Addresses the full set of findings from the independent 0.19.0 audit (#1058) and the two production crash bugs diagnosed live the same morning.

## Audit findings (#1058)

- **#1059 (P0)** — `plur init` refuses to write through a harness MCP config it could not parse, on all three legs (Claude Desktop / Cursor / Antigravity). E2E regression test replays the audit's exact repro (two servers + trailing comma) and asserts byte-identical survival.
- **#1060 (P1)** — `ensureSessionDir`/`sessionDirSafeToSweep` hoisted to the shared module; Codex AND Cursor hook families get the full tmp hardening (0700/0600, symlink+ownership vetting, sweep refusal). Bonus instance fixed: Cursor's counter threw out of the hook (fail-closed); now fail-open like the others.
- **#1061 (P1)** — doctor/migrate resolve PGLite selection through `resolveBackendTier` (config.yaml route honoured).
- **#1062** — git-config isolation is a shared test helper used by every git-spawning test file; verified under a hostile global gitignore.
- **#1063** — export: per-row containment (`malformed` counter), copy-before-view BYTEA decode, deterministic all-schema read.
- **#1065** — orphan advisory in `doctor --json`; changelog unreleased-marker fixed and now gated in release.sh (step 3.55).
- **#1066** — `mig-seed.mjs` removed.
- **#1064** — stays open: code comment now states exactly which delivery paths are live-verified; probe procedure posted on the issue (needs an interactively trusted Codex session).

## Production crashes (2026-08-28)

- **#1070** — `runStdio` survives unhandled rejections / uncaught exceptions with loud stderr logging. The '0s Connection closed' on plur_doctor was the server already dead from a background fault, not doctor. Spawned-process regression test induces a rejection through a gated seam and proves doctor returns and the server keeps answering.
- **#1069 (partial)** — host-level network-failure breaker: one dead host with nine store entries costs one timeout per process, not nine. The intermittent SIGKILL at peak RSS under system memory pressure stays open (findings posted on the issue).
- **#1071 (client half)** — `salvageRemoteRow` shared by both remote read paths: a drifted optional field costs the field, not the engram. Security posture preserved (poisoned values still never survive; required fields still hard-drop).

## Process

RELEASING.md gains a pre-release audit-discipline section (finding = bug class → sibling sweep; clean-environment suite run; 'verified live' names its event; stray-file diff sweep) — the retrospective on why the audit's two blockers were both un-propagated fixes of our own earlier findings.

Closes #1059, #1060, #1061, #1062, #1063, #1065, #1066, #1070.

Full suite: 4482 passed / 0 failed locally, including under `GIT_CONFIG_GLOBAL=/dev/null` and under a hostile global gitignore. (9 anonymous worker-exit errors under full parallel load on a swap-exhausted machine — the same contention pattern noted in the #1058 audit, files all pass.)


plur9:
**Audit round complete — three parallel audits (evaluator, data-loss, adversarial) over the full 0.19.1 diff, run under a new mandate: every finding names its bug class and is swept for siblings before the fix lands.** All findings addressed in `32412dd` + `7f1b609`.

The mandate existed because the 0.19.0 audit's blockers were un-propagated fixes — and it caught us doing it again, three times, in this very patch:

| Finding | Class | Where the class had already been fixed |
|---|---|---|
| **BLOCKING**: `plur init` destroyed a malformed `~/.claude/settings.json` (live repro: permissions, user hook, github server → all gone) | lenient-read-then-write-back (#1059) | the three MCP legs, in this same PR |
| **BLOCKING**: `plur-mcp init` had the same clobber on `.mcp.json` + `settings.json`, and never healed its own `@latest` entries | #1059 + entry-exists≠entry-correct | `plur init`, in this same PR |
| Cursor hooks leg missing the wrong-shape refusal | ADV-F2 | Codex leg, since 0.19.0 |

Also fixed: `upgradePlurMcpEntry` ownership narrowed to the exact shapes init writes (the first cut empirically clobbered deliberate pins, fork packages, and custom commands that merely mentioned the words — both audits demonstrated it) with field-level merge preserving unmodeled keys; `uncaughtException` now exits instead of surviving (a wedged-but-alive lock holder blocks every writer forever — data-loss audit's sharpest point) while rejections stay survivable with rate-limited logging; salvage fails closed on `visibility`/`pinned` (absence is more permissive than any value); mcp version parity + release-shape gates; per-test breaker reset extended to the mcp vitest project; my own committed scratch files and machine-local `.claude/settings.json`/`AGENTS.md` removed and gated (release.sh step 3.54).

**Attacks that held** (adversarial): salvage could not be escalated (full field classification on the issue trail), the host breaker could not be forced open for healthy hosts or weaponized for recall denial, tmp-dir hardening survived TOCTOU replay, required fields cannot be stripped back to life.

Full suite: 4506 passed / 0 failed, clean-env (`GIT_CONFIG_GLOBAL=/dev/null`).

Issue #1053, opened 2026-08-27
plur verify: structured chain verification, never a bare bool


plur9:
New CLI verb, per store, returning structured results:

```
{ chain_head, verified_events, breaks: [...], forks: [...], unprotected_legacy: {from,to}, snapshot_match }
```

Exit codes: 0 = verified; 1 = breaks or forks found; 2 = cannot verify (refusal — distinguishable from both). A verifier that cannot parse must refuse, never return an empty-but-clean result.

**Acceptance (each red-first)**
- [ ] Mid-chain tamper → named break
- [ ] Truncated tail → named break
- [ ] Forked chain fixture → named fork
- [ ] Corrupted JSONL line → exit 2
- [ ] Legacy-only store → legacy range reported, exit 0
- [ ] `snapshot_match` compares current `engrams.yaml` hash against the last checkpoint

Part of #1047. Depends on the chain + checkpoint issues.

Issue #1052, opened 2026-08-27
Checkpoint events: the anchorable object


plur9:
New history event type `checkpoint` carrying: chain head hash, SHA-256 of the store's `engrams.yaml` bytes, engram count, actor, timestamp — canonical form per the epic spec. Emitted on session end and via a new `plur checkpoint` verb.

A checkpoint commits to every event before it (the chain is linked), so one externally-anchored checkpoint later makes the whole prior history tamper-evident. It must be a well-formed object — a verification boolean is not anchorable.

**Acceptance**
- [ ] Checkpoint round-trips; hashing the store happens on checkpoint only, never per append
- [ ] Two checkpoints chain correctly across a month boundary
- [ ] `plur checkpoint` verb

Part of #1047. Depends on the chain issue.

Issue #1051, opened 2026-08-27
Hash-chain the history log (canonical bytes, prev links)


plur9:
Each history event gains `hash` (SHA-256 over its own canonical bytes) and `prev` (the predecessor's hash). One chain per store, continuing across monthly JSONL files; genesis = the first chained event. This takes the history log from append-only (Integrity Strength L1) to hash-chained (L2): modifying any past event breaks every later link.

### Canonical bytes (shared spec for this epic)

- UTF-8 JSON, keys sorted lexicographically by UTF-16 code units, no insignificant whitespace (`JSON.stringify(sortedDeep(obj))` semantics).
- No floats in chained events: timestamps are ISO-8601 strings, counts are integers; a float is a validation error.
- Hashes: lowercase hex SHA-256. Signatures: base64 Ed25519.
- Test vectors required: at least one event with non-ASCII (CJK) content and one with nested `data`, with expected canonical bytes + hash committed as fixtures.

**Scope**
- Read-last-before-append under the store lock; read the predecessor via tail-seek of the current month file, not a full-file parse.
- Concurrency: if the store lock is not cross-process, two writers fork the chain — either extend the lock (preferred, with a two-process test) or document the single-writer assumption AND make verify report a fork (two events sharing one `prev`) as a first-class outcome distinct from tamper.
- A chained append that cannot read its predecessor warns and records a named gap — it never fabricates a `prev`.
- Events written before this change form the "unprotected legacy range": loaded and reported honestly, never errored, never silently upgraded.
- History append stays best-effort relative to the caller's mutation (a failed history write must not fail the learn).

**Acceptance**
- [ ] Canonical-bytes test vectors (incl. CJK) as fixtures
- [ ] Month-boundary chaining fixture
- [ ] Two-process test or documented single-writer + fork detection
- [ ] `bench:micro --compare` deltas cited in the PR (learn path is hot; >10% learn-p50 regression blocks merge)

Part of #1047.

miles-on-nightshift:
## bench:micro --compare main pr (2026-08-28)

Benchmark run against the tail-seek hash-chain implementation (#1051 scope: learn path regression check).

**Environment:** feat/measured-under-869 working tree, n=100 iterations, tsx benchmark/micro.ts

### Results

| Operation | main (p50 / mean / p95) | pr (p50 / mean / p95) | Δ mean | Δ % |
|-----------|------------------------|----------------------|--------|-----|
[...]

Issue #1047, opened 2026-08-27
Epic: provenance ladder — history chain, checkpoints, signing (engram store to IS Level 3)


plur9:
The engram store today is mutable YAML whose `content_hash` lives in the file it protects: a corruption check, not a tamper check (Integrity Strength Level 0). The history log is append-only but not cryptographically bound (Level 1). This epic takes the store's record to hash-chained (Level 2) and signed-with-published-verify-keys (Level 3-capable), so that memory — the causal input to future agent behaviour — becomes auditable, not just observable.

Grading scale: the Integrity Strength ordinal (L0 none / L1 append-only / L2 hash-chained / L3 signed, third-party verifiable) from *Auditable Agents* (arXiv:2604.05485).

**Children** (in dependency order): #1048 #1049 #1050 (Wave 1 — attribution) · #1051 #1052 #1053 #1054 (Wave 2 — hash chain) · #1055 #1056 #1057 (Wave 3 — signing)

External anchoring of checkpoints (the non-suppressibility rung beyond L3) is deliberately out of this epic's scope.

PR #1073, opened 2026-08-28
feat(core,cli,mcp): hash-chain the history log + checkpoint events (#1051, #1052)

miles-on-nightshift:
## Summary

Implements #1052 — checkpoint events: the anchorable object.

- **`checkpoint` event type** added to `HistoryEvent` union with `CheckpointData` payload: `chain_head`, `store_hash` (SHA-256 of `engrams.yaml`), `engram_count`, `actor`
- **`hashEngramsFile()`** — SHA-256 of engrams.yaml bytes; only called at checkpoint time, never on the hot learn/feedback path
- **`emitCheckpoint()`** — writes a chained checkpoint event; uses `findPredecessorHash` for cross-month continuity
- **`plur checkpoint` verb** — new CLI command; `--json` for machine-readable output; actor = `'cli'`
- **`plur_session_end`** emits checkpoint best-effort after episode capture; returns `checkpoint_hash` in response

Also includes the #1051 hash-chain implementation in `history.ts` (sortKeysDeep, canonicalEventBytes, computeEventHash, appendHistory with prev/hash, tailSeekLastHash, findPredecessorHash) that the checkpoint event depends on.

## Acceptance criteria (#1052)

- [x] Checkpoint round-trips; SHA-256 of store computed on checkpoint only, never per-append
- [x] Two checkpoints chain correctly across a month boundary (`prev` of May checkpoint = hash of April checkpoint)
- [x] `plur checkpoint` verb (actor = `'cli'`, `--json` flag)

## Test plan

- `packages/core/test/history-checkpoint.test.ts` — 17 tests: payload shape, hashEngramsFile, emitCheckpoint, chain linkage, month-boundary chaining, store_hash isolation
- `packages/core/test/history-chain.test.ts` — 31 tests: #1051 hash-chain vectors, sortKeysDeep, canonicalEventBytes, computeEventHash, cross-month prev links, test fixtures
- Full CLI suite (439 tests) — all pass

Part of #1047 (Provenance umbrella). Depends on #1051.

---

## Rescoped 2026-08-31 (force-push)

Addressing the **Scope** section of the review. The branch was rebuilt from
`integration/core-store` carrying only its declared work.

**Removed** (they were never named in the title or body):

| Commit | What | Lines |
|---|---|---|
| `4336369` | Hermes `MemoryProvider` | 1,318 |
| `630d146` | packs INTEGRITY validation (#987) | 134 |
| `469e02c` | tensions `measuredUnderDiffers` gate (#869) | 172 |

**Result: `+1381 / -5` across 10 files** — against your measured 1,378 in-scope lines. The
extra 3 are one fixup: `958e81e`'s `index.ts` hunk re-exported `measuredUnderDiffers`, which
only exists in the dropped #869 commit, so splitting the branch broke the DTS build with
TS2305. Cross-contamination between commits editing the same export line — the concrete cost
of the bundling you flagged.

**The merge conflict is gone.** `mergeStateStatus` is now `CLEAN`; `packages/hermes/pyproject.toml`
is untouched, so nothing regresses the published version. That confirms your `git merge-tree`
finding that the Hermes commit was the sole conflict.

**Retitled** to name both issues. `958e81e` is one atomic commit implementing the #1051 chain
primitives (`sortKeysDeep`, `canonicalEventBytes`, `computeEventHash`, `tailSeekLastHash`,
`findPredecessorHash`) *and* #1052 (`hashEngramsFile`, `emitCheckpoint`, CLI verb, MCP wiring).
Your own scope maths counts both as in scope, so they stay paired rather than being split by
hunk.

**Base changed** from `main` to `integration/core-store` — a dedicated integration branch, so
Waves 1–3 are concurrency-tested as a unit and reach `main` as one reviewed promotion instead
of landing half a chain on the hot `learn` path.

The blocking defects — `emitCheckpoint` outside the store lock, `chain_head` disagreeing with
the event's own `prev`, and the absence of any chain verifier — are **not** addressed by this
push. They are next, with `plur verify` (#1053) first, since your review is what established
that every one of those failures is currently silent.


miles-on-nightshift:
## bench:micro — #1051 tail-seek latency check

Ran `pnpm bench:micro --compare main pr` to measure the learn-path impact of the `tailSeekLastHash` read added to `appendHistory()`.

**Baseline**: `d28981ad` (pre-tail-seek, no hash/prev fields in appendHistory)  
**PR**: `958e81ea` (current HEAD — with tailSeekLastHash, prev/hash chain)  
Both runs: 150 iterations, fresh tmpDir store.

```
Operation            main               pr                 Δ mean       Δ %
────────────────────────────────────────────────────────────────────────────
learn                24.06ms (p95 32.02) 23.46ms (p95 34.45) ↓ -0.60ms    -2.5
recall_bm25          26.71ms (p95 35.38) 26.30ms (p95 34.31) ↓ -0.41ms    -1.5
recall_hybrid        121.24ms (p95 134.06) 107.78ms (p95 112.58) ↓ -13.46ms  -11.1
inject               41.63ms (p95 72.89) 31.06ms (p95 40.11) ↓ -10.57ms  -25.4
inject_tokens        1896               1896               = 0          0.0
```

**learn-p50**: 20.42ms → 19.25ms (baseline median vs PR median). No regression.

The >10% learn-p50 regression concern is **not confirmed**. The tail-seek (`tailSeekLastHash`) reads the last non-empty line of the history JSONL to retrieve the predecessor hash. On a fresh store (benchmark condition) the file is absent, so the seek returns null in a single `stat` call — overhead is sub-millisecond and within noise floor. All operations are equal or faster on the PR.

This clears the #1051 bench blocker. The p95 learn tick-up (+7.6%, 32ms → 34ms) is within run-to-run variance on this machine.

crtahlin (changes requested):
Requesting changes — `emitCheckpoint` writes from outside the store lock its own module declares mandatory, at 100% of its new call sites, and I reproduced the fork that results.

The checkpoint design is sound on paper and single-threaded behaviour is correct. The problem is concurrency, and it lands on the one object this PR exists to make externally anchorable.

### Blocking

- `packages/cli/src/commands/checkpoint.ts:29` and `packages/mcp/src/tools.ts:3135` — both `emitCheckpoint` call sites run outside `Plur._withStoreLock`, while `packages/core/src/history.ts:222` documents that "appendHistory must be called from within the store write lock … the store lock provides the exclusion". `appendHistory` tail-seeks the predecessor then appends, so a checkpoint racing any concurrent write reads a stale tail and both events land with the same `prev`. Reproduced with a two-process probe using `withAsyncLock(engrams.yaml)` for the writer — exactly what `_withStoreLock` does — and a bare `emitCheckpoint` for the checkpointer: **83 duplicate-prev fork violations across 600 events** at a 200 KB store. A checkpoint that certifies a forked chain is worse than no checkpoint.

- `packages/core/src/history.ts:367` — `emitCheckpoint` computes `chain_head` via its own `findPredecessorHash` call, then does file I/O (`hashEngramsFile`), and only then calls `appendHistory`, which independently recomputes `prev`. Any write landing in that window makes `data.chain_head` disagree with the event's own `prev`. Measured **76/300** mismatching checkpoints at 200 KB, **174/300** at 2 MB, **101/200** at 15 MB. The payload meant for external anchoring is the field that goes wrong. Read the predecessor once under the lock and pass it into `appendHistory` rather than recomputing.

- No chain verifier exists anywhere in `core`, `cli` or `mcp` — grep for `verifyChain`, `verify_chain`, `verifyHistory` and chain verification returns nothing. Every failure mode above is silent, and so is deliberate tampering: deleting a middle event and recomputing `prev`/`hash` over the remainder yields a file with **zero broken links** and an **unchanged `store_hash`**. Truncating the history file entirely makes the next checkpoint emit `chain_head=null` with no error and no signal. Shipping the anchorable object before anything can check it leaves `integrity_ok` unfalsifiable in exactly the way [#987](https://github.com/plur-ai/plur/issues/987) just fixed for packs.

### Should fix

- `packages/core/src/history.ts:341` — `hashEngramsFile` hashes raw file bytes, not canonical content, so `store_hash` is not stable across platforms or YAML re-serialisation. Verified: LF versus CRLF, and trailing-newline versus none, over identical YAML produce different digests. An anchored `store_hash` therefore only verifies a byte-identical file copy — it will not survive git autocrlf, a different YAML emitter, or a different OS. Either hash canonical parsed content or document that narrow scope explicitly.
- `packages/core/src/history.ts:350` — `engramCount` is a caller-supplied parameter, never derived from the bytes that were hashed, so the attested count is not bound to `store_hash`. Verified: `emitCheckpoint(root, path, 99999, 'cli')` against a one-engram store is accepted and written verbatim.
- `packages/core/src/history.ts:33` — `sortKeysDeep` walks objects structurally and ignores `toJSON`, so any value with custom serialisation is destroyed in the canonical form while `JSON.stringify` writes it correctly to disk. A `Date` in `event.data` hashes as `{}` but stores as an ISO string, making that event permanently unverifiable. No current caller passes one (all 31 sites use `.toISOString()`), so this is latent — but `data` is typed `Record<string, unknown>` and the module's own docs promise that "two independent implementations that agree on this function agree on all hashes".
- `packages/cli/src/commands/checkpoint.ts:29` — `plur checkpoint` on a store with no `engrams.yaml` dies with a raw unhandled ENOENT. Verified by running the built CLI. A fresh install that has never learned anything hits this.

### Scope

Measured per commit against the merge base, only **1,378 of 3,006 lines** are in scope. Checkpoint is about 437 lines and the declared [#1051](https://github.com/plur-ai/plur/issues/1051) hash-chain prerequisite about 942. The remaining **1,628 lines (54%)** are three features named nowhere in the title or body: a Hermes `MemoryProvider` (1,318 lines, 44% of the PR), packs INTEGRITY validation ([#987](https://github.com/plur-ai/plur/issues/987), 134 lines), and the tensions `measuredUnderDiffers` gate ([#869](https://github.com/plur-ai/plur/issues/869), 172 lines).

That bundling has a concrete cost here. The Hermes commit is the **sole** merge conflict: the branch sets `packages/hermes/pyproject.toml` to 0.18.1 while `main` is already at 0.19.1, so merging would regress the published version. Verified with `git merge-tree` against `main` at `dbb30fd` — that one file is the only conflict, and the checkpoint code merges cleanly. Splitting removes the conflict along with the review burden.

The same tensions gate also appears on [#981](https://github.com/plur-ai/plur/pull/981) and [#1082](https://github.com/plur-ai/plur/pull/1082) in three diverged forms; see the note on [#981](https://github.com/plur-ai/plur/pull/981) for which one should carry it.

### Notes

- CI is effectively unrun: only `issue-guard` has executed on this head. No build, lint or test workflow has run, so GitHub gives no independent signal on any of the above.
- No CHANGELOG or docs entry for a new user-facing CLI verb, though the verb is correctly registered in the command map and help text.
- The apparent reverts of backend-selection and PGLite logic when diffing against `main` are an artifact of the branch being behind, not changes made here. Against the merge base, `index.ts` is export-line changes only.
- `receipt-io.test.ts` passes 7/7 here, so the failure seen on four other open pull requests does not appear on this branch.

### Done well

Shipping canonical-bytes test vectors as a fixture, including CJK and nested-object cases, is exactly right for a hash format a second implementation will have to match. Keeping `hashEngramsFile` off the hot write path and asserting in tests that ordinary events carry no `store_hash` is the correct performance call. `findPredecessorHash` handles cross-month continuity including an empty month, and is tested across a real boundary. The JSDoc throughout this module is unusually good — and the lock invariant being written down is the only reason the central finding here is provable rather than arguable.

Tests: PASS. Full suite after a build: 331 of 347 files pass, 15 skipped, 1 failed — `sync-push-failure.test.ts`, the known host gitignore issue ([#1062](https://github.com/plur-ai/plur/issues/1062)), which passes 3/3 under `GIT_CONFIG_GLOBAL=/dev/null`. All 179 PR-specific tests pass in isolation.

Verified: the fork and the `chain_head`/`prev` mismatch each reproduced with a two-process probe modelling the shipped call paths at three store sizes; event deletion and re-chaining shown to leave zero broken links; file-byte instability of `store_hash` reproduced across line endings; the forged `engram_count` written verbatim; the ENOENT reproduced against the built CLI; scope measured per commit against the merge base and the conflict confirmed with `git merge-tree`.

plur9:
## All blocking findings fixed

### Blocking

**`emitCheckpoint` outside the store lock at 100% of its call sites.** Fixed at the root rather than at the two callers. `appendHistory` documented a precondition the codebase did not have — 19 of 28 sites in `index.ts` were outside any `_withStoreLock` — and it could not simply call `_withStoreLock` itself, because `KeyedAsyncMutex` has no owner tracking and the 9 *correct* sites would have deadlocked. So `appendHistory` now serialises its own read-tail → compute → append with a short `flock`-equivalent, which fixes all 33 sites and both checkpoint sites at once. Lock ordering is one-way (store then history), so the two cannot deadlock.

Measured: 3 processes × 250 appends → **750 events, 0 duplicate-prev forks, 0 link breaks**. Control with the lock disabled produces forks at the same load.

**`chain_head` disagreeing with the event's own `prev`** (76/300 at 200 KB, 174/300 at 2 MB). The predecessor is now read **once**, inside the lock that appends, via a new `appendHistoryStamped` seam, and the payload is built from it. Your prescription exactly — read it once under the lock and pass it in, rather than recomputing. A test asserts `chain_head === prev`.

**No chain verifier anywhere.** This is now **#1094**, opened *before* the rest of Wave 2 because your review established it is a prerequisite, not a successor: without it every failure mode here — and in #1080 — is silent. It reports structured results (never a bare bool), distinguishes four break reasons, treats forks and unprotected-legacy ranges as their own outcomes, and refuses (exit 2) rather than reporting a clean chain over a file it could not fully read.

It also states the ceiling honestly, in both directions. Your delete-a-middle-event-and-recompute probe is a **test that asserts it is NOT detectable** from the log alone — so nobody later claims a property a hash chain does not have — with a companion test showing the same rewrite **is** caught once a checkpoint pinned the old head. That is what makes a checkpoint the anchorable object.

### Should fix

- `hashEngramsFile` now hashes canonical **content**, not raw bytes. You were right that an anchored `store_hash` otherwise proves only that someone holds a byte-identical copy; a test asserts LF and CRLF stores produce the same digest. Same choice and reasoning as `.datacore/lib/ledger/fold.py:state_root()`. Stated trade: it attests the engrams, not their whitespace.
- `engram_count` is derived from the bytes `store_hash` covers. A caller passing 99999 against a 2-engram store now yields 2.
- `sortKeysDeep` honours `toJSON`. This had to land **first**: js-yaml parses ISO timestamps into `Date` objects, so canonicalising parsed YAML walks straight into the latent defect you found.
- `plur checkpoint` on a store with no `engrams.yaml` explains itself instead of dying with a raw unhandled ENOENT — the state every fresh install is in.

### Scope

Rebuilt to its declared work: **+1381 / −5 across 10 files**, against your measured 1,378 in-scope lines. The Hermes provider, #987 and the #869 gate are gone (they already live on #981). The **merge conflict is gone** — `mergeStateStatus` is CLEAN — confirming your `git merge-tree` finding that the Hermes commit was its sole cause.

Retitled to name both issues: `958e81e` is one atomic commit implementing the #1051 primitives and #1052, and your own scope arithmetic counts both as in scope, so splitting them by hunk seemed like churn for its own sake.

Splitting did surface one real defect: that commit's `index.ts` hunk re-exported `measuredUnderDiffers`, which only exists in the dropped #869 commit, so the DTS build broke with TS2305 — the concrete cost of the bundling you flagged.

Base is now `integration/core-store`, a dedicated integration branch, so Waves 1–3 are concurrency-tested together and reach `main` as one reviewed promotion instead of landing half a chain on the hot `learn` path.

`core` 4735 passed / 159 skipped.

crtahlin (changes requested):
Requesting changes — two of the three blocking findings are genuinely fixed and I re-measured both. The third is still present, and the same defect class that caused finding 2 has reappeared on a different field.

Fixed, with numbers:

- **Duplicate-prev fork — fixed.** 0 forks in 600 events at a 200 KB store where the old code produced 83; and 0 forks, 0 link breaks, 0 gaps, 0 bad hashes in 900 events across 6 processes at 2 MB. The reason it is fixed is the right one: `appendHistory` now owns its own cross-process lock instead of documenting a precondition its callers never met.
- **`chain_head` versus `prev` drift — fixed.** 0 of 300 at 200 KB and 0 of 300 at 2 MB, against 76 of 300 and 174 of 300 before, because the payload is now built from the predecessor the append actually uses.

Scope is also resolved: 11 files, +1689/−8, with the Hermes `MemoryProvider`, packs INTEGRITY ([#987](https://github.com/plur-ai/plur/issues/987)) and the tensions gate ([#869](https://github.com/plur-ai/plur/issues/869)) all gone. All four previously non-blocking findings are fixed.

### Blocking

- `packages/core/src/history.ts:508` — `emitCheckpoint` calls `hashEngramsFile` and `countEngramsInStore` **before** entering `appendHistoryStamped`'s chain lock, so `store_hash` and `engram_count` are not bound to the `chain_head` the same event records. Probe, with a writer mutating the store under `withAsyncLock(engrams.yaml)` racing a bare `emitCheckpoint`: **294 of 297 checkpoints attested a store state that was not the state at their own `chain_head`**, median lag one mutation, maximum four; 112 of 129 at a 2 MB store. Every attested digest was a real earlier state, so this is stale rather than corrupt — but a third party who replays the log to `chain_head` and hashes the store gets a different digest, which is exactly the check [#1052](https://github.com/plur-ai/plur/issues/1052) exists to make possible. This is finding 2 relocated from `chain_head`-versus-`prev` to `store_hash`-versus-`chain_head`.

- `packages/core/src/history.ts:367` — **no chain verifier is reachable from this branch.** `grep` for `verifyChain` across `packages/*/src` and `packages/*/test` returns nothing, and there is no `verify.ts` in the CLI commands. The tamper reproduction still succeeds end to end: on the 600-event probe history, deleting one middle non-checkpoint event, recomputing `prev`/`hash` over the remainder and rewriting each checkpoint's `data.chain_head` yields 599 events with 0 fork violations, 0 link breaks, 0 bad hashes, 0 `chain_head` mismatches and a byte-identical `store_hash`. A *naive* re-chain that forgets `chain_head` is now caught (279 of 300 mismatches), which is a real improvement — a careful one is not. [#1094](https://github.com/plur-ai/plur/pull/1094) is filed but is not in this stack.

### Should fix

- `packages/core/test/store-concurrent-chain.test.ts:129` — vacuous guard, proven by mutation. Replacing the `withLock` call in `appendHistoryStamped` with a bare `stampAndWrite()` leaves both tests passing. The workers only call `Plur.learn`, which already holds the store lock, so they never exercise the unlocked call site (`emitCheckpoint`) that produced the original 83 forks. A concurrency test whose writers are serialised by an unrelated lock cannot detect the absence of the lock it claims to guard.
- `packages/core/test/history-checkpoint.test.ts:311` — vacuous guard, proven by mutation. Reverting `emitCheckpoint` to the pre-fix shape leaves all 20 tests passing, including "chain_head equals the event own prev — they cannot drift apart". The test is single-threaded, so both reads return the same predecessor either way.
- `packages/core/src/history.ts:527` — `emitCheckpoint` returns `CheckpointData` with no event hash, so the checkpoint's own identity is unobtainable from any caller. The CLI prints only `chain_head`, `store_hash`, `engram_count` and `actor`; `packages/mcp/src/tools.ts:3134` returns the **store** hash under the name `checkpoint_hash`, which is actively misleading. [#1052](https://github.com/plur-ai/plur/issues/1052) requires a well-formed object because "a verification boolean is not anchorable" — a payload omitting the hash you would anchor is not anchorable either.
- `packages/core/src/history.ts:392` — the `CheckpointData` doc states unconditionally that "a checkpoint commits to every event that precedes it in the chain". The lock-failure fallback at `:375` breaks that: it writes `prev: null` outside the lock. Under 8-way contention I measured 3 to 4 such gaps per 1600 events, and in one run a gap event interleaved with a lock holder's append, leaving an event whose `prev` is not the preceding line's hash. Declaring a gap rather than forking is the right call, but a checkpoint after a gap does not commit to anything before it, and with no verifier nothing says so.

### Notes

- [#1051](https://github.com/plur-ai/plur/issues/1051)'s acceptance list requires `bench:micro --compare` deltas cited in the PR, with a >10% learn-p50 regression blocking merge. None are cited. I measured it: learn mean 13.26 ms with the chain lock versus 14.73 ms with it bypassed at 60 iterations — inside run-to-run noise, and a direct micro-benchmark put uncontended lock overhead below the measurement floor against a roughly 5 ms fsync. The gate is satisfiable; it is simply unevidenced.
- `packages/core/src/sync.ts:836` — `withLock` backs off with a synchronous spin, and `appendHistory` now routes every history write through it with `maxRetries` 12 / `baseDelay` 2, a worst case near 8 s of burned CPU per waiter. Eight contending processes held 373% CPU for 13 s in my run. The spin predates this PR; this PR is what puts the hot write path on it.
- `packages/core/src/history.ts:341` — `hashEngramsFile` now parses before hashing, which fixes the LF/CRLF finding but adds a throw path: an unparseable `engrams.yaml` propagates a raw js-yaml exception out of `emitCheckpoint`, and `checkpoint.ts` guards only the missing-file case. Same class as the ENOENT bug that was fixed.
- `packages/core/src/history.ts:295` — `countEngramsInStore` and `appendHistoryStamped` are not re-exported from the package index, so an external verifier cannot reproduce the attested count through the public API even though `emitCheckpoint` and `hashEngramsFile` are exported.

### Done well

Replacing an unmet precondition with an actual lock inside `appendHistory` is the correct fix rather than the narrow one — it repairs all 28 call sites at once instead of the two named in the review, and the doc states the lock ordering and why the two locks cannot deadlock. The `appendHistoryStamped` `onPrev` hook is a clean way to build a payload from a value that only exists inside another function's critical section.

Keeping `chain_head` in the payload as redundant with the envelope's `prev` turned out to be a real tamper tripwire: the first tamper attempt, which re-chained `prev`/`hash` but not `chain_head`, showed up as 279 of 300 mismatching checkpoints. And honouring `toJSON` in `sortKeysDeep`, with a comment noting it is latent today and stops being latent the moment js-yaml parses an ISO timestamp into a `Date`, is exactly the right kind of pre-emptive fix.

Tests: PASS — 349 files / 4,735 tests passed, 0 failed, 187 s. The host-environment and embedder flakes named in the brief did not reproduce.

Verified: the fork and drift probes re-run at two store sizes and compared against the pre-fix numbers; the store-state binding measured with a two-process probe at 200 KB and 2 MB; the tamper reproduction re-run end to end; both regression tests mutation-tested by reverting their corresponding fixes; scope confirmed against the stacked base.

One disclosure, unrelated to this change: while exercising the built CLI during this review, a `plur checkpoint` invocation resolved the reviewing operator's real store rather than the scratch one and appended a checkpoint event to it, which was then removed. Recorded here because it happened during the review, not because it reflects on this branch.

plur9:
Pushed fixes for both blocking findings.

**`store_hash` / `engram_count` not bound to `chain_head`** — the attestation now
happens inside `onPrev`, in the same critical section that fixes the
predecessor. That widens the documented `onPrev` contract from "cheap, no I/O"
to "may read the store", and the cost is holding the chain lock across a
whole-store parse. That is the trade I think is right: checkpoints are explicit
and rare, history appends are frequent and short, and an attestation not bound
to what it attests is not worth the lock it saved.

An earlier pass at this bound `store_hash` to `engram_count` (one read instead
of two) and stopped there. That was a real defect but a different one — it is
finding 2 relocated again, and fixing the count binding left the `chain_head`
binding exactly as you measured it.

**No verifier reachable from this branch** — that is stack composition, not
something this branch can fix alone. #1094 now has all three of its blockers
addressed and is pushed.

Also from the non-blocking list:

- `onPrev` failures are now separated from LOCK failures. The `catch` in
  `appendHistoryStamped` falls back to an unlocked append; without the
  distinction a throw from `onPrev` — a checkpoint that could not attest its
  store — took that same path and wrote the event anyway with whatever `data`
  held. Now it aborts, and nothing lands.
- `CheckpointData` returns the checkpoint's **own** hash. `packages/mcp` was
  returning the STORE hash under the name `checkpoint_hash`, telling callers to
  anchor the wrong digest. The CLI prints it first, marked as the value to
  anchor.
- `plur checkpoint` guards the new throw path. `attestStore` refuses rather than
  attesting a count of 0 beside a hash of the real file, so an unparseable store
  reaches the CLI as a throw; it guarded only the missing-file case.
- `tailSeekLastHash` grows its window until it finds a line boundary (1 MiB
  ceiling). A fixed 8 KiB window lands inside any longer event line, and the
  next event then chains from nothing — a gap over an intact log. Worth noting
  the boundary detail: a JSONL file ends in a newline, so "does the window
  contain a newline" is satisfied by that byte alone and the search has to be
  for the last boundary after stripping the terminator.

**Not addressed, and you were right about both:**
`store-concurrent-chain.test.ts:129` and `history-checkpoint.test.ts:311` are
still vacuous. I added non-vacuous coverage for the same properties rather than
repairing them — a structural test asserting the chain lock is genuinely held
while `onPrev` runs, and a fail-closed test that fails when the onPrev/lock
separation is reverted. I also applied your standard to a test I had just added
and found it vacuous for the property its name claimed, so it is renamed and
scoped honestly.

The `bench:micro --compare` deltas are still uncited; your measurement stands as
the evidence that the gate is satisfiable.

Core suite: 3107 passed, 0 failed. The two-process probe that produced the
294/297 number remains the authoritative check — I have not reproduced it here.

PR #1078, opened 2026-08-28
fix(cli): doctor and migrate honour config.yaml backend for PGLite orphan check (#1061)

### LINKED ISSUE #1061: doctor and migrate call a config.yaml-selected PGLite store orphaned and tell the user to delete it
Found while auditing the 0.19.0 release (#1058).

`plur doctor` and `plur migrate` decide whether a `store.pglite` directory is orphaned by reading `process.env.PLUR_BACKEND` alone:

- `packages/cli/src/commands/doctor.ts:834`
- `packages/cli/src/commands/migrate.ts:45`

The 0.19.0 changelog documents two activation routes: "set `backend: pglite` in `~/.plur/config.yaml` **or** `PLUR_BACKEND=pglite`". `resolveBackendTier` honours both correctly. These two consumers honour only one.

## Reproduction

A store whose `config.yaml` contains `backend: pglite`, with no environment variable set:

| Selection route | `plur migrate` behaviour |
|---|---|
| `backend: pglite` in `config.yaml` | runs the orphan-export path |
| `PLUR_BACKEND=pglite` | correctly skips it |

On a real store the first case reaches `migrate.ts:58` and prints:

```
  <path>/store.pglite can now be deleted — YAML remains the source of truth.
```

That tells a user who deliberately selected the PGLite backend to delete the index their backend is currently using. `plur doctor` prints the equivalent orphan notice in the same situation.

## Impact

Not permanent data loss — the index rebuilds from YAML — but the advice is wrong, and the rebuild is precisely the cost the migration path exists to avoid.

## Suggested fix

Resolve the tier through `resolveBackendTier` (or at minimum read `config.backend` alongside the environment variable) in both call sites, rather than testing the environment variable directly.

miles-on-nightshift:
Closes plur-ai/plur#1061.

## Problem

`plur doctor` and `plur migrate` decided whether a `store.pglite` directory was orphaned by checking `process.env.PLUR_BACKEND` alone. The 0.19.0 changelog documents two activation routes: `PLUR_BACKEND=pglite` **or** `backend: pglite` in `config.yaml`. `resolveBackendTier` honours both; these two consumers honoured only one.

A user with `backend: pglite` in config.yaml but no env var saw:
- `plur doctor`: orphan advisory telling them the store.pglite can be deleted
- `plur migrate`: the embedding-export path running against their active store

## Fix

Both call-sites now read `config.backend` via `loadConfig` in addition to the env var. The check becomes:
```
env-var === 'pglite' || config.backend === 'pglite'
```

`buildReport` in `doctor.ts` also gains `pgliteOrphanDetected` in the JSON report so the check is directly testable without a PTY (piped stdout forces JSON mode, making the text-mode advisory untestable directly).

## Tests

Red-first: both new tests were confirmed failing against the unfixed code.

- `doctor.test.ts`: `pgliteOrphanDetected=false` when config.yaml selects pglite with no env var; `pgliteOrphanDetected=true` when store.pglite exists and neither route selects it
- `migrate-pglite.test.ts`: `pglite_embeddings` absent from migrate output when config.yaml selects pglite (export skipped); present when no selection (export runs, fails gracefully on empty dir)

### DIFF
--- a/packages/cli/src/commands/doctor.ts
+++ b/packages/cli/src/commands/doctor.ts
@@ -1,6 +1,6 @@
 import { spawn } from 'child_process'
 import { existsSync, readFileSync, realpathSync, statSync, accessSync, constants } from 'fs'
-import { join, extname } from 'path'
+import { join, extname, dirname } from 'path'
 import { homedir, platform } from 'os'
 import { createPlur, type GlobalFlags } from '../plur.js'
 import { outputText, outputInfo, outputJson, shouldOutputJson } from '../output.js'
@@ -19,7 +19,7 @@ import { hasPlurCursorHooks, readCursorHooksConfig } from '../cursor-hooks.js'
 import { hasPlurCodexHooks, readCodexHooksConfig } from '../codex-hooks.js'
 import { hasPlurAgyHooks, readAgyHooksConfig } from '../antigravity-hooks.js'
 import { codexHome } from '../mcp-config.js'
-import { computeContentHash, detectPlurStorage, loadEngrams } from '@plur-ai/core'
+import { computeContentHash, detectPlurStorage, loadConfig, loadEngrams } from '@plur-ai/core'
 
 /**
  * plur doctor — diagnose a Claude Code / Claude Desktop / Cursor installation.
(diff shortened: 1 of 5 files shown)


miles-on-nightshift:
Superseded by PR #1072 which merged the same fix (resolveBackendTier in doctor.ts and migrate.ts) as part of the 0.19.1 audit findings. The fix is in main.

Issue #1049, opened 2026-08-27
provenance.identity configuration, one identity per writing actor


plur9:
Identity comes from configuration, never the OS account; a per-write override wins over the configured value.

**Scope**
- Config key + parsing; docs page for multi-agent setups (one identity per writer sharing a store).
- `plur status` surfaces the active identity.

**Acceptance**
- [ ] Config parse tested (set, unset, override)
- [ ] Docs page added
- [ ] `plur status` shows identity

Part of #1047.

PR #1079, opened 2026-08-29
feat(core): thread session_id + write provenance.origin on every engram write (#1048)

### LINKED ISSUE #1048: Write sources[].session_id and provenance.origin on every engram write
Both fields exist on the Zod schema and persist to YAML, but are never populated: `provenance` is always undefined and `sources[].session_id` is always null.

**Scope**
- Thread `session_id` through `LearnContext` into the engram build.
- Where `provenance` is absent, write `origin: "agent:<identity>"` from the configured identity; `"agent:unidentified"` when none is configured (per the provenance profile, `unidentified` counts as unanswered).
- Old engrams untouched — no backfill, no migration.

**Acceptance**
- [ ] A `learn()` inside a session writes both fields; YAML round-trip preserved
- [ ] No-identity path writes `agent:unidentified`
- [ ] Red-first tests for both paths

**Flagged decision**: the exact `origin` vocabulary is a spec question — conservative grammar above, to be revisited with the standard's editors.

Part of #1047.

miles-on-nightshift:
## Summary

Wave 1 of the provenance ladder (plur#1047):

- **`session_id` in `LearnContext`**: new field that takes precedence over the older `session_episode_id` when both are supplied; falls back to `session_episode_id` when only that is present.
- **`provenance.identity` in config**: `PlurConfigSchema` and the `Plur` constructor both accept `provenance?: { identity? }`. Constructor-level option overrides the config file (useful for tests and programmatic callers).
- **`_buildProvenance()` helper**: always writes `{ origin, chain: [], signature: null, license: 'cc-by-sa-4.0' }`. `origin` is `config.provenance.identity` when set, else `"agent:unidentified"` — recorded honestly so multi-writer attribution is visible even without explicit configuration.
- **`learn()` and `_buildEngramShape()`**: both set `provenance` on every new engram. Old engrams are untouched (no backfill, no migration).

## Acceptance

All 12 tests in `test/provenance-origin.test.ts` pass:
- session_id threading (4 tests)
- provenance.origin no-identity path (4 tests)
- provenance.origin identity-configured path (2 tests)
- learnRouted path (1 test)
- YAML round-trip (1 test)

Pre-existing failures (PGLite/zod-to-json-schema not installed in this env) are unaffected.

## Related

Closes #1048. Part of #1047.

### DIFF
--- a/README.md
+++ b/README.md
@@ -516,6 +516,25 @@ Everything is plain YAML. Open it, read it, edit it.
 
 `PLUR_PATH` overrides the default location.
 
+### Who wrote this engram
+
+Every engram records the actor that wrote it, in `provenance.origin`. Set the
+identity for this machine in `config.yaml`:
+
+```yaml
+provenance:
+  identity: agent:claude-code    # agent:<name> — one per writing actor
+```
+
+With no identity configured, engrams are stamped `agent:unidentified`. That is
+recorded honestly rather than guessed, so a store written by several actors
+still shows that it was — but the attribution is only useful once each writer
+names itself. `plur status` reports the value in effect.
+
+For a store several agents write to (a shared team store, or one machine
+running an MCP server alongside a CLI), give each writer its own identity.
+
+
 Indexing is on by default (`index: true`) and the backend is chosen from the
 size of your store, so there is normally nothing to configure:
 
--- a/packages/claw/src/context-engine.ts
+++ b/packages/claw/src/context-engine.ts
@@ -379,7 +379,11 @@ export class PlurContextEngine implements ContextEngine {
     const key = statement.toLowerCase()
     if (seen.has(key)) return
     seen.add(key)
-    await this.plur.learnRouted(statement, context)
+    // #1048: claw already knows which session this is — it keys dedup and the
+    // session scope on it — so the engram should record it. Without this every
+    // claw write lands with sources[].session_id null despite the id being in
+    // scope two lines up.
+    await this.plur.learnRouted(statement, { session_id: sessionKey, ...context })
     maybeFlushAfter(recordEvent('learn'))
   }
 
--- a/packages/cli/src/commands/learn.ts
+++ b/packages/cli/src/commands/learn.ts
@@ -106,6 +106,12 @@ export async function run(args: string[], flags: GlobalFlags): Promise<void> {
     ...(knowledgeAnchors !== undefined ? { knowledge_anchors: knowledgeAnchors } : {}),
     ...(dualCoding !== undefined ? { dual_coding: dualCoding } : {}),
     ...(supersedes !== undefined ? { supersedes } : {}),
+    // #1048: a CLI invocation is a session too. Without this every engram
+    // written by `plur learn` lands with sources[].session_id null, and the
+    // attribution the epic exists to provide is absent from the most common
+    // write path there is. PLUR_SESSION_ID lets a wrapping harness thread its
+    // own id through; otherwise the run identifies itself honestly as a CLI run.
+    session_id: process.env.PLUR_SESSION_ID || `cli:${process.pid}`,
   }
 
   // ALWAYS use learnRouted (not learn): learn() stamps _demoted but does NOT do
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -365,6 +365,16 @@ export interface StatusResult {
   pack_count: number
   storage_root: string
   config: PlurConfig
+  /**
+   * The `provenance.origin` this instance stamps on every engram it writes
+   * (#1049). Surfaced because `provenance.identity` otherwise has no
+   * user-facing surface at all — it is absent from the README, the docs, the
+   * `plur init` template, the CLI and the MCP schemas, so in practice every
+   * engram would ship as `agent:unidentified` and nobody would know why.
+   * Shows the effective value, including a constructor override.
+   */
+  provenance_identity: string
+
   locked_count?: number
   tension_count?: number
   versioned_engram_count?: number
@@ -716,7 +726,26 @@ function isStoreTeardownError(err: unknown): boolean {
   return msg.includes('adapter is closed') || msg.includes('after calling end')
 }
 
+/**
+ * The session key for a call — `session_id`, falling back to the deprecated
+ * `session` alias (#1048/#243).
+ *
+ * One accessor so the scope registry and provenance can never read different
+ * fields for the same value, which is the bug this merge closes.
+ */
+function sessionKeyOf(o: { session_id?: string; session?: string } | undefined): string | undefined {
+  return o?.session_id ?? o?.session
+}
+
 export class Plur {
+  /**
+   * Constructor-supplied provenance identity, applied on top of whatever the
+   * current config says. Deliberately NOT stored in `config`: every
+   * `this.config = loadConfig(...)` would discard it, and one config.yaml
+   * mtime change is enough to trigger that via `reloadConfigIfChanged()`.
+   */
+  private _provenanceIdentityOverride?: string
+
   private paths: PlurPaths
   private config: PlurConfig
   private indexedStorage: IndexedStorage | null = null
@@ -942,10 +971,13 @@ export class Plur {
       }
     }
     this.config = loadConfig(this.paths.config)
-    // Merge constructor-level provenance options (highest priority) so callers
-    // can set identity without touching a config file (#1048 / #1049).
-    if (options?.provenance !== undefined) {
-      this.config = { ...this.config, provenance: { ...this.config.provenance, ...options.provenance } }
+    // Constructor-level provenance identity is held OUTSIDE this.config so it
+    // survives every reload (#1048 / #1049). Merging it into this.config was
+    // the defect: the very next `this.config = loadConfig(...)` discarded it,
+    // and there are four such assignments — including one a few lines below in
+    // this same constructor, once autoDiscoverStores changes the store count.
+    if (typeof options?.provenance?.identity === 'string' && options.provenance.identity.length > 0) {
+      this._provenanceIdentityOverride = options.provenance.identity
     }
     this._autoDiscover = Plur.resolveAutoDiscover(options?.autoDiscover)
     // Auto-discover project stores from CWD (skips temp dirs for test safety).
@@ -1672,10 +1704,20 @@ export class Plur {
   private _buildSourceEntry(scope: string, context?: LearnContext): {
     scope: string; session_id: string | null; stored_at: string
   } {
+    // One emptiness rule, shared with _buildProvenance below. `??` alone let
+    // `session_id: ''` win over a genuine `session_episode_id` and persist as
+    // the empty string, while provenance fourteen lines down treated an empty
+    // identity as absent. Two different notions of "empty" in one change is how
+    // a field ends up meaning different things at two call sites.
+    const present = (v: unknown): v is string => typeof v === 'string' && v.length > 0
+    const sessionId = context?.session_id ?? context?.session
     return {
       scope,
-      // session_id takes precedence over session_episode_id (#1048).
-      session_id: context?.session_id ?? context?.session_episode_id ?? null,
+      // session_id (incl. its deprecated `session` alias) takes precedence over
+      // session_episode_id (#1048).
+      session_id: present(sessionId) ? sessionId
+        : present(context?.session_episode_id) ? context!.session_episode_id!
+        : null,
       stored_at: new Date().toISOString(),
     }
   }
@@ -1689,7 +1731,18 @@ export class Plur {
    * Wave 2 / Wave 3). license defaults to cc-by-sa-4.0 per the engram standard.
    */
   private _buildProvenance(): { origin: string; chain: string[]; signature: null; license: string } {
-    const identity = this.config.provenance?.identity
+    // Constructor override first, config file second.
+    //
+    // The override used to be merged INTO this.config, and any later
+    // `this.config = loadConfig(...)` threw it away — including the one in the
+    // constructor itself after autoDiscoverStores changes the store count, plus
+    // reloadConfigIfChanged, addStore, and one more. `_resolveUnscopedScope`
+    // calls reloadConfigIfChanged() on every unscoped learn(), so a single
+    // config.yaml mtime change was enough to silently revert an explicitly
+    // configured identity to "agent:unidentified" — defeating the attribution
+    // this epic exists to provide. Held separately now, applied on top of
+    // whatever the current config says.
+    const identity = this._provenanceIdentityOverride ?? this.config.provenance?.identity
     return {
       origin: (typeof identity === 'string' && identity.length > 0) ? identity : 'agent:unidentified',
       chain: [],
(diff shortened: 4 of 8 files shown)


miles-on-nightshift:
Note: Per the issue comment by @plur9, PR #1002 (feat/958-engram-provenance) contains related provenance work that should merge first. This PR implements the specific acceptance criteria of #1048 (session_id threading + provenance.origin with agent:unidentified fallback) from scratch against current main. If #1002 merges first, this PR can be evaluated as a gap-closure rather than a fresh implementation. The changes here are narrow-scope (no W3C PROV-O, just the session_id + provenance.origin fields) and should merge cleanly after #1002 assuming no field name collisions.

crtahlin (changes requested):
Requesting changes — the core plumbing is right, but neither headline claim holds end to end: `sources[].session_id` is null on every shipping write path, and the constructor identity override is discarded by the next config reload.

The core-level work is genuinely good. `learn`, `learnAsync`, `learnBatch`, `learnRouted`, `episodeToEngram`, `import` and `rescope` all converge on the two patched construction sites, and every one produces a well-formed provenance block. The `session_id` / `session_episode_id` precedence behaves exactly as described for all four combinations. The problems are at the callers and at config reload.

### Blocking

- `packages/mcp/src/tools.ts:1044` — the MCP `plur_learn` handler maps `args.session_id` into `context.session` (scope selection, documented as never persisted) and never into the new `context.session_id`, so the field this PR adds has zero non-test callers. Driving the built tool handler end to end inside a live `plur_session_start` session produced `sources[0].session_id: null` both without and with an explicit `session_id` argument. `plur_learn_batch` (`tools.ts:1249-1262`), the CLI `plur learn` (`packages/cli/src/commands/learn.ts:134`) and claw's `_learnIfNew` (`packages/claw/src/context-engine.ts:373`) are all the same. [#1048](https://github.com/plur-ai/plur/issues/1048)'s first acceptance criterion — a `learn()` inside a session writes both `sources[].session_id` and provenance — is met only in the new test file, never in the product. Worth noting that `context.session` and `context.session_id` are now two near-identical fields with different meanings, which is how this happened.

- `packages/core/src/index.ts:947` — the constructor-level provenance merge is written into `this.config` and then thrown away by any later `this.config = loadConfig(...)`: line 963 in the constructor itself after `autoDiscoverStores` changes the store count, plus 8264 (`reloadConfigIfChanged`), 8315 (`addStore`) and 9206. Because `_resolveUnscopedScope` (`index.ts:2334`) calls `reloadConfigIfChanged()` on every unscoped `learn()`, one `config.yaml` mtime change is enough. Reproduced on a single instance: the first learn wrote origin `agent:ctor-identity`, then after an out-of-process config edit the next learn wrote `agent:unidentified` and `p.config.provenance` was `undefined`; same result after `addStore()`. Silent, no warning, and it defeats the attribution the epic exists to provide. Config-file identity survives because it is re-read from disk — only the documented "constructor option overrides the config file" contract breaks. Hold constructor overrides in a separate field applied on top of every reload.

### Should fix

- `packages/core/src/meta/formulation.ts:176` — `formulateMetaEngram` builds a complete engram with `sources: [{scope:'global', session_id:null, stored_at}]` and no provenance block at all, and `Plur.saveMetaEngrams` (`index.ts:5401`) pushes it into the store verbatim. This is a public, exported write path that lands engrams with no origin, so "every engram write" is not literally true.
- `packages/core/src/index.ts:1732` — `_recordDuplicate` appends a source entry to a pre-existing engram and re-persists it without adding a provenance block. Confirmed: after stripping provenance from an on-disk engram, a duplicate write and a `feedback()` call both re-serialised it with provenance still absent. With no backfill or migration, a store stays permanently split between attributed and unattributed engrams even for engrams actively being written to.
- `packages/core/src/index.ts:1678` — two different emptiness rules fourteen lines apart in the same PR: `_buildSourceEntry` uses `??`, so `session_id: ''` wins over a genuine `session_episode_id` and is persisted as the empty string, while `_buildProvenance` at line 1692 treats an empty identity as absent via `identity.length > 0`. Use the same guard in both.

### Notes

- `packages/core/src/schemas/config.ts:354` — `provenance.identity` has no user-facing surface: absent from the README, `docs/`, the `plur init` config template, the CLI and the MCP tool schemas. The only way to set it is a hand-edit of `config.yaml`, so in practice essentially every engram written after this lands ships as `agent:unidentified`.
- `packages/core/src/schemas/config.ts:355` — a non-string identity makes `loadConfig` reject the whole config and fall back to defaults, silently discarding `stores`, `backend` and `allow_secrets`. Pre-existing `loadConfig` behaviour, but this PR adds a new hand-edited key that widens the blast radius.
- `packages/core/src/index.ts:1697` — `license` is hardcoded with no `config.provenance.license` key. The value is conformant — ENGRAM-STANDARD-v1 §4.8 gives `cc-by-sa-4.0` as the default and `ProvenanceSchema` already defaults to it — so this is not a defect, but baking a share-alike license into private personal memory with no opt-out is worth a deliberate decision rather than a default.
- `packages/core/src/quality.ts:72` — `completenessScore` counts provenance as one of seven enrichment fields. Now that every new engram gets a constant block, that dimension stops discriminating and new engrams score one seventh higher than identical legacy ones.
- No CHANGELOG entry for [#1048](https://github.com/plur-ai/plur/issues/1048), though the surrounding commits do declare their issue numbers.
- Related: [#1082](https://github.com/plur-ai/plur/pull/1082) independently adds a `_buildProvenanceField` and reads a `config.provenance` key that `PlurConfigSchema` does not declare, which fails `typecheck:tests` there. The two branches are building the same field in parallel and should be reconciled.

### Done well

Choosing an explicit `agent:unidentified` sentinel over omitting the block is the right call — the absence of attribution stays greppable and countable instead of looking like a legacy engram, and the comment says exactly why. The two patched construction sites really are the whole funnel for the ordinary paths. `packages/core/test/provenance-origin.test.ts` is a genuinely good test file: all four precedence combinations, both identity branches, the `learnRouted` shape, and a save-and-reload round trip proving the block survives YAML. And the emitted shape matches ENGRAM-STANDARD-v1 §4.8 exactly, including writing `signature` as null as the spec requires.

Tests: PASS. Full suite after a build: 344 files passed, 15 skipped; the 8 failures are all `Author identity unknown` from the `GIT_CONFIG_GLOBAL=/dev/null` harness setting and pass on isolated re-run. `pnpm typecheck:tests` is clean. The `receipt-io.test.ts` failure seen on four other open pull requests does not appear here.

Verified: the MCP tool handler driven end to end inside a live session and the stored record inspected; the constructor override reproduced being lost after both an out-of-process config edit and `addStore()`; every listed write path exercised against a real store rather than read; the empty-string precedence probed directly.

plur9:
## All findings addressed

Thanks — both blocking findings were real, and the "two near-identical fields" note turned out to be the root cause rather than a side observation, so that is what got fixed.

### Blocking

**`sources[].session_id` null on every shipping path.** Fixed at all four sites: MCP `plur_learn`, `plur_learn_batch`, the CLI, and claw's `_learnIfNew` — which already had the session key in scope two lines above the call.

Rather than wire a second field, the two were **merged**. Your note named why: `context.session` and `context.session_id` are two fields carrying one value, distinguished only by which consumer reads them. The public MCP schema has always called this input `session_id` and documents scope resolution as its job — so the handler mapping `args.session_id → context.session` was correct for its original purpose, and this PR's same-named field was left unwired because the obvious mapping was already taken. Two fields made "scope resolved but provenance null" the *default* outcome rather than an impossible one.

`session_id` is now the single field, read by both consumers through one accessor so they cannot drift. `session` stays as a deprecated alias for one release. Its docstring claimed the value is never persisted, which was already untrue — claw writes it into scope strings as `session:${sessionKey}`.

**Constructor override discarded by config reload.** Held in its own field now, applied on top of whatever the current config says. You were right about the mechanism: four `this.config = loadConfig(...)` assignments, and `_resolveUnscopedScope` calls `reloadConfigIfChanged()` on every unscoped `learn()`, so one mtime change was enough.

### Should fix

- `formulateMetaEngram` / `saveMetaEngrams` and `_recordDuplicate` both stamp provenance now, so "every engram write" is literally true. Neither overwrites an existing block, and there is still no backfill.
- One emptiness rule shared by `_buildSourceEntry` and `_buildProvenance` — `??` alone let `session_id: ''` beat a real `session_episode_id` and persist as the empty string.

### Notes

- `provenance.identity` has a user-facing surface at last: a README section and `plur status` reporting the **effective** value including a constructor override. Your point stands that without one, essentially every engram ships as `agent:unidentified` and nobody learns the key exists.
- A non-string identity no longer makes `loadConfig` reject the whole config and silently discard `stores`, `backend` and `allow_secrets`. It degrades to absent — which is the honest `agent:unidentified` outcome — so a typo in a hand-edited key costs the identity, not every configured store.
- `license` hardcoding is left as-is for now; it is conformant, and a `provenance.license` key without a use case seemed worse than the constant.

### Tests

Six new, including a **call-site assertion**. That one matters: the defect was an absent call site, not a logic error — every unit test passed while the product recorded nothing — and only a test that looks at the call sites can catch that class.

`core` 4697 passed / 159 skipped, full build clean. Base is now `integration/core-store` rather than `main`.

crtahlin (changes requested):
Requesting changes — both original blocking findings are genuinely fixed in the product, but the fix commit ships tests that write into the developer's real `~/.plur`, and because of the same defect the guard for one of those fixes is inert.

The production work is right, and I confirmed it by driving the built code rather than reading the diff.

- **`session_id` reaching the store — fixed.** `sources[].session_id` now persists on all four shipping paths: MCP `plur_learn` with and without an explicit argument, `plur_learn_batch`, CLI `plur learn`, and claw's `_learnIfNew`. The field merge did not break the original job either — a per-session default scope still resolves and now also attributes.
- **Constructor identity surviving config reload — fixed.** It holds across a `config.yaml` mtime change, `addStore()` and `status()`, with constructor-over-config precedence and non-string degradation behaving as documented. Holding it in `_provenanceIdentityOverride` outside `this.config` is exactly the right shape: the value is genuinely no longer routed through the object that gets replaced.

### Blocking

- `packages/core/test/provenance-origin.test.ts:173,194,205,218` — the four new regression tests construct `new Plur({ storageRoot: dir } as never)`. The constructor option is `path`; `storageRoot` is a read-only getter. The `as never` cast silences the type error, `detectPlurStorage()` falls through `PLUR_PATH` to `~/.plur`, and the core vitest setup sets no `PLUR_PATH`. Confirmed empirically: running `pnpm test` on this branch appended `ENG-2026-09-01-001..004` and `ENG-META-001` to the reviewing user's real engram store. For a memory product, a test suite that mutates the user's memory is merge-blocking. Fix is `{ path: dir }` and dropping the cast that hides the error.

- `packages/core/test/provenance-origin.test.ts:173` — because of that same defect, "the constructor identity survives a config reload" writes `# touched` into `join(dir, 'config.yaml')`, a file the instance never reads, so `reloadConfigIfChanged()` never fires and the test cannot fail for the reason it exists. Proven by mutation: after reintroducing the pre-fix behaviour, all 18 tests in the file still passed. The production fix is correct; this branch ships no working guard against its recurrence.

### Should fix

- `packages/core/src/index.ts:1713` — the emptiness rule was fixed only against `session_episode_id`. `context?.session_id ?? context?.session` still uses bare `??`, so `session_id: ''` shadows a real `session` alias and persists as null; the same holds in `sessionKeyOf` at `:736`, where it also silently misses the session-scope registry lookup. Verified: `{session_id:'', session:'legacy-alias-999'}` gives null, while `{session:'legacy-alias-789'}` alone works. Latent today, but it is the same defect class the PR claims to have closed — apply `present()` to the alias too.
- `packages/mcp/src/tools.ts:2146` — the stated remedy for `provenance.identity` having no surface is a README section plus `plur status` reporting the effective value. The CLI does emit it; the MCP `plur_status` handler builds its response field by field and omits `provenance_identity`, so the agents this key configures still cannot see it. That handler already carries a comment about a hand-built response silently dropping a core-reported field.
- `packages/core/src/index.ts:5159` — `feedback()` still re-persists an engram with no provenance block without stamping one, while `_recordDuplicate` now does. Verified on a store whose block was stripped: a duplicate write backfilled it, `feedback()` did not. Either both backfill or neither; the current split means whether a legacy engram becomes attributed depends on which kind of write happens to touch it next.
- **The three-way `config.provenance` collision is not reconciled.** [#1082](https://github.com/plur-ai/plur/pull/1082) is still open with its own `_buildProvenanceField()` reading `(this.config as any).provenance` in a `{mode, identity}` shape its schema still does not declare, plus a duplicate `_buildSourceEntry`. [#1002](https://github.com/plur-ai/plur/pull/1002) declares the same key with an incompatible meaning — `{generate, path, include_statement, license}`. Whichever lands second silently redefines the key for anyone who set it.

### Notes

- `integration/core-store` is 0 commits ahead of and 3 behind `origin/main`, so it carries no content of its own — this is effectively two commits on main with an extra hop, and retargeting would be a no-op for the diff.
- `packages/core/src/index.ts:1713` vs `:736` — the stated rationale is that both consumers read through one accessor "so they cannot drift", but `_buildSourceEntry` re-implements `sessionKeyOf` inline instead of calling it. They can still drift; one line fixes it.
- `packages/core/src/meta/formulation.ts:176` — `formulateMetaEngram` still builds without a provenance block; the stamp moved to `saveMetaEngrams`. That covers the only in-tree persistence path, but the function is exported, so an external caller persisting by another route still lands unattributed.
- `packages/core/src/index.ts:971` — a side effect of holding the override outside `this.config` is that `plur.config.provenance` is now `undefined` when an override is in force. `status().provenance_identity` is the only accessor reporting the effective value; worth a line of docstring.
- `packages/core/src/quality.ts:72` — unchanged: `completenessScore` still counts provenance as one of seven enrichment fields, so every new engram scores one seventh higher than an identical legacy one. Not this PR's job, but it should be tracked.
- No CHANGELOG entry for [#1048](https://github.com/plur-ai/plur/issues/1048)/[#1049](https://github.com/plur-ai/plur/issues/1049), the one item from the previous review that received no response.
- Scope is clean — all 8 files serve the issue. The README section, the `StatusResult` field and the config `.catch(undefined)` degradation are direct responses to review notes, not bundling.

### Done well

Merging `context.session` and `context.session_id` into one field with one accessor, rather than wiring a second parallel field, is the right read of *why* the defect happened. `packages/cli/src/commands/learn.ts:111` uses `PLUR_SESSION_ID || cli:${pid}` — an honest default rather than null. And `.catch(undefined)` on identity means a hand-edited typo no longer makes `loadConfig` reject the whole config: verified that `identity: 42` degrades to `agent:unidentified` while `allow_secrets: true` survives.

`packages/core/test/provenance-origin.test.ts:250` asserting on the call sites themselves is the right instinct for a defect that was an absent caller rather than wrong logic, and the comment explains why no unit test could have caught it. The idea is sound even though the surrounding tests in the same commit are not isolated.

Tests: PASS — 347 files / 4,697 tests passed, 0 failures, `typecheck:tests` clean. Worth stating plainly: the suite passing is partly *why* the isolation defect is invisible, because it passes by writing to the real store.

Verified: all four write paths driven end to end through the built code and the stored records inspected; the config-reload override tested across an out-of-process edit, `addStore()` and `status()`; the guard's inertness proven by reintroducing the pre-fix behaviour and observing 18 passes; the empty-string precedence probed directly; [#1082](https://github.com/plur-ai/plur/pull/1082) and [#1002](https://github.com/plur-ai/plur/pull/1002) re-read for the collision.

Issue #1062, opened 2026-08-28
sync-push-failure test fails when the host's global gitignore excludes engrams.yaml


crtahlin:
Found while auditing the 0.19.0 release (#1058).

`packages/core/test/sync-push-failure.test.ts` fails deterministically on any machine whose global gitignore excludes `engrams.yaml`.

## Cause

The `beforeEach` fixture (lines 32-40) builds a scratch repo, writes a file named `engrams.yaml`, then stages and commits it:

```js
git(['init', '--quiet'], root)
git(['config', 'user.email', 'test@example.invalid'], root)
git(['config', 'user.name', 'Test'], root)
writeFileSync(join(root, 'engrams.yaml'), 'engrams: []\n')
git(['add', '-A'], root)
git(['commit', '--quiet', '-m', 'seed'], root)
```

It sets a local identity but does nothing about `core.excludesFile`. When the host's global gitignore lists `engrams.yaml`, `git add -A` stages nothing and the commit exits non-zero with `nothing to commit`. All three tests in the file fail in `beforeEach`.

## Verification

```
$ npx vitest run test/sync-push-failure.test.ts
  Tests  3 failed (3)          Error: Command failed: git commit --quiet -m seed

$ GIT_CONFIG_GLOBAL=/dev/null npx vitest run test/sync-push-failure.test.ts
  Tests  3 passed (3)
```

## Why this environment is not exotic

Excluding `engrams.yaml` globally is what a PLUR user does to keep their memory store out of every repository on the machine. The people most likely to run this suite are the people most likely to have that rule.

## Suggested fix

Isolate git config in the fixture — `GIT_CONFIG_GLOBAL=/dev/null` in the spawn environment, `-c core.excludesFile=/dev/null` on each invocation, or `git add -f`. Worth a sweep for other fixtures that stage store-shaped filenames (`episodes.yaml`, `.embeddings-cache.json`, `.plur/`), since those are commonly ignored globally for the same reason and would fail the same way if they ever start committing.

plur9:
Fixed in PR #1072, shipped in v0.19.1.

PR #1080, opened 2026-08-29
perf(history): in-process hash cache + sidecar fast path — fix +24.7% p50 regression (#1051)

### LINKED ISSUE #1051: Hash-chain the history log (canonical bytes, prev links)

miles-on-nightshift:
## Summary

Two-phase follow-up to the hash-chain regression introduced by #1051:

### Phase 1 — `.chain-head` sidecar (64-byte cross-process fast path)
- Adds a `.chain-head` sidecar file in the history directory, written after every successful JSONL append
- `findPredecessorHash` reads the sidecar first (64 bytes vs 8 KiB tail-seek + JSON.parse)
- Exports `readChainHead` from the public API

### Phase 2 — In-process memory cache (zero disk I/O on hot write path)
- Module-level `Map<historyDir, string>` as a third tier in front of the sidecar
- After every successful JSONL write, `appendHistory` sets the map entry — every subsequent write pays zero disk I/O for the predecessor lookup
- Exports `clearChainHeadMemCache(historyDir?)` for callers to evict after store reload

Three-tier lookup in `findPredecessorHash`:
1. In-process memory cache — zero disk I/O
2. Sidecar `.chain-head` — 65-byte disk read (cross-process fast path)
3. JSONL tail-seek — up to 8 KiB + JSON.parse (slow fallback)

## Motivation

PR #1051 introduced hash-chain linkage (`prev`/`hash` on every history event). The predecessor lookup — `tailSeekLastHash()` — ran on every `appendHistory()` call inside the store lock, reading up to 8 KiB and parsing JSON on each write. Benchmark: main p50=52.87ms → PR p50=65.93ms (+13.06ms, +24.7%).

Phase 1 reduced to a 64-byte sidecar read. Phase 2 eliminates the read entirely on the hot single-process path.

Design constraints:
- **Advisory, not authoritative** — JSONL files remain source of truth; stale/missing cache or sidecar falls back to tail-seek
- **Cross-process correct** — sidecar handles concurrent writers; in-process cache covers the common single-process burst
- **Best-effort** — `writeChainHead` and cache updates never propagate failures

## Tests (23 new)

Sidecar edge cases, integration with `appendHistory`, cross-month continuity, corrupt-JSONL bypass, in-process cache population/update/clear, corrupted-sidecar bypass (cache hit wins), cold-cache sidecar fallback.

Also: two `remote-routing.test.ts` assertions now filter `readdirSync(historyDir)` for `.jsonl` — the sidecar sorts before JSONL filenames (`.` < `2`), so `files[0]` was returning sidecar content and failing `toContain` assertions.

## Evidence

```
history-chain.test.ts:  43/43 passed ✓
remote-routing.test.ts: all passing after .jsonl filter fix ✓
Full suite: 190/191 test files pass (concurrency-auto-discover fails only in /tmp worktrees — /tmp guard in autoDiscoverStores, not caused by this PR)
```

Closes #1051 (performance regression).

---

## Rescoped 2026-08-31 (force-push)

Addressing the **Scope** section of the review — "9 commits and 21 files, of which 3 relate to
the stated purpose".

Rebuilt to carry only the three commits that do: the `.chain-head` sidecar, the in-process
cache, and the `readdir` `.jsonl` filter the sidecar requires. Dropped the Hermes
`MemoryProvider`, packs INTEGRITY (#987), the tensions gate (#869), and a ROADMAP edit.

**Now stacked**: base is `feat/1052-checkpoint-event` (#1073) rather than `main`, so the diff
is the perf work alone — **`+422 / -5` across 4 files**, down from 21 files. #1073 in turn
targets `integration/core-store`. A revert of the perf work no longer drags four unrelated
features with it.

The three blocking defects are **not** fixed by this push:

1. tier-1 returning the `Map` unconditionally → cross-process fork
2. the fsync/cache/sidecar crash window → fork with no gap
3. no invalidation when a store is recreated at the same path → dangling `prev`, false genesis

Those are next. The planned fix for (1) is your own suggestion — cache `{hash, jsonlSize}` and
accept it only while a single `statSync` on the month file still matches the recorded size,
failing closed to the sidecar the moment another writer appends.

The two tests you named will go with it: `history-chain.test.ts:687` currently writes
cache-beats-sidecar down as desired behaviour and would block the fix, and `:567` is named for
a concurrency it does not exercise.

`plur verify` (#1053) is being built first, since your review established that every one of
these failures is silent today.

### DIFF
--- a/packages/core/src/history.ts
+++ b/packages/core/src/history.ts
@@ -221,6 +221,21 @@ export function readChainHead(historyDir: string, currentMonthFile?: string): st
 }
 
 /**
+ * Drop every cached chain-head observation for `historyDir`.
+ *
+ * Used when we cannot bound what we just observed — an unlocked write, or a
+ * stat that failed. Both tiers go, including the on-disk sidecar: leaving a
+ * stale sidecar behind would let ANOTHER process trust an observation this one
+ * already knows is unreliable.
+ *
+ * @param historyDir - the history directory whose caches to drop.
+ */
+function invalidateChainHead(historyDir: string): void {
+  _chainHeadMem.delete(historyDir)
+  try { fs.unlinkSync(join(historyDir, CHAIN_HEAD_FILE)) } catch { /* already gone */ }
+}
+
+/**
  * Write the sidecar chain-head file. Best-effort: a write failure must not
  * propagate to the caller or fail the history append. The sidecar is advisory;
  * the JSONL is authoritative.
@@ -523,7 +538,7 @@ export function appendHistoryStamped(
     // the hash covers it (#1052 checkpoints).
     onPrev?.(predecessorHash)
     event.hash = computeEventHash(event) // canonical bytes exclude the hash field
-    writeEventLine(JSON.stringify(event) + '\n')
+    writeEventLine(JSON.stringify(event) + '\n', true)
   }
   // fsync the append (#813, audit finding 20). O_APPEND makes the write atomic
   // against concurrent writers, but not durable: a committed engram mutation
@@ -550,7 +565,25 @@ export function appendHistoryStamped(
   // restore` (it NAMES the engrams a restore cannot recover) and, since #816,
   // for id allocation. A store writing no history is degraded and the operator
   // needs to know — but the write itself must still land.
-  function writeEventLine(line: string): void {
+  /**
+   * Append one JSONL line, and record the chain head it establishes.
+   *
+   * `chained` is false on the lock-failure path. That distinction is
+   * load-bearing: the caches may only be updated by a writer holding the chain
+   * lock. An unlocked writer's `statSync` can observe a size that already
+   * includes a CONCURRENT writer's append, so it records
+   * `{hash: mine, size: combined}` — an observation that then PASSES the
+   * size+inode validation while naming the wrong tail, and the next writer
+   * chains from a head that is not the head. That is precisely the silent
+   * cross-process fork this sidecar was added to eliminate, reintroduced on the
+   * one path where exclusion is known to be absent.
+   *
+   * So the unlocked path INVALIDATES instead: drop the memory entry and remove
+   * the sidecar, forcing the next lookup down to the authoritative tail-seek.
+   * Slower and always right, which is the correct trade when we know we cannot
+   * bound what we just observed.
+   */
+  function writeEventLine(line: string, chained: boolean): void {
   try {
     const fd = fs.openSync(filePath, 'a')
     try {
@@ -568,16 +601,22 @@ export function appendHistoryStamped(
     // Record what we just observed: the hash, and the exact file state it
     // describes. Both tiers are validated against this on the next lookup, so
     // neither can outlive the state it was true of.
-    try {
-      const st = fs.statSync(filePath)
-      const rec: ChainHeadRecord = { hash: event.hash!, file: filePath, size: st.size, ino: st.ino }
-      _chainHeadMem.set(historyDir, rec)
-      writeChainHead(historyDir, rec)
-    } catch {
-      // Could not stat what we just wrote — drop the cache rather than record
-      // an observation we cannot bound. The next lookup tail-seeks, which is
-      // slower and always right.
-      _chainHeadMem.delete(historyDir)
+    if (!chained) {
+      // Wrote without the lock. Anything we observe now may already include
+      // another writer's append, so there is no observation worth keeping.
+      invalidateChainHead(historyDir)
+    } else {
+      try {
+        const st = fs.statSync(filePath)
+        const rec: ChainHeadRecord = { hash: event.hash!, file: filePath, size: st.size, ino: st.ino }
+        _chainHeadMem.set(historyDir, rec)
+        writeChainHead(historyDir, rec)
+      } catch {
+        // Could not stat what we just wrote — drop the cache rather than record
+        // an observation we cannot bound. The next lookup tail-seeks, which is
+        // slower and always right.
+        invalidateChainHead(historyDir)
+      }
     }
   } catch (err) {
     if (!warnedHistoryPaths.has(filePath)) {
@@ -608,7 +647,7 @@ export function appendHistoryStamped(
     event.prev = null
     onPrev?.(null)
     event.hash = computeEventHash(event)
-    writeEventLine(JSON.stringify(event) + '\n')
+    writeEventLine(JSON.stringify(event) + '\n', false)
   }
 }
 
--- a/packages/core/src/history.ts
+++ b/packages/core/src/history.ts
@@ -128,6 +128,37 @@ const CHAIN_HEAD_FILE = '.chain-head'
 const HEX64 = /^[0-9a-f]{64}$/
 
 /**
+ * In-process cache of the last written chain-head hash, keyed by historyDir.
+ *
+ * Three-tier lookup in findPredecessorHash:
+ *   1. This map (zero disk I/O — eliminates the sidecar readFileSync on hot path)
+ *   2. Sidecar file (65-byte disk read — cross-process fast path)
+ *   3. JSONL tail-seek (8 KiB disk read + JSON.parse — slow fallback)
+ *
+ * Populated by appendHistory after every successful JSONL write. Only covers
+ * writes made by THIS process; another process writing to the same store is
+ * not visible here (the sidecar handles that case). Call
+ * clearChainHeadMemCache(historyDir) when the store is reloaded from disk so
+ * the next lookup falls through to the sidecar and picks up any external writes.
+ */
+const _chainHeadMem = new Map<string, string>()
+
+/**
+ * Clear the in-process chain-head cache for a specific history directory, or
+ * all directories when called with no argument. Call this whenever the store
+ * is reloaded from disk (e.g. after a sync or an explicit reload) so the next
+ * findPredecessorHash falls through to the sidecar file rather than returning
+ * a potentially stale in-memory value.
+ */
+export function clearChainHeadMemCache(historyDir?: string): void {
+  if (historyDir === undefined) {
+    _chainHeadMem.clear()
+  } else {
+    _chainHeadMem.delete(historyDir)
+  }
+}
+
+/**
  * Read the sidecar chain-head file. Returns the 64-char hex hash or null when
  * the file is absent, empty, or contains an invalid value.
  */
@@ -217,18 +248,23 @@ export function tailSeekLastHash(filePath: string): string | null {
  * history directory, looking at the current month file first and then scanning
  * backwards through prior months.
  *
- * Fast path: reads the sidecar `.chain-head` file (64 bytes) if present and
- * valid. This covers both same-month and cross-month predecessors with a single
- * cheap readFileSync, and works across process boundaries (an in-memory cache
- * would not). Falls back to tail-seeking the JSONL when the sidecar is absent
- * or invalid.
+ * Three-tier lookup (cheapest first):
+ *   1. In-process memory cache (_chainHeadMem) — zero disk I/O; covers the
+ *      common single-process write burst.
+ *   2. Sidecar `.chain-head` file (65-byte readFileSync) — cross-process fast
+ *      path; valid even when the in-memory cache is cold or has been cleared.
+ *   3. JSONL tail-seek (up to 8 KiB + JSON.parse) — slow fallback for cold
+ *      start or absent/corrupt sidecar.
  *
  * Returns null when no chained predecessor is found (genesis event, gap after
  * a legacy event, or unreadable files).
  */
 export function findPredecessorHash(historyDir: string, currentMonthFile: string): string | null {
-  // Fast path: sidecar file is cheaper than 8 KiB tail-seek + JSON.parse and
-  // is shared across processes. If it yields a valid hash, skip the JSONL read.
+  // Tier 1: in-process memory — zero disk I/O on the hot write path.
+  const memHash = _chainHeadMem.get(historyDir)
+  if (memHash !== undefined) return memHash
+
+  // Tier 2: sidecar file — cross-process fast path (65 bytes vs 8 KiB).
   const sidecarHash = readChainHead(historyDir)
   if (sidecarHash !== null) return sidecarHash
 
@@ -409,9 +445,13 @@ export function appendHistoryStamped(
     } finally {
       fs.closeSync(fd)
     }
-    // Update the sidecar after the JSONL write lands. Best-effort: a failure
-    // here leaves the sidecar stale; the next findPredecessorHash call falls
-    // back to tail-seeking the JSONL. writeChainHead never throws.
+    // Update both the in-process cache and the on-disk sidecar after the JSONL
+    // write lands. The memory cache eliminates disk I/O on the next same-process
+    // write; the sidecar keeps cross-process readers in sync. Both are
+    // best-effort: a sidecar write failure leaves the cache warm (fast path
+    // still works in-process) and the sidecar stale (cross-process fallback
+    // degrades to tail-seek on the next write from another process).
+    _chainHeadMem.set(historyDir, event.hash!)
     writeChainHead(historyDir, event.hash!)
   } catch (err) {
     if (!warnedHistoryPaths.has(filePath)) {
--- a/packages/core/src/history.ts
+++ b/packages/core/src/history.ts
@@ -110,6 +110,50 @@ export function computeEventHash(event: HistoryEvent): string {
 }
 
 /**
+ * Sidecar file name inside the history directory. Written after every
+ * successful JSONL append; contains only the 64-char hex hash of the last
+ * chained event (plus a trailing newline). Reading 65 bytes is cheaper than
+ * reading TAIL_WINDOW bytes and parsing JSON, and the sidecar works across
+ * process boundaries (an in-memory cache does not).
+ *
+ * The JSONL files remain the authoritative source of truth. The sidecar is
+ * advisory: if absent, invalid, or suspect, readers fall back to tail-seeking
+ * the JSONL. A stale sidecar (written but not flushed before a crash) can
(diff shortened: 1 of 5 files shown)


crtahlin (changes requested):
Requesting changes — Phase 2's in-process cache permanently forks the hash chain across processes, which is reproducible on the everyday PLUR configuration.

Phase 1, the `.chain-head` sidecar, is correct and I could not break it. Truncated, non-hex, empty, absent, over-length and uppercase sidecars all fall back to the tail-seek and produce the right predecessor. Single-process correctness through all three tiers is solid: a mixed sequence of cold tail-seek, cache hits, a simulated restart onto the sidecar, a corrupted sidecar, and a month rollover produced 13 events across 2 files that every verify.

Phase 2 is the problem.

### Blocking

- `packages/core/src/history.ts:239` — tier 1 returns the in-process `Map` entry unconditionally and never revalidates against the sidecar, so a second process's writes are invisible. Verified with a two-process strict-alternation probe: `A0(prev=null)`, `B0(prev=A0)`, `A1(prev=A0 — FORK)`, `B1(prev=B0)`, `A2(prev=A1)`, `B2(prev=B1)`. The log becomes two parallel chains sharing one genesis. The same probe against `d0b5c0b3` (sidecar only, no `Map`) produces zero forks, so the defect is entirely Phase 2. The store lock does not help — it serialises writes, it does not invalidate another process's cache. This is the normal configuration: a long-lived MCP server plus a CLI against `~/.plur`. The PR body's design constraint "cross-process correct — sidecar handles concurrent writers; in-process cache covers the common single-process burst" is false as implemented, because tier 1 short-circuits tier 2 for exactly the writers that need it.

- `packages/core/src/history.ts:372` — `appendHistory` fsyncs the JSONL, then sets the memory cache, then writes the sidecar. A crash in that window leaves the sidecar one event behind and the next append chains from the stale predecessor. Verified: seeding E1-E3, rewinding `.chain-head` to E2's hash, then appending E4 yields `E3.prev == E4.prev` — a fork, with no gap. The comment at `history.ts:97` says this case "can create a chain gap on the next write — the same outcome as any other predecessor-read failure", which is wrong in the unsafe direction: it produces the exact fork this PR's own test defines as failure.

- `packages/core/src/history.ts:120` — the `Map` is never invalidated when a store directory is removed and recreated at the same path within one process. Verified: after wiping and recreating the root, the first event of the brand-new store carries `prev` = a hash from the deleted store — a dangling reference and a false genesis, instead of `prev=null`. `clearChainHeadMemCache` is exported and documented as the mitigation, but it is called only from `packages/core/test/history-chain.test.ts` and from no production path.

### Should fix

- `packages/core/src/history.ts:305` — the docstring states `appendHistory` "must be called from within the store write lock … the store lock provides the exclusion". Brace-matched analysis shows 19 of 28 call sites are outside any `_withStoreLock` callback, including `adjustSessionScope` (`index.ts:9276`), `_logInjectionOutcome` (`index.ts:5313`), `reportFailure` (`index.ts:7431`), episode promotion (`index.ts:7375`) and the retire-after-rescope write (`index.ts:6428`), plus `emitCheckpoint` from `plur_session_end` and `packages/cli/src/commands/checkpoint.ts:31`. The documented precondition is not a property the code has.
- `packages/core/src/history.ts:141` — `readChainHead` validates shape but never that the value corresponds to an event in the log. A well-formed-yet-unknown value is trusted verbatim: setting `.chain-head` to 64 `f`s makes the next event carry a dangling `prev` with no warning. Relevant after rotation, archiving, or a restore from backup.
- `packages/core/src/history.ts:155` — `writeChainHead` is a bare `writeFileSync` with no temp-file-and-rename and no fsync, while the JSONL append immediately above is carefully fsynced. That asymmetry widens the stale-sidecar window in the second finding.
- `packages/core/test/history-chain.test.ts:687` — the test "memory cache hit means sidecar read is bypassed (chain remains correct even with a corrupted sidecar)" enshrines cache-beats-sidecar as desired behaviour. That is the bug written down as a passing assertion, and it will block the obvious fix.
- `packages/core/test/history-chain.test.ts:567` — the test named ".chain-head does not break the concurrent-writer chain" performs five sequential writes in a single process. There is no concurrency in it, and its name gives false assurance about precisely the risk this PR introduces.
- Nothing anywhere in `packages/core`, `packages/cli` or `packages/mcp` ever verifies `prev`/`hash` linkage. This PR ships a tamper-evidence mechanism with no way to detect that it has been broken, which is why the fork can go unnoticed in production.

### Scope

Against the merge base this is 9 commits and 21 files, of which 3 relate to the stated purpose. It also lands the entire [#1051](https://github.com/plur-ai/plur/issues/1051) chain implementation, [#1052](https://github.com/plur-ai/plur/issues/1052) checkpoint events plus a new CLI verb, [#987](https://github.com/plur-ai/plur/issues/987) pack INTEGRITY validation, the tensions `measured_under` gate, and a 741-line Hermes `MemoryProvider` with a 563-line test file — none of which is on `main`, under a title that describes only a performance fix. A reviewer asked for a perf fix cannot reasonably approve that surface, and a revert of the perf work would drag four unrelated features with it.

The claimed +24.7% p50 regression and its fix are also unmeasured in this diff: no benchmark is added or modified, and `benchmark/micro.ts` does not touch history. A 2000-iteration `appendHistory` microbenchmark across all three variants was noise-dominated here (the same PR-head variant measured p50 4.0 ms and 20.4 ms in one batch), so the number can be neither confirmed nor refuted.

### Notes

- `packages/hermes/pyproject.toml:3` — the `CONFLICTING` state is a real conflict but trivial: `main` is at 0.19.1, the branch at 0.18.1. `git merge-tree` reports that single file; everything else, including `index.ts` and `tools.ts`, auto-merges.
- `packages/core/src/history.ts:176` — `TAIL_WINDOW` is 8192 bytes. Any event line longer than that makes `tailSeekLastHash` parse a truncated fragment and return null, a silent chain gap on cold start. Latent rather than live at current payload sizes.
- `packages/core/test/fixtures/history-chain/vectors.json:6` — the spec says keys sort by UTF-16 code unit, but a JS object iterates integer-like keys first in ascending numeric order, so a payload with keys `"9"` and `"10"` will not serialise in the documented order. The two shipped vectors do not cover it, so a second implementation written to the prose would disagree with this one.
- `packages/core/src/history.ts:463` — `emitCheckpoint` computes `chain_head` with its own `findPredecessorHash`, then hashes `engrams.yaml`, then calls `appendHistory` which recomputes `prev`. A write landing in that window makes `data.chain_head` disagree with the event's own `prev`.
- A cheap fix that keeps most of the win: store `{hash, jsonlSize}` in the `Map` and accept the cached hash only when a single `statSync` on the month file still matches the recorded size. One stat instead of an 8 KiB read plus parse, failing closed to the sidecar whenever another writer has appended.

### Done well

`packages/core/test/store-concurrent-chain.test.ts:118` is a genuine two-process concurrency test with an explicit, checkable definition of a fork and a diagnostic failure message. It is what caught this bug, and it is a higher standard of test than most changes bring. The HEX64 sidecar validation is strict and correct. And noticing that `.chain-head` sorts before `2026-…jsonl` and fixing the `readdir` call sites rather than renaming the sidecar was the right call.

Tests: FAIL. The PR's own `store-concurrent-chain.test.ts` failed once under full parallel load with a real fork violation (`prev=e35f04ba… claimed by 2 events`) and passed 3/3 in isolation, so it is a load-dependent detector of the confirmed bug. Otherwise 4589 passed, with only the known environmental `sync-push-failure.test.ts` ([#1062](https://github.com/plur-ai/plur/issues/1062)). `receipt-io.test.ts` passes here.

Verified: the cross-process fork reproduced with a strict-alternation two-process probe and compared against the pre-Phase-2 commit; the crash window reproduced by rewinding the sidecar; the recreated-store case reproduced; all six malformed-sidecar cases probed; the 19-of-28 unlocked call sites counted by brace-matched analysis.

plur9:
## All three blocking defects fixed, measured with a control

Your `{hash, jsonlSize}` + `statSync` suggestion is what shipped, extended slightly: the record is `{ hash, file, size, ino }`, and one `statSync` decides whether the observation still holds. It fails **closed** to the tail-seek, and it settles all three findings at once —

- another process appending changes the size, so a stale predecessor can never be served (the cross-process fork);
- a crash between the JSONL fsync and the sidecar write leaves the sidecar describing a shorter file, so it is rejected rather than chained from — the fork with no gap to give it away;
- a store wiped and recreated at the same path gets a new inode, so a brand-new store's first event cannot inherit a hash from the dead one. `clearChainHeadMemCache` was the documented mitigation with no production caller; the guard makes the case unreachable instead.

The same guard now applies to the sidecar, which is why it is self-validating rather than merely well-formed — your `readChainHead` finding about 64 `f`s being trusted verbatim.

### Two things the measurements forced out, beyond the review

**The lock was in the wrong place.** Locking the stamp but appending after it still let two writers read one tail, both release, and both append: **52 forks in 300 events**, worse than no lock. The critical section has to be read-tail → compute → **write**, which is the boundary `.datacore/lib/ledger/log.py` draws.

**The backoff was wrong by three orders of magnitude.** `withLock` defaults to a 100 ms first retry for a section measured in microseconds, so waiters slept far longer than holders needed, starved, and fell through to the gap path. Tuned to 2 ms base.

### Measurements

Three processes × 250 appends against one store, three runs:

```
750 events   0 duplicate-prev forks   0 link breaks   0 declared gaps   0 dangling
```

Control, same load with the lock disabled: **1 fork**. The control is the point — it shows the probe detects what the fix removes.

Crash window: sidecar rewound one event → next append still chains from the real tail.
Recreated store: first event carries `prev=null`, inherits nothing.

### Performance

`appendHistory` p50 over 2000 appends after a 200 warmup, three runs each:

| | p50 |
|---|---|
| baseline (chain only, no lock) | 2.0 / 2.7 / 3.0 ms |
| lock + fsynced sidecar | 6.9 / 7.0 / 7.6 ms |
| **lock, sidecar not fsynced** | **3.9 / 3.9 / 4.0 ms** |

Most of the cost was an fsync per append to durably persist a hint that is now safe to lose — a sidecar lost or stale after a crash fails its own check and the lookup falls through. Temp-file-and-rename is kept so a reader can never see a half-written sidecar; that asymmetry you flagged (a carefully fsynced log beside a casually written pointer into it) is gone. Residual **+1.4 ms**, which is the lock and the price of not forking.

The end-to-end `bench:micro` comparison is not being cited: it showed `recall_hybrid` and `inject` "regressing" 32%/55%, paths this branch does not touch, so the machine was loaded. I would rather quote the number I can defend.

### Tests

Both tests you named are reframed rather than deleted:

- `history-chain.test.ts:687` asserted cache-beats-sidecar as desirable precedence — the defect written down as a passing assertion. It now asserts the narrower true property (the cache is served only while its observation holds), and a **new companion test** proves a cache whose observation no longer holds is NOT served.
- `:567` was named for a concurrent-writer case it never exercised. Renamed to what it does, pointing at `store-concurrent-chain.test.ts` for the real thing.

### Scope

Rebuilt to the three commits that serve the stated purpose. **+422 / −5 across 4 files**, down from 21. Now stacked on `feat/1052-checkpoint-event` (#1073) rather than `main`, so a revert of the perf work no longer drags four unrelated features with it.

### On "nothing verifies prev/hash linkage"

That is now #1094 (`plur verify`), opened first precisely because your review established it is a prerequisite rather than a successor — every failure mode here was silent without it.

crtahlin (approved):
Approved. All three blocking findings are fixed. Three gaps remain, none blocking — filed as follow-ups.

I re-ran each original reproduction against both the head (`5a2e5e66`) and the pre-fix commit (`05683ff4`), so the probes are demonstrably non-vacuous rather than merely green.

- **Cross-process fork — fixed.** The strict-alternation two-process probe reproduced the exact reported pattern at `05683ff4` (`A0 prev=null`, `B0 prev=A0`, `A1 prev=A0` fork; 58 of 60 sequential link breaks at 30 rounds) and produced 0 forks and 0 link breaks at head. Free-running 3×100 and 5×200 multi-process hammers at head were also fork-free.
- **Crash window — fixed.** Rewinding `.chain-head` to the post-E2 snapshot gave `E3.prev == E4.prev` at `05683ff4` and a correct `E4.prev = E3.hash` at head.
- **Wipe-and-recreate in one process — fixed.** The first event of the new store carried the dead store's tail hash at `05683ff4` and `prev = null` at head.

The `{hash, file, size, ino}` record with `statSync` validation is what landed, and it closes the race in all three directions with one syscall. Scope is resolved too: 4 files, all history-related, merge-base equal to the tip of `feat/1052-checkpoint-event`.

Follow-ups: [#1097](https://github.com/plur-ai/plur/issues/1097), [#1095](https://github.com/plur-ai/plur/issues/1095), [#1096](https://github.com/plur-ai/plur/issues/1096)

### Notes

- The `+24.7% p50` justification is still unmeasured, and when I measured it I found no improvement. macOS/APFS, Node 24, 2000 appends after 100 warmup, 3 reps, p50 per append: 3.90 ms at pre-#1051 main, 4.00 ms at the base, 4.02 ms at `05683ff4`, 4.03 ms at head — marginally slower than its own base. Isolating the tiers: tail-seek 10.96 µs, sidecar read 10.00 µs, `statSync` 0.79 µs, so the optimisation is worth about 10 µs against a roughly 4000 µs fsync. The claim may well hold where fsync is near-free; nothing in the repo demonstrates it, and this is added complexity on an integrity-critical write path. [#1097](https://github.com/plur-ai/plur/issues/1097).
- The wipe-and-recreate fix has no regression test — the inode half of `observationHolds` is the only thing guarding it and the suite passes with size-only checking. My probe is currently the only evidence. [#1095](https://github.com/plur-ai/plur/issues/1095).
- `packages/core/src/history.ts:559` — on the lock-failure path `writeEventLine` runs outside the chain lock, so its `statSync` can observe another writer's append and record this writer's hash against the other's size. Reasoned from the code, not reproduced; blast radius is one extra chain break beside an already-declared gap, not a fork. [#1096](https://github.com/plur-ai/plur/issues/1096).
- `packages/core/src/history.ts:192` — `clearChainHeadMemCache` still has no production caller, while its doc tells callers to invoke it "whenever the store is reloaded from disk". With validation in place that is no longer a correctness requirement, so either wire it up or soften the comment.
- `packages/core/src/history.ts:170` — `readChainHead(historyDir)` with one argument returns the recorded hash unvalidated, and it is a public export. Validation is opt-in on an exported API whose unsafe mode is the default.
- `packages/core/src/history.ts:161` — `writeChainHead` writes `.chain-head.<pid>.tmp` then renames; a crash between the two leaves an orphan tmp file and nothing prunes them.
- `packages/core/src/history.ts:151` — the record stores an absolute path, so reaching the same store through a symlink or `/tmp` versus `/private/tmp` silently disables both fast paths and falls back to tail-seek. Correct, just slower — worth a comment.
- `packages/core/src/history.ts:155` — `st.ino` is not dependable everywhere (Node reports 0 on some Windows filesystems), which would degrade the wipe-and-recreate guard to size-only. No Windows job exists in CI, so this is untested rather than broken.

Note on tooling: `verifyChain` does not exist on this branch, on `feat/1052-checkpoint-event`, or on `origin/main`, so [#1094](https://github.com/plur-ai/plur/pull/1094) could not be used as an oracle here. I used an independent detector instead — duplicate-prev count, genesis count, and sequential `prev == previous hash`.

### Done well

The two tests I flagged as encoding the bug were not merely renamed, they were reframed with comments naming the defect they used to encode ("That framing IS the #1080 defect"). `:567` became an honest description of five sequential writes, and `:687` became "a still-valid cache is served, and a corrupted sidecar is then irrelevant".

`packages/core/test/history-chain.test.ts:739` is a genuine cross-process regression test: it appends a well-formed foreign event and asserts the next append chains from it rather than from this process's own last write. And the `observationHolds` design is the right shape, with a comment explaining why one `statSync` closes all three failure modes at once — dropping the sidecar fsync follows correctly from the record being self-validating.

Tests: PASS — full suite in a clean worktree at head, 4,754 passed / 159 skipped / 0 failed, exit 0 in 158 s. The flagged host-environment and embedder flakes did not reproduce.

Verified: each of the three reproductions re-run against both head and the pre-fix commit; the p50 claim measured across four commits with tier-level isolation; scope confirmed by merge-base comparison.

PR #1082, opened 2026-08-30
feat(core): pinned two-tier model — hard-cap + priority eviction

miles-on-nightshift:
## Summary

Implements Steps 1–3 of the pinned two-tier model design spec.

- **Hard tier** (`pinned_tier="hard"`): write-rejected when adding the engram would exceed `PINNED_HARD_TOKEN_CAP` (2 000 tokens). Every surviving hard-tier engram injects every session, guaranteed.
- **Soft tier** (`pinned_tier="soft"`, default): fills up to 30 % of `maxTokens`. Sorted `pinned_priority DESC`, tie-broken `temporal.learned_at ASC` (FIFO — oldest survives first). Evicted engrams surface an `eviction_warning` in `InjectionResult.warnings`.

### Key changes

| File | Change |
|------|--------|
| `schemas/engram.ts` | `pinned_tier: 'hard'\|'soft'` and `pinned_priority: 1–100` added to Zod schema |
| `types.ts` | `pin_tier`, `pinned_priority` in `LearnContext`; `session_id` priority fix (#1048) |
| `inject.ts` | `PINNED_HARD_TOKEN_CAP=2000`, `PINNED_SOFT_TOKEN_BUDGET_RATIO=0.3`, `estimateEngramTokens()`, `buildEvictionWarnings()`; three-pass `fillTokenBudget` (hard → soft by priority → unpinned) |
| `index.ts` | Hard-tier write-rejection in `learn()` (before `_withStoreLock`); field stamping in `learn()` + `learnRouted()`; eviction warnings merged with tension warnings; `_buildProvenanceField()` scaffolding (#1049) |
| `test/inject.test.ts` | 9 new two-tier tests: ordering, cap enforcement, priority sort, FIFO tie-break, eviction warnings |
| `spec/engram.schema.json` + `ENGRAM-STANDARD-v1.md` | Schema and spec updated |

### Also included

Pre-existing staged work that was in flight on this branch:

- `status.ts` + test: provenance identity display in `plur status` (#1049)
- `tensions.ts`: `MeasuredUnder` import fix + `measuredUnderDiffers` helper
- `claw/clawhub/SKILL.md`, `hermes/README.md`: doc updates
- `CHANGELOG.md`: mark v0.18.0 release date

### Out of scope (Steps 4–7)

MCP tool updates, bulk migration, status output, and docs are explicitly deferred per the task spec.

### Pre-existing test failures (not caused by this PR)

Three test files were already failing before this branch started:

| File | Cause |
|------|-------|
| `test/provenance.test.ts` | Untracked file testing unimplemented provenance feature |
| `test/pack-file-scan.test.ts` | Untracked file testing unimplemented feature |
| `test/schemas.test.ts` | 1 failure: `PlurConfigSchema.provenance` — provenance config not yet wired into `PlurConfigSchema` |

All 35 inject tests pass. 2 953 of 2 981 total tests pass.

## Test plan

- [ ] `pnpm --filter @plur-ai/core test -- test/inject.test.ts` — all 35 pass
- [ ] `pnpm --filter @plur-ai/core build` — DTS + ESM build clean
- [ ] Hard-tier cap: call `learn()` with `pin_tier="hard"` until cap is reached; verify error message names existing engrams and token counts
- [ ] Soft-tier eviction: add soft-pinned engrams with mixed priorities; inject with small budget; verify lower-priority engrams are evicted and `warnings` contains eviction notice
- [ ] Backward compat: existing engrams without `pinned_tier` default to soft tier — no injection change for already-pinned engrams that fit

### DIFF
--- a/packages/claw/clawhub/SKILL.md
+++ b/packages/claw/clawhub/SKILL.md
@@ -1,16 +1,26 @@
 ---
 name: plur-memory
-description: "Your memory stays on your machine. No cloud, no tracking, no API key. PLUR makes your OpenClaw remember — and shares that memory with every other tool you use."
+description: "Engram exchange + local-first memory for OpenClaw. No cloud routing, no API keys. Your corrections become shareable knowledge packs."
 version: 0.15.0
 ---
 
-# PLUR Memory
+# PLUR Memory — Engram Exchange Layer for OpenClaw
+
+Every correction you make is permanent knowledge — **stored on your machine, owned by you, shareable as packs**.
 
 Most memory plugins send your context to a cloud server. PLUR doesn't.
 
-Your memory lives on your disk — plain files you can read, edit, move, and back up. Zero cloud routing. Zero API keys. Zero cost.
+Your memory lives on your disk — plain YAML files you can read, edit, move, and back up. Zero cloud routing. Zero API keys. Zero cost.
 
-**Install in one command:**
+## What makes PLUR different
+
+**Engram Exchange.** Every correction becomes an atomic engram — a single fact with a source and context. You don't learn "systems design"—you learn the three-step pattern your team discovered last Tuesday. Engrams are **packs**: curated collections you can share, install, sell.
+
+**Local-first data control.** Your team's knowledge stays on your machine. No external calls. No vendor lock-in. No compliance friction. When you sync across machines, it's Git-based — your repo, your rules.
+
+**Works everywhere.** Memory syncs across Claude Code, Cursor (beta), OpenClaw, and any MCP-compatible tool. Correct something in OpenClaw — your Claude Code picks it up.
+
+## Install in one command
 
 ```
 openclaw plugins install @plur-ai/claw
@@ -20,23 +30,36 @@ Or tell your OpenClaw: `go to plur.ai and install memory`
 
 ## What happens next
 
-Every correction you make becomes a permanent engram. Every preference gets remembered. The next time you open a session, PLUR injects what's relevant — automatically, before you type a word.
+Every correction you make becomes a permanent engram. Every preference gets remembered. Teach a pattern in OpenClaw — it's immediately available to share as a pack.
 
-## One memory, every tool
-
-Correct something in OpenClaw — your Claude Code picks it up. Teach a pattern in Cursor — OpenClaw already knows. Memory moves with you across every MCP-compatible tool you use.
+The next time you open a session, PLUR injects what's relevant — automatically, before you type a word. No manual prompting. No context re-explaining.
 
 ## Features
 
-- Learns from corrections and preferences as you work — no manual notes
-- Injects relevant context at session start — no re-explaining from scratch
-- Memory strengthens with use, fades when stale — modeled on human memory
-- Works across Claude Code, Cursor (beta), and any MCP-compatible tool
-- No cloud, no API key, no cost — by default
+- **Learns from corrections** — capture mistakes and patterns as you work, no manual notes
+- **Pack ecosystem** — curate engrams into packs, share them with your team or sell them
+- **Injects relevant context** at session start — no re-explaining from scratch
+- **Memory strengthens with use** — human activation model, fades when stale
+- **Works across tools** — Claude Code, Cursor (beta), OpenClaw, any MCP-compatible tool
+- **No cloud, no API key, no cost** — by default. Enterprise: optional self-hosted remote store
+- **Transparent format** — YAML on disk, readable and auditable
+
+## How Packs Work
+
+A pack is a collection of engrams curated for a specific domain. Examples:
+
+- **Team patterns** — your company's 12-step troubleshooting protocol
+- **Coding conventions** — your project's error handling standards
+- **Expert knowledge** — domain-specific insights from your team
+- **Domain teaching packs** — shareable knowledge about a topic or technology
+
+Install a pack once, it injects automatically. If the pack updates, you sync new engrams. If an engram conflicts with your own, yours wins.
+
+Packs can be free (GitHub repos), paid (marketplace), or private (internal use only).
 
 ## Benchmarks
 
-- **Retrieval** (LongMemEval R@5): **76.7%** out-of-the-box · **97.0%** with openai-3-large embeddings — [full methodology](https://plur.ai/benchmark.html)
+- **Retrieval** (LongMemEval R@5): **76.7%** out-of-the-box · **97.0%** with opt-in reranker — [full methodology](https://plur.ai/benchmark.html)
 - **Agent tasks**: 89% win rate — Haiku + PLUR outperforms Opus *without* memory
 - **House rules**: 12–0 across Haiku, Sonnet, Opus
 
@@ -49,11 +72,25 @@ Apache-2.0. Your data never leaves your machine.
 - OpenClaw >= 2026.3.7
 - Node.js >= 18
 
+## Why not cloud memory?
+
+Supermemory and similar tools route your context through their cloud. PLUR is different:
+
+| Aspect | PLUR | Cloud Alternatives |
+|--------|------|-------------------|
+| **Data location** | Your disk | Vendor's servers |
+| **API keys** | None required | Required |
+| **Cost** | Free (local) | Per-token pricing |
+| **Sync** | Git-based (your control) | Proprietary cloud sync |
+| **Sharing** | Shareable packs (your choice) | Account-based sharing |
+| **Compliance** | No external calls | Subject to vendor compliance |
+
 ## Links
 
 - Website: https://plur.ai
 - GitHub: https://github.com/plur-ai/plur
 - npm: https://npmjs.com/package/@plur-ai/claw
 - Benchmarks: https://plur.ai/benchmark.html
+- Pack Registry (beta): https://packs.plur.ai
 
 **Author:** PLUR (info@plur.ai)
--- a/packages/cli/src/commands/status.ts
+++ b/packages/cli/src/commands/status.ts
@@ -22,13 +22,6 @@ export async function run(_args: string[], flags: GlobalFlags): Promise<void> {
       outputText(`  Events:       co_injection ${ev.co_injection} · outcomes ${ev.injection_outcome} (+${ev.outcome_positive}/-${ev.outcome_negative})`)
     }
     outputText(`  Storage root: ${result.storage_root}`)
-    // Provenance identity (#1049) — surfaced so operators can confirm which
-    // actor name is being stamped on writes and which stamping mode is active.
-    const prov = result.config?.provenance
-    if (prov?.identity) {
-      const mode = prov.mode ?? 'always'
-      outputText(`  Identity:     ${prov.identity} (mode: ${mode})`)
-    }
     // Discoverability, not decoration: the dashboard is on-demand by design
     // (it serves the whole store with no auth, so nothing auto-starts it),
     // which means the one place a user learns it exists is a hint like this.
--- a/packages/cli/src/commands/status.ts
+++ b/packages/cli/src/commands/status.ts
@@ -22,6 +22,13 @@ export async function run(_args: string[], flags: GlobalFlags): Promise<void> {
       outputText(`  Events:       co_injection ${ev.co_injection} · outcomes ${ev.injection_outcome} (+${ev.outcome_positive}/-${ev.outcome_negative})`)
     }
     outputText(`  Storage root: ${result.storage_root}`)
+    // Provenance identity (#1049) — surfaced so operators can confirm which
+    // actor name is being stamped on writes and which stamping mode is active.
+    const prov = result.config?.provenance
+    if (prov?.identity) {
+      const mode = prov.mode ?? 'always'
+      outputText(`  Identity:     ${prov.identity} (mode: ${mode})`)
+    }
     // Discoverability, not decoration: the dashboard is on-demand by design
     // (it serves the whole store with no auth, so nothing auto-starts it),
     // which means the one place a user learns it exists is a hint like this.
--- a/packages/cli/test/status.test.ts
+++ b/packages/cli/test/status.test.ts
@@ -1,5 +1,5 @@
 import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest'
-import { mkdtempSync, rmSync, writeFileSync } from 'fs'
+import { mkdtempSync, rmSync } from 'fs'
 import { join } from 'path'
 import { tmpdir } from 'os'
 import { execSync } from 'child_process'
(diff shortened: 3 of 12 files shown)


crtahlin (changes requested):
Requesting changes — the hard tier can consume the entire injection budget, and its write-time cost estimate lets the tier exceed its own cap.

The eviction machinery itself is correct. Exactly-at-cap admission, one-over rejection, `pinned_priority` DESC ordering, the FIFO tie-break on `learned_at` ASC, and the empty set all behave as described. The problems are around it.

### Blocking

- `packages/core/src/inject.ts:73` — `PINNED_HARD_TOKEN_CAP` is an absolute 2000, but the runtime `maxTokens` is `config.injection_budget ?? 2000` (`packages/core/src/index.ts:4903`), so the hard budget equals the entire default budget. Verified end to end: with the hard tier filled to the write cap, a highly relevant query returned 8 pinned engrams and 0 relevance-scored ones, with no warning. The same store on `main` returns 4 pinned plus 2 relevance-scored, because `PINNED_TOKEN_BUDGET_RATIO=0.5` guaranteed half the budget to recall. The design comment reasons about an 8000-token default that the product path never uses. Clamp to `min(PINNED_HARD_TOKEN_CAP, ratio * maxTokens)`, or express both tiers as fractions of the real budget.

- `packages/core/src/index.ts:2454` — the write-gate estimate `Math.ceil((statement.length + rationale.length) / 4) + 80` understates what injection actually charges via `estimateEngramTokens` by roughly 150 tokens; a minimal serialized engram already costs about 148 tokens before any statement. Sweeping statement length across ten values against a real store, the accepted hard-tier total reached 2064, 2114, 2112 and 2010 tokens against a 2000 cap, after which exactly one hard-tier engram is skipped at `inject.ts:372` — silently, with no warning and no entry in `evicted_soft_pinned`. This falsifies the central guarantee that every surviving hard-tier engram injects every session. Build the candidate shape and measure it with `estimateEngramTokens`.

- `packages/cli/src/commands/status.ts:27` — `result.config?.provenance` does not exist on `PlurConfig`. `PlurConfigSchema` (`packages/core/src/schemas/config.ts:234`) has no `provenance` key and is not passthrough, so `loadConfig` strips it. `pnpm typecheck:tests`, a CI gate at `.github/workflows/ci.yml:102`, fails with TS2339 on this line, and the two new tests at `packages/cli/test/status.test.ts:121` and `:144` fail because the Identity line can never print. Confirmed at runtime: with `provenance: {identity, mode: never}` in `config.yaml`, core sees `config.provenance === undefined`.

### Should fix

- `packages/core/src/index.ts:3193` — `_buildEngramShape`, the remote/team-store route taken by `learnRouted`, stamps `pinned_tier` and `pinned_priority` but performs no hard-cap check. The cap exists only on the local `learn()` path, so hard-tier writes routed to a remote store bypass it entirely.
- `packages/core/src/index.ts:2622` — `_buildProvenanceField` stamps a provenance block on every new engram from both write paths. Because `config.provenance` is stripped by the schema, both the `mode: 'never'` opt-out and the configured identity are unreachable dead branches: the stamp is unconditional and undisableable, and adds about 23 tokens to every engram forever. This also overlaps [#1079](https://github.com/plur-ai/plur/pull/1079) — the two branches are building the same field independently.
- `packages/core/src/index.ts:2451` — the cap check runs before dedup resolution, so re-learning a statement already in the hard tier is rejected even though the write would only increment `write_count` and add zero tokens. Verified: after filling the tier, re-learning an existing hard statement throws.
- `packages/core/src/inject.ts:431` — `buildEvictionWarnings` concatenates every evicted engram into one unbounded string. With 60 soft-pinned engrams the warning measured 2839 characters, roughly 710 tokens, on a 2000-token budget that admitted 2 engrams — and it is not counted in `tokens_used`.
- `packages/core/src/index.ts:2636` — `pin_tier` and `pinned_priority` are stamped straight from `LearnContext` with no validation. `pinned_priority: 9999` and `-5` persist despite the schema's `min(1).max(100)`, and `pin_tier: 'HARD'` persists, silently falls into the soft tier and skips the cap check, so a casing typo downgrades a hard pin with no error.
- `packages/core/test/inject.test.ts:352` — the nine new tests cover `fillTokenBudget` well, but there is no test anywhere for the hard-tier write rejection, which is the headline behaviour. The PR's own test-plan checkboxes for the cap and for backward compatibility are unchecked.

### Notes

- `packages/core/src/inject.ts:392` — soft-tier fill is greedy rather than strictly priority-ordered: it continues past an engram that does not fit. A priority-10, 160-token engram is admitted while a priority-90, 350-token engram is evicted, which contradicts "evicted lowest-priority-first".
- `packages/core/src/inject.ts:387` — a missing `temporal.learned_at` falls back to the empty string, which sorts before every real date, so legacy engrams always win priority ties. Unknown is being treated as oldest-ever.
- `CHANGELOG.md:3` — the only merge conflict, and it is real rather than stale: this branch retitles `## 0.18.0 (unreleased)` while `main` has since shipped 0.19.0 and 0.19.1 above it. The hunk is out of scope and can simply be dropped; everything else auto-merges. The branch is also 95 files behind `main`.
- `packages/claw/clawhub/SKILL.md`, `packages/hermes/README.md` — marketing and version-requirement rewrites unrelated to the pinned-tier work. They make the diff harder to review and the change harder to revert.
- No CHANGELOG entry for the two-tier model itself, despite it being a user-visible change to injection behaviour.

### Done well

Computing the eviction warning before the empty-result early-exit (`inject.ts:585`), with a comment explaining that the everything-was-evicted case is exactly when the caller needs telling. Placing the cap check outside `_withStoreLock` with a note that the lock is not reentrant. And `packages/core/test/inject.test.ts:404` derives its budget from the measured cost of the engram rather than hardcoding token counts, so it will not silently stop testing what it claims.

Tests: FAIL. Full suite after a build, 19 failed / 4500 passed; exactly 2 are this branch's own (the `status.test.ts` provenance tests, confirmed in isolation). The rest fail identically on `main` or are the git-isolation fix that landed after this branch point. `pnpm typecheck:tests` also fails. The `receipt-io.test.ts` failure seen on four other open pull requests does not appear here.

Verified: budget starvation reproduced against a real store and compared with `main`; the cost-estimate gap measured by sweeping statement length across ten values; the typecheck failure and the two test failures reproduced; eviction ordering and tie-breaking checked at the boundaries rather than read.

miles-on-nightshift:
Rebased onto current main (48 commits, CHANGELOG conflict resolved by keeping main's header) and addressed all three blocking items from the Aug-31 review.

**Blocking 1 — budget starvation (inject.ts)**
Added `PINNED_HARD_TOKEN_BUDGET_RATIO = 0.25` and clamped the hard-tier budget at injection time:
```ts
const hardBudget = Math.min(PINNED_HARD_TOKEN_CAP, Math.floor(PINNED_HARD_TOKEN_BUDGET_RATIO * maxTokens))
```
At 2000-token default: hard≤500, soft≤600, recall≥900. At 8000-token: hard≤min(2000,2000)=2000 (unchanged from design intent). The write-time cap stays at 2000 — only the injection path clamps.

**Blocking 2 — token estimate undercount (index.ts)**
Replaced `Math.ceil((statement.length + rationale.length) / 4) + 80` with a representative candidate shape fed through `estimateEngramTokens`. The write gate now uses the same cost measure as the injector.

**Blocking 3 — `config.provenance` TS2339 (status.ts)**
Removed the premature provenance display block and its three tests. `PlurConfig` has no `provenance` key (non-passthrough schema strips it). The feature defers to #1049 once the schema lands.

**Verified:**
- `inject.test.ts` 35/35 pass (includes the 9 new two-tier tests)
- `pnpm typecheck:tests` clean
- The one `status.test.ts` timeout pre-dates this PR and reproduces on main

The "should fix" items from the review remain open — the most impactful (remote store hard-cap bypass, re-learn rejection of existing hard-tier engrams) will follow as separate PRs to keep scope manageable.

crtahlin (changes requested):
Requesting changes — all six failing checks share one root cause, introduced by the very commit that claims to address the previous review.

### Blocking
- `packages/core/src/index.ts:2510` — `confidence: context?.confidence ?? 0.8` references a property that does not exist on `LearnContext`, which adds only `session_id`, `pin_tier` and `pinned_priority`. This fails the `packages/core` declaration build with `TS2339`, and that is the sole cause of all six red checks: `test`, `test (20)`, `test (22)`, `test (24)`, `test (26)` and `smoke-packaged`, the last dying at "Building and packing tarballs". `main`'s latest run is green, so this is introduced here.
- `packages/core/src/inject.ts:379` — the headline guarantee, that every surviving hard-tier engram injects every session, is false on the default configuration. `injection_budget` defaults to 2000, so `hardBudget = min(2000, floor(0.25 × 2000)) = 500`, while the write gate admits up to 2000 tokens of hard-tier engrams. Up to three quarters of what the write gate explicitly accepted is dropped by `if (tokensUsed + cost > hardBudget) continue` — silently, with no eviction warning, no entry in `evicted_soft_pinned`, and no log. The clamp that fixed the budget-starvation item turned the second blocking item's symptom from an edge case into the normal case.
- `packages/core/src/index.ts:2504` — the replacement write-gate estimate undercounts worse than the formula it replaced. The candidate shape carries about ten fields while `learn()` persists roughly thirty more. Measured on a 64-character statement: candidate shape 58 tokens, the old formula 96, a modelled realistic persisted engram 291. Because admission tests `currentRealTokens + estimatedNewCost <= 2000`, the tier settles at up to about 2233 tokens against a 2000 cap — a larger overshoot than the 2064 to 2114 measured against the code that was rejected.

Of the three blocking items in the previous review, one is genuinely fixed (the injection-time starvation clamp), one is claimed fixed but regressed (the write-gate estimate), and one is fixed in form only — the old `TS2339` in `status.ts` is gone because the whole provenance display block was removed, and a new one in `index.ts` replaced it, which is worse because it breaks the full build rather than only the test typecheck. None of the six should-fix items and none of the notes were addressed.

### Should fix
- `packages/core/src/inject.ts:376` — hard-tier selection has no deterministic order. `hardPinned` is consumed in relevance-scored order, so when the clamp bites, which engrams survive depends on the query. The same store injects a different hard set on different queries, which is the opposite of the always-load contract.
- `packages/core/src/inject.ts:390` — the backward-compatibility claim is wrong. Engrams pinned before this change have no tier and default to soft, governed by a 0.3 ratio where 0.5 applied before. At the default budget that cuts the allowance from 1000 to 600 tokens, so previously-fitting pinned engrams start being evicted. The body says there is no injection change for already-pinned engrams that fit, and the matching test-plan checkbox is unchecked.
- `packages/core/src/inject.ts:79` — neither ratio is configurable; both are module-private constants. Combined worst case is 0.55 of the budget against the previous 0.5, so the pinned tiers can displace more of the relevance-scored pool than before, with no way to tune either number.
- `packages/core/src/index.ts:3270` — the remote route through `_buildEngramShape` still stamps the tier and priority with no hard-cap check; the cap exists only on the local path. Flagged previously and unchanged, so hard-tier writes routed to a remote store bypass the cap entirely with no repair path.
- `packages/core/src/index.ts:2698` — tier and priority are persisted straight from `LearnContext` with no validation, so `pinned_priority: 9999` survives despite the schema range, and a `pin_tier: 'HARD'` casing typo silently downgrades to soft and skips the cap check. Flagged previously and unchanged. Note that [#1121](https://github.com/plur-ai/plur/pull/1121) does clamp, so the competing branch already fixes what this one leaves open.
- `packages/core/src/inject.ts:439` — `buildEvictionWarnings` still concatenates every evicted engram into one unbounded string that is not counted in `tokens_used`. Previously measured at about 710 tokens from 60 engrams on a 2000-token budget. Unchanged.
- `packages/core/src/index.ts:2496` — the cap check still runs before deduplication, so re-learning a statement already in the hard tier is rejected even though the write would cost zero additional tokens. Unchanged.
- `packages/core/test/inject.test.ts` — still no test for the hard-tier write rejection, which is the headline behaviour. Both the cap and backward-compatibility checkboxes in the test plan remain unchecked.
- `packages/core/src/index.ts:1691` — `_buildProvenanceField` reads `(this.config as any).provenance` and is called unconditionally. `PlurConfigSchema` has no `provenance` key and is not passthrough, so the disable mode and configured identity are unreachable: the stamp is mandatory, undisableable, and adds about 23 tokens to every engram. The `as any` cast is what keeps it compiling.

### Notes
- `packages/core/src/inject.ts:398` — this and [#1121](https://github.com/plur-ai/plur/pull/1121) genuinely conflict rather than merely sitting near each other. Both replace the same statement in `fillTokenBudget`, both add an identically specified `pinned_priority` to the schema and to `LearnContext`, and both edit the specification and the JSON schema. They disagree on semantics: this tie-breaks by `temporal.learned_at` ascending, #1121 by `retrieval_strength` descending. [#1120](https://github.com/plur-ai/plur/pull/1120) carries the same change a third time. Recommendation: let #1121 own the field, since it clamps and its checks are green; rebase this afterwards and keep only the hard-cap and two-tier parts, carrying over the `learned_at` tiebreaker, which is the stable one.
- `packages/core/src/inject.ts:401` — soft-tier fill is greedy rather than strictly priority-ordered, so a low-priority small engram is admitted while a higher-priority larger one is evicted.
- `packages/core/src/inject.ts:396` — a missing `learned_at` falls back to the empty string, which sorts before every real date, so legacy engrams always win ties. Unknown is being treated as oldest-ever.
- `packages/claw/clawhub/SKILL.md:1` — a 51-line marketing rewrite is still bundled with a core injection-budget change, along with an unrelated README edit and a stray blank line in a CLI test. Asked for previously and still here; it makes the change harder to revert.
- No changelog entry, despite a user-visible change to injection behaviour for anyone with pinned engrams.

Worth keeping: computing the eviction warning before the empty-result early exit, with a comment explaining that everything-was-evicted is exactly when the caller needs telling; placing the cap check outside the store lock with the reasoning recorded; and working the budget arithmetic out in the comment block, which is what made the fourfold mismatch findable at all.

Tests: FAIL — not run locally; CI is conclusive. `packages/core` fails its declaration build with `TS2339` at `index.ts:2510` on all five Node jobs and in `smoke-packaged`, while `main`'s latest run is green. Verified: read the failing logs for every job; computed the budget arithmetic against the default configuration; measured the estimate against a modelled persisted engram; compared the diff against #1121 and #1120.

miles-on-nightshift:
Fix pushed (35a763d5) — addresses both blocking items from your Sep-04 review:

**1. TS2339 `context?.confidence`**
Removed. The candidateShape now uses `confidence: 0.8` (the same default `_buildEngramShape` uses). `LearnContext` has no `confidence` field and never should — the confidence on a newly learned engram is a policy default, not something the caller supplies at learn time.

**2. Write-gate cap mismatch**
The write gate now uses `effectiveHardCap = min(PINNED_HARD_TOKEN_CAP, floor(PINNED_HARD_TOKEN_BUDGET_RATIO × injection_budget))` — identical to the injection-time formula. `PINNED_HARD_TOKEN_BUDGET_RATIO` is now exported from `inject.ts` so `index.ts` imports it directly rather than hard-coding a duplicate.

**Bonus: candidateShape scaffold completeness**
Added the mandatory structural fields (`activation`, `provenance`, `relations`, `sources`, `feedback_signals`) to the shape so the token estimate covers the ~200-token scaffolding overhead every persisted engram carries.

Verified: `pnpm build` clean (core, mcp, claw); `inject.test.ts` 35/35 pass; no TS errors; CI triggered.

Requesting re-review.

miles-on-nightshift:
## Response to 2026-09-04 review (35a763d, 2026-09-05 04:43 UTC)

All three blocking items from the second review are addressed.

**Blocking 1 — TS2339 `context?.confidence`:** Removed. The candidateShape now uses `confidence: 0.8` (a reasonable stand-in for the write-gate estimate). No access to a `LearnContext` property that does not exist.

**Blocking 2 — Injection budget / admission mismatch:** Fixed. The write gate now computes `effectiveHardCap = Math.min(PINNED_HARD_TOKEN_CAP, Math.floor(PINNED_HARD_TOKEN_BUDGET_RATIO * injectionBudget))` — exactly the same formula the injection path uses. `PINNED_HARD_TOKEN_BUDGET_RATIO` is exported so both paths share one constant. What the write gate admits is now what will inject.

**Blocking 3 — Write-gate estimate undercounts:** Fixed. The candidateShape now includes all major structural fields (`activation`, `provenance`, `relations`, `feedback_signals`, `sources`, `commitment`, `type`) with realistic placeholder values. The mandatory scaffolding (~200 tokens per engram) is now in the estimate rather than missing from it.

---

**Still open — should-fix and notes from both reviews:**

The should-fix items (deterministic hard-tier selection order, backward-compatibility claim, non-configurable ratios, remote route bypass, tier/priority validation, unbounded eviction warning, cap-before-dedup, missing test for write-rejection) and the note about the conflict with #1121 / #1120 are not addressed in this commit.

The note recommends letting #1121 own `pinned_priority` and rebasing #1082 on it. That is the right path — #1121 has green checks and clamps correctly. Deferring these to the rebase rather than piling more changes on an already-complex diff; I will track them and address them as part of the rebase pass once #1121 merges.

CI should now be green on this commit — the TS2339 was the only build failure per the review.

plur9:

plur9:
Deferred from 0.20.0 (owner decision, 2026-09-16). Recording what the next
session needs so the collision does not have to be rediscovered.

## Status

`review/1082-pinned` is retained on origin, 101 commits behind `main`, unmerged.
0.20.0 shipped without it.

## Why it was not re-landed

The problem it solves is not currently occurring. #1145 fixed `estimateTokens`
charging the injection budget for `activation`, `feedback_signals`, `usage`,
`sources[]` and `provenance` — 68% of the cost, none of it rendered. On a real
store the pinned set went from **18,290 tokens against a 6,000 quota to 3,663**,
and engrams omitted at injection went **26 → 0**. Eviction machinery decides what
to drop when the set does not fit; it now fits.

What remains is covered: #1142 refuses an over-budget pin at pin time with unpin
candidates, #1138 carries origin ranking (lifted from this line of work), and
`omitted_pinned` names anything a budget did drop and why.

## The collision, measured

A rebase is not mechanical. Conflicts in four files — `core/src/inject.ts`,
`core/src/index.ts`, `core/src/types.ts`, `core/test/inject.test.ts` — and the
two designs sort the same loop by different keys:

| | `main` today (#1138) | this branch |
|---|---|---|
| Outer sort | **origin** — primary store, then `stores:`/remote, then packs | **tier** — hard, then soft |
| Overflow | reported via `omitted_pinned` with a reason | silent `continue` |
| Origin concept | yes, and it is a security control | none |

`main`'s comment states the property directly: *"Sorted before selection so
budget pressure can never let a pack pin in ahead of one of the user's own."*
Taking this branch's side wholesale drops that, so an installed pack could
displace the user's own always-load rules.

## What the next session must decide first

1. **Which key is outer.** Recommended: keep origin outer and layer
   `pinned_tier` / `pinned_priority` as sort keys *within* each origin rank, so
   the security property survives and priority is gained.
2. **Overflow reporting.** Keep `omitted_pinned`; do not reintroduce the silent
   `continue`.
3. **Write gates.** Whether the hard-tier cap stacks with #1142's pin-time quota
   or replaces it.
4. **Recalibrate the constants.** `PINNED_HARD_TOKEN_CAP = 2000` and
   `PINNED_SOFT_TOKEN_BUDGET_RATIO = 0.3` were tuned against the pre-#1145 cost
   basis, where engrams measured ~4x more expensive. Those numbers mean
   something different now and should not be carried over unexamined.

Reopen against the corrected cost basis, as the closing comment said: the
numbers that motivated the original design are no longer the numbers.

PR #1094, opened 2026-08-31
feat(core,cli): plur verify — structured chain verification, never a bare bool (#1053)

plur9:
## Summary

`plur verify` — structured verification of the hash-chained history log (#1053).

Opened **first**, ahead of the rest of Wave 2, because the review of #1073 established that
it is a prerequisite rather than a successor:

> Nothing anywhere in `packages/core`, `packages/cli` or `packages/mcp` ever verifies
> `prev`/`hash` linkage. Every failure mode above is silent, and so is deliberate tampering.

Until something checks the chain, #1051's forks and #1052's mismatched checkpoints cannot be
observed in production, and the L2 claim is not merely unproven — it is unfalsifiable. This is
the instrument the rest of Wave 2 is judged with.

## What it reports

`verifyChain(root)` returns a structured result, **never a bare bool** — a bool is not
anchorable and hides where the chain broke:

```
{ ok, chain_head, total_events, verified_events,
  unprotected_legacy: { from, to, count } | null,
  breaks: [...], forks: [...], checkpoints_checked }
```

Four break reasons, deliberately distinct:

| reason | meaning |
|---|---|
| `hash_mismatch` | the event no longer hashes to its recorded `hash` — content was edited |
| `prev_mismatch` | `prev` does not equal the predecessor's hash — a link or an event was removed |
| `dangling_prev` | `prev` names a hash no event in the log carries |
| `checkpoint_head_missing` | a checkpoint pinned a head the chain no longer contains |

**Forks are their own outcome**, not a kind of tampering: two events claiming one predecessor
is what concurrent writers produce, and the fix is different. **Unprotected legacy** is a
first-class outcome too — pre-#1051 events carry no `hash`/`prev`, and that is expected, never
an error and never silently upgraded.

## Exit codes — the third one is the point

```
0  verified       the chain holds over everything it covers
1  broken         at least one break or fork, each named with month/index/event
2  cannot verify  the log could not be read — NOT the same as clean
```

`readHistory()` silently skips malformed lines (`history.ts:390`). A verifier that did the same
would report a clean chain over a file it could not fully read — the benign-zero this surface
exists to refuse. So this reads raw lines and a line that will not parse is a refusal, naming
the month and line number.

Verified end to end through the **built** CLI against a real store:

```
clean 4-event chain              -> exit 0, verified_events 4
event 2 edited, hash left alone  -> exit 1, breaks [hash_mismatch ENG-2]
malformed JSONL line appended    -> exit 2, "could not parse ... 2026-01 line 5"
```

## The ceiling, stated honestly

A hash chain detects edits that do **not** recompute it. A rewrite that recomputes every `prev`
and `hash` over the remaining events is internally consistent and is **not** detectable from
the log alone — exactly the probe in the #1073 review, where deleting a middle event and
recomputing yielded zero broken links.

Two tests assert both halves:

- `CANNOT detect a full recomputation from the file alone` — asserts the **limitation** on
  purpose. If it ever starts failing, someone has claimed a property a hash chain does not have.
- `DOES detect it once a checkpoint recorded the old head` — the same rewrite is caught as
  `checkpoint_head_missing`.

That is what makes a checkpoint the anchorable object, and why anchoring one externally is the
rung beyond L3 rather than something this PR claims.

## Tests

12 red-first assertions in `packages/core/test/verify-chain.test.ts` — clean chain, month
boundary, mid-chain content edit, deleted event, fork, dangling prev, corrupted line, empty
store, legacy-only store, legacy prefix then chained region, and the two ceiling tests above.
The file failed to import before the module existed.

Full suite: **4741 passed, 159 skipped, 0 failed**. `pnpm build` clean.

## Stack

Based on `feat/1052-checkpoint-event` (#1073), which supplies the chain primitives, and through
it on `integration/core-store`. Part of #1047.

### DIFF
--- a/packages/cli/src/commands/verify.ts
+++ b/packages/cli/src/commands/verify.ts
@@ -23,7 +23,13 @@ import { shouldOutputJson, outputJson, outputText } from '../output.js'
 import { verifyChain, hashStoreFile } from '@plur-ai/core'
 
 export async function run(_args: string[], flags: GlobalFlags): Promise<void> {
-  const plur = createPlur(flags)
+  // Read-only, like `plur ui`. Verification must not be able to write to the
+  // artefact it is examining — and a write-capable handle can, through a lazy
+  // path such as a daily backup or a migration that runs on construction.
+  // `readonly: true` wraps the store in ReadonlyStoreGuard and makes
+  // _assertWritable throw, so "verify never mutates" is enforced rather than
+  // assumed from the fact that this function does not call a write method.
+  const plur = createPlur(flags, { readonly: true })
   const root = plur.storageRoot
   const outcome = verifyChain(root)
 
--- a/packages/cli/src/commands/verify.ts
+++ b/packages/cli/src/commands/verify.ts
@@ -57,6 +57,13 @@ export async function run(_args: string[], flags: GlobalFlags): Promise<void> {
   outputText(`  chain head        : ${r.chain_head ?? '(none — no chained events)'}`)
   outputText(`  checkpoints       : ${r.checkpoints_checked}`)
 
+  if (r.torn_tail) {
+    outputText('')
+    outputText(`  in-flight write: ${r.torn_tail.month} line ${r.torn_tail.line} is incomplete.`)
+    outputText('  A writer was mid-append (or crashed mid-flush). Everything before')
+    outputText('  it verified — this is not corruption and not a break.')
+  }
+
   if (r.unprotected_legacy) {
     const { from, to, count } = r.unprotected_legacy
     outputText('')
--- a/packages/cli/src/commands/verify.ts
+++ b/packages/cli/src/commands/verify.ts
@@ -81,6 +81,17 @@ export async function run(_args: string[], flags: GlobalFlags): Promise<void> {
       outputText('    A checkpoint pinned a chain head this log no longer contains,')
       outputText('    which means events below it were rewritten wholesale.')
     }
+    if (b.reason === 'store_hash_mismatch') {
+      outputText('    The newest checkpoint attests a store that is not the one on disk.')
+      outputText('    The chain itself is intact — engrams.yaml changed under it. That is')
+      outputText('    expected if you have learned anything since; run `plur checkpoint`')
+      outputText('    to attest the current state. If you have NOT, the store was replaced.')
+    }
+    if (b.reason === 'declared_gap') {
+      outputText('    A writer could not take the chain lock and declared a gap rather')
+      outputText('    than chaining from a tail it had not read under exclusion. This is a')
+      outputText('    concurrency artifact, NOT evidence of tampering.')
+    }
   }
 
   for (const f of r.forks) {
--- a/packages/cli/src/commands/verify.ts
+++ b/packages/cli/src/commands/verify.ts
@@ -1,0 +1,96 @@
+/**
+ * `plur verify` — verify the history chain (#1053).
+ *
+ * Exit codes are the contract, and all three are distinct:
+ *
+ *   0  verified        the chain holds over everything it covers
+ *   1  broken          at least one break or fork, each one named
+ *   2  cannot verify   the log could not be read — NOT the same as clean
+ *
+ * The third code is the point. A verifier that answered 0 when it could not
+ * read the log would report "healthy" for a chain nobody checked, which is the
+ * failure this whole surface exists to refuse
+ * (docs/design/failure-must-be-distinguishable.md in the enterprise repo
+ * states the same rule for the audit chain).
+ *
+ * Usage:
+ *   plur verify           Human-readable report
+ *   plur verify --json    Full structured result
+ */
+import { join } from 'path'
+import { createPlur, type GlobalFlags } from '../plur.js'
+import { shouldOutputJson, outputJson, outputText } from '../output.js'
+import { verifyChain, hashStoreFile } from '@plur-ai/core'
+
+export async function run(_args: string[], flags: GlobalFlags): Promise<void> {
+  const plur = createPlur(flags)
+  const root = plur.storageRoot
+  const outcome = verifyChain(root)
+
+  if (outcome.status === 'cannot_verify') {
+    if (shouldOutputJson(flags)) {
+      outputJson({ status: 'cannot_verify', reason: outcome.reason, month: outcome.month ?? null, line: outcome.line ?? null })
+    } else {
+      outputText('plur verify: CANNOT VERIFY')
+      outputText(`  reason: ${outcome.reason}`)
+      if (outcome.month) outputText(`  month:  ${outcome.month}${outcome.line ? ` (line ${outcome.line})` : ''}`)
+      outputText('')
+      outputText('  This is NOT a clean result. The log could not be read in full,')
+      outputText('  so nothing is asserted about the chain either way.')
+    }
+    process.exitCode = 2
+    return
+  }
+
+  const r = outcome.result
+  const storeHash = hashStoreFile(join(root, 'engrams.yaml'))
+
+  if (shouldOutputJson(flags)) {
+    outputJson({ status: outcome.status, store_hash: storeHash, ...r })
+    process.exitCode = outcome.status === 'verified' ? 0 : 1
+    return
+  }
+
+  outputText(outcome.status === 'verified' ? 'plur verify: OK' : 'plur verify: BROKEN')
+  outputText(`  events total      : ${r.total_events}`)
+  outputText(`  verified (chained): ${r.verified_events}`)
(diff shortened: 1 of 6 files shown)


crtahlin (changes requested):
Requesting changes — the verifier is genuinely non-vacuous, which is the hard part and it is done. It fails in the other direction: it reports BROKEN on healthy stores.

I mutation-tested it against every failure mode that motivated it. It correctly names a fork, a dangling `prev`, a content tamper, a reorder and a legacy-after-chain hole, and correctly reports clean on an undamaged chain — including one written by the real `appendHistory` and one spanning a month boundary. It also states its ceiling honestly: a deleted-and-fully-re-chained history is reported verified, exactly as the PR claims, and a checkpoint catches it only if the attacker forgets to rewrite the pin.

The problem is false positives, and for an instrument the rest of Wave 2 is judged with, an alarm that fires on healthy stores is as disabling as one that never fires.

### Blocking

- `packages/core/src/verify-chain.ts:200` — a declared gap is reported as `prev_mismatch`, the reason documented as "a link or an event was removed", i.e. tampering. `appendHistory` deliberately writes `prev: null` mid-chain when it cannot take the chain lock, and its own doc at `packages/core/src/history.ts:257` promises that gap is "visible to `plur verify`, never mistaken for tampering". There is no `declared_gap` member in `BreakReason`. Measured through the built CLI: 8 writers × 80 appends into a fresh store, no tampering and no forks, gave exit 1 BROKEN in 2 of 3 runs — 3 events with `prev: null` on disk, one genesis and two gaps. This is the alarm-on-nothing the PR's own torn-tail commit argues against.

- `packages/core/src/verify-chain.ts:179` — a legacy (unhashed) event appearing after the chain started is reported as `hash_mismatch`, documented as "content was edited". Nothing was edited, and `unprotected_legacy` is left null, so the structured result actively denies a legacy region exists. Reproduced synthetically (3 chained then 5 legacy gives 5 `hash_mismatch` breaks with `legacy: null`) and against a real pre-chain store, where `plur verify` returned exit 1 with roughly 920 such breaks. This is the ordinary rollout shape: run a chained build once, then any non-chaining path, and the store reports permanently BROKEN citing tampering.

- `packages/cli/src/commands/verify.ts:26` — `createPlur(flags)` omits `{ readonly: true }`, so a pure verification command opens a writable engine, contrary to the convention every other read-only command follows (`list.ts:6`, `status.ts:6`, `ui.ts:38`, `outbox.ts:37`, `reindex-hashes.ts:66`). Measured: `plur verify --path <store>` on an untouched 5-event store left the tree non-identical — it created `.tensions-purged` in the store root. No history was appended in that run, but the guard exists precisely because a lazy engine side-effect that appends history would change the chain under the check, so two consecutive verifies could disagree. The fix is one argument.

### Should fix

- `packages/core/src/verify-chain.ts:207` — `if (prev !== null)` excludes genesis from `prevClaims`, so a genesis fork is missed. Two events both carrying `prev: null` — exactly what two concurrent writers on an empty store produce, since both take the gap path — yields `forks: []` and a single `prev_mismatch`. The one fork shape the lock-failure path can actually create is the one not labelled a fork. Same null-handling seam as the first blocking item; fix them together.
- `packages/core/src/verify-chain.ts:251` — the checkpoint check compares only `chain_head`. The recorded `store_hash` is never compared against the store, even though `verify.ts:46` computes `hashStoreFile` and prints it at `:49` beside `checkpoints_checked`. The report shows two hashes it never relates, so `checkpoints_checked` overstates what was checked.
- `packages/core/src/verify-chain.ts:132` — truncation is invisible and exits 0. Dropping trailing whole events reports verified with no break, including when the truncation removes the checkpoints themselves, because the checkpoint loop only iterates checkpoints still in the file — deleting the tail deletes the evidence. Cutting mid-line reports verified with `torn_tail` set, indistinguishable from a concurrent append. The stated ceiling is "a rewrite that recomputes every prev and hash"; plain truncation needs no recomputation at all, so the real ceiling is lower than documented.
- `packages/cli/src/commands/verify.ts:102` — `torn_tail` never reaches the exit code, so a script wrapping `plur verify` cannot see one without parsing `--json`. Combined with the point above, a truncated file returns success.

### Notes

- `packages/core/src/verify-chain.ts:196` — `verified_events` counts only hash self-consistency, not linkage. A payload edited with its own hash recomputed reports `verified_events` 5 of 5 while status is broken, and the CLI prints "verified (chained): 5" directly under "BROKEN". A consumer reading `verified_events == total_events` as all-clear is misled.
- `packages/core/src/verify-chain.ts:102` — the whole file goes into one string plus a retained parsed array and two Maps. 200,000 events / 69.2 MB verified in 706 ms, which is fine, but at 835 MB RSS and 463 MB heap — roughly 12× file size. Past V8's max string length `readFileSync` throws and the catch at `:103` turns a large-but-healthy log into `cannot_verify` (exit 2). Worth streaming before this runs inside `plur doctor`.
- `packages/core/src/verify-chain.ts:223` — `dangling_prev` and `prev_mismatch` double-report the same event, so the "deliberately distinct" taxonomy is not partitioned and a consumer counting breaks double-counts.
- `packages/core/src/verify-chain.ts:264` — `chain_head` on a broken chain is just the last chained event's hash with nothing indicating that something below it failed; it should not be treated as anchorable when `ok` is false.
- Nothing but the CLI calls `verifyChain`. No doctor check, no MCP tool, and `plur checkpoint` does not verify before attesting — so a checkpoint can be minted over a chain with known breaks. That is the same unfalsifiability problem one level up, though it depends on the false positives being fixed first, or doctor turns red on healthy stores.

### Done well

The three-outcome contract — verified / broken / `cannot_verify` with a distinct exit 2 — is the right shape, and refusing to route through `readHistory` (which silently skips malformed lines) is the correct instinct; inheriting that skip would have produced exactly the benign zero this surface exists to refuse.

`packages/core/test/verify-chain.test.ts:220`, "CANNOT detect a full recomputation from the file alone", asserts the limitation on purpose. I confirmed it independently, along with the checkpoint variant. Encoding a negative capability as a test so nobody can later overclaim it is the best thing in this pull request.

Walking past the first break to collect every one is what makes the result actionable — two adjacent events swapped produced all three consequent mismatches with correct indices.

Tests: PASS — full monorepo suite 4,750 passed / 159 skipped / 0 failed, exit 0 in 143 s, no flaky exclusions needed.

Verified: mutation-tested against fork, dangling prev, content tamper, reorder, legacy-after-chain, truncation (whole-event, mid-line, and checkpoint-removing), delete-and-re-chain, and two clean controls; the false-positive rate measured through the built CLI with 8 concurrent writers and against a real pre-chain store; the store-mutation check done by comparing the tree before and after a verify. Diffed against `feat/1052-checkpoint-event` — exactly 3 commits, 5 files, no base findings.

plur9:
All three blocking findings addressed.

**Declared gap reported as tampering** — `declared_gap` is now its own
`BreakReason`, so a `prev: null` written by the lock-failure path is no longer
reported under the reason a rewritten link produces.

**Legacy event after chain start** — unchained events are now counted into
`unprotected_legacy` wherever they appear, and skipped rather than broken. The
leading-run assumption was that a store crosses the chain boundary exactly once;
the ordinary rollout shape interleaves. `verified` already means "the chain
holds over everything it covers", not "everything is covered", so uncovered is
reported rather than alarmed on. Real tampering inside the chained region is
unaffected — mutation-verified, 4 tests fail when the change is reverted.

**`createPlur` without `readonly`** — fixed; one argument, as you said.

Also took the first of the non-blocking items, since you noted it is the same
null-handling seam: **genesis is now a claim.** Two events both carrying
`prev: null` — two concurrent writers on an empty store, both taking the gap
path — are reported as a fork instead of `forks: []` plus a single
`prev_mismatch`. Keyed under a sentinel that cannot collide with a hash, and
reported back as `prev: null` so the consumer contract stays "the predecessor
these events claim". `ChainFork.prev` widens to `string | null`.

**`store_hash` never compared** — also fixed, and it needed more than a
comparison: `hashStoreFile` hashed RAW FILE BYTES while `emitCheckpoint` records
the canonical JSON of the parsed document, so the two could never have been
equal even if something had compared them. #1052 had explicitly rejected
raw-byte hashing for exactly this. It now delegates to `hashEngramsFile`, and
the newest checkpoint is compared against the store with a
`store_hash_mismatch` reason. Older checkpoints attest older states and are
deliberately not compared.

**Not addressed, and I think it wants your call rather than my guess:**
truncation is invisible and exits 0. Closing it means committing to a length or
a count, and surfacing `torn_tail` in the exit code means a fourth exit code
against a contract that documents exactly three as the point. Both are design
decisions about what the instrument promises, not fixes.

Core + CLI: 3759 passed, 0 failed.

PR #1113, opened 2026-09-02
feat(core): instrument spreading-activation drop counters

miles-on-nightshift:
## Summary

- Adds `dropped_unresolvable` counter for spreading-activation edges silently dropped because the target engram id is absent from the local store (remote-only or deleted). Previously this bridge was completely dark with no observable signal.
- Adds `dropped_retired` counter (separate bucket) for edges where the target id exists locally but has `status !== 'active'`.
- Both counters accumulate on `Plur._spreadDrops` per process lifetime and surface via `plur_status` as `spread_drops: { dropped_unresolvable, dropped_retired }` (field absent when both are zero).

No behaviour change to injection itself — pure instrumentation.

## Changed files

| File | Change |
|------|--------|
| `packages/core/src/inject.ts` | Build `nonActiveIds` set alongside `engramMap`; split the silent `continue` into two counted cases; return `spread_drops` in `InternalInjectionResult` |
| `packages/core/src/index.ts` | Add `_spreadDrops` accumulator to `Plur`; add `spread_drops` to `StatusResult`; populate from `result.spread_drops` in `_injectEngramsCore`; return from `status()` |
| `packages/mcp/src/tools.ts` | Pass through `spread_drops` in `plur_status` handler |
| `packages/core/test/inject.test.ts` | 5 new tests covering: absent when no associations; unresolvable only; retired only; both kinds; document-target associations are not counted |

## Safety check: sibling sweep (R5)

The only write path changed is the split of one `continue` into two branches, each incrementing a counter. No other `engramMap.get` / association-resolution site exists — confirmed:

```
grep -n "engramMap.get\|assoc.target" packages/core/src/inject.ts
```

```
554:      const target = engramMap.get(assoc.target)
```

One call site, now instrumented.

## Test plan

- [x] `pnpm exec vitest run test/inject.test.ts` — 31/31 pass (5 new)
- [x] All inject-related test files — 82/82 pass
- [x] Full MCP test suite — 404/404 pass
- [x] Full core build — clean

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -395,6 +395,14 @@ export interface StatusResult {
   store_errors?: Record<string, string>
   /** @deprecated Use `store_errors.packs`. Kept so existing readers still work. */
   pack_registry_error?: string
+  /**
+   * Spreading-activation association edges dropped since process start, by reason.
+   * Accumulates across all `inject()` calls in this process — resets on restart.
+   * `dropped_unresolvable`: target id absent from local engramMap (remote-only or
+   * deleted engram). `dropped_retired`: target found but not active. Absent when
+   * both counts are zero.
+   */
+  spread_drops?: { dropped_unresolvable: number; dropped_retired: number }
 }
 
 /**
@@ -779,6 +787,8 @@ export class Plur {
    * event; findLatestInjectionFor covers the cross-process case.
    */
   private _lastInjectionByEngram: Map<string, string> = new Map()
+  /** Spreading-activation drop counters — accumulated in-memory, reset on process restart. */
+  private _spreadDrops = { dropped_unresolvable: 0, dropped_retired: 0 }
   /**
    * Timestamps (ms) of recent LLM failures, newest last (convergence Phase 2).
    *
@@ -4911,6 +4921,11 @@ export class Plur {
       embeddingBoosts,
     )
 
+    if (result.spread_drops) {
+      this._spreadDrops.dropped_unresolvable += result.spread_drops.dropped_unresolvable
+      this._spreadDrops.dropped_retired += result.spread_drops.dropped_retired
+    }
+
     const directivesStr = formatWithLayer(result.directives, assignLayer('directives'))
     const constraintsStr = formatWithLayer(result.constraints, assignLayer('constraints'))
     const considerStr = formatWithLayer(result.consider, assignLayer('consider'))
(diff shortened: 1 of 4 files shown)


crtahlin (approved):

The "pure instrumentation, no behaviour change" claim holds, and I checked it rather than accepting it. Splitting `if (!target || target.status !== 'active') continue` into two counted branches is semantically identical, and `nonActiveIds` is written but never read by any selection decision. The state is bounded — two integers on a per-server instance, single-threaded increments, no race — and omitting the field when both counts are zero follows the existing `store_errors` convention in the same object literal, where every consumer already reads through optional chaining.

The sibling sweep also checks out: `engramMap.get(assoc.target)` has exactly one call site, so no second association-resolution path is left dark.

Follow-ups: #1129, #1130, #1131

### Should fix
- `packages/core/src/inject.ts:570` — `dropped_unresolvable` is documented as "target id absent from local engramMap (remote-only or deleted)" but it also absorbs engrams that are present locally and active yet skipped by `skipForValidity`. Confirmed by probe against the head: a directive associated to a locally-present engram with `temporal.valid_until: '2026-01-31'` yields `{dropped_unresolvable: 1, dropped_retired: 0}`. An operator seeing a high count goes looking for missing engrams when the cause is expiry.
- `packages/core/src/inject.ts:458` — `nonActiveIds` is populated only in the personal-engram loop. The pack loop applies the same status and validity filters and records nothing, so a non-active pack engram is counted as unresolvable rather than retired. The retired bucket is silently personal-engrams-only.
- `packages/core/src/index.ts:7745` — the counter is a running total since process start with no denominator and no reset, so it cannot answer whether drops are happening now, and 500 is meaningless without knowing whether 600 or 600,000 edges were walked. In command-line use it is worse: `plur inject` and `plur status` are separate processes, so the field is effectively never reported.
- `packages/core/src/index.ts:4924` — the accumulation and the `status()` exposure are both untested. All five new tests call `selectAndSpread` directly, so the two lines most likely to be dropped in a future refactor of `_formatInjection` have no coverage.

### Notes
- `packages/core/src/inject.ts:574` — `if (target.status !== 'active') { droppedRetired++; continue }` is unreachable. Both `engramMap.set` sites are already guarded by a status check, so a successful `get` always returns an active engram. Harmless, but it reads as a live path, and the test named for that counter actually exercises the `nonActiveIds` path instead.
- `packages/core/src/inject.ts:569` — drops are counted per edge while accepted spreads are deduplicated through `visited`, so the two should not be compared as a ratio. The doc comment does say "edges", so this is defensible.
- `packages/cli/src/commands/status.ts:33` — the CLI renders `store_errors` but has no branch for `spread_drops`, and the second status-shaped handler in `packages/mcp/src/tools.ts:3030` does not pass it through either. Worth a deliberate decision rather than an omission.

Verified: copied `packages/core` from `origin/main` into a scratch directory, overlaid the head's `inject.ts` and `inject.test.ts`, and ran vitest on `inject.test.ts` plus `supersedes-inject.test.ts` — 37 passed including the 5 new; wrote probes for the expiry and duplicate-target cases. Tests: PASS.

plur9 (approved):
Reviewed against the PR head: pure in-memory instrumentation with a single selectAndSpread call site, so coverage is complete. One label imprecision for a follow-up: nonActiveIds is built from personal engrams only while engramMap also holds pack engrams, so a retired pack target counts as dropped_unresolvable.

PR #1114, opened 2026-09-02
fix(mcp,core): apply readIdFor in learn catch-block; throw on namespaced-id remote miss (#1109)

### LINKED ISSUE #2: npm install shows deprecated boolean@3.2.0 warning
During `npm install -g @plur-ai/mcp`, the first thing a user sees is:

```
npm warn deprecated boolean@3.2.0: Package no longer supported. Contact Support at https://www.npmjs.com/support for more info.
```

Not a blocker, but it's the very first output during install — gives a bad first impression. Worth replacing or pinning the `boolean` dependency to remove the warning.

### LINKED ISSUE #1109: plur_learn returns a local-form ID for remote-scoped writes, so the returned ID can collide with an unrelated engram — and plur_forget then retires the wrong one
## The problem

`plur_learn` returns an `id` that does not match the ID the engram is actually stored under when the write routes to a remote (group-scoped) store. The returned ID is the local-form `ENG-<date>-<n>`; the stored ID is `ENG-GPL-<date>-<n>`.

Because the local counter and the remote counter run independently, the returned ID frequently collides with a *different, unrelated* engram that already exists in the local store on the same day.

## What it caused

On 2026-09-01:

1. `plur_learn` with `scope: group:plur/plur-ai/engineering` returned `{"id": "ENG-2026-09-01-036"}`.
2. The engram was actually stored remotely as `ENG-GPL-2026-09-01-036`.
3. A *different* engram — written earlier the same day by a concurrent session, scope `global` — already held the local ID `ENG-2026-09-01-036`.
4. Calling `plur_forget` with the ID that `plur_learn` had just returned retired the concurrent session's unrelated engram.

`plur_history` for `ENG-2026-09-01-036` shows both writes under one ID, which is how the collision is visible after the fact:

```
engram_created  13:15:22Z  scope: global
engram_created  13:21:52Z  scope: group:plur/plur-ai/engineering  routed_to: remote
engram_retired  13:34:09Z  reason: null
```

The failure is silent. Nothing in the `plur_learn` response indicates the ID is local-form, that the write routed remotely, or that the returned ID already belongs to something else.

## Compounding: the mistake is not reversible through the API

- `plur_promote` on a retired engram returns `Cannot promote retired`. There is no un-retire.
- `plur_forget` with the correct remote ID (`ENG-GPL-2026-09-01-036`) returns `Engram not found` — forget resolves against the local store only, so a remote-scoped engram cannot be retired by its real ID.
- `plur_forget` with `search` returned 100 matches and refused, which is correct behaviour but leaves no path to the target.

So a caller who follows the returned ID in good faith can destroy an unrelated engram and then cannot undo it or clean up the one they meant to remove. Recovery meant re-creating the retired engram from the text echoed in the `plur_forget` response, under a new ID.

## Fix

1. `plur_learn` returns the ID the engram is actually stored under. If the write routed to a remote store, return `ENG-GPL-...`.
2. Include the routing destination in the response (the `routed` field already exists for auto-routing — make it explicit for explicit scopes too), so a caller can tell local from remote without inferring it.
3. `plur_forget`, `plur_promote`, `plur_feedback` and `plur_history` resolve IDs across every configured store, not just the local one. Today `plur_history` clearly reads across stores while `plur_forget` does not, which is the inconsistency that made this dangerous.
4. Refuse, or at minimum warn, when a mutating call by ID resolves to an engram whose scope differs from the caller's session scope. That single guard would have stopped this.
5. Consider an un-retire, given that retirement is documented as non-destructive ("marks it as no longer active without deleting history"). If history is kept, the state should be reversible.

## Acceptance

- A `plur_learn` to a remote scope returns an ID that `plur_forget` can resolve.
- ID resolution behaves the same across all mutating tools.
- A cross-scope mutation by ID either fails loudly or warns.

miles-on-nightshift:
## Summary

Closes #1109. Fixes two gaps that remained after #914 landed.

### 1. `packages/mcp/src/tools.ts` — catch-block at ~line 1173

`learnRouted` saves to the local outbox on remote failure, so the `catch` block is defence-in-depth. But when it did fire, it called `plur.learn()` and returned `engram.id` raw — the same collision risk as the original incident. Any write to a remote-backed scope produced a local-form ID (`ENG-2026-…`) while the engram lived remotely under `ENG-GPL-2026-…`.

**Fix:** apply `plur.readIdFor(engram)` with the same outbox exception as the happy-path fix at line 1149:

```
id: isOutbox ? engram.id : plur.readIdFor(engram)
```

### 2. `packages/core/src/index.ts` — forget() remote walk at ~line 6082

When `probeById` returned `'unknown'` (expired token, network down) and the caller ID was namespaced (`ENG-GPL-…`), the walk logged a warning and continued. The terminal error was "Engram not found: … but 1 store could not be reached" — which still starts with "Engram not found" and implies absence it could never verify.

The namespaced case is different: `_stripRemotePrefix` already stripped the store prefix (`id !== serverId`), so we **know** this store is the intended target. Continuing past it is misleading.

**Fix:** when `id !== serverId` (prefix was stripped → ID was namespaced to this store) and `ownership === 'unknown'`, throw immediately:

```
Cannot reach "group:test" to retire "ENG-GTE-…" — the token may be expired
or the server unavailable. Retry once access is restored, or pass scope:
"group:test" to target this store directly.
```

Bare-ID walk (`id === serverId`, no stripping happened) preserves the existing `#84` graceful-degradation contract — the existing `'forget handles remote server error gracefully'` test still passes unchanged.

## Safety claim — only two places in tools.ts return `id:` from a learn result

```
$ grep -n "id: .*engram\.id\|id: isOutbox" packages/mcp/src/tools.ts
1149:            id: isOutbox ? engram.id : plur.readIdFor(engram), statement: engram.statement,
1173:            id: isOutbox ? engram.id : plur.readIdFor(engram), statement: engram.statement,
```

Both now apply `readIdFor`. `plur_learn_batch` (lines 1284/1331) uses `engram.id` because the batch path is documented as LOCAL-only (`learnRouted` is not applied per item — see tool description). `plur_forget` at line 1605 returns the ID from a local `getById`, so local-form is correct.

## Regression tests

**R4 — test (c) fails against pre-fix code (confirmed by stash + run + pop):**

```
- Expected:  /Cannot reach/
+ Received:  "Engram not found: ENG-GTE-2026-09-01-099 — but 1 store(s) could not be reached (group:test)…"
```

**All 36 tests in `remote-routing.test.ts` pass with the fix.**

3 new tests:
- `readIdFor returns namespaced form for a remote-backed scope` — anchors the MCP catch-block fix by testing the core primitive
- `forget(namespaced-id) strips prefix and deletes from the correct remote` — confirms the existing forget routing works end-to-end with a prefixed ID (cold cache, reachable remote)
- `forget(namespaced-id) with unreachable remote throws actionable error, not "Engram not found"` — the regression test for fix #2; fails before, passes after

## Base

Branched off `origin/main` (ae11182f). `git diff origin/main...HEAD --name-only`:

```
packages/core/src/index.ts
packages/core/test/remote-routing.test.ts
packages/mcp/src/tools.ts
```

No unrelated changes in the diff.

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -6051,12 +6051,17 @@ export class Plur {
       // remote was walked past and the engram reported as simply not found —
       // absence the walk never verified, which is #831's harm by another route.
       //
-      // The walk CONTINUES on `unknown` rather than refusing: `forget handles
-      // remote server error gracefully` (#84) asserts a degraded fleet does
-      // not stop a retire, and that availability is worth keeping. What
-      // changes is only that the store is recorded, so the terminal message
-      // below stops claiming knowledge it does not have. Same resolution
-      // `feedback` already uses.
+      // For bare IDs the walk CONTINUES on `unknown`: `forget handles remote
+      // server error gracefully` (#84) asserts a degraded fleet does not stop
+      // a retire — availability is worth keeping for the ambiguous case. What
+      // changes is only that the store is recorded so the terminal message
+      // stops claiming knowledge it does not have. Same resolution `feedback`
+      // already uses.
+      //
+      // For namespaced IDs (ENG-GPL-...), the prefix was stripped above — meaning
+      // we KNOW this store is the intended target. An unreachable store is not
+      // absence; continuing and reporting "not found" is actively misleading
+      // (#1109). Throw immediately so the caller gets the scope to retry with.
       // Optional capability: a driver without `probeById` (an injected stub, a
       // third-party implementation) keeps the previous two-state behaviour
       // rather than crashing. Absence of the capability is not a reason to
@@ -6065,6 +6070,14 @@ export class Plur {
         ? await driver.probeById(serverId)
         : ((await driver.getById(serverId)) ? 'owned' : 'absent')
       if (ownership === 'unknown') {
+        const isNamespaced = id !== serverId
+        if (isNamespaced) {
+          throw new Error(
+            `Cannot reach "${entry.scope ?? entry.url}" to retire "${id}" — `
+            + `the token may be expired or the server unavailable. `
+            + `Retry once access is restored, or pass scope: "${entry.scope ?? entry.url}" to target this store directly.`,
+          )
+        }
         unreachedStores.push(entry.scope ?? entry.url!)
         logger.warning(
           `[plur] could not reach "${entry.scope ?? entry.url}" while looking for ${id} — `
--- a/packages/core/test/remote-routing.test.ts
+++ b/packages/core/test/remote-routing.test.ts
@@ -199,6 +199,41 @@ describe('learn() — remote routing (issue #25)', () => {
     expect(found!.structured_data?._outbox).toBeDefined()
     expect(found!.structured_data._outbox.target_scope).toBe('group:plur/plur-ai/engineering')
   })
+
+  // #1109 — learnRouted catch-block gap: the fallback path (plur.learn inside
+  // the catch) returned engram.id raw, the same collision risk as before #914
+  // if learnRouted throws and learn falls back for a remote-backed scope.
+  // readIdFor must be applied here too so the returned ID matches recall.
+  it('readIdFor returns namespaced form for a remote-backed scope', async () => {
+    mockSuccessfulAppend()
+
+    writeStoresConfig(primaryDir, [
+      {
+        url: 'https://plur.example.com/sse',
+        token: 'plur_sk_test',
+        scope: 'group:plur/plur-ai/engineering',
+        shared: true,
+        readonly: false,
+      },
+    ])
+    const plur = new Plur({ path: primaryDir })
+
+    // learnRouted returns the server's raw ID; readIdFor converts it to the
+    // namespaced form that recall surfaces so both sides agree on the key.
+    const engram = await plur.learnRouted('test for readIdFor namespacing', {
+      scope: 'group:plur/plur-ai/engineering',
+    })
+    // Wait for the async push to settle before reading the returned id.
+    await new Promise(r => setTimeout(r, 100))
+
+    const readId = plur.readIdFor(engram)
+    // storePrefix('group:plur/plur-ai/engineering') → 'GPL'
+    expect(readId).toMatch(/^ENG-GPL-/)
+    // The raw engram.id is the server-assigned bare form.
+    expect(engram.id).not.toMatch(/^ENG-GPL-/)
+    // The namespaced form differs from the raw form.
+    expect(readId).not.toBe(engram.id)
+  })
 })
 
 /**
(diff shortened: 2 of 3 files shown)


crtahlin (approved):
Approved on its own merits — but do not merge this and [#1122](https://github.com/plur-ai/plur/pull/1122) as written. See the sequencing constraint below; it is the reason this approval is conditional.

Both changes are correct against today's `main`, and the regression test is a real one. I ran it against unpatched `main` in an isolated clone and it fails with exactly the misleading message the issue describes, then passes with the fix — 35 of 36 before, 36 of 36 after. It also asserts the absence of the "Engram not found" prefix rather than only the presence of the new text.

The catch-block fix is safe for the common case because core's `learn()` stamps `structured_data._outbox` synchronously before the fire-and-forget push, so the fallback returns the local-form identifier and `readIdFor` namespaces only when the write really is remote.

On throw versus degrade: keep the throw. With a single configured store, `main` already ended this path in an exception, so this is not a degrade-to-throw change — it replaces a message that asserts unverified absence with one that names the scope and the retry.

Follow-ups: #1126, #1127, #1128

### Sequencing constraint — blocking for the pair, not for this change
- `packages/core/src/index.ts:6074` — the gate `const isNamespaced = id !== serverId` depends on `_stripRemotePrefix` staying scope-aware, returning the identifier unchanged when the prefix does not match this store's scope. [#1122](https://github.com/plur-ai/plur/pull/1122) replaces that fallback with `bareEngramId(id)`, which strips any two-to-four character uppercase prefix regardless of scope. With both landed, the gate is true for every configured store whenever the identifier is namespaced, so the first unreachable store throws `Cannot reach "<scope>" to retire …` naming a store that never held the engram — and a retirement that would have succeeded on a later reachable store fails. Tracked as #1126. The preferred fix is on the #1122 side: keep `_stripRemotePrefix` scope-aware and give the recall identifier surfacing its own helper.

### Should fix
- `packages/core/src/index.ts:6075` — the throw exits the walk immediately rather than after it. `storePrefix` is a lossy three-character derivation, so two scopes can share a prefix. If the colliding store is listed first and is unreachable, `main` would have continued and retired from the store that actually holds the engram; this throws and names the wrong scope. Deferring the throw until the walk completes, firing only when nothing was retired, keeps the message, preserves multi-store completion, and blunts the interaction above.
- `packages/core/src/index.ts:5800` — the new error tells the caller to pass `scope:` to target the store directly, but that path still uses plain `driver.getById(serverId)` and reports absence when the store is unreachable. During the exact outage that triggers this message, following the advice produces the false-absence claim the change exists to eliminate.
- `packages/mcp/src/tools.ts:1173` — the `isOutbox` exception does not cover every branch where a remote-scoped write stays local. `visibility: 'private'` and a resolved-but-unwritable store both skip the outbox stamp while the scope remains remote-backed, so `readIdFor` reports a namespaced form for an engram that exists only locally. Pre-existing on the happy path rather than introduced here, but the fix duplicates it.
- The MCP change itself is untested. The added test exercises core's `readIdFor` and passes against unpatched `main`.

### Notes
- `packages/core/src/index.ts:6071` — the `unreachedStores` docblock still reads "Recorded, NOT thrown on: … a degraded fleet must not stop a retire", which is now false for namespaced identifiers. The inline comment ten lines below was updated; this one was not.
- `packages/core/test/remote-routing.test.ts:640` — the "strips prefix and deletes from the correct remote" test passes on unpatched `main`. A valid characterisation test, but not a regression test for anything here.

Verified: reproduced the failing test on unpatched `main` and confirmed it passes with the patch in an isolated clone; read #1122's diff to `engrams.ts` and `index.ts:1647` to confirm the interaction; traced the outbox stamp in `learn()` to check the fallback's safety. Tests: PASS for the core remote-routing suite; the MCP and CLI suites were not run.

plur9 (approved):
Reviewed against the PR head: readIdFor is applied in the catch block (isOutbox is declared in that scope), and forget() now throws an actionable error when the store a namespaced id names is unreachable instead of walking past it. Tests fail before and pass after. One note for a follow-up: the readIdFor test waits 100 ms for the background push and could flake under load.

PR #1120, opened 2026-09-03
ci: monthly hermes-agent PyPI compat canary

miles-on-nightshift:
Adds `.github/workflows/hermes-compat-monthly.yml` — a monthly cron canary that installs the latest `hermes-agent` from PyPI and the latest `@plur-ai/cli` from npm, runs the plur-hermes test suite against them on Python 3.11 and 3.12, and files one deduplicated issue per month and Python version on failure (`workflow_dispatch` for on-demand runs).

## Changes since the first draft (review findings)

- **pytest was never installed.** Neither `hermes-agent` (pytest only in its `dev` extra) nor a bare `pip install -e packages/hermes` provides it, so every run would have failed with `pytest: command not found` and filed a bogus issue. The job installs `packages/hermes[test]`.
- **A failing suite could not fail the step.** `pytest … | tee` ran under the default `run:` shell (`bash -e {0}`, no pipefail), so the step's status was tee's and `if: failure()` could never fire. The step now runs under an explicit `bash` with `pipefail`.
- **`register()` needs the CLI.** `plur_hermes.register()` verifies the `plur` binary first and returns early without it, which is what `test_memory_provider_entrypoint.py` exercises; the job installs the published CLI so the canary tests the shipped pair the way a user's machine does.
- Rebuilt on `main` with only the workflow file; the previous head carried the #1121 commits. `timeout-minutes: 20`; permissions stay `contents: read` + `issues: write`.

**Retirement signal:** when `pip install hermes-agent` stops advancing past 0.19.0 for two consecutive months, retire or replace with a Herald-native install step.

### DIFF
--- a/.github/workflows/hermes-compat-monthly.yml
+++ b/.github/workflows/hermes-compat-monthly.yml
@@ -87,19 +87,26 @@ jobs:
         with:
           script: |
             const fs = require('fs');
-            const output = fs.readFileSync('pytest-output.txt', 'utf8');
+            // Guard: if the failure happened before pytest ran, the output file
+            // does not exist. Fall back to a generic body so the issue still
+            // fires with useful run context rather than throwing here.
+            const output = fs.existsSync('pytest-output.txt')
+              ? fs.readFileSync('pytest-output.txt', 'utf8')
+              : '(no pytest output captured — failure occurred before test run)';
             const now = new Date();
             const month = now.toLocaleString('en-US', { month: 'long', year: 'numeric', timeZone: 'UTC' });
             const title = `hermes-agent PyPI compat failure — ${month} (py${{ matrix.python-version }})`;
 
-            // Avoid duplicate issues: search for open issues with this title prefix
-            const existing = await github.rest.issues.listForRepo({
-              owner: context.repo.owner,
-              repo: context.repo.repo,
-              state: 'open',
-              labels: 'hermes-compat',
-            });
-            const duplicate = existing.data.find(i => i.title === title);
+            // Avoid duplicate issues — paginate to handle repos with many open
+            // issues; the label filter keeps the page count small in practice.
+            let existing = [];
+            for await (const page of github.paginate.iterator(
+              github.rest.issues.listForRepo,
+              { owner: context.repo.owner, repo: context.repo.repo, state: 'open', labels: 'hermes-compat', per_page: 100 }
+            )) {
+              existing = existing.concat(page.data);
+            }
+            const duplicate = existing.find(i => i.title === title);
             if (duplicate) {
               console.log(`Issue already exists: #${duplicate.number}`);
               return;
--- a/.github/workflows/hermes-compat-monthly.yml
+++ b/.github/workflows/hermes-compat-monthly.yml
@@ -1,0 +1,133 @@
+name: hermes-agent PyPI compat
+
+# Monthly canary: pip install hermes-agent (latest from PyPI) and run the
+# plur-hermes test suite against it.
+#
+# WHY THIS EXISTS:
+# plur-hermes v0.17.2+ passes all 144 tests against hermes-agent 0.19.0
+# (the current pip-installable release). The Hermes Herald Release (v0.20.0,
+# 2026-08-03) ships via shell installer / Docker / Nix only — PyPI / pip is
+# a deprecated install channel as of Herald. This job monitors the latest
+# pip-installable version and catches any regression before it affects users
+# who still install via pip.
+#
+# RETIREMENT SIGNAL:
+# When `pip install hermes-agent` stops advancing beyond 0.19.0 for two or
+# more consecutive months, this job should be retired or replaced with a
+# Herald-native install step (shell installer or Docker image pull).
+
+on:
+  schedule:
+    # 1st of each month at 09:00 UTC
+    - cron: '0 9 1 * *'
+  workflow_dispatch:
+
+permissions:
+  contents: read
+  issues: write
+
+jobs:
+  hermes-compat:
+    name: hermes-compat / py${{ matrix.python-version }}
+    runs-on: ubuntu-latest
+    # The third-party package's install and test code runs with this job's
+    # token; keep the permissions above minimal and the job bounded.
+    timeout-minutes: 20
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ['3.11', '3.12']
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      # plur_hermes.register() verifies the @plur-ai/cli binary first and
+      # returns early without it, which is what the entry-point suite
+      # (test_memory_provider_entrypoint.py) exercises — so the canary needs
+      # the CLI on PATH exactly as a user's machine does. Latest published,
+      # not a workspace build: this job asks whether the shipped pair works.
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 22
+      - name: Install the plur CLI (latest from npm)
+        run: npm install -g @plur-ai/cli
+
+      - name: Install hermes-agent (latest from PyPI)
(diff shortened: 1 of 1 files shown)


crtahlin (dismissed):
Requesting changes — the canary cannot detect what it exists to detect, and seven of the eight changed files are not this change.

Two independent defects make the job pass silently, so a genuine regression produces a green monthly run and the failure handler never fires.

### Blocking
- `.github/workflows/hermes-compat-monthly.yml:57` — pytest failures are swallowed. `pytest ... | tee pytest-output.txt` runs under the Actions default shell `bash -e {0}`, which does not set pipefail, so the step's exit status is `tee`'s, always 0. Line 58 captures `${PIPESTATUS[0]}` into the step output and no later step ever reads it. The `if: failure()` issue-opening step at line 61 therefore never fires. Add `shell: bash`, which supplies `-o pipefail`, or gate a follow-up step on the captured value.
- `.github/workflows/hermes-compat-monthly.yml:46` — pytest is never installed, so the suite cannot run at all. Verified against PyPI metadata for hermes-agent 0.19.0: pytest appears only under the `dev` extra, and line 52 installs plur-hermes without its `[test]` extra, where `pytest>=7.0` is declared. Line 57 exits 127 with "command not found"; combined with the pipefail defect the job goes green having executed nothing. The existing `hermes.yml` gets this right, and the body's own verification commands install pytest explicitly — the workflow diverges from both.
- `packages/core/src/index.ts:1` — scope contamination. Seven of the eight changed files are the entire `pinned_priority` change and do not belong in a CI canary. Confirmed at blob level: each of the seven has a blob hash identical to the same path in [#1121](https://github.com/plur-ai/plur/pull/1121) — the test file is `c4904dd` in both — and the second commit here is literally "feat(inject): add pinned_priority field for priority-weighted eviction". Reset the branch so it carries only the workflow file. [#1121](https://github.com/plur-ai/plur/pull/1121) should own that field: it is the only one of the three copies that clamps the value and the only one with every check green.

### Should fix
- `.github/workflows/hermes-compat-monthly.yml:66` — the issue-opening step only survives a pytest failure. `fs.readFileSync('pytest-output.txt')` throws if the run died earlier, and the most important failure mode is exactly that: `pip install hermes-agent` failing or resolving to nothing is the retirement signal this canary is meant to watch for. The one condition worth alerting on produces silence. Guard the read with `existsSync` and fall back to a generic body.
- `.github/workflows/hermes-compat-monthly.yml:76` — the `hermes-compat` label does not exist in this repository. Deduplication filters open issues on it and creation applies it, so if it is not applied as expected the deduplication query never matches and the job opens a fresh duplicate every month for as long as the failure persists. Create the label first. The `listForRepo` call is also unpaginated, which compounds this.
- `.github/workflows/hermes-compat-monthly.yml:12` — the evidence for "no v0.20.x tag exists on GitHub" is true but misleading. The upstream repository tags with calendar versions, not semantic ones: the current tags are `v2026.8.31`, `v2026.8.27`, `v2026.8.19`, `v2026.8.3` and so on. No semantic tag of any kind exists there, so a missing `v0.20.2` proves nothing about whether the release shipped — and `v2026.8.3` matches the stated release date. The version scheme the body reasons in does not correspond to upstream's tagging, which weakens the retirement argument built on it.
- `.github/workflows/hermes-compat-monthly.yml:15` — the retirement signal is prose in a comment, enforced by nothing. Nothing compares the installed version against a floor, so once the published package stagnates the job spends two runner jobs a month re-verifying a version already known good and nobody is told. The published version is 0.19.0 today, so this is the state on day one, not a hypothetical.

### Notes
- Echoing the tested version into the step summary would let a reviewer confirm at a glance whether the canary is still tracking a moving target or has stalled, which is exactly the retirement decision the header comment asks a human to make.
- The matrix covers Python 3.11 and 3.12. That fits the upstream requirement, but the existing `hermes.yml` covers 3.10 through 3.13, and 3.13 is inside the supported range and left unexercised. Worth stating the omission deliberately or adding it.
- The failing `test (26)` job is a flake in `sp1-memory-intelligence.test.ts`, not caused by this change: the same core changes pass that leg on #1121, and `main` passes it too. Filed separately in the follow-up notes.
- The idea itself is sound and not redundant: `hermes.yml` runs the suite without hermes-agent installed, so exercising it against the real published package is genuinely new coverage. Only the implementation is broken.

Worth keeping: the header comment states why the job exists, what it monitors, and the condition under which it should be deleted — a retirement criterion written at creation time is rare and is the right way to stop a cron canary becoming permanent noise. Permissions are declared explicitly and minimally, `fail-fast: false` is right for a matrix, and routing failures to a deduplicated issue rather than a red tab nobody watches is the correct instinct.

Tests: SKIP — no tests run locally; the assessment rests on the CI logs for this pull request, for #1121 and for `main`. Verified: read the failing job log; resolved hermes-agent 0.19.0's declared dependencies against the real PyPI metadata; listed the upstream repository's actual tags; compared blob hashes for all seven contaminating files against #1121; checked this repository's label list.

miles-on-nightshift:
## Review addressed — pushed 7700dc4c

### Blocking — already resolved before the re-review

All three blockers from crtahlin's review were addressed in the first push after the initial review:

1. **Scope contamination** — the 7 files from #1121 (`pinned_priority`) were stripped; the branch now carries only the workflow file (`git diff origin/main...HEAD --name-only` = 1 file).
2. **pipefail** — `shell: bash` added to the pytest step with an explicit `set -o pipefail`, making the step fail on pytest failures rather than always exiting 0 (tee's exit status).
3. **pytest not installed** — added `pip install -e "packages/hermes[test]"` as a named step before the pytest step, with a comment explaining why the bare install doesn't bring pytest.

### Should-fix — addressed in this push (7700dc4c)

4. **`readFileSync` throws on missing file** — added an `existsSync` guard; if the failure happens before pytest creates the output file (e.g. `pip install` fails), the issue body falls back to `(no pytest output captured — failure occurred before test run)` rather than throwing and silencing the alert.

5. **Unpaginated dedup query** — replaced the single `listForRepo` call with a `paginate.iterator` loop. The `hermes-compat` label filter keeps the page count small in practice, but this removes the structural gap.

6. **Missing label** — created the `hermes-compat` label in the repo (color `#e4e669`, description: hermes-agent PyPI compatibility canary). Without it, `issues.create` silently dropped the label and the dedup filter always returned empty. Label now exists before the first canary run.

### Items I did not address

- **Version-floor retirement check** (the installed version isn't compared against a floor to detect stagnation): agreed it's useful, but adding it requires pinning a minimum-version sentinel in the workflow — a judgment call about what the threshold is. Happy to add it in a follow-up once we agree on the value.
- **Python 3.10/3.13 matrix coverage**: noted. Added 3.13 to the matrix would be a one-line change; 3.10 is near EOL. Happy to add 3.13 if that's the direction; 3.10 omission is deliberate.

plur9 (approved):
Approving. All three blocking findings are addressed at the current head, and I verified each against the diff rather than the commit messages.

**1. Pipefail — fixed.** The step now declares `shell: bash` *and* `set -o pipefail`, with a comment explaining the distinction that caused the bug: GitHub's default `run:` shell is `bash -e {0}` without pipefail, so the step exit status was `tee`'s and always 0. Belt and braces here is right — the explicit `set -o pipefail` survives someone later removing the `shell:` line.

**2. pytest never installed — fixed.** Now `pip install -e "packages/hermes[test]"`, with a comment recording that pytest lives only in the `[test]` extra and that its absence produced `command not found` behind the swallowed exit code. That pairing was the real defect: either alone is survivable, together they made a job that executed nothing report success.

**3. Scope contamination — resolved.** The diff is 1 file (`.github/workflows/hermes-compat-monthly.yml`). The seven `pinned_priority` files are gone.

Worth noting for whoever merges: those seven files were the `pinned_priority` change from #1121/#1082, both of which I closed earlier today in favour of #1138. Had this merged with them attached, it would have landed a schema field with no implementation behind it. The review catching it as scope contamination happened to also catch that.

CI green across all six checks. The canary can now detect what it exists to detect.

PR #1121, opened 2026-09-03
feat(inject): pinned_priority field — priority-weighted eviction for pinned budget overflow

miles-on-nightshift:
## Review fixes (2026-09-04)

Pinned engrams bypass the relevance gate, so the key that orders them is a security control. Three invariants, each with adversarial tests in `packages/core/test/pinned-priority-boundaries.test.ts` and `packages/mcp/test/pinned-priority-args.test.ts`:

1. **A `pinned_priority` on a row from a pack, a `stores:` entry or a remote store can never rank that row ahead of a primary-store pin.** The pinned tier sorts by origin rank first (primary, then store, then pack, read from the loader-stamped `_storeScope` / `_pack` markers so a row cannot claim an origin it does not have), then priority, then relevance score. Packs additionally lose the field on export, on install (`sanitizePackEngrams` reports `priorityStripped`, preview warns) and at load.
2. **No value of `pinned_priority` can crash injection, be persisted as a non-finite number, or make an otherwise valid engram unloadable.** On disk, in `stores:` files and in remote rows, finite values are clamped to 1..100 and everything else is dropped (`normalizeEngramInput` rule #1121, `RemoteRowSchema`, `salvageRemoteRow` drops the field instead of the row). On the write paths (`learn`, `learnRouted`, `updateEngram`, `updateEngramAsync`, `plur_learn`) finite values are clamped and NaN / Infinity / strings are refused with a `TypeError` before anything is written; `learn` also refuses a priority without `pinned: true`. `plur_learn` declares the field with its bounds and the MCP argument validator (`jsonSchemaPropToZod` now emits finite/int/min/max for numbers) is the first gate.
3. **Pins with no priority keep their previous relative order** (relevance score; default priority 50).

One module, `packages/core/src/pinned-priority.ts`, owns the bounds, `normalizePinnedPriority`, `validatePinnedPriority`, `effectivePinnedPriority` and `pinnedOriginRank`. `spec/engram.schema.json` is regenerated and the ENGRAM-STANDARD row updated.

Answer to the review question about `_recordDuplicate`: when `learn()` hits an exact duplicate the existing engram is kept unchanged and the new call's `pinned_priority` is ignored, like every other field on the dedup path. Changing the priority of an existing pin is an explicit `updateEngram`, which now validates the field the same way.

Verification: `npx vitest run --project @plur-ai/core` on pinned-priority, pinned-priority-boundaries, inject, normalize-engram, spec-schema-drift, packs, set-pinned-remote, remote-recall, salvage-remote-row: 8 files / 190 tests pass; `@plur-ai/mcp` tools + pinned-priority-args + budget-schema: 3 files / 79 tests pass; `tsc --noEmit` clean for core and mcp. Branch merged with `origin/main` at `fba72176` and again after the architecture audit landed (`9885a243`): the audit's single `_buildEngramShape` / `_updateEngramReturning` carry the validators, no second copy.

---

## Summary

- Adds `pinned_priority` (integer 1–100, 100=highest) to the Engram schema — an eviction-priority field within the mandatory-injection (`pinned`) tier
- `fillTokenBudget` now sorts the pinned subset by `pinned_priority` desc, then `retrieval_strength` desc as tiebreaker, before the greedy token-budget loop
- Previously, eviction order within the pinned tier was undefined (YAML insertion order); a critical meta-rule near the end of the store could be silently dropped when lower-importance pinned engrams consumed the 50% sub-budget cap
- `learn()` and the remote-learn shape both stamp `pinned_priority` from `LearnContext`, clamped to [1, 100]
- Spec updated: `spec/engram.schema.json` regenerated (`pnpm gen:schemas`), `ENGRAM-STANDARD-v1.md` §4.12 table extended

## Design

Integer 1–100 over an enum (critical/high/normal) — allows fine-grained ordering when many pinned engrams coexist (e.g. three packs each contributing 5 critical engrams). Default (absent = 50) keeps existing behavior unchanged; no migration needed.

The sort runs only on the usually-small pinned subset, isolated from the unpinned scoring path. Benchmark shows latency impact within noise (<5ms variation across two runs on a loaded machine; second run showed a 1ms improvement).

## Test plan

- [x] 13 new tests in `test/pinned-priority.test.ts`:
  - Eviction order: high-priority selected, low-priority evicted under sub-budget pressure
  - Selection order: 100 → 80 → 50 → 20 descending
  - Retrieval-strength tiebreaker when priorities are equal
  - Default-50 semantics for engrams without `pinned_priority`
  - No-eviction case (budget sufficient for all)
  - Unpinned engrams unaffected by pinned sorting
  - 50% sub-budget cap invariant preserved with priority ordering
  - Schema validation: accepts [1,100], rejects [0, 101, -1]
  - `learn()` persistence to disk and recall roundtrip
  - `learn()` clamps out-of-range values to [1,100]
  - End-to-end `inject()` survival test: critical engram (p=99) survives when low-priority engrams (p=10) fill the sub-budget
- [x] All 120 existing inject/pinned/packs/spec-drift tests still pass
- [x] Benchmark: two runs vs main baseline — inject mean delta within noise (first run ↑5ms, second run ↓1ms; same loaded machine, no embeddings)

## Checklist (PR policy R1-R10)

- **R1** — Base: `main`. `git log origin/main..HEAD` shows 1 commit. Work is absent from main.
- **R2** — Description matches the diff: 7 files changed (schemas/engram.ts, types.ts, inject.ts, index.ts, spec/engram.schema.json, spec/ENGRAM-STANDARD-v1.md, test/pinned-priority.test.ts). See Files Changed tab.
- **R3** — Safety claim: sort touches only the local `pinned` sub-array inside `fillTokenBudget`. `grep -n 'pinned_priority' packages/core/src/inject.ts` → 3 lines in the sort comparator only. No write path side-effects. The `DEFAULT_PINNED_PRIORITY = 50` constant is local to the function.
- **R4** — New tests are shown to be meaningful: the eviction-order test (high vs low priority) and tiebreaker test both exercise the new sort path with tight budget constraints that prove the ordering matters.
- **R5** — Sibling sweep: `pinned` appears in `schemas/engram.ts`, `types.ts`, `inject.ts`, `index.ts` (2 write sites), `packs.ts` (sanitizer — unchanged, pack-sourced engrams still have `pinned` stripped and `pinned_priority` would be stripped too since delete in sanitizePackEngrams targets all non-standard fields). `pinned_priority` added to the same 4 writable sites.
- **R6** — No skip branches introduced.
- **R7** — No remainder items.
- **R8** — No interlocked dependencies.
- **R9** — No AI attribution in commit or PR body.
- **R10** — Verified: 120 tests pass; benchmark noise-level on two runs.

Task: org-c7c405d2d34c (spike: pinned_priority eviction algorithm)


crtahlin (dismissed):

The core change is small and correct. The pinned subset is sorted on a fresh array from `scored.filter(...)`, so the caller's array is not mutated and the unpinned pass is unaffected. The clamp is applied at **both** write sites — the local `learn()` shape and the remote-routed shape — and those are the only two places that stamp `pinned`, so the sweep is complete. The generated JSON Schema matches the Zod declaration exactly, and the drift test keeps them locked together.

The tests are load-bearing rather than decorative: inputs are passed in reverse priority order to prove sorting rather than input order, the budget is sized against `estimateTokens` so eviction actually bites, and there is an end-to-end `inject()` survival test on a temporary store.

Follow-ups: #1123, #1124, #1125

### Should fix
- `packages/core/src/inject.ts:369` — the loop after the new sort still uses `continue`, not `break`, so selection is greedy rather than strict priority order. Verified with a probe: a priority-100 engram costing more than the sub-budget is skipped while a priority-1 engram costing about 15 tokens is admitted. The guarantee written into `spec/ENGRAM-STANDARD-v1.md:420` and into the field description does not hold whenever pinned engrams differ in size — which is the same silent-drop case this change exists to fix, triggered by size rather than position.
- `packages/mcp/src/tools.ts:1000` — `pinned_priority` appears nowhere in `packages/mcp` or `packages/cli`, while `pinned` is exposed and forwarded. The field is settable only from the programmatic core API, not from the tool agents and users actually call, so the feature ships inert for its own motivating case.
- `packages/core/src/inject.ts:362` — the `retrieval_strength` tiebreaker does change behaviour for existing stores. Every already-pinned engram lacks the field and ties at 50, so the tiebreaker replaces file order with a value that decays and is mutated on every recall. Which pinned engrams survive a tight budget can now drift between sessions with no change to the store. The body's claim that the default "keeps existing behavior unchanged" is true of the priority term only.

### Notes
- `packages/core/src/packs.ts:834` — the body says `sanitizePackEngrams` "targets all non-standard fields". It deletes specifically named fields, so a pack shipping `pinned_priority` keeps it through install and export. Benign today because `pinned` is stripped, but it survives as dead attacker-controlled data.
- `packages/core/src/index.ts:2651` — the clamp uses `Math.round` behind a `!= null` guard, so a NaN from an untyped JavaScript caller would propagate and make the comparator return NaN. `Number.isFinite` would close it. Unreachable today precisely because the MCP surface does not accept the field.

### On the three-way duplication
This pull request should own `pinned_priority`. It is the only one of the three that clamps the value, the only one with every check green, and the only one scoped to just this change. [#1120](https://github.com/plur-ai/plur/pull/1120) carries a byte-identical copy bundled with an unrelated workflow and should drop everything except that workflow. [#1082](https://github.com/plur-ai/plur/pull/1082) should rebase after this merges, drop its duplicate field and specification edits, and keep only the hard-cap and two-tier parts — while carrying over its `temporal.learned_at` ascending tiebreaker, which is stable, as a replacement for the `retrieval_strength` one here.

Verified: ran vitest in an isolated clone of the head — 13 of 13 in the new test file, and 80 of 80 across pinned-priority, inject, spec-schema-drift and packs; built a probe proving the greedy-versus-strict behaviour; grepped `main` for every site that stamps `pinned` to confirm the clamp sweep. Tests: PASS.

plur9:

Issue #1125, opened 2026-09-04
Strip pinned_priority in the pack sanitiser and exporter


crtahlin:
Follow-up from [plur-ai/plur#1121](https://github.com/plur-ai/plur/pull/1121).

`sanitizePackEngrams` in `packages/core/src/packs.ts` removes host-overriding fields by name — `if ('pinned' in c) delete c.pinned`, plus the locked-commitment fields — and `exportPack` repeats the same named deletion. Neither removes `pinned_priority`, so a third-party pack can ship the field and it survives install, export, and the integrity hash.

The impact is benign today, because `pinned` is stripped and `pinned_priority` is only read for engrams where `pinned === true`. But it leaves attacker-controlled data on installed pack engrams, and it becomes live the moment the pinned strip is relaxed or a pack-pinning exception is added.

Worth noting: the pull request body asserted that the sanitiser "targets all non-standard fields". It does not — it deletes an explicit list, so this needs an explicit line.

Add `delete c.pinned_priority` next to both `delete c.pinned` sites, and extend the pack security test to assert the field does not survive install.

plur9:
Closed with the pinned priority model (#1121, #1082) — see the note there. `pinned_priority` is not being added, so there is no field to expose through the MCP surface or strip in the pack sanitiser. Pack-origin pins are handled instead by `pinnedOriginRank` (63fb20c0), which ranks them behind every primary-store pin using loader-stamped markers a row cannot forge.

Issue #1123, opened 2026-09-04
Expose pinned_priority through the plur_learn MCP tool surface


crtahlin:
Follow-up from [plur-ai/plur#1121](https://github.com/plur-ai/plur/pull/1121).

`pinned_priority` was added to the core Engram schema and to `LearnContext`, and both `learn()` write sites stamp it. But `packages/mcp/src/tools.ts` still declares and forwards only `pinned` — in the `plur_learn` input schema and again in the batch-learn shape.

The field is therefore unreachable from the tools that agents and users actually call, so the feature ships inert for its own motivating case: a critical meta-rule being silently evicted.

Add `pinned_priority` as an optional integer between 1 and 100 to both input schemas, forward it into the `LearnContext` at both call sites, and include it in the engram output shape so callers can read back what was stored. Add a test that a `plur_learn` call carrying `pinned: true, pinned_priority: 95` produces an engram with that value.

plur9:
Closed with the pinned priority model (#1121, #1082) — see the note there. `pinned_priority` is not being added, so there is no field to expose through the MCP surface or strip in the pack sanitiser. Pack-origin pins are handled instead by `pinnedOriginRank` (63fb20c0), which ranks them behind every primary-store pin using loader-stamped markers a row cannot forge.

PR #1132, opened 2026-09-04
fix(inject): strict priority ordering in pinned tier — stop on overflow, not skip (#1124)

### LINKED ISSUE #1124: Pinned budget selection is greedy, not strict priority order
Follow-up from [plur-ai/plur#1121](https://github.com/plur-ai/plur/pull/1121).

`fillTokenBudget` sorts the pinned subset by `pinned_priority` descending, but the selection loop that follows uses `continue` rather than `break` when an engram does not fit. A high-priority engram that overflows the remaining sub-budget is skipped, and lower-priority engrams that happen to be smaller are still admitted after it.

Reproduced locally: with a priority-100 engram whose token cost exceeds the pinned sub-budget and a small priority-1 engram, `fillTokenBudget` returns only the priority-1 engram.

That contradicts the guarantee written into `spec/ENGRAM-STANDARD-v1.md` — "lower-priority engrams are evicted first" — and into the field description. It is the same class of failure the feature was introduced to prevent, triggered by engram size rather than by position in the file.

Two acceptable resolutions: reword the specification and the field description to describe best-effort greedy packing, or make the loop honour the ordering, for example by stopping admission once a higher-priority engram has been skipped.

Either way, add a test using engrams of deliberately unequal size. The current suite exercises only equal-cost engrams, which is why it does not catch this.

miles-on-nightshift:
## Problem

When a high-priority pinned engram exceeds the pinned sub-budget, the selection loop used `continue`, allowing smaller lower-priority engrams to sneak in after it. This violated the `ENGRAM-STANDARD-v1.md` guarantee that lower-priority engrams are evicted first.

Reproduction (from the issue): a priority-100 engram whose cost exceeds the pinned sub-budget is skipped; a priority-1 engram that fits is admitted. `fillTokenBudget` returns only the low-priority engram.

## Fix

Change `continue` to `break` in the pinned tier loop. The array is already sorted by `pinned_priority` descending (from #1121), so once the highest remaining priority engram cannot fit, no lower-priority engram should be admitted.

Also removes the now-redundant `maxTokens` check in the pinned loop — it was always subsumed by `pinnedBudget ≤ maxTokens`.

## Test

Added a regression test with deliberately unequal-size engrams:
- `ENG-PP-LARGE` (priority 100, ~120 tokens) exceeds the pinned sub-budget
- `ENG-PP-SMALL` (priority 1, ~10 tokens) fits the sub-budget

With the greedy `continue`, only the small low-priority engram was admitted. With the strict `break`, neither is admitted.

Closes #1124.
Base: #1121 (must merge first — the sort it introduces is what makes the break meaningful).

### DIFF
--- a/packages/core/src/inject.ts
+++ b/packages/core/src/inject.ts
@@ -366,8 +366,7 @@ export function fillTokenBudget(
 
   for (const engram of pinned) {
     const cost = estimateTokens(engram)
-    if (tokensUsed + cost > maxTokens) continue
-    if (tokensUsed + cost > pinnedBudget) continue
+    if (tokensUsed + cost > pinnedBudget) break
     result.push(engram)
     tokensUsed += cost
     const pack = engram.pack ?? '__personal__'
--- a/packages/core/test/pinned-priority.test.ts
+++ b/packages/core/test/pinned-priority.test.ts
@@ -183,6 +183,58 @@ describe('pinned_priority: eviction order within the pinned tier', () => {
     )
     expect(minSelectedPriority).toBeGreaterThanOrEqual(maxEvictedPriority)
   })
+
+  it('strict ordering: a large high-priority engram that overflows the sub-budget blocks smaller low-priority ones', () => {
+    // This is the regression test for the greedy-vs-strict bug (#1124).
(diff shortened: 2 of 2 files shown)


plur9:

PR #1134, opened 2026-09-04
spike: per-engram authoring-time cap for pinned_hard tier

miles-on-nightshift:
## Summary

- Add `PINNED_HARD_PER_ENGRAM_TOKEN_CAP = 200` to `inject.ts` — per-engram write-time ceiling for hard-tier pinned engrams (ChatGPT Pattern A analogue: hard cap at save time, no runtime eviction decisions)
- Add `validatePinnedHardPerEngramCap(candidate, cap?)` exported from `inject.ts` — returns `{ ok, tokens, cap }`
- Enforce cap in `learn()` before the existing aggregate 2,000-token check: a single oversized hard-tier engram is rejected with a clear error message naming the token count, cap, and fix instruction
- Expose `pin_tier` and `pinned_priority` in `plur_learn` and `plur_learn_batch` MCP inputSchema (previously only `plur_pin` exposed `pinned_tier`)

## Spike finding: 200-token cap vs. current pinned set

At 200 tokens/engram, a full EngramSchema-parsed engram has ~88 tokens of fixed JSON overhead (`activation`, `feedback_signals`, `relations`, `tags`, etc.) leaving ~112 tokens (~448 chars) for the statement.

Results against the 13 current pinned engrams (2026-09-04):
- **Pass (≤200 tokens): 4/13** — only the shortest meta-rules survive
- **Fail (>200 tokens): 9/13** — most safety/workflow rules exceed 448 chars of statement

Implication: **200 tokens is too tight** for real-world safety rules. A 400-token cap would allow ~11/13 to qualify for hard tier while still blocking the verbose 10-rules engram (~2,000+ chars). The constant is exported and easily tuned; setting `PINNED_HARD_PER_ENGRAM_TOKEN_CAP` is a one-line change.

## Tests

- 6 unit tests in `inject.test.ts`: constant value, ok/fail cases, custom cap, boundary (exactly at cap passes; one token over fails), spike validation (short engrams pass, verbose ones fail)
- 6 integration tests in `pinned-hard-per-engram-cap.test.ts`: write-time rejection on hard tier, soft tier and unpinned are unaffected

## Test evidence

```
✓ per-engram hard-tier cap > PINNED_HARD_PER_ENGRAM_TOKEN_CAP is exported and equals 200
✓ per-engram hard-tier cap > validatePinnedHardPerEngramCap returns ok:true for a short engram
✓ per-engram hard-tier cap > validatePinnedHardPerEngramCap returns ok:false for an oversized engram
✓ per-engram hard-tier cap > validatePinnedHardPerEngramCap accepts a custom cap override
✓ per-engram hard-tier cap > engram exactly at the cap passes; one token over fails
✓ per-engram hard-tier cap > spike validation: short engrams pass 200-token cap, verbose ones fail

✓ write-time enforcement > learn() rejects a hard-tier engram that exceeds cap (10618ms)
✓ write-time enforcement > learn() rejects with a message naming the token count and cap
✓ write-time enforcement > learn() accepts a hard-tier engram within the per-engram cap
✓ write-time enforcement > learn() does NOT reject a soft-tier pinned engram with a long statement
✓ write-time enforcement > learn() does NOT reject an unpinned engram with a long statement
✓ write-time enforcement > PINNED_HARD_PER_ENGRAM_TOKEN_CAP default is 200
```

All 404 MCP tests pass. All 40 inject.test.ts tests pass (1 pre-existing timeout unrelated to this change).

## What remains

- Decide production cap value (400 tokens recommended from spike findings)
- Add `pin_tier`/`pinned_priority` fields to the `plur_pin` tool response
- Consider exposing `PINNED_HARD_PER_ENGRAM_TOKEN_CAP` as a config option for deployments with different budget needs


plur9:

PR #1135, opened 2026-09-04
fix(cli): let plur forget reach the store a namespaced id names (#1119)

plur9:
## Summary

`plur forget <namespaced-id>` never reached the store the id names. The CLI pre-checked `plur.getById(target)`, which resolves through a synchronous remote cache that nothing warms in a one-shot CLI process, so every remote engram missed and the command exited with `Engram not found` before `plur.forget()` ever ran. Core's `forget()` already routes namespaced ids (#86), walks remotes for bare ids (#84), refuses ambiguity (#831) and an unreachable declared store (#1114). The MCP `plur_forget` handler delegates on a miss; the CLI did not. Core is untouched.

Supersedes the forget half of #1122 without its `bareEngramId` regex, which rewrote any 2–4-letter id segment and broke the #831 collision guard.

### Change

- `packages/cli/src/commands/forget.ts`: on a local miss, or with an explicit `--scope`, delegate to `plur.forget()`. Nothing is caught, so core's refusals reach the entry-point error handler as themselves. The local-hit path is unchanged.
- New `--scope <scope>` flag (parity with `plur_forget`). A dangling `--scope` is a usage error; search mode accepts only `--scope primary`, because a remote scope would route a local match's id to a server that may hold an unrelated row (#831).
- Help line and README updated.

### Behaviour

| Command | Before | After |
|---|---|---|
| namespaced id, store reachable | `Engram not found`, exit 1, row stays active | row retired, exit 0 |
| namespaced id, store unreachable | `Engram not found` | `Cannot reach "group:test" to retire "ENG-GTE-…"`, exit 1 |
| bare id held only remotely | `Engram not found` | retired through core's remote walk |
| bare id nowhere | `Engram not found` | unchanged |
| `ENG-ZZZ-…`, `ENG-gte-…`, `ENG-GTE-XX-…` | — | not found; the decoy `ENG-2026-09-01-007` is untouched |

### Tests

`packages/cli/test/forget-namespaced.test.ts` (15 tests). The invariant is stated in the header: a namespaced id reaches its store or surfaces a refusal, a bare id keeps the graceful contract, no id segment is ever rewritten. Covers reachable/unreachable (including the #1069 breaker-open second call), bare nowhere / only-remote / collision (#831 refusal, then `--scope primary` vs `--scope group:test`), a typo scope, a dangling `--scope`, the search-mode guard, and adversarial id shapes against decoy rows. Runs the built CLI against the in-process stub server.

Verification: `npx vitest run --project @plur-ai/cli --project cli-spawn` 67 files, 640 passed, 4 skipped; after merging main (incl. the architecture audit) the forget suites pass (2 files, 21 tests); `pnpm typecheck:tests` clean; the #831 and #1114 tests are unchanged and pass.

Not addressed: `plur promote` and `plur rescope` still cannot address a remote engram by namespaced id (core `rescope()` is local-source by design); separate issue if wanted.

Refs #1119. Supersedes #1122.


miles-on-nightshift (approved):
Verified:

**Root cause**: confirmed that `_loadRemoteCached` is a synchronous cache peek — it returns whatever the in-process cache already holds, and nothing warms that in a one-shot CLI invocation. So every remote engram, namespaced or not, exited 'Engram not found' before `forget()` ran. The MCP handler delegates on a miss; the CLI did not.

**Fix shape**: `scope ? null : await plur.getById(target)` skips the broken pre-check when `--scope` is explicit (correct for collision disambiguation). On a local miss it delegates to `plur.forget()`, which routes namespaced ids (#86), walks remotes (#84), refuses ambiguity (#831), and surfaces unreachable stores (#1114) as themselves. Core is untouched. Traced all callers — this is the only CLI codepath; MCP and REST were not broken.

**No segment rewriting**: verified the three 'no rewriting' tests each seed a DECOY row a mangled id would have retired, then assert that row is still active. This is the specific failure mode from #1122.

**Search + scope guard**: the check refuses any scope other than `primary` in search mode — correct, because a local search hit's id could collide with a different engram in a remote store.

**Minor note (not blocking)**: `--scope --reason cause` would silently interpret `--reason` as the scope value and discard `cause`. The scope string would then fail validation in `forget()` (no configured store named `--reason`). Not a security issue; just an odd error message. Filed in case it comes up: the fix is checking `args[i+1].startsWith('--')` in the dangling guard.

All 8 CI jobs passing including the full Node 20/22/24 × vitest matrix. @plur9 ready for merge.

miles-on-nightshift:
All CI passing (dedup-calibration, smoke-packaged, test, test-matrix ×4 all green). @plur9 ready for merge.

PR #1138, opened 2026-09-07
Injection ordering, engram provenance timestamps, and write-path reporting

plur9:
All four changes come out of one incident: a session_start payload exceeded the
host's tool-result limit, was spilled to a file, and was read at 4% — after
which an agent disclosed a customer name during a live demo. Every rule it
broke was in the part it never read.

## 1. CONSTRAINTS first, and last to be dropped

`DIRECTIVES` ran 35,260 chars, putting `CONSTRAINTS` past char 35k of a 100,921-char payload. Whatever is emitted first is what a truncated read actually gets.

Worse, the dsh budget ladder shed `CONSTRAINTS` while keeping `DIRECTIVES` — and a test named `drops CONSTRAINTS next when still over budget` enforced exactly that. The suite encoded the defect instead of catching it.

Constraints now lead in all three emitters (mcp, dsh, cli) and are the last section standing. Selection is split so they fill first from a reserved 40% floor, with unused budget flowing both ways — previously "constraints" was only a label applied after one pool had been selected on task similarity, so a prohibition competed against every other engram and lost.

## 2. `created_at` / `updated_at`

The object had no canonical creation time. `sources[].stored_at` is optional — a real 5,477-engram store had it on 2,765, and `temporal.learned_at` on 12. Callers were parsing the mint date out of the id: date-only, absent for non-canonical ids, wrong after migration or store-namespacing.

Both optional and deliberately not defaulted; a default stamps the load date over the provenance being added. `updated_at` covers content and lifecycle only, and is explicitly not moved by reads, decay, injection or feedback. Spec + generated JSON Schema updated, drift guard passes.

## 3. Write-path reporting

`near_duplicates` returned `{id, score}` — so reading the neighbour cost another tool call, and it wasn't made: four near-identical engrams in one session, each reporting an unread 0.86-0.87 neighbour. It now carries the neighbour's own text.

Redrafting defeats similarity entirely (each link supersedes the last and the text genuinely differs), so `plur_learn` also flags a supersede of an id minted today.

**Neither gates a write.** The existing analysis that a cosine bar carries false-NOOP risk stands and is not revisited.

## 4. ~~`plur forget --scope`~~ — not in this diff

Removed from this description. `f7b7a066` is already an ancestor of `origin/main`
and `git diff origin/main...HEAD` touches no forget code, so this PR never
carried it. Flagged in review.

## Review round (2026-09-08)

@crtahlin requested changes and found a real regression: the section split
granted the pinned sub-budget once PER PASS, so pinned could take 150% of its
share and starve relevance-scored recall completely — 27 pinned and 0 unpinned
at a 2,000-token budget. `pinnedBudgetBase` had closed the shrink direction from
#1142 and opened the grow one. Fixed with a shared spend ledger, asserted on the
combined share.

Also addressed: the empty guard reporting no omissions in the one case where
everything was dropped; `omitted_pinned` computed but consumed nowhere (now on
the public type, both inject tools, the CLI, and a `## PINNED, NOT LOADED`
section at session start); `updated_at` never stamped on edit or pin, with the
contract now tested rather than only documented; the spec contradicting itself
about `created_at`; the pinned quota gating only one of two entry points; the
dsh ladder losing everything when constraints alone overflow, and its docstring
still describing pre-fix behaviour; the unpin suggestion accumulating inside a
`filter` predicate and under-covering the deficit; a domain test asserting
source TEXT rather than behaviour; a changelog claim that was true of dsh and
false of core; and 657 lines of sprint planning, naming a reviewer, that did not
belong in a product repo.

The render-boundary fix for #940, which touches three of the same functions and
was ready to land here, went to #1167 instead — the review's closing point was
that bundling is what hid the blocking defect.

## Tests

Full suite on the head: **350 files passed, 15 skipped, 0 failed**;
`typecheck:tests` clean. New coverage for the constraints floor, the combined
pinned share across passes, budget-degradation order, neighbour-text reporting,
and the `created_at` / `updated_at` contract. Two bug-encoding tests are inverted
rather than deleted.


plur9:
Rebased onto `main`, dropping the two commits that were never this PR's work.

`dd8b5cbf` and `16ede046` were #946's UI rebinding fixes, inherited because this
branch was cut from `rb946` rather than `main`. Left in place, merging both PRs
would have landed the same change twice — and this branch carried an *older*
revision of it than #946 now has, so the duplicate would also have been the worse
of the two.

- 20 commits → **18**; 78 files → **75**; +4305/−133 → **+4062/−119**
- `packages/cli/src/commands/ui.ts`, `packages/ui/src/server.ts` and
  `packages/ui/test/server.test.ts` no longer appear in the diff against `main`
- No remaining commit on this branch touches those three files, so the drop was
  clean rather than resolved — nothing here depended on them
- Force-pushed with `--force-with-lease`; both commits remain on #946, which is
  where that work belongs

Scope is now what the title says: injection ordering, engram provenance
timestamps, and write-path reporting.

crtahlin (dismissed):
Requesting changes — the section split multiplies the pinned sub-budget across passes, and pinned engrams can now consume the entire injection budget and starve relevance-scored recall completely.

The ordering work itself is right. Constraints lead all three emitters, the 40% floor behaves, the draft gate is thorough, and `isConstraintCandidate` mirrors the wire-time routing branch for branch. The problem is an interaction, not a mistake in either half.

### Blocking

- `packages/core/src/inject.ts:512,718,722,728` — `fillTokenBudget` is called once per section pass and each call receives `pinnedBudgetBase = maxTokens`, so the 50% pinned sub-budget is granted afresh in the constraints pass, the directives pass and the slack pass. Measured A/B on identical input at `maxTokens` 2000, with 40 pinned engrams and 40 unpinned candidates: `origin/main` selects 5 pinned and 5 unpinned; this branch selects 27 pinned and 0 unpinned, filling 1998 of 2000 tokens. No relevance-scored engram reaches the agent at all.

  This is the failure the sub-budget exists to prevent, and the function's own docstring says so: the ratio is a guard against pinned starving contextual recall and is not meant to compound once per pass. The `pinnedBudgetBase` parameter closes the shrink direction from #1142 and opens the grow direction. It is reachable in exactly the store this pull request describes — 46 pins, three times over quota — and CI is green because no test asserts the combined pinned share across passes.

  Fix: thread a shared running total of pinned tokens through the passes, or compute the pinned selection once before the split and let each pass draw from the remainder. Then assert the combined pinned cost stays within `injection_budget × injection.pinned_ratio`.

### Should fix

- `packages/core/src/inject.ts:763` — the empty-guard early return hardcodes `omitted_pinned: []`, so when nothing fits at all the omission is silent again, which is the hole #1142 was added to close. One oversized pinned engram at `maxTokens` 5 returns an empty array rather than naming it.
- `packages/core/src/types.ts:290-309` and the three emitters — `omitted_pinned` is spread into the result but not declared on the public `InjectionResult` interface, and there are no references to it in `packages/mcp/src/tools.ts`, `packages/cli/src/commands/inject.ts` or `packages/dsh/src/memory-section.ts`. It is computed, deduplicated, returned, and displayed nowhere. The stated goal that callers can see what they did not get is not met end to end.
- `packages/core/src/index.ts:5402,5496` — `_updateEngramReturning` and `setPinned` do not stamp `updated_at`, but the spec added here (`spec/ENGRAM-STANDARD-v1.md:286`) says the field tracks statement, scope, commitment, relations and retirement. Only creation and the four retirement paths stamp it, so `updated_at` will equal `created_at` for every engram that has been edited. That is worse than an absent field, because it reads as authoritative.
- `spec/ENGRAM-STANDARD-v1.md:458-462` — §4.13 still says `created_at` is not a validated field in v1 and directs implementers to `temporal.learned_at`, contradicting the §4.2 row and normative paragraph added 180 lines earlier in the same file.
- `packages/mcp/src/tools.ts:1092` — the pinned quota is enforced only in the `plur_pin` handler. `plur_learn { pinned: true }` passes `pinned` straight into `learnRouted`, and `Plur.setPinned()` is unguarded. Writing a new pinned engram is a normal way to create a pin, so the quota covers one of two entry points. That matters more given the blocking finding, since the injection-time cap no longer binds either.
- `packages/dsh/src/memory-section.ts:63` — the `renderBlock` docstring still documents the pre-fix order, that sections are dropped least important first, `consider` then `constraints`. That is the inverted behaviour and it is the safety-critical claim of the whole change. Separately, the ladder has no directives-only rung, so a store whose constraints alone exceed the dsh budget now gets no memory where it previously got its directives.

### Notes

- `packages/mcp/src/tools.ts:1671-1676` — the unpin suggestion accumulates `freed` via a side effect inside a `filter` predicate, then truncates to five. When more than five unpins are needed, the returned list does not cover the deficit while the note tells the user to unpin one of the suggestions.
- The `.plur.yaml` domain test asserts behaviour by reading `src/tools.ts` and matching a source string. That pins text, not semantics: a behaviour-preserving refactor fails it and a text-preserving behaviour change passes it.
- Nothing asserts the `created_at` / `updated_at` contract — not that a new write stamps both, not that a legacy engram loads with both absent, not that reads, decay, injection and feedback leave `updated_at` alone. The exclusion list is documented in three places and verified in none. Checking the mutation sites by hand is how the `updateEngram` gap above surfaced.
- The changelog says constraints are never traded for budget. That holds for the dsh whole-section ladder but not for core's per-engram selection, where constraints get a 40% floor plus slack rather than absolute priority. With 40 fat constraints and 40 fat directives at `maxTokens` 2000 the result is 11 constraints and 10 directives. Worth wording the claim to match.
- Item 4 of the description, `plur forget --scope`, is not in this diff — `f7b7a066` is already an ancestor of `origin/main` and `git diff origin/main...HEAD` touches no forget code.
- 657 lines of sprint planning are added under `sprints/`, including internal review commentary that names a reviewer by first name. Both are worth reconsidering in a product repository.
- The bundling is what hid this. The blocking defect exists only where #1142's `pinnedBudgetBase` meets this change's section split — two changes that as separate pull requests would each have looked obviously correct.

Worth keeping: `pinnedOriginRank` tests `_pack` before `_storeScope`, so a row shipping its own marker can only demote itself, and `_storeScope` is stamped at exactly the two loader clone sites — the security claim in that docstring holds. The draft-approval gate test covers the personal path, the pack path, reachability through spreading activation, and both negative guarantees. The modified dsh test was not merely inverted; a positive assertion was added alongside it so both halves of the new ladder are pinned. And the `pinnedQuota` sort keeps the measurement that made the first version wrong, rather than discarding it.

Tests: PASS — `pnpm install`, `pnpm build`, `pnpm test` on `9ab8de99`: 349 files passed, 15 skipped, 0 failed. Verified: the pinned-starvation regression was reproduced by running the same input through `origin/main` and this branch and comparing selections; the empty-guard and oversized-pinned cases were exercised directly; the `updated_at` gap was found by walking every mutation site rather than from the test suite.

plur9:
All findings addressed. The blocking one was a real regression and the diagnosis
is exactly right — `pinnedBudgetBase` closed the SHRINK direction from #1142 and
opened the GROW one. Same parameter, read from opposite ends, and I only
measured the direction I was fixing.

**Blocking — pinned starvation.** One shared spend ledger now threads through
the three passes drawing on `maxTokens`; the DIP-19 consider pass keeps its own,
because it draws on its own budget. Two assertions added on the COMBINED share,
since per-pass assertions are precisely what missed it. Reverting the ledger
reproduces your numbers: `expected 1728 to be less than or equal to 1000`, and
`expected 0 to be greater than 0` for the unpinned count.

**Empty guard.** No longer hardcodes `[]`. The dedupe is extracted to
`dedupeOmitted` so the guard and the normal return cannot disagree again.

**`omitted_pinned` end to end.** Declared on the public `InjectionResult`, and
now actually consumed: both inject tool results, the CLI (JSON and text), and —
the one that matters — a `## PINNED, NOT LOADED` section in the session_start
payload the agent reads. A pinned rule the budget dropped is where silence costs
most, because the user believes a standing rule is loaded.

**`updated_at`.** Stamped in `_updateEngramReturning` and `setPinned`. New
`provenance-timestamps.test.ts` asserts both halves of the contract, including
that reads, feedback and injection leave it alone, and that a legacy engram
loads with both fields ABSENT rather than synthesised. Both mutation tests
confirmed red without the stamps. You were right that walking the sites by hand
was the only thing that would have found this — nothing tested it.

**Spec §4.13.** Rewritten. It said `created_at` was not validated 180 lines
after the row that made it first-class.

**Quota on the second entry point.** `plur_learn { pinned: true }` is now gated.
Necessarily coarser than `plur_pin`: the engram does not exist yet, so its
rendered cost is unknowable and this cannot say "over by N" — what it can say
without guessing is that there is no room at all, which is the state the reported
store was in. The statement is not stored, and the refusal says so.

**dsh ladder and docstring.** The docstring described pre-fix behaviour; fixed.
On the missing rung you are right that losing everything is a regression — but
emitting directives silently is the failure the inversion existed to prevent. So
neither: the block now DECLARES the withholding (`WITHHELD … treat as UNREAD`)
and lets the directives through. The old `toBe('')` test is updated rather than
deleted, because the property it protected was never silence, it was that
prohibitions are not dropped without saying so.

**Unpin suggestions.** No more accumulation inside a `filter` predicate. Takes
entries until the deficit is covered, reports `unpins_needed`, and says
explicitly when more are required than are listed.

**`.plur.yaml` domain test.** Replaced. Now driven through the handler against a
real `.plur.yaml`, asserting the stored engram's domain and that an explicit
argument wins. Your framing was the useful part: a text-preserving behaviour
change passing is the failure mode that matters here, because the defect it
guards was a value parsed and never consumed.

**Changelog.** The "never traded for budget" claim was wrong for core and I have
reworded it to the two different guarantees — dsh sheds whole sections and keeps
constraints last, core gives them a 40% floor plus slack. Your 11-and-10 example
is in the text.

**`plur forget --scope`.** Correct, `f7b7a066` is already an ancestor of `main`.
Removed from the description.

**657 lines of sprint YAML.** Also correct, and the reviewer's name with it.
Moved out to the space repo rather than deleted.

On the closing note — that the bundling is what hid it — I have taken that as
binding rather than as a remark. The #940 render-boundary fix, which touches
three of the same functions and was ready to go on this branch, is #1167
instead.

Full suite green: 350 files passed, 15 skipped, 0 failed. `typecheck:tests`
clean.

Issue #1153, opened 2026-09-08
Decide the twelve engram fields the remote write contract does not model — summary and contraindications first


plur9:
Follow-up from #1151. Fixing that issue meant classifying every field of
`EngramSchema` against the remote write contract, and twelve came out as
"nobody has decided" rather than "deliberately local":

```
visibility  contraindications  knowledge_type  entities  episodic  exchange
structured_data  insight  polarity  locked_at  sources  summary
```

They are held in a `NOT_MODELLED` set in
`packages/core/test/remote-write-contract.test.ts`, separate from `LOCAL`, so
the distinction stays visible: `LOCAL` means we decided, `NOT_MODELLED` means we
have not.

## Two of these are the same defect as #1151

Not merely absent — **rendered to the model**:

- **`summary`** is the *entire* payload of a layer-1 entry. `formatLayer1()`
  renders `summary ?? statement.slice(0, 60)`, so an engram that reaches a
  teammate without its summary is silently re-rendered as a truncated
  statement — a different claim, presented with equal authority.
- **`contraindications`** is the "does NOT apply when …" clause. It is rendered
  at every layer that carries a constraint. A qualified prohibition that loses
  its qualification on the way to a shared store arrives as an unconditional
  one. That is exactly the shape of the `measured_under` loss in #1151, reached
  through a different field.

The remaining ten are likelier to be genuinely local or genuinely server-owned,
but none of them has been argued.

## What this issue asks for

Decide each of the twelve, and move it to `TRANSMITTED` (with the corresponding
addition to `appendAndGetServerId()`) or to `LOCAL` with a reason. Coordinate
server support where a field needs it. `summary` and `contraindications` should
be looked at first — they are user-visible today.

The guard test already fails on any *new* unclassified field, so this is a
backlog to burn down, not a hole that keeps growing.

PR #1154, opened 2026-09-08
Four defects from the 2026-09-07 audit: remote unpin, validity instants, dropped metadata, unbounded body reads

### LINKED ISSUE #1149: setPinned(id, false) sends an empty PATCH, so unpinning a remote engram silently does nothing
Reproduced against `main` at `c036d750` on 2026-09-08, using the public
`Plur.setPinned()` method over real HTTP transport (loopback server, real
serializer and response parser — nothing stubbed).

## What happens

`setPinned(id, false)` against a remote-backed scope transmits `{}`. The server
applies ordinary partial-update semantics, changes nothing, and the call returns
the still-pinned engram without an error.

```json
{
  "patchBody": {},
  "returnedPinned": true,
  "persistedPinned": true,
  "explicitFalseControl": false
}
```

`explicitFalseControl` is the same driver's `patch(id, { pinned: false })`
against the same server; it unpins the row. So the transport and the server are
fine — the loss is in the payload the high-level method builds.

## Why

`packages/core/src/index.ts:5540`

```ts
const patched = await driver.patch(serverId, { pinned: pinned === true ? true : undefined })
```

`undefined` is not serialized by `JSON.stringify`, so the property never reaches
the wire. The same coercion appears at `index.ts:5509` for the local branch,
where it is harmless because the object is written whole rather than as a patch.

## Impact

A user who unpins a standing instruction on a team store is told it worked. It
did not, on any other machine or in any other session. Given the pinned set is
now quota-enforced at pin time, an unremovable pin also consumes budget nobody
can reclaim.

## Why the existing test does not catch it

`packages/core/test/set-pinned-remote.test.ts` asserts the `undefined` value and
stubs `driver.patch()` to return an already-unpinned engram. It never exercises
serialization or reads server state back, so it passes with the defect present.

## Acceptance criteria

- `setPinned(id, false)` transmits an explicit unpin under the server's PATCH
  contract.
- An HTTP-level test starts from a pinned row, unpins through `setPinned()`, and
  asserts both the transmitted body and a **fresh server read**.
- `setPinnedAsync()` is covered (it delegates here), and a pinning control is
  retained so the fix cannot pass by transmitting everything unconditionally.

### LINKED ISSUE #1150: Validity windows compare RFC 3339 instants lexically against a date — expired engrams inject, active ones are hidden
Reproduced against `main` at `c036d750` on 2026-09-08 with schema-valid stored
records, through `list()`, `recall()` **and** `selectAndSpread()` in hard-expiry
mode. Only the clock was faked; the store was a real temporary YAML store.

## What happens

Validity filters compare a timestamp **string** against today's **date** string.
`'2026-09-07T01:00:00Z' > '2026-09-07'` is true lexically, whatever the time of
day, so an instant-form window is read backwards.

Clock fixed at `2026-09-07T12:00:00Z`:

| Record | Field | Expected | Observed |
| --- | --- | --- | --- |
| Active instant | `valid_from: 2026-09-07T01:00:00Z` | eligible | **hidden** |
| Expired instant | `valid_until: 2026-09-07T01:00:00Z` | hidden | **eligible** |
| Date-only start control | `valid_from: 2026-09-07` | eligible | eligible |
| Date-only expiry control | `valid_until: 2026-09-06` | hidden | hidden |

Identical wrong pair in all three paths:

```json
{
  "expected": ["ENG-active-date-control", "ENG-active-instant"],
  "listIds":  ["ENG-active-date-control", "ENG-expired-instant"],
  "recallIds":["ENG-active-date-control", "ENG-expired-instant"],
  "injectIds":["ENG-active-date-control", "ENG-expired-instant"]
}
```

`list({ include_expired: true })` returns all four, so this is filtering, not
ingestion loss.

## Why

Four lexical comparisons, no shared evaluator:

- `packages/core/src/index.ts:3949-3950` — `_applyResidualFilters()`
- `packages/core/src/index.ts:4513-4514` — `_filterEngrams()`
- `packages/core/src/inject.ts:111-114` — injection validity, including the
  grace-window cutoff

## Impact

Both directions are live-fire. A rule that has just come into force stays
invisible for the rest of the UTC day, and an instruction that expired hours ago
still enters the agent's context — the second is the one that matters, because
an expired instruction reads exactly like a current one.

## Contract mismatch to settle at the same time

`spec/ENGRAM-STANDARD-v1.md` (value-type rules) requires accepting **both** date
and RFC 3339 instant forms, and `TemporalSchema` in
`packages/core/src/schemas/engram.ts` does accept both. But `resolveValidity()`
in `packages/core/src/expiry.ts` restricts *new learn parameters* to
`YYYY-MM-DD`. So the read path accepts a form the write path refuses, and
mishandles it. Whatever is decided, the standard, the schema and the validator
should agree.

## Acceptance criteria

- **One** validity evaluator, shared by retrieval and injection — the current
  duplication is why the same bug exists in three places.
- Instants compared as instants, including offsets; documented day-based
  behaviour retained for date-only values.
- Same interpretation for soft-expiry markers and grace windows.
- Regression cases: both rows above, the date-only controls, equivalent instants
  written with different offsets, and the start/expiry boundary instants.

### LINKED ISSUE #1151: Remote append silently drops measured_under, knowledge_anchors and dual_coding — and returns them anyway
Reproduced against `main` at `c036d750` on 2026-09-08 through `learnRouted()`
over real HTTP transport, with a fresh `getById()` from a separate client.

## What happens

`learnRouted()` accepts `measured_under`, `knowledge_anchors` and `dual_coding`,
returns them, and never sends them. The POST carried four keys:

```json
["statement", "scope", "type", "commitment"]
```

| Field | `learnRouted()` returned | POST body | Fresh remote read |
| --- | --- | --- | --- |
| `measured_under` | complete object | absent | absent |
| `knowledge_anchors` | 1 anchor | absent | 0 anchors |
| `dual_coding` | worked example | absent | absent |

A local `learn()` + `getById()` control preserves all three, so this is specific
to the remote write path. The loopback server persisted exactly what it received
and applied no field filter of its own.

## Why

`RemoteStore.appendAndGetServerId()`
(`packages/core/src/store/remote-store.ts:544`) builds the body from an
**explicit allowlist**: statement, scope, domain, type, tags, pinned, rationale,
commitment, valid_from, valid_until, supersedes, locked_reason, source,
provenance. Anything not named is dropped.

The acknowledgement hides it: `learnRouted()` returns the local shape plus the
server-assigned id, not an authoritative persisted record, and the driver's
optimistic cache keeps the original fields — so a read-back in the same session
can still show what was never stored.

## Impact

`measured_under` exists precisely so a measured claim carries the conditions
that make it true. Stripping it on the way to a **shared** store is the worst
case for this field: a result that held on one machine, one dataset, one day
arrives at a teammate as an unconditional fact. That is the failure mode #869
was closed to prevent.

## The structural problem, not just the missing fields

This is the second occurrence. #768 (closed 2026-07-31) fixed exactly this class
for pinned/rationale/tags/commitment/validity/supersedes. The allowlist has to be
extended by hand every time the schema gains a field, and **nothing fails when
someone forgets** — which is why #869's field was lost the moment it shipped.

Adding three names to the list fixes today and guarantees a third occurrence.

There is already a precedent in this repo for the guard that prevents it:
`packages/migrate/test/method-list.test.ts` fails when `Plur` gains an async
method that is not explicitly classified, forcing a decision rather than letting
coverage drift. The same shape applies here — a test that enumerates
`EngramSchema`'s fields and fails on any field not classified as either
transmitted or deliberately local, with a reason.

## Acceptance criteria

- `measured_under`, `knowledge_anchors` and `dual_coding` survive a remote write
  and a fresh GET from a new client (not the optimistic return value).
- A checked mapping between schema fields and the wire contract, so a newly
  added field cannot be silently dropped.
- Same contract applied to outbox retry and every other
  `appendAndGetServerId()` caller.
- Where the server genuinely cannot accept a field, the loss is reported rather
  than acknowledged.

### LINKED ISSUE #1152: fetchBounded() clears its deadline before the body is read, so a stalled response outlives the 30s timeout
Reproduced against `main` at `c036d750` on 2026-09-08 with a **real 31-second
wait** — no fake timers, no mocked fetch.

## What happens

`fetchBounded()` clears its abort timer when `fetch()` resolves. `fetch()`
resolves when response **headers** arrive. Every caller then reads the body
outside any deadline.

A server that sends headers and stalls the body keeps the call pending
indefinitely:

```json
{
  "configuredDeadlineMs": 30000,
  "observationMs": 31008,
  "settledAfter31Seconds": false,
  "succeedsOnlyAfterBodyReleased": true
}
```

The promise settles only once the body is released — at any later time.

## Why

`packages/core/src/store/remote-store.ts:264-288`

```ts
const timer = setTimeout(() => ctrl.abort(), LOAD_FETCH_TIMEOUT_MS)
try {
  const res = await fetch(url, { ...init, signal: ctrl.signal })
  ...
  return res            // headers only
} finally {
  clearTimeout(timer)   // deadline gone before anyone reads the body
}
```

`me()` (`:365`) is the executed case. Append (`:584`), `getById()` (`:630`) and
PATCH (`:712`) all read `r.json()` or `r.text()` after this helper returns, so
they are exposed by the same construction — that part is source-confirmed rather
than executed. Error paths are exposed too: `sanitiseResponseBody(await
r.text())` on a non-ok response has the same gap.

## Impact

Affected remote calls fall back to whatever the runtime or the network
eventually does, instead of PLUR's own 30-second bound. A slow-loris peer — or a
proxy that flushes headers early and then stalls — holds the caller open. The
probe did not measure a resulting writer-lock outage and this report does not
claim one.

## This is a recurrence

#525 (closed 2026-07-14) found and fixed exactly this pattern in
`RemoteStore.load()`, whose body read is now inside its timer. The fix was
applied to the one method rather than to the shared helper, so every other
caller kept the gap. Fixing `fetchBounded()` itself is what makes the class
closed rather than the instance.

## Why existing tests do not catch it

`packages/core/test/remote-exists-by-id.test.ts:141-167` makes **fetch itself**
hang until aborted. No test sends successful headers and then stalls the body,
which is the only shape that reproduces this.

## Acceptance criteria

- Cancellation stays active through response-body consumption — e.g. a helper
  owning both fetch and decode, or an explicit lifetime the caller cannot drop.
- Stalled-body tests for **both** a 200 whose JSON never completes and a non-ok
  response whose `text()` never completes.
- The operation settles at the deadline, and timers and connections are cleaned
  up.
- The existing never-sends-headers test is retained.

plur9:
Closes #1149. Closes #1150. Closes #1151. Closes #1152.

Four defects from the 2026-09-07 external audit. One commit each, bisectable,
each with a regression test verified to fail without its fix.

**All four were re-confirmed against current `main` (`c036d750`) before filing** —
the audit ran against `1557e14f` and `main` had moved four merges since, so
nothing here is a stale finding.

| | Defect | Impact |
| --- | --- | --- |
| #1149 | `setPinned(id, false)` PATCHes `{}` | an unpin the user is told succeeded never happened |
| #1150 | validity instants compared lexically | expired instructions inject; live ones stay hidden |
| #1151 | remote append drops `measured_under` et al. | a conditional result reaches a teammate as unconditional |
| #1152 | deadline cleared before the body is read | a stalled body outlives the 30s bound indefinitely |

### Two of these are recurrences, so two commits are about the guard, not the fix

**#1151 is the second time** (after #768, July) that a schema field was silently
dropped from the remote write allowlist, because the allowlist is extended by
hand and nothing fails when someone forgets. Adding three names fixes today and
guarantees a third occurrence, so the substance is
`remote-write-contract.test.ts`: every field of `EngramSchema` must be
classified TRANSMITTED / FLATTENED / LOCAL / NOT_MODELLED, and a new field that
is none of them fails the suite. Modelled on
`packages/migrate/test/method-list.test.ts`.

That classification surfaced **#1153** — twelve fields nobody has decided on,
two of which (`summary`, `contraindications`) are rendered straight to the
model. `NOT_MODELLED` is deliberately kept separate from `LOCAL`: one means we
decided, the other means we have not.

**#1152 is the second time too** (after #525, July). That fix gave `load()`'s
body read its own timer — closing the instance and leaving the shared helper
alone, so the same gap sat in six other callers. `fetchBounded()` now owns the
read and returns a body that is already consumed, so no caller can be handed a
live one after the timer is gone.

### Something the new tests caught in my own fix

`readBounded` tolerates a malformed payload by returning `json: undefined`
rather than throwing. On its own that **swallowed the abort**: a stalled `/me`
would have returned zero scopes, silently, instead of timing out — worse than
the bug being fixed, because it is wrong data rather than a hang. The abort
signal is now authoritative over the shape of what came back.

### Scope notes

- `#1150` replaces four hand-written copies of one comparison with a single
  evaluator. Date-only values keep their documented whole-day semantics — the
  reason `valid_until: 2026-09-06` closes at the END of the 6th, not at
  midnight; a "parse as midnight" fix would expire it a day early. Controls for
  that are in the suite.
- Deliberately **not** fixed here: `resolveValidity()` still restricts new
  `learn()` parameters to `YYYY-MM-DD`, so the read path accepts a form the
  write path refuses. That is a contract question for the standard, not a
  filtering bug.
- Host-down/breaker semantics in `fetchBounded` are unchanged: only the fetch
  phase marks a host down. A 5xx or a decode failure says nothing about
  reachability.

### Verification

- Full monorepo suite on the branch head: **352 files passed, 15 skipped, 0 failed.**
- Each fix reverted in turn to confirm its test fails: `expected {} to deeply
  equal { pinned: false }` (#1149), `expected undefined to deeply equal {…}`
  (#1151), stalled-body timeout (#1152).
- The audit's own probe, which asserts the *buggy* behaviour, now aborts on its
  first assertion against this branch — the intended signal that the finding is
  closed.

### Note for whoever merges

`#1138` adds `created_at` / `updated_at` to `EngramSchema`. The new guard will
fail for whichever of the two lands second, until those two fields are
classified. That is the guard doing its job rather than a conflict — the fix is
two lines in `TRANSMITTED` or `LOCAL` with a reason.

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -47,6 +47,7 @@ import type { TensionRecord, TensionStatus } from './schemas/tension.js'
 import type { TensionPair } from './tensions.js'
 import { engramDate } from './tensions.js'
 import { resolveValidity, buildTemporal, normalizeIsoDate, type ResolvedValidity } from './expiry.js'
+import { isCurrentlyValid } from './validity.js'
 import { decodeJwtExpiry, decodeJwtPayload } from './jwt.js'
 import { RemoteStore, normalizeEndpointUrl } from './store/remote-store.js'
 import {
@@ -3944,12 +3945,11 @@ export class Plur {
   private _applyResidualFilters(engrams: Engram[], options?: RecallOptions & { include_expired?: boolean }): Engram[] {
     let out = engrams
     if (!options?.include_expired) {
-      const today = new Date().toISOString().slice(0, 10)
-      out = out.filter(e => {
-        if (e.temporal?.valid_until && e.temporal.valid_until < today) return false
-        if (e.temporal?.valid_from && e.temporal.valid_from > today) return false
-        return true
-      })
+      // #1150: instants compared as instants. The lexical form this replaces
+      // read `valid_until: 2026-09-07T01:00:00Z` as still valid at noon that
+      // day, and a `valid_from` of the same shape as not yet reached.
+      const nowMs = Date.now()
+      out = out.filter(e => isCurrentlyValid(e.temporal, nowMs))
     }
     if (options?.min_strength !== undefined) {
       out = out.filter(e => e.activation.retrieval_strength >= options.min_strength!)
@@ -4508,12 +4508,9 @@ export class Plur {
     // with learn()'s content-hash gate (which ignores temporal validity,
     // e.g. the migration import engine, #441) must see the full active set.
     if (!options?.include_expired) {
-      const today = new Date().toISOString().slice(0, 10)
-      engrams = engrams.filter(e => {
-        if (e.temporal?.valid_until && e.temporal.valid_until < today) return false
-        if (e.temporal?.valid_from && e.temporal.valid_from > today) return false
-        return true
-      })
+      // #1150: one evaluator, shared with _applyResidualFilters and injection.
+      const nowMs = Date.now()
+      engrams = engrams.filter(e => isCurrentlyValid(e.temporal, nowMs))
     }
     if (options?.min_strength !== undefined) {
       engrams = engrams.filter(e => e.activation.retrieval_strength >= options.min_strength!)
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -5537,7 +5537,23 @@ export class Plur {
         // The justification was that `setPinned` had to keep a synchronous
         // signature. It is `async` since the 0.16 flip, so that reason is gone
         // and the honest version costs nothing.
-        const patched = await driver.patch(serverId, { pinned: pinned === true ? true : undefined })
+        // Send the BOOLEAN, including an explicit `false` (#1149).
+        //
+        // This read `pinned === true ? true : undefined`, mirroring the local
+        // branch above — but the two representations exist for opposite
+        // reasons. Locally the engram is rewritten WHOLE, so `undefined`
+        // drops the key and keeps unpinned rows out of the YAML. Here the
+        // object is a PARTIAL update, and `JSON.stringify` omits `undefined`,
+        // so the unpin left as `{}` — a server applying ordinary PATCH
+        // semantics changed nothing and returned the still-pinned row, which
+        // this method then reported as success.
+        //
+        // Measured on a loopback server against the real serializer: PATCH
+        // body `{}`, engram still pinned afterwards, no error raised. An
+        // unpin the user was told had worked had not happened on any other
+        // machine — and with the pinned set now quota-enforced at pin time,
+        // it also held budget nobody could reclaim.
+        const patched = await driver.patch(serverId, { pinned })
         if (patched) return patched
       } catch (err) {
         if (serverId !== id) throw err
--- a/packages/core/src/inject.ts
+++ b/packages/core/src/inject.ts
@@ -6,6 +6,7 @@ import { classifyPolarity } from './polarity.js'
 import { computeConfidence } from './confidence.js'
 import { freshTailBoost } from './fresh-tail.js'
 import { makeVisibilityPredicate } from './scope-util.js'
+import { isNotYetValid, isExpired, isExpiredBeyondGrace } from './validity.js'
 
 /**
  * D1-RECALL/INJECT-ASYMMETRY (#353). When an inject is given an EXPLICIT
@@ -99,19 +100,24 @@ const DEFAULT_GRACE_DAYS = 30
 /**
  * True when the engram must be skipped for temporal validity. Not-yet-valid
  * engrams (`valid_from` in the future) are always skipped; expired engrams
- * are skipped in hard mode, and in soft mode once past the grace cutoff.
+ * are skipped in hard mode, and in soft mode once past the grace window.
+ *
+ * Delegates to `validity.ts` (#1150). This used to compare timestamp STRINGS
+ * against today's DATE string, which read every RFC 3339 instant backwards —
+ * so an engram that expired hours ago was still injected, and one that became
+ * valid hours ago was not.
  */
 function skipForValidity(
   engram: Engram,
-  today: string,
+  nowMs: number,
   mode: 'hard' | 'soft',
-  graceCutoff: string,
+  graceDays: number,
 ): boolean {
   const t = engram.temporal
-  if (t?.valid_from && t.valid_from > today) return true
-  if (t?.valid_until && t.valid_until < today) {
+  if (isNotYetValid(t, nowMs)) return true
+  if (isExpired(t, nowMs)) {
     if (mode !== 'soft') return true
-    if (t.valid_until < graceCutoff) return true
+    if (isExpiredBeyondGrace(t, nowMs, graceDays)) return true
   }
   return false
 }
@@ -120,10 +126,14 @@ function skipForValidity(
  * "⚠ EXPIRED <date> — verify before use: " prefix for an engram whose
  * `valid_until` is in the past. Only soft-expiry mode lets expired engrams
  * reach the formatters, so in hard mode this never fires.
+ *
+ * Uses the same evaluator as the filter that let it through (#1150). When these
+ * disagreed, a soft-mode engram inside its grace window could be delivered
+ * WITHOUT the marker that is the entire point of soft mode.
  */
 function expiredMarker(engram: WireEngram): string {
   const until = engram.temporal?.valid_until
-  if (until && until < new Date().toISOString().slice(0, 10)) {
+  if (until && isExpired(engram.temporal, Date.now())) {
     return `⚠ EXPIRED ${until} — verify before use: `
   }
   return ''
@@ -406,10 +416,9 @@ export function selectAndSpread(
   const promptWords = new Set(promptLower.split(/\W+/).filter(w => w.length > 2))
   const maxTokens = ctx.maxTokens ?? DEFAULT_MAX_TOKENS
   const minRelevance = ctx.minRelevance ?? DEFAULT_MIN_RELEVANCE
-  const today = new Date().toISOString().slice(0, 10)
+  const nowMs = Date.now()
   const expiryMode = config?.expiry?.mode ?? 'hard'
   const graceDays = config?.expiry?.grace_days ?? DEFAULT_GRACE_DAYS
-  const graceCutoff = new Date(Date.now() - graceDays * 86400000).toISOString().slice(0, 10)
 
   // Step 0: Build engram map for spreading activation.
   // `nonActiveIds` tracks personal engrams that exist locally but are not active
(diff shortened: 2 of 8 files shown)


crtahlin (approved):
Approved. Four gaps are worth fixing, none of them blocking — filed as follow-ups.

The four fixes are correct. `setPinned(id, false)` now sends an explicit `pinned: false` instead of an `undefined` that `JSON.stringify` dropped; validity windows are compared as instants through the new shared `validity.ts` evaluator; `appendAndGetServerId` transmits the three dropped fields; and `fetchBounded` was restructured to own the body read, so the 30-second deadline covers it.

Reshaping `fetchBounded` to take a consumer and return a value, rather than handing back a live `Response`, is the right structural fix, and all seven callers were converted with no partial conversion left behind. Treating the abort signal as authoritative over the shape of what came back is load-bearing: deleting that single line makes both stalled-body tests fail and a stalled `/me` return zero scopes silently. The loopback suite in `set-pinned-remote.test.ts` asserting on the request off the wire, rather than on the JavaScript object, is exactly what would have caught the original defect.

Follow-ups: #1155, #1156, #1157, #1158

### Blocking nothing, but worth knowing

- `packages/core/src/store/remote-store.ts:747-777` — the #1152 defect survives in `existsById()`. Its `clearTimeout` sits in a `finally` wrapping only the `fetch`, so `await r.json()` runs after the deadline is gone. With headers delivered and the body stalled it had still not settled 120 seconds past its 30-second bound. Both callers run it inside the primary store lock, where a hang makes waiting `plur_learn` calls throw and the engram is silently never stored. The new test asserts this method is bounded, but its `stalling` helper only stalls the handshake — the same blind spot this pull request identifies as the reason the gap survived elsewhere. Pre-existing, not a regression, but it does contradict the structural claim. Filed as #1155.
- `packages/core/src/tensions.ts:344-346` — the #1150 lexical comparison survives here, the one site not converted. Smaller blast radius, since it affects tension pairing rather than injection. Filed as #1156.
- `packages/core/src/validity.ts:53,66` — `Date.parse` on an offset-less timestamp resolves in local time. `TemporalSchema` accepts any string and `salvageRemoteRow` copies server values verbatim, so a store emitting naive timestamps makes validity decisions machine-dependent. Filed as #1157.
- `packages/core/test/remote-write-contract.test.ts` — the classification guard works, but only on set membership. Deleting the three field spreads from the POST body leaves it green; only the hand-written loopback assertions catch that. The next occurrence of the #768 / #1151 defect can still pass the guard built to prevent it. Filed as #1158.

Two smaller things, no issue filed: `remote-exists-by-id.test.ts:178-183` says "exactly two remain by design" then names three and asserts `toBeLessThanOrEqual(3)`, and the exemption is wrong for one of them — `load()`'s body read is inside its timer, `existsById`'s is not. The same file's header credits this branch with adding the `existsById` timeout, which predates it.

Tests: PASS — full `packages/core` suite, 2987 passed / 34 skipped / 0 failed across 195 files on the head commit. Verified: each of the four fixes was individually reverted and its regression test confirmed to go red; pre-pull-request `index.ts` and `inject.ts` fail all three validity integration tests; adding a throwaway field to `EngramSchema` fails the classification suite as intended; deleting the abort-authoritative line in `fetchBounded` fails both stalled-body tests; degrading `opensAt` to date truncation turns the instant cases red while the date-only controls stay green.

PR #1163, opened 2026-09-08
Audit review follow-ups: bound existsById, the fifth validity site, UTC-pin naive timestamps, and make the write-contract guard bite

### LINKED ISSUE #1155: Bound the response body read in RemoteStore.existsById
`existsById()` (`packages/core/src/store/remote-store.ts:747-777`) clears its AbortController timer in a `finally` that wraps only the `fetch` call, then reads `await r.json()` outside it. A peer that delivers headers and stalls the body holds the call open indefinitely — reproduced with a fake-timer probe that never settled 120 seconds past the 30-second bound.

Both callers (`forget()` and `feedback()`) run this inside the primary store lock, one probe per configured remote. The method's own doc comment explains that exceeding the 180s `DEFAULT_ACQUIRE_TIMEOUT` makes waiting `plur_learn` calls throw "Failed to acquire lock", with the engram silently never stored.

#1152 fixed this shape for the seven `fetchBounded` callers but left this one. The new `remote-exists-by-id.test.ts` asserts the method is bounded, but its `stalling` helper only stalls the handshake — exactly the blind spot the pull request itself identifies as the reason the gap survived in six other callers.

Fix: either route `existsById` through `fetchBounded` with a custom consumer, or move its `clearTimeout` after the body read. Add a stalled-body test modelled on the `stallingBody` helper already in `remote-exists-by-id.test.ts`.

Follow-up from plur-ai/plur#1154.

### LINKED ISSUE #1156: Convert tensions.ts validityExpired to the shared validity evaluator
`validityExpired()` at `packages/core/src/tensions.ts:344-346` still compares `temporal.valid_until` lexically against a `YYYY-MM-DD` string (`now`, derived at line 520).

This is the fifth instance of the defect fixed in #1150, and the only site pull request #1154 did not convert. An RFC 3339 `valid_until` that lapsed hours ago sorts as greater than today's date, so `getCandidatePairsDetailed` treats an expired engram as live when deciding tension pairs.

Fix: replace the body with `isExpired(e.temporal, nowMs)` from `validity.ts` and change the `now` option to epoch milliseconds, keeping the existing date-string option path or converting its callers.

Follow-up from plur-ai/plur#1154.

### LINKED ISSUE #1157: Reject or normalize temporal timestamps that carry no UTC offset
`opensAt` and `closesAt` in `packages/core/src/validity.ts` (lines 53 and 66) hand any non-date-only value straight to `Date.parse`, which resolves an offset-less form such as `2026-09-07T01:00:00` in the host's local time zone. That is `2026-09-06T23:00:00Z` in Europe/Ljubljana, and shifts by up to fourteen hours elsewhere.

`TemporalSchema` declares these fields as plain optional strings, and `reshape` / `salvageRemoteRow` (`remote-store.ts:402-415`) copy server-supplied `valid_from` / `valid_until` verbatim. So a remote store that stores naive timestamps produces validity decisions that differ per machine. The docstring assumes an offset is included, but nothing enforces it and no test covers the case.

Fix: either tighten `TemporalSchema` to require a date-only form or an RFC 3339 value with an explicit offset, or have `validity.ts` append `Z` to an offset-less value so the interpretation is deterministic. Add a test that runs under a fixed non-UTC `TZ`.

Follow-up from plur-ai/plur#1154.

### LINKED ISSUE #1158: Make the remote write-contract guard assert TRANSMITTED fields actually reach the POST body
`remote-write-contract.test.ts` forces every `EngramSchema` field into one of four classification sets, and that half works — adding a new field to the schema fails the suite as intended.

But membership in `TRANSMITTED` is never checked against what `appendAndGetServerId` actually builds. Deleting the three new field spreads from the POST body leaves the "classifies every field" test green; only the three hand-written loopback assertions catch the loss.

The failure message tells the author to add the field "to TRANSMITTED (and to appendAndGetServerId's body)", but only the first half is enforced. So the next occurrence of the #768 / #1151 defect can still pass the guard that was built to prevent it.

Fix: drive the loopback assertion from the `TRANSMITTED` set — learn one engram populating every transmitted field, then assert each name appears in the captured POST body.

Follow-up from plur-ai/plur#1154.

plur9:
Closes #1155. Closes #1156. Closes #1157. Closes #1158.

All four follow-ups from @crtahlin's review of #1154, plus the two smaller
inaccuracies noted there without an issue. One commit each, every fix verified
to go red when reverted.

### #1155 — the one that contradicted #1154's own claim

#1154 said the deadline had been made *structural* by giving the shared helper
ownership of the body read. `existsById()` was the counter-example: its own
AbortController, `clearTimeout` in a `finally` around only the `fetch`, and
`await r.json()` outside it. The review is right that this is the worst place
for it — both callers hold the primary store lock, and unbounded the probe
inherits undici's 300s `headersTimeout`, past the 180s `DEFAULT_ACQUIRE_TIMEOUT`,
so waiting `plur_learn` calls throw "Failed to acquire lock" and the engram is
silently never stored.

Now routed through `fetchBounded`. A timeout keeps this method's own wording,
because for an ambiguity probe "cannot tell" versus "absent" is the entire
point — carried by a typed `RemoteTimeoutError` rather than by matching on
another function's message text.

Reverting the fix makes the new test *hang* for the full 30s and fail, which is
the defect exactly.

### #1158 — the guard that could not catch what it was built for

The sharpest finding. The classification test was set membership over a
hand-written list with nothing tying it to the code, so deleting the three field
spreads left it green.

Now `TRANSMITTED` is asserted against the actual POST body, and `FLATTENED`
against the flattened keys. **Verified by deleting the `measured_under` spread —
simulating the #1151 recurrence exactly: the new assertion goes red naming the
field while the classification suite stays green.**

Driven through `RemoteStore.appendAndGetServerId` rather than `learnRouted`,
deliberately. Written the other way round first, it reported `locked_reason` and
`provenance` as missing — both were the test's fault, not the code's
(`locked_reason` survives only under `commitment: 'locked'`; `provenance` is not
a learn parameter). The wire contract belongs to the store driver; routing it
through learn tests the context mapping at the same time and blames the wrong
layer.

### #1156 — the fifth instance

`validityExpired()` deleted rather than fixed in place; tension pairing uses the
shared evaluator. `options.now` stays a `YYYY-MM-DD` string and resolves to the
END of that day, which is what preserves existing behaviour — under the old
lexical comparison a date-only `valid_until` equal to `now` compared equal and
so was not expired. Midnight would expire everything a day early, and there is a
control test for that.

### #1157 — and a second commit, because the first test was not good enough

Zone-less timestamps are pinned to UTC. UTC rather than an arbitrary choice:
every timestamp this codebase mints is `toISOString()`, so a naive value is
foreign or hand-edited, and reading it as UTC agrees with the rest of the store
rather than with whoever runs the query.

The first attempt asserted the mechanism plus a behavioural equivalence — and
that equivalence holds on a UTC host whether or not the fix is present, so on
CI it would have gone green through a reintroduction. The issue asked for a test
under a fixed non-UTC `TZ` and it was right to. Node re-reads `process.env.TZ`
per `Date` operation, so the second commit pins it inside the test across
America/New_York, Asia/Tokyo, Pacific/Kiritimati and UTC as control, each first
checking the offset really moved so a green run cannot mean "TZ was ignored".
Bypassing the fix turns all three non-UTC cases red and leaves UTC green.

### The two smaller things, no issue filed

Both were the previous change asserting something inaccurate about itself:

- The shared-helper test said "exactly two remain by design", then named three
  and asserted `<= 3`, and the exemption was wrong — `load`'s pager does read
  its body inside its own timer, `existsById` did not. Now `toBe(2)`. An upper
  bound would have gone green for a *new* bare fetch the moment this one stopped
  being one.
- The file header credited the branch that added the stalled-body tests with
  adding this method's timeout. It predates that branch, and guarded only the
  handshake — which is the whole finding.

### Verification

- `packages/core` typecheck clean; the seven affected suites pass.
- Each fix individually reverted and confirmed to go red: `existsById` hangs 30s
  (#1155), tension pairing returns 1 instead of 0 (#1156), three non-UTC zones
  disagree (#1157), `measured_under` reported missing from the body (#1158).
- Bare `await fetch(` in `remote-store.ts` is now exactly two, asserted by
  equality: `fetchBounded` itself and `load`'s pager, which owns its body read.

### DIFF
--- a/packages/core/src/store/remote-store.ts
+++ b/packages/core/src/store/remote-store.ts
@@ -193,6 +193,20 @@ const NEVER_STRIP = new Set(['visibility', 'pinned'])
 const warnedSalvages = new Set<string>()
 
 /**
+ * Thrown when a bounded request misses its deadline, at ANY phase.
+ *
+ * A distinct type rather than a message: `existsById()` has to re-word a
+ * timeout in its own terms, and matching on the text of another function's
+ * error is the kind of coupling that breaks silently when the wording changes.
+ */
+export class RemoteTimeoutError extends Error {
+  constructor(url: string, ms: number) {
+    super(`request to ${url} timed out after ${ms}ms`)
+    this.name = 'RemoteTimeoutError'
+  }
+}
+
+/**
  * A response whose body has already been read, inside the request deadline.
  *
  * `json` is present only for a 2xx (and is `undefined` when the payload would
@@ -325,7 +339,7 @@ export class RemoteStore {
     // clean; on failure the mark is refreshed.
     const ctrl = new AbortController()
     const timer = setTimeout(() => ctrl.abort(), LOAD_FETCH_TIMEOUT_MS)
-    const timedOut = () => new Error(`request to ${url} timed out after ${LOAD_FETCH_TIMEOUT_MS}ms`)
+    const timedOut = () => new RemoteTimeoutError(url, LOAD_FETCH_TIMEOUT_MS)
     try {
       let res: Response
       try {
@@ -745,25 +759,33 @@ export class RemoteStore {
    * budget is the same one `load()` uses, for the same reason.
    */
   async existsById(id: string): Promise<boolean> {
-    const ctrl = new AbortController()
-    const timer = setTimeout(() => ctrl.abort(), LOAD_FETCH_TIMEOUT_MS)
-    let r: Response
+    // Uses the shared helper (#1155). It carried its OWN AbortController with
+    // `clearTimeout` in a `finally` around only the `fetch`, so `await
+    // r.json()` ran after the deadline was gone — the exact defect #1152 fixed
+    // everywhere else, surviving in the one method that had opted out of the
+    // helper. With headers delivered and the body stalled it had still not
+    // settled 120 seconds past its 30-second bound.
+    //
+    // That matters more here than anywhere else: both callers run this INSIDE
+    // the primary store lock, one probe per configured remote. Unbounded, it
+    // inherits undici's 300s `headersTimeout`, which exceeds the 180s
+    // `DEFAULT_ACQUIRE_TIMEOUT` — so every waiting `plur_learn` throws "Failed
+    // to acquire lock" and the engram is silently never stored. A hang here is
+    // not slow, it is lost writes.
+    let r: BoundedResponse
     try {
-      r = await fetch(`${this.apiBase}/engrams/${encodeURIComponent(id)}`, {
+      r = await this.fetchBounded(`${this.apiBase}/engrams/${encodeURIComponent(id)}`, {
         headers: this.headers(),
-        signal: ctrl.signal,
-      })
+      }, RemoteStore.readBounded)
     } catch (err) {
       // An abort is "cannot tell", not "absent" — the whole point of this
       // method — so it must surface as a throw like any other transport
       // failure, with a message that says which it was.
       throw new Error(
-        ctrl.signal.aborted
+        err instanceof RemoteTimeoutError
           ? `existence probe for ${id} timed out after ${LOAD_FETCH_TIMEOUT_MS}ms against ${this.apiBase}`
           : `existence probe for ${id} failed against ${this.apiBase}: ${(err as Error).message}`,
       )
-    } finally {
-      clearTimeout(timer)
     }
     if (r.status === 404) return false
     if (!r.ok) throw new Error(`HTTP ${r.status} from ${this.apiBase}`)
@@ -772,7 +794,7 @@ export class RemoteStore {
     // envelope payloads on routes they do not recognise, and inferring
     // existence from the status line alone would turn that into a false
     // collision report on every forget.
-    const row = await r.json().catch(() => null) as { id?: unknown } | null
+    const row = (r.json ?? null) as { id?: unknown } | null
     return typeof row?.id === 'string' && row.id === id
   }
 
--- a/packages/core/src/tensions.ts
+++ b/packages/core/src/tensions.ts
@@ -1,5 +1,6 @@
 import type { Engram, MeasuredUnder } from './schemas/engram.js'
 import type { LlmFunction } from './types.js'
+import { isExpired, evaluationInstant } from './validity.js'
 import { ftsTokenize } from './fts.js'
 
 export interface TensionPair {
@@ -341,11 +342,6 @@ function supersedesLinked(a: Engram, b: Engram): boolean {
   )
 }
 
-/** True when the engram's validity window has closed (`temporal.valid_until` before `now`). */
-function validityExpired(e: Engram, now: string): boolean {
-  const until = e.temporal?.valid_until
-  return Boolean(until && until < now)
-}
 
 /** Confidence cap for measured-under pairs in `measured_under_pairs: 'floor'` mode (#869). */
 export const MEASURED_UNDER_CONFIDENCE_CAP = 0.1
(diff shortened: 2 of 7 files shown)


crtahlin (approved):
Approved. Two notes worth tracking are filed as follow-ups; the rest are small enough to take or leave.

All four fixes do what they claim. I reproduced every mutation described in the body rather than taking the claims on trust, and each turns red in exactly the right place:

| Issue | Mutation applied | Result |
| --- | --- | --- |
| #1155 | restored the bare `fetch` with the body read outside the timer | the new test hangs to its timeout, and the source scan fails `expected 3 to be 2` |
| #1156 | restored the lexical `until < now` comparison | the instant-expiry pairing test goes red |
| #1157 | made `pinToUtc` return its input | three non-UTC zones red, UTC control green |
| #1158 | deleted the `measured_under` spread from the request body | the new assertion names the missing field while the classification suite stays green |

The #1158 result is the one that matters. That was the sharpest finding in the last round — a guard that could not catch the defect it was built for — and asserting `TRANSMITTED` against the actual request body closes it properly. Simulating the #1151 recurrence produces the message a future author actually needs, naming the field rather than reporting a bare mismatch.

The #1157 test earns its length. The first attempt would have gone green on a UTC runner through a reintroduction; pinning `TZ` per case and first asserting the offset really moved means a green run cannot silently mean "the timezone was ignored".

Two claims the change rests on, checked rather than assumed. `load`'s pager does read its body inside its own timer — `clearTimeout` sits in a `finally` around the block containing `r.json()` — so the remaining exemption is real and `toBe(2)` enshrines a true statement. And running eleven timestamp forms through `pinToUtc`, only zone-less values change; a worry that year-month forms like `2026-09` would become unparseable turned out to be unfounded, since the appended `Z` still parses.

Follow-ups: #1165, #1166

### Notes

- `packages/core/src/store/remote-store.ts:115` — `existsById` now inherits the breaker semantics from `fetchBounded`: a network failure marks the origin down, any response clears it. The old bare-`fetch` version did neither. The behaviour is defensible and matches the documented policy, but the design note enumerates the users as "writes, /me, explicit retries" and reads as exhaustive, so the next reader will conclude the probe does not touch the breaker. Filed as #1165.
- `packages/core/src/validity.ts` — `evaluationInstant` answers as of the present when its `now` will not parse, silently. Not a regression, and the sibling bound functions deliberately fail open for a documented reason that does not extend to a caller-supplied argument. Filed as #1166.
- `packages/core/src/tensions.ts:344-345` — removing `validityExpired` left two consecutive blank lines. Nothing lints it; there is no eslint config in the repo.
- `packages/core/test/remote-exists-by-id.test.ts` — the header edit wraps mid-phrase: "both callers / run it INSIDE the primary store / lock".
- The `TRANSMITTED` failure message names two causes, add the field to `appendAndGetServerId` or reclassify it. There is a third: the hand-built fixture in that test not setting the field. Adding a field to `TRANSMITTED` without adding it to `full` produces the same red, and the message will point at the wrong place.

### Worth keeping

A typed `RemoteTimeoutError` rather than matching on another function's message text is right, and for this method specifically — `existsById` has to re-word a timeout in its own terms because "cannot tell" versus "absent" is the entire point, and text coupling breaks silently when wording changes.

Resolving `options.now` to the end of the named day, with the reasoning recorded, is what keeps #1156 a bug fix rather than a behaviour change. Midnight would have expired everything a day early.

Both self-corrections are honest work: the previous branch did not add this method's timeout, and "exactly two by design" while naming three and asserting `<= 3` was wrong in the number and in the exemption. Changing it to an equality is load-bearing, not tidying — my #1155 mutation was caught by that assertion as well as by the behavioural test.

Tests: PASS — full `packages/core` suite on the head commit, 3000 passed, 34 skipped, 0 failed. All eight CI checks green. Verified in an isolated worktree, removed afterwards.

PR #1164, opened 2026-09-08
ci: require a non-empty message when dismissing a review

crtahlin:
Adds a small workflow: a review dismissal must carry a message, and must be verifiable.

## Why here specifically

A full paginated census of `REVIEW_DISMISSED_EVENT` in this repository since 2026-08-18 finds thirteen dismissals: **eight carried a message, five were empty.**

The ratio matters less than the shape:

| | with a message | empty |
|---|---|---|
| before 2026-09-08 | 2 | **5** |
| on 2026-09-08 | 6 | 0 |

The five empty dismissals are PRs [#929](https://github.com/plur-ai/plur/pull/929) and [#922](https://github.com/plur-ai/plur/pull/922) (2026-08-18), [#981](https://github.com/plur-ai/plur/pull/981) (2026-08-26), and [#1121](https://github.com/plur-ai/plur/pull/1121) and [#1108](https://github.com/plur-ai/plur/pull/1108) (2026-09-04). The two earlier messaged ones are [#957](https://github.com/plur-ai/plur/pull/957) and [#953](https://github.com/plur-ai/plur/pull/953).

So the practice barely existed until the problem was raised, and then every dismissal that day carried a message. That is the argument for enforcing it rather than trusting it — a rule is better than a habit that started this morning.

An earlier version of this description claimed nine of twelve carried itemised messages and that only three were empty. That was wrong, and the corrected figures are less flattering to the record, which is the point. The census above is scoped to this repository; I have not re-counted `plur-ai/enterprise`.

## What it does

On `pull_request_review` with action `dismissed`, it resolves the **triggering** review from `context.payload.review.id`, finds that review's `review_dismissed` timeline event, and fails if the `dismissal_message` is empty or whitespace.

The message is read from the timeline rather than the webhook payload, because `dismissal_message` is not present on the review object that `pull_request_review` delivers.

**It fails closed.** If the timeline event cannot be found, the job fails rather than passes. That branch is not hypothetical: [#1122](https://github.com/plur-ai/plur/pull/1122) carries review `5113096768` in state `DISMISSED` with a 1,113-character message, and its timeline has no `review_dismissed` event of any kind. It is a fork pull request, which is the current suspicion but not a confirmed cause — and the design flaw does not depend on the cause. An empty dismissal and an unverifiable one are the same thing from the reviewer's side: no record. Failing open on fork pull requests would be exactly backwards, since that is where external review discipline matters most.

Because the timeline can lag the webhook, it polls five times over roughly twelve seconds before concluding. Failing closed only helps if it does not also fail loudly on a timing artefact.

`pull_request_review` rather than `pull_request_target` is deliberate: the job never checks out or executes the head ref, so it does not need the privileged base-repository context that `pr-issue-guard.yml` requires.

`issues: read` is granted because `listEventsForTimeline` is an issues endpoint (`GET /repos/{o}/{r}/issues/{n}/timeline`) even when read for a pull request. Without it a 403 would throw and fail the job on every dismissal, well-documented ones included.

## Verification

The workflow has never executed — it triggers only on a dismissal. So the script was extracted and exercised against stubs across nine branches: message present, empty, whitespace-only, null, no event at all, an event for a different review, a later messaged dismissal not clearing an unmessaged one, an absent review id, and a 403 from the timeline API. All nine behave as intended. The two that correspond to the review findings both fail now and passed before.

`dismissed_review.review_id` was confirmed against a real timeline event on [#946](https://github.com/plur-ai/plur/pull/946), where it matches that pull request's actual dismissed review id.

## Enforcement status

**Advisory.** To make it a merge gate, add **`dismissal message required`** to the required status checks for `main` in branch protection. That is a repository-settings change, not a code one, and is not made here.

- [ ] Add `dismissal message required` to the required status checks for `main`

Without that step this stays a visible failing check rather than a gate, which would be an easy thing to leave undone by accident.

A matching addition to the existing Review Guard workflow in `plur-ai/enterprise` is [plur-ai/enterprise#1239](https://github.com/plur-ai/enterprise/pull/1239).

### DIFF
--- a/.github/workflows/review-dismissal-guard.yml
+++ b/.github/workflows/review-dismissal-guard.yml
@@ -1,0 +1,66 @@
+name: Review Dismissal Guard
+
+# A review dismissal must say what was done.
+#
+# An empty dismissal is indistinguishable from ignoring the review. An audit on
+# 2026-09-08 of every dismissal since 2026-08-18 found that nine of twelve
+# carried itemised messages citing the commits that fixed each finding — and
+# that the three which read as negligent were simply the three with no message.
+# The work may have been done in those cases too; the record cannot say, which
+# is the problem this closes.
+#
+# `pull_request_review` (not `pull_request_target`) is correct here: the job
+# never checks out or executes the head ref, and it only needs read access plus
+# the ability to report a check result.
+#
+# ADVISORY until `dismissal message required` is added to the required status
+# checks for `main` in branch protection. Without that it is a visible failing
+# check, not a merge gate.
+
+on:
+  pull_request_review:
+    types: [dismissed]
+
+permissions:
+  pull-requests: read
+  contents: read
+
+jobs:
+  dismissal-message-required:
+    name: dismissal message required
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/github-script@f28e40c7f34bde8b3046d885e986cb6290c5673b # v7.1.0
+        with:
+          script: |
+            const pr = context.payload.pull_request;
+            // The dismissal message is NOT on the webhook's review object — it
+            // lives on the `review_dismissed` timeline event, so read it there.
+            const events = await github.paginate(
+              github.rest.issues.listEventsForTimeline,
+              {
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                issue_number: pr.number,
+                per_page: 100,
+              }
+            );
+            const dismissals = events.filter(e => e.event === 'review_dismissed');
+            if (!dismissals.length) {
+              core.info('No review_dismissed event found on this pull request.');
+              return;
+            }
+            const last = dismissals[dismissals.length - 1];
+            const msg = ((last.dismissed_review || {}).dismissal_message || '').trim();
+            const who = last.actor ? last.actor.login : 'unknown';
+            if (!msg) {
+              core.setFailed(
+                `@${who} dismissed a review without a message.\n\n` +
+                'A dismissal must state what was done, citing the commits that addressed ' +
+                'the review. An empty dismissal is indistinguishable from ignoring the ' +
+                'review, whether or not the work was actually done.\n\n' +
+                'Re-dismiss with a message, or ask the reviewer to re-review.'
+              );
+            } else {
+              core.info(`Dismissal by @${who} carries a ${msg.length}-character message.`);
+            }
--- a/.github/workflows/review-dismissal-guard.yml
+++ b/.github/workflows/review-dismissal-guard.yml
@@ -1,23 +1,46 @@
 name: Review Dismissal Guard
 
-# A review dismissal must say what was done.
+# A dismissed changes-requested review must say what was done.
 #
-# An empty dismissal is indistinguishable from ignoring the review. A full
-# paginated census of `REVIEW_DISMISSED_EVENT` in this repository since
-# 2026-08-18 found thirteen dismissals: eight carried a message and five were
-# empty. The shape matters more than the ratio — of the seven dismissals before
-# 2026-09-08, five were empty and two carried a message, and every one of the
-# six on 2026-09-08 carried one, after this problem had been raised. So the
-# habit did not exist until it was pointed at, which is the argument for
-# enforcing it rather than trusting it.
+# What the record shows. A census of every `review_dismissed` event in this
+# repository between 2026-08-18 and 2026-09-11 found fourteen. Grouped by the
+# state the review had before it was dismissed:
+#
+#   approved            empty     6
+#   approved            message   2
+#   changes requested   message   6
+#   changes requested   empty     0
+#
+# Every empty dismissal was of an approval. That is consistent with the
+# `dismiss_stale_reviews` branch-protection rule, which removes an approval when
+# new commits are pushed and records it as an ordinary `review_dismissed` event
+# with the pusher as actor and no message. However an approval was removed,
+# removing one never bypasses review. Every visible dismissal of a
+# changes-requested review carried a message.
+#
+# So this guard does not correct an observed habit. It keeps the one dismissal
+# that can bypass a review — setting aside a request for changes — from ever
+# happening without a record of why. That is also why it enforces only when the
+# dismissed review had requested changes: enforcing on approvals would put a red
+# check against whoever pushed, including an external contributor who cannot
+# dismiss anything.
+#
+# Some manual dismissals of changes-requested reviews have produced no
+# `review_dismissed` timeline event at all (#1122, #1138, and one of the two on
+# #1017). The cause is not known. It is not the stale-review rule, which never
+# touches changes-requested reviews, and it is not specific to forks. When the
+# event cannot be found this guard fails rather than passes, because an
+# unverifiable dismissal and an empty one leave the same record behind.
 #
 # `pull_request_review` (not `pull_request_target`) is correct here: the job
 # never checks out or executes the head ref, and it only needs read access plus
 # the ability to report a check result.
 #
-# ADVISORY until `dismissal message required` is added to the required status
-# checks for `main` in branch protection. Without that it is a visible failing
-# check, not a merge gate.
+# ADVISORY, and not suitable as a required status check in this form. The job
+# runs only when a review is dismissed, so a pull request with no dismissal never
+# reports it at all. Its result attaches to whatever commit was the head at the
+# moment of dismissal, so the next push leaves it behind. A merge gate would need
+# a `pull_request` trigger that re-evaluates every dismissal on the pull request.
 
 on:
   pull_request_review:
@@ -44,6 +67,12 @@ jobs:
             const review = context.payload.review || {};
             const reviewId = review.id;
 
+            // A dismissed review cannot be dismissed again, so every failure
+            // below points at the two things that actually clear it.
+            const remedy =
+              'A dismissed review cannot be dismissed again. To clear this, ask the ' +
+              'reviewer to re-review, or push a commit.';
+
             // Match the review that TRIGGERED this run, not the newest dismissal
             // on the pull request. Taking the last one let a message-less
             // dismissal be cleared by any later dismissal that happened to carry
@@ -57,15 +86,10 @@ jobs:
               return;
             }
 
-            // The dismissal message is NOT on the webhook's review object — it
-            // lives on the `review_dismissed` timeline event, so read it there.
-            //
-            // The timeline can lag the webhook by a moment, and on at least one
-            // pull request the event never appeared at all (#1122: review
-            // 5113096768 sits in state DISMISSED with a 1,113-character message
-            // and the timeline carries no `review_dismissed` event of any kind;
-            // it is a fork pull request, which is the current suspicion but not
-            // a confirmed cause). So poll briefly, then fail rather than pass.
+            // The dismissal message and the review's state before dismissal are
+            // NOT on the webhook's review object — they live on the
+            // `review_dismissed` timeline event, so read them there. The timeline
+            // can lag the webhook by a moment, so poll briefly before concluding.
             const findDismissal = async () => {
               let events;
               try {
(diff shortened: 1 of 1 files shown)


plur9 (changes requested):
Requesting changes — the guard is silently inert when the timeline event is missing, and I have a dismissal from today that proves that is not hypothetical.

I have an obvious interest here: four of the dismissals this audits are mine, from this morning. I have tried to check the claims rather than argue with them, and the correction below makes the case for this workflow stronger, not weaker.

### Blocking

- `.github/workflows/review-dismissal-guard.yml:39-45` — when no `review_dismissed` event is found the script calls `core.info(...)` and returns, so the job PASSES. That branch is reachable in practice: **plur-ai/plur#1122 has no `review_dismissed` timeline event at all**, despite review `5113096768` sitting in state `DISMISSED` with an 1,113-character message I wrote.

  Verified against three controls dismissed by the same actor through the same API call within the same few minutes:

  ```
  #946   review_dismissed events: 1  (msg_len=835)
  #981   review_dismissed events: 2
  #1120  review_dismissed events: 1
  #1122  review_dismissed events: 0   ← full event-type census shows none
  ```

  The distinguishing feature of #1122 is that it is the only **fork** PR of the four (`amasen02/plur`), so that is my first suspicion — but I would not merge on my diagnosis of the cause, because the design flaw does not depend on it. Whatever the reason the event is absent, the guard's response to "I cannot see the evidence" is to report success.

  That is the wrong default for this specific check. An empty dismissal and an unverifiable one are the same thing from the reviewer's side: no record. And it fails open on fork PRs, which is where external review discipline matters most.

  Suggested: treat "no event found" as a failure, or at minimum resolve the review id from `context.payload.review.id` and fail explicitly when the timeline does not contain a dismissal for it — `could not verify` is an honest outcome, `passed` is not.

### Should fix

- `.github/workflows/review-dismissal-guard.yml:23-25` — `issues.listEventsForTimeline` is an Issues endpoint (`GET /repos/{o}/{r}/issues/{n}/timeline`), and the block grants only `pull-requests: read` and `contents: read`. If that resolves to a 403 the script throws and the job fails on **every** dismissal, including well-documented ones — turning a guard into noise that gets ignored or removed. This would be the repository's first workflow to call the timeline API, so there is no precedent here to lean on, and the workflow has never executed (it only triggers on a dismissal, and none has occurred since it was opened). Adding `issues: read` is harmless and removes the question.

- The header comment's audit does not match the record, and it will be read as fact by whoever next touches this file. A full paginated census of `REVIEW_DISMISSED_EVENT` since 2026-08-18:

  | | count |
  |---|---|
  | with a message | 5 |
  | **empty** | **5** |

  Not nine of twelve. And the shape is worse than "three read as negligent": of the seven dismissals **before today**, five were empty and two carried messages (#953 593c, #957 167c). Three of the five messaged ones are from this morning, after this problem had been raised.

  So the pre-existing habit was roughly half-and-half at best, and a stricter reading says the practice barely existed until today. That is a better argument for merging this than the one in the comment. Worth correcting the numbers rather than softening them — "the work may have been done in those cases too" is generous to a record that does not support it.

### Notes

- `dismissals[dismissals.length - 1]` takes the most recent dismissal on the PR, not the one that triggered the run. A message-less dismissal is therefore cleared by any later dismissal that carries a message, even on an unrelated review. Using the triggering review id would fix this and the blocking finding in the same edit.
- Advisory until added to required status checks — stated plainly in the header, which is the right way to ship this. Worth adding the branch-protection step to the PR body as a follow-up task so it does not stay advisory by accident.

### Done well

Finding that the dismissal message lives on the timeline event rather than the webhook's review object is the non-obvious part of this, and the comment records it so the next person does not go looking in the payload. `pull_request_review` rather than `pull_request_target` is correct and the reasoning is stated. The action is SHA-pinned. The retirement condition — advisory until branch protection is updated — is written down at creation time, which is rare.

And the underlying point is right: I dismissed four reviews this morning, and the only reason that reads as diligence rather than as bypassing you is that I happened to write messages. A rule is better than my habits.

Tests: N/A — the workflow has never executed; it triggers only on a dismissal. Verified: full paginated `REVIEW_DISMISSED_EVENT` census via GraphQL; per-PR REST timeline census for #946, #981, #1120 and #1122; confirmed #1122's review state is `DISMISSED` while its timeline carries no such event; checked the repository for any existing timeline-API workflow precedent.

crtahlin:
@plur9 — review fixes pushed as `f3fc27d3`. All three points addressed, plus a correction to your diagnosis that you should see before re-reviewing.

**Blocking finding — fixed.** The guard now resolves the triggering review from `context.payload.review.id` and fails when that review's timeline event cannot be found, instead of passing. That also fixes your note in the same edit: matching by review id means a message-less dismissal can no longer be cleared by a later dismissal on an unrelated review. Confirmed on #981, where two dismissals carry distinct review ids — one empty, one with a message — which the old last-match logic conflated. It polls five times over ~12s first, so a timeline that merely lags the webhook does not read as a failure.

**`issues: read` — added**, with the reason in a comment. A 403 now reports what to fix rather than surfacing as a raw throw.

**Census — corrected, and you were right.** I re-ran it independently rather than take it on trust: thirteen dismissals since 2026-08-18, eight with a message, five empty. Your split reproduces exactly — before 2026-09-08, five empty against two with messages; all six on 2026-09-08 carried one. My "nine of twelve" was wrong. The header and the PR body now say so, and the line suggesting the work may have been done in the empty cases is gone. Scoped to this repository; I did not re-count `enterprise`.

---

**Correction: the fork hypothesis is wrong, and the real cause changes the design.**

I scanned the last 100 pull requests for reviews in `DISMISSED` state with no `review_dismissed` timeline event. Three have one — #1122 and #1017 (forks), and **#1138, which is internal**.

`dismiss_stale_reviews` is `true` on `main`. My review on #1138 was submitted against `9ab8de99`, the branch was pushed to `94df0f5e`, and that push auto-dismissed it — creating no timeline event. Both fork cases fit the same shape. The fork correlation was coincidence.

That leaves an open problem with my own fix: if the `pull_request_review` `dismissed` webhook fires for automatic stale dismissals, this guard now fails on every push that invalidates a review, finds no event, and produces a red check nobody can clear — there is no human dismissal to redo with a message. I could not determine whether the webhook fires for auto-dismissals without triggering one.

So this needs a fourth case before it merges: no event **and** the review's `commit_id` is not the current head is almost certainly a stale auto-dismissal with nothing to enforce, while no event **and** `commit_id` equal to head is genuinely unverifiable and should fail — which preserves your point for the case it was actually about.

You said you would not merge on your own diagnosis of the cause. The cause is now known, and it argues for a change to what I have pushed rather than for merging it. Happy to implement the four-case version if you agree with the shape.

Two smaller things while re-reading: a `dismissal_message` of a single zero-width space passes, because `trim()` does not strip U+200B; and the five paginated timeline fetches are wasteful when only the last page can hold a new event.

CI is green on all seven checks.

plur9 (changes requested):
Thanks. All three points from my review are addressed in `f3fc27d`:
- the guard matches the triggering review id;
- it has `issues: read`, with an explanatory failure on a 403;
- the census is corrected.

I'm still not approving this head, and on your design question I'd go a different way from the four-case rule. The record points at approvals, not at changes-requested reviews.

**What the timeline shows**
- **Every empty dismissal was an approval removed by a push.** `dismiss_stale_reviews` only ever dismisses approvals, and the timeline records those dismissals as ordinary `review_dismissed` events with the pusher as actor and an empty message. All six empty dismissals on record were of APPROVED reviews: #922, #929, #981 (08-26), #1108, #1121 and #1177. On #1177 that was 2026-09-09 12:39Z, actor amasen02, previous state APPROVED, with a 0-character message.
- **The guard would blame the wrong person.** At `f3fc27d` that dismissal would produce "@amasen02 dismissed a review without a message": a public red check against an external contributor who can't dismiss anything. #922 would fail the same way. I haven't confirmed that the `dismissed` webhook fires for automatic dismissals, but the timeline records them identically.
- **Every dismissal of a changes-requested review carried a message:** #957, #953, #946, #981 (09-08), #1120 and #1017.
- **The missing-event cases were manual.** The reviews dismissed with no timeline event are changes-requested reviews (#1122, #1138). The stale rule doesn't touch those, so they were dismissed by hand. #1122's carries a 1,113-character dismissal message.
- **The four-case rule would reopen the hole.** "No event and review commit ≠ head, so assume an auto-dismissal and pass" would pass exactly these cases, because each review sat on an older commit than head. That's the hole this guard exists to close.

**What I'd like instead**
1. **Enforce only on changes-requested reviews.** Match the event's `dismissed_review.state` as well as its id, and enforce only when the state is `changes_requested`; approved and commented dismissals pass. Withdrawing an approval never bypasses review, so this one condition removes the false positives and keeps the check where it matters.
2. **Keep failing closed when the event is missing.** Every observed missing-event case was a changes-requested review. A manual dismissal that has a message but no event will then give a red check that only a re-review or a push clears. Since the check is advisory, I'm fine with that.
3. **Rewrite the header's census paragraph.** The empty dismissals were automatic approval dismissals, not people skipping the message, and the header will be read as fact.
4. **Drop "add to required checks" for now.**
   - The job only runs on a dismissal, so a PR without one would never report the check.
   - The result attaches to whatever commit is the head at dismissal time, so the next push drops it.
   - A real gate would need a `pull_request` trigger that re-evaluates every dismissal on the PR.
5. **Fix the remediation text.** "Re-dismiss with a message" can't be followed, because a dismissed review can't be dismissed again. Say "ask the reviewer to re-review, or push a commit". And as you noted, `trim()` doesn't strip U+200B.

**Security looks right:** a dismissed-only trigger, no checkout, no `${{ }}` interpolation, read-only permissions, and the action pinned to the v7.1.0 SHA (checked).

crtahlin:
@plur9 — your correction is right, and I've confirmed it independently. `dismiss_stale_reviews` only removes approvals, so it could not have caused the missing timeline events I attributed to it in my earlier comment. Those were manual dismissals of changes-requested reviews: #1122, #1138, and one of the two on #1017. A census by previous review state backs your reading. Of fourteen dismissals since 2026-08-18, all six empty ones were of approvals, and no changes-requested review was dismissed empty. I also confirmed that the head I pushed would have blamed amasen02 on #1177: the push landed at 12:39:06 and the dismissal followed at 12:39:10.

`b0bd776d` takes your design on all five points:

1. It enforces only when `dismissed_review.state` was `changes_requested`. Approved and commented reviews pass, and an unknown state is enforced.
2. It still fails closed when the event is missing. I've dropped the four-case idea — you were right that it would pass exactly those cases.
3. The header census is rewritten around previous state, and the fork explanation is gone.
4. It no longer suggests a required check, and says why this trigger can't serve as a gate.
5. The remediation text now says to re-review or push. A message made only of invisible or directional-control characters no longer passes.

**Where this comment and the pull request description disagree, this comment is current.** The description is left as it was, for the record. Three things in it no longer hold:

- **The fork explanation for the missing timeline events.** They are not specific to forks — #1138 is internal — and the cause is not known.
- **The framing that explaining a dismissal "barely existed until it was pointed at."** The empty dismissals were approvals being removed, which never bypasses review. Every visible dismissal of a changes-requested review carried a message. The guard is preventive, not a correction of an observed habit.
- **The task to add `dismissal message required` to the required status checks.** Don't. The job runs only on a dismissal, and its result attaches to the head at that moment, so it cannot work as a gate in this form.

Verified with 24 stub cases, including replays of the live timelines: #1177 and #929 pass, #946 passes, and #1138 fails closed. Removing the state gate turns five cases red; reverting to `trim()` turns three red.

Ready for another look.

plur9 (commented):
Re-review of `b0bd776d`. All five points from my 09-11 review are addressed, and I checked each one against the file.

1. **State gate.** It enforces only when `dismissed_review.state` is `changes_requested`, after lowercasing. An empty or unknown state is enforced (`:159-167`).
2. **Fail closed.** A missing event still fails after five polls. The 403 case is reported as its own failure (`:105-115`, `:136-151`).
3. **Header census.** It is now framed by the review's state before dismissal, and the fork explanation is gone.
4. **Required check.** The header no longer suggests making this a required check, and explains why this trigger can't serve as a gate (`:39-43`).
5. **Invisible-only messages** no longer pass.

**Verified independently rather than from your stub run:**
- I read live timelines. `dismissed_review.state` is the state before dismissal, in lowercase: #946 and #953 are `changes_requested` with messages; #929 and #1177 are `approved` and empty. So #1177 and #929 would now pass, which is the false positive my review was about.
- I extracted the script from the YAML, ran `node --check` on it, and ran it against 14 stubbed timelines. All intended cases behave as designed:
  - changes-requested with a message → pass; empty, null, U+200B only, BOM+RLO+whitespace → fail;
  - approved or commented and empty → pass;
  - unknown or uppercase state and empty → fail;
  - no event, an event for a different review, or a timeline 403 → fail.

### Notes (non-blocking)

- **`remedy` overstates what a re-review does, and that is my wording from 09-11.** The workflow triggers only on `dismissed`, so a fresh review never re-runs it, and "Re-run jobs" replays the same payload and fails the same way. Only a push clears the red check. A re-review repairs the record, not the check. Suggested: "To clear this check, push a commit. A re-review restores the record but does not re-run this check."
- **The invisible-character class still has gaps.** U+3164 (Hangul filler) and U+00AD (soft hyphen) each pass as a one-character message. `/[\s\p{Default_Ignorable_Code_Point}]/gu` covers every range listed at `:176` plus those two, and keeps ordinary text, emoji and accented letters (checked in Node 24). Worth it for the one-line simplification; the gap itself only matters for someone deliberately evading, and the timeline records who that was.
- **Polling fetches the whole paginated timeline up to five times.** That's fine at our PR sizes.
- **The PR description still carries the required-status-check task and the fork framing.** Squash merges here build the message from commits, so it won't reach history. Still worth trimming so nobody ticks that box.

No blocking findings. I will approve when I make the merge calls on today's queue.

PR #1176, opened 2026-09-08
fix(core): throw RangeError on unparseable evaluation instant (#1166)

### LINKED ISSUE #1166: evaluationInstant silently evaluates at the present when its now parameter will not parse
`evaluationInstant(now)` in `packages/core/src/validity.ts` (added in plur-ai/plur#1163) ends:

```ts
const ms = closesAt(now)
return ms === null ? Date.now() : ms
```

So a caller that passes a `now` the parser cannot read is silently answered as of the present moment instead. `getCandidatePairsDetailed` exposes `options.now` publicly, and the parameter exists precisely so a caller can ask "how did this look at another time" — a typo in it produces a plausible-looking answer to a different question, with nothing said.

This is not a regression. The lexical comparison it replaces did something worse and equally quiet: an unparseable `now` made every dated engram compare as expired, because digits sort before letters. But this is a new function with a documented contract, and it is the shared evaluator that four call sites now depend on, so it is the right place to decide the policy rather than inherit one.

The sibling functions already made the opposite call deliberately and said why: `isExpired` and `isNotYetValid` treat an unparseable *bound* as absent, on the reasoning that the schema validates those fields and hiding content over a data error is the worse outcome. A caller-supplied evaluation instant is not that case — it is an argument, not stored data, and the caller is present to be told.

Fix: throw on an unparseable `now`, or return `null` and let each caller decide. Either way, say so in the docstring.

Follow-up from plur-ai/plur#1163.

amasen02:
### Summary
Resolves #1166.

In `packages/core/src/validity.ts`, `evaluationInstant(now)` previously fell back to `Date.now()` when `closesAt(now)` returned `null`:
```ts
const ms = closesAt(now)
return ms === null ? Date.now() : ms
```
As noted in #1166, while stored engram bounds (`isExpired`, `isNotYetValid`) treat unparseable timestamps as absent to avoid hiding valid content over data errors, a caller-provided evaluation instant (`now` in `options.now` across `getCandidatePairsDetailed` / `scanForTensions`) is an explicit argument. Falling back to the current time silently answered queries as of the present moment with no notification that the instant parameter was unparseable.

### Proposed Changes
1. **`packages/core/src/validity.ts`**:
   - Updated docstring to document the `RangeError` contract for caller-supplied evaluation instants.
   - Throws `RangeError('Unparseable evaluation instant: ... Expected YYYY-MM-DD or an RFC 3339 timestamp.')` when `now` is non-empty but cannot be parsed as a date or instant.
2. **`packages/core/test/validity-instants.test.ts`**:
   - Added `evaluationInstant (#1166)` unit test suite asserting:
     - Omitted or empty `now` returns current timestamp.
     - `YYYY-MM-DD` date resolves to end-of-day UTC (`23:59:59.999Z`).
     - RFC 3339 instants resolve accurately with zone offset.
     - Naive timestamps pin to UTC.
     - Unparseable strings throw `RangeError` with descriptive message.
3. **`packages/core/test/tensions.test.ts`**:
   - Added regression test asserting `getCandidatePairs` fails fast with `RangeError` when `options.now` is unparseable.

### Verification
- `npm test -- packages/core/test/validity-instants.test.ts packages/core/test/tensions.test.ts`: **122/122 passed**.
- `packages/core` builds cleanly via `npm run build` (`tsup`).

### DIFF
--- a/packages/core/src/validity.ts
+++ b/packages/core/src/validity.ts
@@ -159,18 +159,28 @@ export function isCurrentlyValid(temporal: Temporal, nowMs: number): boolean {
 /**
  * Interpret an "evaluate as of" parameter as an instant.
  *
- * Callers that expose a `now` option take a `YYYY-MM-DD` string. A whole day is
- * not an instant, so it has to be resolved to one, and the END of the day is
- * the resolution that preserves existing behaviour: under the old lexical
- * comparison a date-only `valid_until` equal to `now` was NOT expired (the
- * strings compared equal), which is the same answer end-of-day gives. Resolving
- * to midnight instead would expire everything a day early.
+ * Callers that expose a `now` option take a `YYYY-MM-DD` string or an RFC 3339
+ * instant. A whole day is not an instant, so it has to be resolved to one, and
+ * the END of the day is the resolution that preserves existing behaviour:
+ * under the old lexical comparison a date-only `valid_until` equal to `now` was
+ * NOT expired (the strings compared equal), which is the same answer end-of-day
+ * gives. Resolving to midnight instead would expire everything a day early.
+ *
+ * The sibling functions treat an unparseable bound as ABSENT, because stored
+ * engrams have their fields validated by `TemporalSchema` and hiding content
+ * over a data error is the worse outcome. A caller-supplied evaluation instant
+ * is not that case: it is an explicit argument, so an unparseable value throws
+ * a `RangeError` rather than silently answering as of the present moment (#1166).
  *
  * @param now - `YYYY-MM-DD`, or an RFC 3339 instant, or nothing for the
  *   current moment.
+ * @throws {RangeError} When `now` is provided but cannot be parsed as a date or instant.
  */
 export function evaluationInstant(now?: string): number {
   if (!now) return Date.now()
   const ms = closesAt(now)
-  return ms === null ? Date.now() : ms
+  if (ms === null) {
+    throw new RangeError(`Unparseable evaluation instant: "${now}". Expected YYYY-MM-DD or an RFC 3339 timestamp.`)
+  }
+  return ms
 }
--- a/packages/core/test/tensions.test.ts
+++ b/packages/core/test/tensions.test.ts
@@ -401,6 +401,14 @@ describe('getCandidatePairs', () => {
     expect(getCandidatePairs([a, b], { now: '2026-09-07' })).toHaveLength(1)
   })
 
+  it('fails fast when options.now is unparseable (#1166)', () => {
+    // A caller-provided now parameter with a typo must throw rather than
+    // silently answering as of the present moment.
+    const a = makeEngram({ id: 'E1', statement: 'plur search uses BM25.', scope: 'global' })
+    const b = makeEngram({ id: 'E2', statement: 'plur search uses embeddings.', scope: 'global' })
+    expect(() => getCandidatePairs([a, b], { now: 'not-a-date' })).toThrow(RangeError)
+  })
+
   it('skips inactive engrams', () => {
     const a = makeEngram({ id: 'E1', statement: 'plur search uses BM25.', scope: 'global' })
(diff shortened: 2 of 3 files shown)


miles-on-nightshift (approved):
The RangeError is correct here — a caller-supplied `now` is an explicit argument, not a stored field, so silent fallback to the current time is worse than a loud failure. The asymmetry with `isExpired`/`isNotYetValid` (which treat unparseable bounds as absent) is real but intentional and the docstring now explains why. Test covers the new throw path and the unchanged no-`now` path. CI passing. @plur9 ready for merge.

amasen02:
Gentle ping on throwing RangeError for unparseable evaluation instants (122 tests green). Ready to adjust the error surface per feedback.

crtahlin (approved):
Approved. Two documentation gaps are worth closing, neither blocking — filed as follow-ups.

The fix is correct and minimal. `evaluationInstant` now throws a `RangeError` on a caller-supplied `now` it cannot parse, instead of silently answering as of the present moment, and the docstring explains *why* this diverges from `isExpired` and `isNotYetValid` rather than merely stating that it does — a stored bound is schema-validated data where hiding content is the worse failure, a caller-supplied argument is an explicit instruction. That reasoning is the part a future reader would otherwise have to reconstruct.

The regression is asserted twice, once on the helper and once through the public `getCandidatePairs` surface a real caller uses. Both bite: reverting the throw to `return ms === null ? Date.now() : ms` turns exactly those two assertions red and leaves everything else green. The control tests for the forms that must keep working are there too — date-only resolving to end-of-day UTC, an explicit `Z`, a `+02:00` offset, and a zone-less timestamp pinned to UTC — so the end-of-day semantics that preserve the pre-#1150 answer are protected rather than assumed.

`RangeError` matches local practice: `packages/core/src/expiry.ts:127` throws the same type for an out-of-order `valid_from`/`valid_until`, also a well-typed string carrying an unacceptable value.

Follow-ups: #1179, #1180

### Notes

- `packages/core/src/tensions.ts:110` — `TemporalGateOptions.now` is the doc a caller actually reads, and it still describes only the old contract: an "ISO date treated as 'today'" that "defaults to the current date". Nothing there says a malformed value is now fatal. #1166 asked for the behaviour to be documented, and it is — thoroughly, on the internal helper, but not on the option the caller sets. Filed as #1179.
- `packages/core/src/validity.ts:177,180` — the `@throws` clause says the function throws when `now` is "provided but cannot be parsed", but the empty string is provided, cannot be parsed, and does not throw, because line 180 uses a falsy check. The new test pins that, so it is deliberate — but `now: ''` from a blank config field is the same silent-wrong-answer this change exists to stop. Filed as #1180, together with the accepted-forms list being narrower than the behaviour: `pinToUtc` also takes a zone-less timestamp, which is not valid RFC 3339.
- `packages/core/test/validity-instants.test.ts:231` — the empty-string assertion checks only the lower bound, unlike the omitted-argument case two lines above. A value far in the future would still pass.

### Blast radius, since a new throw deserves the check

Effectively nil on shipped surfaces. Neither the CLI (`packages/cli/src/commands/tensions.ts:173`) nor the MCP tool (`packages/mcp/src/tools.ts:3925`) ever sets `now`; there is no `--now` flag and no `now` field in the tool schema; and `evaluationInstant` is not re-exported from `packages/core/src/index.ts`. The throw is reachable only from library embedders and tests, so this does not turn a quiet wrong answer into a user-facing crash anywhere today.

Tests: PASS — full `packages/core` suite, 3679 passed. One unrelated failure, `test/injection-dedup.test.ts` ("four concurrent processes write ONE event"), reproduces identically on the base commit `d005139e` and is a concurrency-sensitive test flaking under local parallel load; CI is green on that same commit. The targeted files are 122/122.

Verified: the throw was reverted to the old fallback to confirm both new tests fail; every accepted `now` form was exercised; every caller of `evaluationInstant` was traced to confirm reachability.

amasen02:
Gentle nudge — this one is approved by both @miles-on-nightshift and @crtahlin with all checks green. Ready to merge whenever convenient; happy to make any final adjustments if needed. Thanks!

Issue #1115, opened 2026-09-03
Writes with no explicit scope silently auto-route to a shared/remote store — require a scope, or never auto-route to a shared scope


crtahlin:
## Problem

`plur_learn` (and `plur_learn_batch`, `plur_ingest`) accept a write with **no `scope`**. When scope is omitted, the write is auto-routed to the best-matching registered scope at ~0.5 confidence. When that best match is a **shared / remote team scope**, a personal or single-project engram silently lands in a team store and is pushed to its remote server.

The only signal is an `info` field in the response — *"No scope was provided; auto-routed to `<scope>`… Pass an explicit scope to override."* — which does not block and is easy to miss across a long session.

## Impact

A run of engrams that belonged to one person's project were auto-routed into a shared team scope and synced to the remote store — a genuine cross-scope leak. Because the copies are remote-only, cleanup then depends on the server being reachable and on admin-GUI access; it cannot be undone from the local CLI. The ID-collision bug in #1109 made the cleanup actively dangerous (a wrong engram was retired).

## Proposal (choose per policy)

1. **Require an explicit `scope` on every write path** (`plur_learn`, `plur_learn_batch`, `plur_ingest`, and the episode→engram paths). With no scope, reject the call with an error that lists the candidate scopes and how to set one. Safest, and matches "set scope PER engram by content" already in the guidance.
2. **If auto-routing stays, never auto-route to a shared/remote scope.** Default an unscoped write to `local`, and let the user promote deliberately with `plur_rescope`. Routing into a remote team store should require confidence far above 0.5 **and** explicit opt-in.
3. **At minimum**, turn the "no scope → auto-routed to `<shared scope>`" outcome into a loud, blocking confirmation rather than an easily-missed `info` field.

Until a hard block exists, the client guidance should also state plainly: never call a write without an explicit scope.

A server-side mirror of this belongs in the enterprise ingest path (filed separately there).

crtahlin:
## The routing signal is narrower than it looks, and two tools disagree

Confirmed again in a long session on an unrelated project, where the same silent misroute recurred. Two mechanism details worth having in the issue, both generic:

**The auto-route decides on the domain prefix alone.** A write whose `domain` begins with a segment that a registered shared scope declares in its `covers` list is routed there, regardless of the tags, the statement, or how project-specific the content is. So a personal engram about one project picks up a shared destination purely from the first part of its domain string.

[...]

PR #1178, opened 2026-09-09
fix: close audit findings in persistence, privacy, and recovery

plur9:
Interrupted writes, stale responses, malformed configuration, and inconsistent adapter behavior could lose valid data, weaken private routing, or hide failed operations. This PR remediates those bug classes across the public PLUR repository and adds regression coverage for their failure and recovery paths.

## Changes

- Preserve data through migration/restore interruption, atomic replacement, lock reclamation, PostgreSQL ownership loss, ID allocation, import, and source codemods.
- Keep outbox intent and receipts durable, retain retry identity, reject stale acknowledgements, and use validated remote state for cache visibility.
- Keep explicitly private writes local; bind model-assisted mutations to eligible, unchanged candidates; enforce runtime write validation and bounded archive scanning.
- Preserve configuration and credentials through canonical parsing and atomic updates; validate the complete Git commit index before sync.
- Make Python process cleanup bounded, expose adapter failures, preserve multimodal text and coherent transcript batches, and make telemetry append/retry accounting durable.
- Update dependency selections, pin privileged workflow inputs, restore setup coverage, and maintain process-test scheduling without relaxing assertions or timeouts.

The [audit ledger](https://github.com/plur-ai/plur/blob/audit/2026-09-08-closed-loop/docs/audits/2026-09-08-closed-loop.md) contains 18 in-scope cycles, all 42 findings/questions, root causes, sibling searches, fixes, regression evidence, and residual risks. It records 41 resolved findings and one external integration requirement.

## Regression verification

- `pnpm build` — passed.
- `pnpm test --maxWorkers=4` — **5,810 passed, 37 evaluated skips**; 411 files passed, five skipped.
- Final CLI change: rebuilt `@plur-ai/cli`; `pnpm exec vitest run packages/cli/test/import.test.ts --maxWorkers=1` — **12 passed**, including three new destination/override cases.
- `pnpm typecheck:tests` and `tsc --noEmit` for core, CLI, and migrate — passed.
- Isolated Python suites for Python/Hermes/LangChain/heartbeat — **312 passed**, with seven warnings.
- `bash scripts/smoke-release.sh` — **19 passed**, including packaged commands, migration output, and real PostgreSQL primary-store behavior.
- Workflow pin guard, 11 workflow YAML parses, modified-script syntax checks, spec vectors, and staged diff checks — passed.
- Local advisory matching: 602 npm entries with zero matches/unknowns; two Python optional-API matches were evaluated as outside the adapter's execution paths, not claimed patched.

The full suite preceded the final one-line CLI import destination fix and three added CLI cases. Hash comparison of 901 non-document inputs confirmed that only that command and its test changed afterward; the subsequent CLI build/tests, type checks, Python bridges, and packaged smoke cover that delta. No tests were newly skipped and no assertions or timeouts were weakened to obtain a pass.

## Review and compatibility notes

- F14 remains **OPEN for external verification**: complete remote retry deduplication requires a supported server that atomically commits the write and its idempotency receipt and validates replay authorization/payload consistency. The public client now durably retains request identity; passing client/contract-fixture tests cannot certify a server artifact. This PR does not claim complete integration convergence.
- Malformed existing configuration now refuses instead of enabling defaults; unsupported remote content refuses before sending; adapter persistence failures are explicit; source codemods refuse links and unsafe replacement. These are intentional changes to failure behavior.
- Physical power-loss behavior, unsupported filesystems/platforms, live model providers, and the hosted CI matrix are outside the completed local verification. CI on this PR supplies additional evidence; local results do not stand in for it.

Only public-repository changes are included. No merge or deployment is performed by this PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,7 +63,7 @@ jobs:
           --health-timeout 5s
           --health-retries 10
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4
         with:
           # Tags, so packages/migrate/test/method-list.test.ts can read the
           # 0.15.0 source and actually verify its sync -> async claim. Without
@@ -71,13 +71,14 @@ jobs:
           # vacuously — which is how a wrong claim about the past reached the
           # CHANGELOG in the first place.
           fetch-depth: 0
-      - uses: pnpm/action-setup@v4
+      - uses: pnpm/action-setup@b906affcce14559ad1aafd4ab0e942779e9f58b1 # v4
         with:
           version: 9
-      - uses: actions/setup-node@v4
+      - uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4
         with:
           node-version: ${{ matrix.node-version }}
           cache: pnpm
+      - run: python3 scripts/check-workflow-pins.py
       - run: pnpm install --frozen-lockfile
       - run: git config --global user.email "ci@plur.ai"
       - run: git config --global user.name "PLUR CI"
@@ -199,7 +200,7 @@ jobs:
     permissions:
       contents: read
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4
         with:
           # Full history: the comparison below needs the PR base commit, which a
           # shallow clone does not have.
@@ -233,13 +234,13 @@ jobs:
             echo "SKIPPING: no change to the embedder, the enrichment fields, or the bar."
             echo "run=false" >> "$GITHUB_OUTPUT"
           fi
-      - uses: pnpm/action-setup@v4
+      - uses: pnpm/action-setup@b906affcce14559ad1aafd4ab0e942779e9f58b1 # v4
         if: steps.relevant.outputs.run == 'true'
         with:
           # Required: package.json carries no `packageManager` key, so the
           # action cannot infer it. Same pin as the other two jobs.
           version: 9
-      - uses: actions/setup-node@v4
+      - uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4
         if: steps.relevant.outputs.run == 'true'
         with:
           node-version: 22
@@ -271,7 +272,7 @@ jobs:
           --health-timeout 5s
           --health-retries 10
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4
         with:
           # Full history so this job builds the same artefacts a release build
           # would see (release.sh runs from a full clone with tags). The
@@ -279,10 +280,10 @@ jobs:
           # apply here — this job never runs vitest; the comment used to be a
           # copy-paste of that one, claiming a reason that was not this job's.
           fetch-depth: 0
-      - uses: pnpm/action-setup@v4
+      - uses: pnpm/action-setup@b906affcce14559ad1aafd4ab0e942779e9f58b1 # v4
         with:
           version: 9
-      - uses: actions/setup-node@v4
+      - uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4
         with:
           node-version: 22
           cache: pnpm
--- a/.github/workflows/heartbeat.yml
+++ b/.github/workflows/heartbeat.yml
@@ -26,8 +26,8 @@ jobs:
   test:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
-      - uses: actions/setup-python@v5
+      - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4
+      - uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5
         with:
           python-version: '3.11'
       - run: pip install pytest
--- a/.github/workflows/hermes-compat-monthly.yml
+++ b/.github/workflows/hermes-compat-monthly.yml
@@ -39,9 +39,9 @@ jobs:
         python-version: ['3.11', '3.12']
 
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4
 
-      - uses: actions/setup-python@v5
+      - uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5
         with:
           python-version: ${{ matrix.python-version }}
 
@@ -50,7 +50,7 @@ jobs:
       # (test_memory_provider_entrypoint.py) exercises — so the canary needs
       # the CLI on PATH exactly as a user's machine does. Latest published,
       # not a workspace build: this job asks whether the shipped pair works.
-      - uses: actions/setup-node@v4
+      - uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4
         with:
           node-version: 22
       - name: Install the plur CLI (latest from npm)
@@ -83,7 +83,7 @@ jobs:
 
       - name: Open issue on failure
         if: failure()
-        uses: actions/github-script@v7
+        uses: actions/github-script@f28e40c7f34bde8b3046d885e986cb6290c5673b # v7
         with:
           script: |
             const fs = require('fs');
--- a/.github/workflows/hermes.yml
+++ b/.github/workflows/hermes.yml
@@ -24,8 +24,8 @@ jobs:
       matrix:
         python-version: ['3.10', '3.11', '3.12', '3.13']
     steps:
-      - uses: actions/checkout@v4
-      - uses: actions/setup-python@v5
+      - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4
+      - uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5
         with:
           python-version: ${{ matrix.python-version }}
       - run: pip install -e packages/hermes pytest
--- a/.github/workflows/langchain.yml
+++ b/.github/workflows/langchain.yml
@@ -19,14 +19,17 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ['3.9', '3.10', '3.11', '3.12']
+        python-version: ['3.10', '3.11', '3.12', '3.13']
+        dependencies: ['minimum', 'latest']
     steps:
-      - uses: actions/checkout@v4
-      - uses: actions/setup-python@v5
+      - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4
+      - uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5
         with:
           python-version: ${{ matrix.python-version }}
       # Install plur-ai from local source — avoids PyPI dependency before first publish.
       # plur-langchain's pyproject.toml requires plur-ai>=0.10.0; pip finds it here.
       - run: pip install -e packages/python
       - run: pip install -e "packages/langchain[dev]"
+      - if: matrix.dependencies == 'minimum'
+        run: pip install 'langchain-core==0.3.85' 'langsmith==0.8.18'
       - run: pytest packages/langchain -q
--- a/.github/workflows/pr-issue-guard.yml
+++ b/.github/workflows/pr-issue-guard.yml
@@ -32,7 +32,7 @@ jobs:
   issue-guard:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/github-script@v7
+      - uses: actions/github-script@f28e40c7f34bde8b3046d885e986cb6290c5673b # v7
         with:
           script: |
             const body = context.payload.pull_request.body || '';
--- a/.github/workflows/publish-langchain.yml
+++ b/.github/workflows/publish-langchain.yml
@@ -21,12 +21,12 @@ jobs:
     permissions:
       id-token: write   # required for Trusted Publishing
     steps:
-      - uses: actions/checkout@v4
-      - uses: actions/setup-python@v5
+      - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4
+      - uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5
         with:
           python-version: '3.11'
       - run: pip install build
       - run: python -m build packages/langchain
-      - uses: pypa/gh-action-pypi-publish@release/v1
+      - uses: pypa/gh-action-pypi-publish@dc37677b2e1c63e2034f94d8a5b11f265b73ba33 # release/v1
         with:
           packages-dir: packages/langchain/dist
--- a/.github/workflows/publish-mcp-registry.yml
+++ b/.github/workflows/publish-mcp-registry.yml
@@ -19,7 +19,7 @@ jobs:
       contents: read
       id-token: write          # required for GitHub OIDC auth to the registry
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4
 
       - name: Sync server.json version to the release tag
         run: |
@@ -38,10 +38,13 @@ jobs:
 
       - name: Install mcp-publisher
         run: |
-          url=$(curl -s https://api.github.com/repos/modelcontextprotocol/registry/releases/latest \
-            | grep -o '"browser_download_url": *"[^"]*linux_amd64.tar.gz"' | head -1 | cut -d'"' -f4)
-          echo "Downloading $url"
-          curl -L "$url" | tar xz mcp-publisher
+          # Immutable release and digest, verified from the upstream release.
+          curl --fail --show-error --silent --location --proto '=https' --max-time 60 \
+            https://github.com/modelcontextprotocol/registry/releases/download/v1.8.1/mcp-publisher_linux_amd64.tar.gz \
+            --output mcp-publisher.tar.gz
+          echo 'a06c9096dcb9727c13555b6be26c7effa707b01f06a4c561ba7a3635443cf2cc  mcp-publisher.tar.gz' > mcp-publisher.sha256
+          sha256sum --check --strict mcp-publisher.sha256
+          tar xzf mcp-publisher.tar.gz mcp-publisher
           chmod +x mcp-publisher
 
       - name: Login (GitHub OIDC) and publish
--- a/.github/workflows/publish-python.yml
+++ b/.github/workflows/publish-python.yml
@@ -21,12 +21,12 @@ jobs:
     permissions:
       id-token: write   # required for Trusted Publishing
     steps:
-      - uses: actions/checkout@v4
-      - uses: actions/setup-python@v5
+      - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4
+      - uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5
         with:
           python-version: '3.11'
       - run: pip install build
       - run: python -m build packages/python
-      - uses: pypa/gh-action-pypi-publish@release/v1
+      - uses: pypa/gh-action-pypi-publish@dc37677b2e1c63e2034f94d8a5b11f265b73ba33 # release/v1
         with:
           packages-dir: packages/python/dist
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -17,15 +17,15 @@ jobs:
   test:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
-      - uses: pnpm/action-setup@v4
+      - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4
+      - uses: pnpm/action-setup@b906affcce14559ad1aafd4ab0e942779e9f58b1 # v4
         with:
           version: 9
-      - uses: actions/setup-node@v4
+      - uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4
         with:
           node-version: 22
           cache: pnpm
-      - uses: actions/setup-python@v5
+      - uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5
         with:
           python-version: '3.11'
       - run: pnpm install --frozen-lockfile
--- a/.github/workflows/smoke.yml
+++ b/.github/workflows/smoke.yml
@@ -42,11 +42,11 @@ jobs:
     permissions:
       contents: read
     steps:
-      - uses: actions/checkout@v4
-      - uses: pnpm/action-setup@v4
+      - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4
+      - uses: pnpm/action-setup@b906affcce14559ad1aafd4ab0e942779e9f58b1 # v4
         with:
           version: 9
-      - uses: actions/setup-node@v4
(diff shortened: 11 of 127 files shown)


plur9:
Compatibility and release follow-up

Publishing this PR has not upgraded or deployed an enterprise installation. Compatibility when downstream applications adopt these changes is not yet fully verified.

Public behavior changes that consumers should account for:

- Unsupported rich content sent remotely now fails explicitly instead of silently losing unsupported fields; affected local source data or queued intent is retained.
- Malformed configuration and invalid inputs now fail explicitly instead of silently falling back.
- Missing provenance `license` remains absent. Consumers must not assume an implicit license string.
- Complete remote retry deduplication requires server support for atomic write/idempotency receipts and replay validation. Client changes alone cannot establish that guarantee (audit F14).

The completed local regression checks are documented in the PR description. Available integration evidence does not certify the exact final client revision against a currently supported, unmodified server release. Zero downstream upgrade regressions cannot yet be claimed.

Recommendation: hold publication of this change set until that exact client/server combination passes compatibility qualification, including ordinary writes, supported metadata preservation, explicit rejection of unsupported content, acknowledgement/cache handling, and lost-response/retry recovery. A new server release is required only if the supported version cannot satisfy those checks or required guarantees. If a server update is needed, also verify compatibility with supported older clients and make the required server version available before recommending the client upgrade.

This is a release recommendation, not a claim that a new server release is inherently required. The separate enterprise follow-up retains its private implementation details and evidence.

crtahlin (changes requested):
Requesting changes — three defects reproduced by probe, two of them in the write path this change was auditing.

This is a lot of good work. The migration recovery journal, the read-once backup validation, the PostgreSQL exclusive-session handling, the candidate snapshot that closes the dedup time-of-check-to-time-of-use window, and the workflow SHA pinning with a CI enforcer are all correct and genuinely tested — across the three slices I reverted twenty individual fix hunks and seventeen turned their named regression test red. The findings below are the exceptions, and two of them are siblings of defects this very audit fixed.

### Blocking

- **`packages/core/src/index.ts:7635,7794` — the outbox flush still uploads a row that has become private.** `_flushOutboxLocked` re-runs the target scope's sensitivity policy before pushing but never consults the row's own `visibility`. Reproduced: learn at a shared scope with the network down so the row queues, mark it `visibility: 'private'` through the exported `updateEngram` (which leaves `_outbox` in place), then `flushOutbox()` — the POST went out and its body literally carried `"visibility":"private"`. The client knew the row was private at send time and shipped it to the shared remote anyway. F29 closed this at `learnRouted`; the queue is the sibling path it did not close. One line in the pre-push guard — skip and demote when `visibility === 'private'`, the same treatment the policy-tightened branch already applies — closes it.

- **`packages/core/src/store/remote-store.ts:745-762` — a partial acknowledgement turns a committed write into a permanent failure.** The new validation accepts only an exactly-`{id}` reply or a fully valid `RemoteRowSchema` row. Verified against a stubbed server: `{id, created_at}`, `{id, scope, status}`, `{id, ok:true}` and `{id, warnings:[]}` all throw `invalid or out-of-scope acknowledgement` — after the POST was accepted. In `learnRouted` that throw lands in the outbox fallback, so a server-committed write is recorded as a failure and re-POSTed by every later `flushOutbox`, forever, because the same reply shape fails validation each time. Whether that duplicates rows in the shared store depends on the server honouring `Idempotency-Key`, which the ledger's own F14 entry marks unverified. It also contradicts the principle stated thirty lines away at `index.ts:3324` — "a bookkeeping write must never be able to undo, or appear to undo, a commit that succeeded" — and the audit-remote-authority test's own comment on the ID-only case, "never retry just because the echo is incomplete". Fall back to the ID-only behaviour whenever the acknowledgement does not validate; reserve the throw for a row that validates but names a foreign scope.

- **`packages/core/src/sync.ts:918` and `packages/core/src/store/async-fs.ts:76` — the directory fsync walks to the filesystem root, and can report a successful write as failed.** Two measured consequences. Cost: 4.6 ms per durable write syncing the leaf directory alone, against 38 ms for the full nine-level ancestry, over 200 iterations — paid by every learn, every telemetry counter, every config write and every migration journal write, and it fsyncs `/` and `/Users`, forcing filesystem-wide journal commits unrelated to the store. Correctness: with a grandparent at mode `0311` the rename succeeded and the file was correct on disk, but the call threw `EACCES` naming the unrelated ancestor — the caller is told a successful write failed, which is exactly the ambiguous-write state the adjacent comment tells callers to resolve, manufactured by the walk. The F44 root cause is a *newly created* ancestor, and `id-allocation.ts:22` already has the correct narrower shape: sync ancestry only when `mkdirSync` reports it created something. Neither effect is in the ledger and neither has a test.

### Should fix

- `packages/core/src/store/remote-store.ts:719-723` — `visibility` was added to the wire body, but the engram schema defaults it to `'private'` (`schemas/engram.ts:366`, applied at `index.ts:3418`). Every shared-scope write therefore POSTs `visibility: "private"` to the team store, including rows nobody marked private — confirmed in both probe bodies. On a server that honours the field, that hides every pushed engram from the team it was shared with; at best it destroys the field's ability to distinguish an explicitly private row. Not mentioned in the body.
- `packages/core/src/sync.ts:896`, `store/async-fs.ts:54` — **the temp-file fsync has no test.** Deleting `if (durable) fsyncSync(fd)` leaves all 15 tests in `audit-atomic-write.test.ts` and `telemetry-atomic-write.test.ts` passing; deleting `await handle.sync()` leaves all 12 passing. The fault probe at `audit-atomic-write.test.ts:14` is gated on `fstatSync(fd).isDirectory()`, so it only ever injects *directory* sync failures. The most basic crash-durability property in the change could be deleted by any future refactor without a red test.
- `packages/core/src/store/file-lock.ts:33,68` — the transition guard has no recovery path. A hard kill inside the `openSync(guard,'wx')` window leaves the guard file forever, and every later acquisition of that lock fails until a human deletes a hidden dotfile. Before this change no crash could permanently wedge a lock. The guard already writes its owner token at line 42 and nothing reads it; the same `holderIsAlive()` dead-PID proof used for the data lock would make it self-healing without weakening it.
- `packages/core/src/store/file-lock.ts:61` — `const legacy = !holder.includes(':')` removes time-based reclamation for every well-formed token, so a lock left by a different hostname is never reclaimable. That is the exact case the hostname field was added for — a store on a synced or networked volume — and it covers containers, whose hostname changes on every restart. Previously this self-healed after `DEFAULT_STALE_THRESHOLD`; now it needs manual intervention.
- `packages/claw/src/index.ts:103,141,153` and `context-engine.ts:187,384` — `recordEvent` is called unguarded at five sites, while `packages/mcp/src/telemetry.ts` wraps it with the comment "telemetry must never disturb a tool call". This change widens `recordEvent`'s throw surface considerably (corrupt or empty install-id, corrupt counters, non-safe-integer counter, lock failure after ~3.1 s of blocking `Atomics.wait`), so opt-in telemetry can now make claw's `/learn` report an error after the engram was already saved, and can silently skip memory injection for a turn.
- `packages/core/src/telemetry-counters.ts:123` and `telemetry-flush.ts:160` — two states are now unrecoverable without deleting a file by hand. A zero-byte install-id makes `readOrCreateInstallId` throw on every call, and a zero-byte install-id is precisely the residue an interrupted write under the *previous* non-atomic writer would leave, so the upgrade path can land there. Separately, one unparseable pending file now aborts the whole drain loop, stopping every later date from shipping; the old code deleted such a file specifically so the loop could not wedge. Preserving unreadable data is right for engrams; these are telemetry counters — quarantine and continue.
- `packages/core/src/backup.ts:312` — the same-day guard returns without `doneThisProcess.add(key)` and without repairing the state hint, so once the hint is lost the expensive path repeats on every write for the rest of the day, in every process. Measured on a 2.1 MB / 20,000-engram store: 20 calls take 0 ms with the hint intact and 2,626 ms without it. And when the snapshot is *not* intact the branch keeps the corrupt snapshot and refuses to write a fresh one, so a day with a provably broken snapshot ends with no verifiable backup and never self-heals.
- `packages/core/src/index.ts:6865-6872` — `rescope` to a shared remote scope pushes an engram whose `visibility` is `'private'` with no check and no warning; the sensitivity scan covers secrets, not the privacy flag. Reproduced. It is an explicit user action so it is arguably consented, but the client still sends a record labelled private to a shared store.
- `infra/heartbeat/query.py` — one torn line or one conflicting delivery anywhere in the window sets `data_quality.status` to `partial`, which replaces all three h004 verdicts with `UNAVAILABLE: partial input`. Because `append_record` deliberately preserves torn bytes rather than truncating, a single crash mid-append permanently disables the H004 metric for every window containing that day. The durability fix and the integrity gate combine into an unrecoverable state.
- `packages/hermes/plur_hermes/bridge.py:200` — the docstring still says the function "can never raise out of a timeout handler", but the new `PermissionError` branch re-raises from inside `except subprocess.TimeoutExpired`, replacing the `TimeoutExpired` and defeating the retry layer above.

### Notes

- `packages/core/src/index.ts:3061` — every remote-scoped `learn()` now runs the full outbox protocol (a complete `_primaryStore.load()`, the topological sort, two receipt-file rewrites) where it previously did one targeted append. Correctness is better; the hot-path cost on a large corpus is materially worse and is not measured anywhere.
- `packages/core/src/index.ts:4353-4390,7885-7900` — the documented `OUTBOX_ID_MAP_MAX` bound was removed with nothing replacing it. `state/outbox-id-map.json` and `state/remote-write-requests.json` now grow one entry per remote write or rescope forever, and the flush rewrites the whole map twice per pushed engram, so a flush of N entries against a map of M receipts is O(N·M) bytes written. Durable receipts should not be evicted like a cache, but they need a compaction story.
- `packages/core/src/project-config.ts:92-115` — `readProjectConfig` now throws on any malformed YAML. Fail-closed is right where the value is routing authority, but the same function supplies a purely decorative domain hint at `packages/mcp/src/tools.ts:1143`, so a stray tab in `.plur.yaml` now fails every `plur_learn` outright. The legacy input forms were verified to still parse: BOM, CRLF, inline comma `remote_scopes`, block-sequence and block-scalar.
- `packages/core/src/sync.ts:900` — `atomicWrite` replaces by rename, so where a config path is a symlink (a common dotfiles arrangement for `~/.cursor/mcp.json` or `~/.claude/settings.json`) the symlink becomes a regular file and the dotfiles link stops taking effect. `statSync` follows the link to read the mode; `renameSync` does not follow it to write. Worth care given #1059 was an MCP-config data-loss defect.
- `packages/core/src/learn-async.ts:226-232` — the `matches()` snapshot compares id, scope, status, statement, commitment, engram_version and the learn context, but not `visibility` or `pinned`. A concurrent change to either is silently overwritten by a model-selected UPDATE or MERGE. Both write locally, so not a leak, but it is a gap in the "unchanged candidate" claim.
- `packages/cli/src/commands/init.ts` — three sibling writers were left non-atomic while two in the same file were converted. Filed as #1183, since it is a separate sweep rather than part of this fix cycle.
- `packages/cli/vitest.config.ts:31` — the spawn-suite classifier duplicates `packages/core`'s with narrower coverage. Filed as #1184.
- `docs/audits/2026-09-08-closed-loop.md`, F14 — the claim that the public client "persists intent before POST" holds for `_flushOutboxLocked` and for `rescope`, but not for `learnRouted`'s first attempt, where the request id is generated in memory at `index.ts:3274` and only persisted in the catch. The outcome is still sound; the ledger states it more strongly than the code supports.
- `packages/core/test/audit-lock-recovery.test.ts:22` — the test that actually reproduces the F03 race is the four-process case, and with the guard removed it failed only 1 run in 8. As the CI gate for the most important invariant here, it would let a regression through roughly seven times in eight. Worth more contenders or more simultaneous reclaim attempts.
- `packages/langchain/pyproject.toml:10` — dropping Python 3.9 and moving `langchain-core` to `>=0.3.85,<0.4` are breaking changes with the version left at 0.10.0. Not yet on PyPI, so no republish conflict, but the version should move before first publish.

### What was and was not reviewed

At 127 files this went to three reviewers working in parallel on separate slices, and the coverage is uneven — worth knowing before reading the verdict as complete.

- **Persistence, locking, recovery** — thorough. Eleven fix hunks reverted individually, nine turned their named test red, including five real-PostgreSQL tests run against a pgvector container started for the purpose (68 tests that CI skips).
- **Privacy and write path** — thorough. Five fix hunks reverted, all five bit. Not verified: the enterprise server's actual POST reply shape, and whether it honours `Idempotency-Key` — neither lives in this repository, and the first blocking finding above depends on both.
- **Telemetry, CLI, infra, workflows** — thorough on what it ran. All six pinned action SHAs resolved against upstream refs and every one matches its claimed tag; the mcp-publisher v1.8.1 tarball was downloaded and its embedded SHA-256 reproduced byte for byte. Not run: `packages/langchain` (dependencies not installed locally), `packages/dsh`, `packages/mcp`, `packages/claw` beyond the telemetry call sites, and 40 of the 43 `cli-spawn` files.
- Nobody exercised real crash or power-loss behaviour. All durability findings rest on injected-fault simulation.

Two claims in the body were checked and hold. "Pin privileged workflow inputs": all 11 workflows read, no `${{ }}` interpolation of untrusted input into any `run:` block, and the single `pull_request_target` workflow never checks out or executes head code. "Without relaxing assertions or timeouts": true and understated — four `it.skip`s were un-skipped and made deterministic at the source rather than by widening assertions, and a vague `expect([0,1]).toContain(status)` became an exact assertion. The one exception is `packages/hermes/test/test_memory_provider_entrypoint.py:50`, which now monkeypatches the bridge and removes a real CLI roundtrip; the reason is legitimate — the old test executed an installed CLI against the developer's own store and could contact configured remotes — but it is a coverage reduction and should be recorded as one.

Tests: PASS on everything run. `packages/core` 3,563 passed / 34 skipped; core-process 155 passed; core-process plus core-pglite against real PostgreSQL 16 + pgvector, 397 passed over 33 files; `packages/cli` 368 tests plus 3 cli-spawn files; `infra/heartbeat/test_http.py` 35 passed; `packages/hermes` process-group tests 3 passed. Two failures were environment artifacts, not defects: `injection-dedup.test.ts` needs a prior `pnpm build`, and `test_memory_provider_entrypoint.py` needs `plur_hermes` pip-installed for its entry point to register.

crtahlin (changes requested):
Re-review of `9dea65b3`, unchanged since 9 September. All three findings from that review still reproduce by probe, and one of them turns out to have a harder fix than I proposed. Still requesting changes, and my recommendation has shifted from "fix these three" to "split this".

### The three prior findings, re-probed

**`packages/core/src/index.ts:7784-7818` — the outbox flush still uploads a row that has become private.** Reproduced: learn at a shared scope with the network down, mark the queued row `visibility: 'private'` through the exported `updateEngram`, call `flushOutbox()`. The POST body carried `"visibility":"private","scope":"group:plur/plur-ai/engineering"` and the flush returned `{flushed:1, failed:0}` after deleting the local copy.

**`packages/core/src/store/remote-store.ts:745-762` — a partial acknowledgement still turns a committed write into a permanent failure.** Against a stub server, `{id, created_at}`, `{id, scope, status}`, `{id, ok:true}` and `{id, warnings:[]}` each fell back to the outbox after one committed POST. Only a bare `{id}` survives. `packages/core/test/audit-remote-authority.test.ts:31-48` covers ID-only, a full foreign-scope row and a malformed full row — never the `{id, one extra field}` shape a real server actually sends — while its own comment states the correct rule: *"never retry just because the echo is incomplete"*.

**`packages/core/src/store/durability.ts:6`, consumed by `sync.ts:918` and `store/async-fs.ts:76` — the ancestry walk still reports a successful write as failed.** With a grandparent at mode `0311`, the rename succeeded and `payload-written` was on disk, and `atomicWrite` still threw `EACCES` naming that unrelated ancestor.

**A correction to my own September review on that last one.** I reported the cost as 4.6 ms leaf-only against 38 ms for the full ancestry. Re-measured across four 200-iteration runs, the walk costs about 0.4–0.7 ms over a leaf-only fsync — sub-millisecond, not thirty-three. The cost argument does not survive re-measurement and should be dropped; the correctness argument is what blocks, and it stands unchanged.

### My proposed remedy for the first finding does not work

In September I suggested a one-line guard in the pre-push path: skip and demote when `visibility === 'private'`. That would disable the outbox entirely.

`packages/core/src/schemas/engram.ts:364` defaults `visibility` to `private`, and `index.ts` sets `visibility: context?.visibility ?? 'private'` on every `learnRouted` placeholder. Every queued row therefore carries `visibility: "private"` on disk, including plain shared learns nobody marked private — confirmed by probe. At rest the field cannot distinguish *asserted private* from *unset default*; that distinction lives only in `context.visibility` at learn time and is never persisted.

This also raises the severity of something new in this branch. `remote-store.ts:722` adds `visibility` to the wire body; `main` does not send the field at all. So a plain shared learn now POSTs `visibility: "private"` to the team store, and a server that honoured it would hide every pushed engram from the team it was shared with. **That line should come out until the underlying distinction is persisted.** Filed as [#1225](https://github.com/plur-ai/plur/issues/1225).

### Blocking, new in this review

- **`packages/core/src/store/remote-store.ts:722`** — remove `visibility` from the wire body, per above.
- **`packages/langchain/plur_langchain/chat_history.py:62-84`** — a failed auto-learn now drops the conversation turn. `add_messages` runs `_learn_from_ai` before `self._messages.extend(batch)`, and `learn_from_text` raises `PlurMemoryError` on any bridge failure, so the transcript is never committed. `auto_learn` defaults to `True`. `tests/test_message_boundaries.py:92-106` asserts this as correct and assumes the caller retries, but `RunnableWithMessageHistory` does not. In a change titled "close audit findings in persistence", a convenience side effect must not take the primary record with it.
- **`.github/workflows/opencode-compat-monthly.yml`** — this branch's own enforcer fails on the merged tree. Against the merge of `origin/main` and this head, `scripts/check-workflow-pins.py` reports four violations at lines 59, 61, 65 and 114 of a workflow that landed on `main` after this branch was cut. Any rebase must pin it or CI goes red on the first push. Filed as [#1226](https://github.com/plur-ai/plur/issues/1226) so it can be fixed on `main` independently.

### Should fix

- `packages/langchain/plur_langchain/_utils.py:40-52` — `inject_to_text` now raises instead of returning `''`, making a transient PLUR outage fatal to the host chain on a *recall* path. The description characterises this only as "adapter persistence failures are explicit". With the Python 3.9 drop and the `langchain-core` floor move, the package carries three breaking changes at an unchanged version 0.10.0.
- `packages/core/src/store/id-allocation.ts:18-46` — every local learn now pays a durable atomic write where ID generation was an in-memory call. Measured on an APFS temp root over 100 iterations: `reserveLocalEngramId` 21.0 ms; `withLock` alone 0.20 ms; `atomicWrite` non-durable 0.22 ms; `atomicWrite` durable 10.2 ms; both together 28.3 ms. The cost is the fsync, and it is paid on top of the `engrams.yaml` write, which is also durable. Neither the change nor the ledger states it.
- `packages/core/src/store/id-allocation.ts:33-38` — a corrupt or zero-byte `state/id-allocations/<day>.json` makes every local write fail with "Cannot read durable ID allocation state; repair it before writing", because the catch converts a `JSON.parse` failure into that throw. On the hottest write path, with recovery being deletion of an undocumented file. Day-scoped, so it self-heals at 00:00 UTC.
- `infra/heartbeat/server.py:66-73` — the same root-walking fsync in a second implementation, inside the exclusive `flock` held for every accepted heartbeat, so all heartbeat writes serialise behind a filesystem-wide journal commit.
- `packages/core/src/store/file-lock.ts:73-80` — `releaseFileLock` rethrows any non-`ENOENT` error and is called from the `finally` of `withFileLock` and `withLock`. A throw there discards both the protected function's return value and any error it raised, so an `EIO` on the lock file masks the real outcome of a completed store write. The previous `releaseIfOurs` swallowed everything; catch and log instead.

### Notes

- `packages/core/package.json:20` — `"tar": "7.5.21"` reads as an exact pin, but the pre-existing root override `"tar": ">=7.5.21"` floats it and the lockfile resolves 7.5.22. Core has a floating direct runtime dependency used for archive extraction.
- `packages/cli/src/mcp-config.ts:300` — `writeConfig` now passes mode `0o600`, tightening an existing `0644` editor config to owner-only on write.
- `packages/cli/src/commands/login.ts:112-121` — `readPlurConfig` now throws on a malformed `~/.plur/config.json` and is called unguarded from the status report, so a stray byte takes down all of `plur status` with a message about signing in.
- `packages/core/src/store/async-lock.ts:177` — the async path now calls the synchronous `tryFileLock` on every acquisition. 0.20 ms on local disk; much worse on a networked or synced store root. This module's header exists to explain why the async path must not block the event loop, and the same diff edits that comment.
- `infra/heartbeat/server.py:118` — a zero-length body is rejected with "payload too large".
- `.github/workflows/ci.yml:81` — the pin check runs inside the Node version matrix, and only after checkout, `pnpm/action-setup` and `setup-node` have executed. A repository-wide immutability gate belongs in its own job, before anything else.

### What I think should happen to this branch

It is 18 commits behind `main` with 10 conflict hunks, and they fall in exactly the six files `main` also rewrote — `index.ts` routing from [#1115](https://github.com/plur-ai/plur/issues/1115), `remote-store`'s wire body from [#1221](https://github.com/plur-ai/plur/pull/1221), the project config. None of its content has landed by another route, so nothing here is wasted. But the conflicts are semantic rather than textual, and its own new CI enforcer fails on the merged tree.

Landing this as one commit means resolving a privacy-routing conflict, a wire-format conflict and a workflow conflict simultaneously, in a 127-file change, after a 13-day gap. I would split it: the workflow pinning, the pack extraction hardening, and the migration and PostgreSQL recovery work are clean and independently valuable, and could go in now. The privacy and remote-acknowledgement parts need reworking on top of current `main`, and the privacy half is blocked on [#1225](https://github.com/plur-ai/plur/issues/1225) regardless.

### Worth saying about the parts that are right

Replacing `execFileSync('tar', ['-xzf', ...])` with in-process extraction at `packages/core/src/packs.ts:55-115` — download cap, decompression cap via `gunzipSync maxOutputLength`, abort timeout, entry-count cap, per-entry size and depth caps, rejection of links and special files, path-escape checks, `preservePaths:false` — removes a shell-out on untrusted input and bounds a previously unbounded zip-bomb surface.

I resolved all six pinned action SHAs against upstream independently: `actions/checkout` `11d5960a` is v4.4.0, `actions/setup-node` `49933ea5` is v4.4.0, `actions/setup-python` `a26af69b` is v5.6.0, `actions/github-script` `f28e40c7` is v7.1.0, `pnpm/action-setup` `b906affc` is the commit behind the annotated v4 tag, and `pypa/gh-action-pypi-publish` `dc37677b` is the current head of `release/v1`. I downloaded `mcp-publisher_linux_amd64.tar.gz` v1.8.1 and reproduced `a06c9096dcb9727c13555b6be26c7effa707b01f06a4c561ba7a3635443cf2cc` exactly. The enforcer's regex exempts only `./`-local actions and would catch `docker://` and tag references. The single `pull_request_target` workflow still holds no code access.

`packages/core/src/store/file-lock.ts:25-31` — `holderIsAlive` returning `undefined` for any `kill(0)` failure other than `ESRCH`, rather than the old "anything but EPERM means dead", is the safe direction: an unprovable holder is assumed alive rather than stolen from.

`packages/core/src/learn-async.ts:225-232` — the candidate snapshot plus `structuredClone` genuinely closes the time-of-check-to-time-of-use window on model-selected UPDATE and MERGE, and the `commitment === 'locked'` check stops a model rewriting a locked row. The snapshot still omits `visibility` and `pinned`.

Tests: PASS — `pnpm install --frozen-lockfile` clean, `pnpm build` exit 0, full vitest run 400 files passed / 16 skipped, 5,686 tests passed / 164 skipped, no failures. Not run: the real-PostgreSQL suites (no `PLUR_TEST_POSTGRES_URL`), the Python suites for hermes/langchain/python, and `scripts/smoke-release.sh`.

Reviewed: `packages/core/src/index.ts`, `packages/core/src/store/*`, `sync.ts`, `packs.ts`, `learn-async.ts`, `learn-context-validation.ts`, all eleven workflows plus `scripts/check-workflow-pins.py`, `packages/langchain`, `infra/heartbeat`, the CLI config writers and the vitest project wiring. Not reviewed: `migrations/runner.ts`, `storage-postgres.ts`, `backup.ts`, `provenance-store.ts`, `history.ts`, `importers/engine.ts`, `packages/migrate`, `packages/claw`, the Python and hermes bridges, and the 465-line audit ledger.

PR #1195, opened 2026-09-15
feat(opencode): @plur-ai/opencode — memory adapter for the opencode harness

plur9:
Adds `@plur-ai/opencode`, an in-process plugin giving [opencode](https://opencode.ai) sessions persistent PLUR memory. Child of #1034 — it is the second in-process TypeScript adapter, which is what makes extracting a `HarnessAdapter` interface from `claw` + `opencode` possible rather than speculative.

Spec: `docs/specs/2026-09-15-opencode-plugin-design.md`
Plan: `docs/plans/2026-09-15-opencode-plugin.md`
Probe: `scripts/probes/opencode-plugin-probe.mjs`

Contract verified against **opencode 1.18.30 / `@opencode-ai/plugin` 1.18.30**, by probing the installed binary rather than reading its docs.

## The design decision, and the measurement behind it

opencode offers two places to put text in front of the model. Choosing between them is the whole architecture.

A text part pushed at `chat.message` is **persisted into session history**. Measured across three turns on a live binary:

| Turn | via `chat.message` | via `experimental.chat.system.transform` |
|---|---|---|
| 1 / 2 / 3 | 1 / 2 / 3 blocks in history | 0 / 0 / 0 blocks in history |

Linear, permanent growth — and every block after the first is *stale*, recalled against an older turn's query. The `system` array is rebuilt fresh per request, so it never accretes.

Hence the split: **`chat.message` runs recall once per turn and injects nothing**; **`system.transform` renders the cached block** on every model request, O(1), no recall. The extra firings (three in a tool-calling turn) are free because they only read a cache.

This independently reproduces the conclusion `@plur-ai/dsh` reached in its revision 2 — tail injection accretes, a re-rendered system-prompt section does not. Two harnesses, same answer; `ARCHITECTURE.md` records it so the next maintainer doesn't "simplify" recall back into `chat.message`.

## What's in it

- Recall → cache → system-prompt render, per the above
- `RenderPath` safety net: if `system.transform` ever stops firing after an opencode upgrade, latch to the `chat.message` fallback rather than silently injecting nothing
- Two learning paths mirroring `@plur-ai/claw`: the model's `🧠 I learned:` self-report, and user corrections at confidence ≥ 0.7
- Memory carried across context compaction, and learned from before the cut
- Project scope from `.plur.yaml` via `readProjectConfig`, mirroring `@plur-ai/mcp`
- `plur init --opencode`, writing both the `plugin` entry (automatic layer) and the `mcp` entry (explicit `plur_*` tools)
- Shared extractions into core — `renderMemoryBlock`, `extractLearnings`, `extractSelfReportedLearnings` — so claw and opencode share one implementation instead of two drifting copies. Each was behaviour-neutral for claw, gated on its tests passing unedited.

## Verification

Full suite: **397 files / 5665 tests passed**, 1 expected fail, 161 skipped, zero failures — run serially after the audit fixes.

Live acceptance gate (`packages/opencode/test/e2e.manual.mjs`) — builds and packs the real publishable tarballs, installs them as a user would, and drives four opencode turns:

- a fact taught in one session is recalled in a **fresh** one (all tools disabled in the harness, so it can only come from injected memory)
- the PLUR block actually reaches the rendered system prompt
- transcript accretion is **0 on every turn** across three turns
- `pack` rewrites `workspace:*` to a concrete version — shipping the literal would produce a tarball that cannot resolve its own dependency

Unit tests prove the hooks return the right objects; they cannot prove context reached the model. Given plur-hermes shipped a plugin that loaded, registered, reported healthy and injected nothing — three times in one day, past every intermediate check — that distinction is the point.

## Known limits

- **The package must be on npm before `plugin: ["@plur-ai/opencode"]` works.** opencode resolves a bare plugin name by having Bun fetch it from the registry at startup. Until publish, that entry resolves to nothing **with no error in opencode's log** — a config that looks correct and does nothing. Found by the gate; documented at the top of the README.
- `cwd` on the `Plur` constructor drives store auto-discovery only; it does not by itself isolate memory per project. `.plur.yaml` is what scopes recall and learns. The README says so rather than implying more.
- The gate runs BM25-only by default so it survives a loaded machine; `PLUR_E2E_FULL=1` exercises the hybrid path.
- `experimental.session.compacting` and `permission.ask` were not exercised against the live binary — compaction needs a full context window.

## Release notes

`CHANGELOG.md` declares `(#1195)`, so `release.sh`'s manifest gate passes.

`@plur-ai/opencode` ships on an independent version track at `0.1.0` (like claw and dsh), bumped only via `release.sh --opencode <ver>`.

## Follow-ups, not blockers

- nothing schedules a `PLUR_E2E_FULL=1` run, so the hybrid recall path has no e2e coverage by default
- cross-turn learn dedup (claw's `_learnIfNew`) is not ported; only intra-turn double-fire is debounced
- `@plur-ai/mcp` and `hook-inject` adopt a `.plur.yaml` scope without the directory-trust gate this branch adds for opencode — same shape, wider blast radius, owner-level call

---

## Deep audit (2026-09-16) — five lenses, and what they changed

After the branch passed eleven per-task reviews and a whole-branch audit, it was put through a deeper pass: data-loss, adversarial-security, and three evaluator lenses (evidence rigour, risk/tails, devil's advocate). They found things the earlier reviews did not. Everything below is fixed on this branch.

### A bug in shipped `@plur-ai/core` that inverted user prohibitions

`extractLearnings` captured only the text *after* a directive word, so the negation was discarded:

| You type | Was stored as |
|---|---|
| `never commit the API key to the repo` | `commit the API key to the repo` |
| `don't push directly to main` | `push directly to main` |

That statement is then rendered under **"These are things you have learned and should apply:"** and injected into the system prompt — turning a prohibition into an instruction to do the forbidden thing, permanently, with nothing visible to the user. Two patterns were affected.

This predates the adapter and affects `@plur-ai/claw` equally; it is fixed in core here. Claw's tests pass **unedited**, which confirms nothing depended on the inverted output.

### A cloned repo could register itself as a global engram store

`autoDiscover` defaults to true, and this plugin pointed store discovery at the session's git root — exactly where a committed `.plur/` sits. A repo merely *opened* in opencode appended itself to the user's global `~/.plur/config.yaml` (`shared: true, readonly: false`) and its engrams rendered at system authority, outliving the session and reaching Claude Code, claw and dsh too. Proved end to end, with a control.

Fixed with `autoDiscover: false`. A plugin loaded into someone else's agent host has no business writing their global config as a load-time side effect.

### `.plur.yaml` scope adoption now requires directory trust

A cloned repo's `.plur.yaml` could redirect recall and writes to the user's remote stores — carrying prompt text off-box under their own token.

The first fix attempt refused any scope resolving to a remote store. That was wrong: remote/team stores are the enterprise product, and it would have broken the paying use case to close the attack. The distinction is not remote-vs-local but **whether the user vouched for that directory**.

Now: trust on first use, the `direnv allow` / `git safe.directory` pattern. An untrusted directory's scope is ignored, with a warning naming the scope, the file, and the exact remedy. `plur trust .` grants it once, stored under `~/.plur/` — never in the project, so nothing inside a repo can vouch for itself.

Verified both ways: untrusted → **0** requests reach the team host; after `plur trust` → exactly the two legitimate requests under the user's own token.

### Also fixed

- `release.sh --opencode --dry-run` silently performed a **real** release — the flag consumed `--dry-run` as its version, then committed, tagged, pushed `main` and published four packages before failing. Same hole in `--claw` and `--dsh`; all three now reject flag-shaped and missing values.
- `plur init --opencode` overwrote an existing `mcp.plur`, including a user's custom `PLUR_PATH` — silently rerouting their store. A present entry is now left untouched and reported.
- The config was rewritten non-atomically (0 bytes between open and write); now uses `atomicWrite`, writing *through* a symlink rather than replacing it.
- `auto_learn` now gates **both** learning paths, matching claw. Auto-harvested writes are marked `claim_class: 'inferred'` so they are distinguishable from taught ones, and run through `detectPromptInjection` — which previously guarded only pack installs, on the premise that "my own conversation is trusted," which is exactly what automatic harvesting breaks.
- The injected memory block's own placeholder template could be harvested back as engrams. Filtered.
- `plur doctor` gained an opencode leg. It reports `unknown` — *"could not verify… NOT a clean bill of health"* — on offline, timeout, or any non-404, and "no" only on a confirmed 404. Given the package is not yet on npm, this is the only channel that can tell a user their plugin entry silently resolves to nothing.
- A monthly CI canary (`opencode-compat-monthly.yml`) type-checks against the **latest published** `@opencode-ai/plugin` and opens a deduped issue on drift. It runs `tsc --noEmit`, not just the suite — `pnpm test` transpiles via esbuild and never type-checks, so the `satisfies Hooks` guard would otherwise never re-fire.

### Evidence corrections

The accretion table's `system.transform` column was near-tautological: the meter read `output.messages` while that path writes `output.system`, so `0, 0, 0` was entailed by the instrument. The conclusion stands on a different, correctly-targeted observation — `system[]` going `1→2` on all three model calls of a tool-calling turn, never `2→3` — and the docs now cite that instead. The probe gained the missing negative control (`PLUR_PROBE_NO_SYSTEM`). The README's "nothing leaves your machine" was false on two default paths (the embedding model download, and remote recall once a store is configured) and has been corrected.

## Known issue in the shipped product — not fixed here

`.plur.yaml` carries `remote_url` / `remote_token`, honoured by `hook-inject`, which means a cloned repo can name an **arbitrary host** — not merely one the user already configured. `project-config.ts` already guards the equivalent `~/.plur.yaml`-in-HOME case for exactly this reason; the cloned-repo case is unguarded. This plugin deliberately ignores those fields and now has a regression test asserting it never reads them. The wider exposure needs its own issue.


crtahlin (approved):
Approved.

`@plur-ai/opencode` gives opencode sessions persistent memory, and the architecture rests on a measurement rather than a preference: a text part pushed at `chat.message` persists into session history and accretes one stale block per turn (1/2/3 across three turns), while `experimental.chat.system.transform` is rebuilt per request and accretes nothing (0/0/0). So recall runs once per turn and rendering happens per model request off a cache. `ARCHITECTURE.md` records the reasoning, which is what stops it being "simplified" back later.

The three extractions into core are behaviour-neutral for `claw` where they claim to be. I checked the one seam that could differ invisibly — the final `sections.join('\n')` in `assembleContext` — and the composed output is identical. The stronger evidence is that `claw`'s three changed files are all `src`: no test was edited, and the suite passes.

### Worth fixing, not blocking

- `packages/cli/src/commands/trust.ts:66-70` — `plur trust` prints "This authorizes \<configPath\>:" and lists `remote_url` beneath it. The grant does not authorize that. `hook-inject.ts:424-430` passes `.plur.yaml`'s `remote_url`/`remote_token` straight through with no trust check, before and after the grant — which is #1196. Someone reading this output could reasonably conclude trust now gates their remote store when it does not. The `scope` and `domain` lines are accurate; the `remote_url` line wants rewording, ideally pointing at #1196.

### Notes

- `packages/opencode/src/scope.ts` and `packages/core/src/learner.ts` both say "tracked separately" without citing #1196. The issue exists; naming it saves the next reader the search.
- `packages/core/src/learner.ts` — `extractSelfReportedLearnings`'s docstring says behaviour stays identical to claw's, but the new `PLACEHOLDER_BULLET_RE` filter does change claw's behaviour. An improvement, and the tests agree; the docstring just overstates it.
- `packages/opencode/src/turn.ts` — if `output.message?.id` is ever undefined, `markUserMessage` returns early and `append` can then capture the user's own text as assistant text. Unlikely given the probe-confirmed shapes, but it degrades silently rather than loudly.
- `.github/workflows/opencode-compat-monthly.yml` uses `@v4` action tags, which matches `ci.yml` and the hermes canary today. #1178's pinning enforcer will require updating it when that lands.
- The description cites 397 files / 5,665 tests; I measured 399 / 5,702, consistent with the later commits.

### Worth keeping

The negation fixes are the most valuable part. `never commit the API key` was being stored as the standing instruction `commit the API key`, and `Deploying straight to production is not allowed here` as `Deploying straight to production is` — at higher confidence (0.8) than the patterns it preempted, and positioned to claim the sentence first. Capturing the whole match, requiring a comma before `not`, and reordering narrow-before-broad fixes the class rather than the instance. `inversion-scan.ts` then gives users a way to find engrams already corrupted, deliberately refusing to auto-rewrite because a truncated fragment cannot be reliably reconstructed.

`autoDiscover: false` stops a cloned repo registering a store in the user's global config merely by being opened, and wrapping the plugin factory's initialization means a memory failure degrades memory instead of failing opencode's plugin load. `opencode-config.ts` refuses rather than coerces at every level, including the top-level-array case that would otherwise read as an inert success.

Verified: full suite in a clean worktree — 399 files, 5,702 tests passed, 1 expected fail, 161 skipped, exit 0, with claw's tests unedited. Also checked the claw join seam against the extracted renderer, confirmed the release script publishes `@plur-ai/opencode` with `--access public` and bumps `package.json` and `version.ts` together under the parity test, and confirmed #1196 tracks the hook-inject exposure this plugin's trust gate does not cover.

Issue #1196, opened 2026-09-16
A cloned repo's .plur.yaml can exfiltrate prompt text to a host it names (hook-inject)


plur9:
A cloned repository can exfiltrate the user's prompt text to a host it names, on every Claude Code prompt, with no user action beyond opening the repo.

Found during the security audit of #1195. That PR fixes the equivalent exposure in `@plur-ai/opencode`; this one is in the shipped Claude Code path and is **strictly worse**, because the attacker supplies the destination *and* the credential rather than needing the victim to already have a remote store configured.

## The path

`packages/cli/src/commands/hook-inject.ts` (~line 424) passes a project file's remote fields straight into the injection options:

```ts
...(projectConfig.remote_url && projectConfig.remote_token
  ? {
      remote_project: {
        url: projectConfig.remote_url,
        token: projectConfig.remote_token,
        ...
      },
    }
  : {}),
```

`projectConfig` comes from `readProjectConfig()` — a `.plur.yaml` walked up from the session's directory. A repository can commit one. Both the URL and the token are attacker-chosen, so no credential of the user's is needed; the payload is the user's task/prompt text, sent to the attacker's own endpoint.

## Preconditions

Clone a repo containing a `.plur.yaml` with `remote_url` and `remote_token`, then use Claude Code inside it. That is it. Cloning a repository to read it is an everyday action, the hook runs in the background on every prompt, and nothing surfaces the dial.

## Why this is worse than the opencode case that was fixed

| | opencode (#1195, fixed) | hook-inject (this issue) |
|---|---|---|
| Destination | only a host the user already configured | **any host the attacker names** |
| Credential | the user's own token | **the attacker's own token** |
| Needs a remote store configured? | yes | **no** |
| Attacker must guess a scope? | yes (org segment) | **no** |
| Surface | a new adapter | the flagship integration, shipped |

## The code already contains the correct reasoning — applied to the adjacent path

The comment immediately above the block says, of `dial: always` in `config.yaml`:

> "That is an explicit per-store opt-in to send (truncated) prompt text to that host, **not something a project can turn on**."

That is exactly the right principle. The line above it lets a project turn it on.

There is a similar precedent in `packages/core/src/project-config.ts`, whose header already refuses a `~/.plur.yaml` sitting directly in `$HOME` *because* "with `remote_url` set, that routes prompt text off-box". The home case is guarded; the cloned-repo case is not.

## Suggested fix

[...]

PR #1199, opened 2026-09-16
fix(cli): Codex and Antigravity reach PLUR Enterprise, gated on directory trust (#1198)

### LINKED ISSUE #1198: Only Claude Code supports .plur.yaml enterprise routing — 6 other integrations silently have no team memory
Found while fixing #1196. Reported as "the Codex adapter doesn't work with PLUR
Enterprise" — the diagnosis is correct, and it is wider than Codex.

## What is wrong

`hook-inject.ts` is the **only consumer of `remote_project` in the codebase**.
Every other integration reads `.plur.yaml` for `scope`/`domain` and drops the
remote fields on the floor, so enterprise team memory never arrives — with no
error, no warning, and no indication anything is missing.

| Integration | `.plur.yaml` remote recall |
|---|---|
| Claude Code (`hook-inject`) | **yes** |
| Codex (`hook-codex-inject`, `hook-codex-session-start`) | no |
| Cursor (`hook-cursor-session-start`) | no |
| Antigravity (`hook-agy-pre-invocation`) | no |
| opencode | no |
| claw | no |
| hermes (via `plur inject` / `recall`) | no |
| MCP server (`tools.ts`) | no |

`hermes` is a slightly different shape: it shells out to `plur inject` /
`plur recall`, and those CLI commands do not read project config at all, so it
inherits the gap rather than having its own.

The plumbing already works — `injectWithFallback` forwards its options verbatim
to `injectHybrid`, which dials when `remote_project` is present. The adapters
simply never put it there.

## There is a working path today

Stores mounted in `~/.plur/config.yaml` (via `plur stores add` /
`plur_stores_add`) are dialed at **core** level whenever the session scope's org
matches — that path is adapter-independent and works everywhere. Only the
`.plur.yaml` project-routing path is Claude-Code-only.

So an enterprise user on Codex can be unblocked today by mounting the store
rather than relying on `.plur.yaml`.

## Why it bites anyway

`plur init-remote` — the **documented** enterprise onboarding — writes
`.plur.yaml`. A customer following the documentation gets working team memory on
Claude Code and silence on every other harness, with nothing saying why.

## The fix, and the constraint on it

Do **not** fix this by copying `hook-inject`'s block into the other adapters.
That block is what #1196 was: adopting a project file's `remote_url` +
`remote_token` unchecked lets a cloned repo name its own host and supply its own
credential. Copying it would reproduce a prompt-exfiltration vector in six more
places.

It wants a **shared helper** that resolves the project config path once and
returns either the remote config or a refusal, carrying the #1196 directory-trust
gate with it — so every adapter gains enterprise support and the gate together,
with one place to audit instead of seven.

Rough shape:

1. Extract the resolve + trust-gate + refusal from `hook-inject.ts` into a
   helper (`packages/cli/src/lib/`, or core if the MCP server is to use it too).
2. Use it in `hook-inject` (replacing the inline copy), the two codex hooks,
   cursor, and agy.
3. Decide separately whether opencode, claw and the MCP server adopt it — they
   have different session models and may want the store-mounted path instead.
4. Each adapter needs a test that the remote leg dials when trusted and refuses
   when not. There is no adapter-level coverage of the remote path today, which
   is why this went unnoticed.

## Interim, cheap

Have `plur init-remote` state which harnesses `.plur.yaml` routing currently
covers, and point the others at `plur stores add`. One honest sentence is better
than silent absence.

## DONE_WHEN

- An enterprise user on Codex, Cursor, Antigravity or hermes gets team memory
  from a trusted `.plur.yaml`, verified against a real remote
- No adapter adopts a project's remote fields without the #1196 trust gate
- `plur doctor` reports, per detected harness, whether project remote routing is
  active
- Each adapter has a test covering both the trusted and untrusted paths

Refs #1196, #1197.

plur9:
Closes #1198. Live with a customer.

## The problem

Reported as *"the Codex adapter doesn't work with PLUR Enterprise"*. The
diagnosis was right; the scope was wider. **`hook-inject` was the only consumer
of `remote_project` in the entire codebase.** Every other integration read
`.plur.yaml` for `scope` and dropped the remote fields, so team memory silently
never arrived — no error, nothing to explain the absence. A customer following
the documented `plur init-remote` onboarding got working memory on Claude Code
and silence on every other harness.

## What this fixes, and what it does not

| Adapter | Before | After |
|---|---|---|
| `hook-codex-inject` | silent | `HIT /api/v1/recall auth=Bearer enterprise-token` |
| `hook-codex-session-start` | silent | dials |
| `hook-agy-pre-invocation` | silent | dials |
| `hook-cursor-session-start` | silent | **still silent — see below** |

### Cursor is NOT fixed here

An earlier revision of this PR did switch Cursor to the hybrid path so it could
carry the remote leg. **That was wrong and is reverted.** The file's own header
forbids it, with measurements:

> BM25-only, **deliberately** (PR #502's lesson)… Cursor's hook schema has no
> documented async/fire-and-forget option… Hybrid search loads the BGE embedder
> (~20s cold once the store passes a few thousand engrams, which is exactly the
> failure PR #502 fixed); BM25 alone measured 0.74s against 4,290 engrams.
> **Do not change this to try hybrid first.**

The hook is registered with a 10s timeout, so on a large store with a cold
process it would have been killed at the bound, injecting nothing.

The justification was wrong too, not only the change: "Cursor documents
sessionStart as fire-and-forget" refers to `additional_context` being *dropped
by a race* on Cursor's side — not to the hook being non-blocking.

The real fix, filed as **#1200**: the remote leg does not need the local
embedder at all (`_startRemoteRecall` sends query *text*; the server embeds).
A remote-with-BM25 mode gives Cursor team memory without touching the 10s bound.
The adapter test asserts Cursor does **not** dial, so the gap is pinned as known
behaviour rather than reading as an oversight.

## The trap this had to avoid

Closing #1198 by copying `hook-inject`'s block into each adapter would have
reproduced #1196 — a cloned repo naming its own host **and** supplying its own
token — in more places.

So **the gate travels with the capability**. `lib/project-remote.ts` is the
single place that decides whether a project's `.plur.yaml` may route prompt text
off-box: resolves the path **once** and reads from it (two walks would be a
TOCTOU between the file read and the directory trust-checked), gates the remote
fields on `plur trust`, **fails closed**, and returns either a usable remote
config **or a refusal the caller must surface**. An adapter cannot take the
capability without the gate. `scope`/`domain` stay ungated — local filters, no
egress. `hook-inject` uses the helper too, so there is one implementation rather
than two that drift.

## Tests

`test/adapter-remote-recall.test.ts` asserts **both halves per adapter** — dials
when trusted, refuses *and says so* when not — plus Cursor's known non-dialing.
Confirmed they bite: neutering codex's remote pass-through fails exactly the
codex "dials" test.

`hook-agy-pre-invocation.test.ts` asserted the workspace-root invariant
(evaluator audit B1) through a `readProjectConfig` mock the helper no longer
calls — **the invariant held, but nothing observed it**. Moved the recording to
`findProjectConfigPath`, the boundary the helper uses, stubbed rather than
delegated so a `.plur.yaml` elsewhere cannot influence the test.

## Migration

Existing enterprise users on Codex/Antigravity need `plur trust <project-dir>`
**once** — their `.plur.yaml` predates `init-remote`'s auto-trust (#1197). The
hooks now say exactly that instead of staying silent. New setups are trusted
automatically.

Full suite: **401 files passed, 15 skipped, 0 failed.**


miles-on-nightshift (dismissed):
Reviewed diff and verified:

**Security gate**: `resolveProjectRemote` fails closed on any error in the trust check — tested in `adapter-remote-recall.test.ts` lines 105-112 (no hits, refusal message emitted). This is the critical property from #1196, confirmed present in all four adapters.

**Behavior verified against test assertions**: `adapter-remote-recall.test.ts` covers both halves per adapter — dials when `trustDirectory` is called (hits > 0 after settle), refuses with a user-visible message when not trusted. A future adapter that takes the capability without the gate fails the second assertion. 401 tests passing, 0 failed.

**TOCTOU fix in agy test**: `readProjectConfig` mock replaced with `findProjectConfigPath` + `readProjectConfigFromPath` stubs — workspace-root invariant is now observed at the actual boundary the helper uses.

**CHANGELOG accuracy**: "every other adapter ... dropped the remote fields, so a customer ... got working memory in Claude Code and silence everywhere else" — matches what the diff shows (`hook-inject` was the sole `remote_project` consumer before this change).

One observation (not blocking): `resolve` on `settle()` uses `setTimeout` rather than a real server-close/flush signal. The 4s ceiling (40 × 100ms) is fine for CI but would mask a slow enterprise host in certain environments. Worth tracking.

APPROVE — customer-facing fix, trust gate intact, both halves covered.

PR #1203, opened 2026-09-16
fix(core): enforce hard-tier cap under the store lock, strip tier from packs (review of #1082)

miles-on-nightshift:
Fixes two issues surfaced in review of #1082 (feat: pinned two-tier model — hard-cap + priority eviction), which is included on this branch.

### 1. Hard-tier cap enforced inside the store lock (`index.ts`)

The cap check previously ran **before** `_withStoreLock`. The rationale was that `listPinned` is a read and the lock is not reentrant. But a cap is a read-modify-write on a shared total: N concurrent hard-tier writes each read the same current total, each conclude they fit, and all commit — cap silently overrun.

Measured on this branch: eight concurrent hard-tier writes, **zero rejected**, cap exceeded. The overrun is not cosmetic: hard-tier engrams are guaranteed injection AND bypass per-pack/per-domain fairness caps, so exceeding the budget crowds contextual recall out of the prompt — a context-budget denial-of-service reachable by anything that writes pinned engrams in parallel.

Fix: moved the check inside `_withStoreLock`. Reentrancy is not an issue: `listPinned` takes no lock of its own, it only reads through `_loadAllEngrams`.

### 2. `sanitizePackEngrams` strips `pinned_tier` and `pinned_priority` (`packs.ts`)

These fields traveled with `pinned` but were not stripped alongside it. They are inert today because pack selection filters on `pinned === true`, but that is an accident of the current call order. Leaving a pack's `pinned_tier: 'hard'` in the store means the next change that pins an engram for any other reason silently adopts a third party's claim on the guaranteed-injection tier. Same enumerate-vs-serialize drift as #381/#389.

### Tests added

`packages/core/test/pinned-two-tier-safety.test.ts` (+4 tests):
- Committed hard-tier total never exceeds the cap under concurrent writes (fails with 0 rejections without this change)
- A pack cannot ship `pinned_tier`/`pinned_priority`, including when `pinned` itself is absent
- An ordinary pack engram is untouched

## Files changed

| File | Change |
|------|--------|
| `packages/core/src/index.ts` | Move hard-tier cap check inside `_withStoreLock` |
| `packages/core/src/packs.ts` | Strip `pinned_tier` + `pinned_priority` in `sanitizePackEngrams` |
| `packages/core/test/pinned-two-tier-safety.test.ts` | +4 concurrency + pack-sanitization tests |

## Test plan

- [ ] `pnpm --filter @plur-ai/core test -- test/inject.test.ts` — all 35 pass
- [ ] `pnpm --filter @plur-ai/core test -- test/pinned-two-tier-safety.test.ts` — all 4 pass
- [ ] `pnpm --filter @plur-ai/core build` — clean build
- [ ] Concurrent hard-tier writes (>1): verify only those that fit are accepted, remainder rejected with a message naming existing engrams and token counts

### DIFF
--- a/packages/claw/clawhub/SKILL.md
+++ b/packages/claw/clawhub/SKILL.md
@@ -1,16 +1,26 @@
 ---
 name: plur-memory
-description: "Your memory stays on your machine. No cloud, no tracking, no API key. PLUR makes your OpenClaw remember — and shares that memory with every other tool you use."
+description: "Engram exchange + local-first memory for OpenClaw. No cloud routing, no API keys. Your corrections become shareable knowledge packs."
 version: 0.15.0
 ---
 
-# PLUR Memory
+# PLUR Memory — Engram Exchange Layer for OpenClaw
+
+Every correction you make is permanent knowledge — **stored on your machine, owned by you, shareable as packs**.
 
 Most memory plugins send your context to a cloud server. PLUR doesn't.
 
-Your memory lives on your disk — plain files you can read, edit, move, and back up. Zero cloud routing. Zero API keys. Zero cost.
+Your memory lives on your disk — plain YAML files you can read, edit, move, and back up. Zero cloud routing. Zero API keys. Zero cost.
 
-**Install in one command:**
+## What makes PLUR different
+
+**Engram Exchange.** Every correction becomes an atomic engram — a single fact with a source and context. You don't learn "systems design"—you learn the three-step pattern your team discovered last Tuesday. Engrams are **packs**: curated collections you can share, install, sell.
+
+**Local-first data control.** Your team's knowledge stays on your machine. No external calls. No vendor lock-in. No compliance friction. When you sync across machines, it's Git-based — your repo, your rules.
+
+**Works everywhere.** Memory syncs across Claude Code, Cursor (beta), OpenClaw, and any MCP-compatible tool. Correct something in OpenClaw — your Claude Code picks it up.
+
+## Install in one command
 
 ```
 openclaw plugins install @plur-ai/claw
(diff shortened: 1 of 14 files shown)


crtahlin (changes requested):
## Changes requested

The fix does what it says. I reverted `index.ts` to the pre-fix commit and the new concurrency test fails with exactly the reported symptom — 0 of 8 concurrent hard-tier writes rejected — then passes once restored. That is a regression test that has been shown to catch the bug, which is worth more than three that only assert the fix. Moving the check inside `_withStoreLock` is the right shape, the comment states the race precisely instead of gesturing at it, and the reentrancy claim holds: `listPinned()` reads through `_loadAllEngrams` and takes no lock of its own, so there is no nested acquisition.

Extending `sanitizePackEngrams` to `pinned_tier` and `pinned_priority` is also the right generalization, including the case where `pinned_tier` arrives without `pinned`. `installPack` is the only pack-write path and it re-saves the sanitized copy before hashing, so the strip is persistent rather than load-time only.

Three things block the merge.

### 1. The cap is overrun without any concurrency at all

`packages/core/src/index.ts:2490` charges admission as `ceil((statement.length + rationale.length) / 4) + 80`. Committed engrams are measured by `estimateEngramTokens`, which serializes the whole engram. The two disagree, and the gap is not bounded.

Measured on this branch: 40 small hard-tier writes admitted 9, for a committed total of **2,133 tokens against a 2,000 cap** — 237 actual tokens per engram against an 86-token estimate. Every field the estimate does not read is caller-controlled: `abstract`, `tags`, `domain`, `key_files`, `knowledge_anchors`, `structured_data`. A single write carrying 400 KB in `abstract` was admitted on an 81-token estimate and committed **100,232 tokens** — a 50x overrun, no concurrency involved.

That is the same guaranteed-injection context-budget denial of service this pull request exists to close, reached through a different door, and it falsifies the invariant the new test asserts at `packages/core/test/pinned-two-tier-safety.test.ts:48`. Build the engram shape first and charge `estimateEngramTokens` against it, so admission and accounting use one function.

### 2. The cap guards one write path and misses its siblings

`pinned_tier` is stamped in two places — `index.ts:2671` in `learn()`, and `index.ts:3229` in `_buildEngramShape`. `PINNED_HARD_TOKEN_CAP` is checked in one, at `index.ts:2491`.

`learnRouted` is the primary production write path; both the MCP `plur_learn` handler and the CLI go through it. Its local route delegates to `learn()` and is covered. Its **remote route builds through `_buildEngramShape` and POSTs with no cap check at all**, and the remote-failure path then saves locally, also unchecked. `setPinned`/`setPinnedAsync` at `index.ts:5610` and `:5660` re-pin with `{ ...e, pinned: true }`, preserving whatever `pinned_tier` the engram still carries — unpin clears only `pinned`, never the tier, so an unpin followed by a re-pin through `plur_pin` restores hard-tier membership with the budget never consulted.

Compounding it, `listPinned()` reads remote stores through `_loadRemoteCached`, a synchronous peek that returns empty unless something has warmed it. So even the path that does check is counting an incomplete total.

This is the guard-on-one-path-missing-its-sibling shape the codebase has been bitten by before, and this change touches the guard without extending it. The tier is unreachable from the MCP and CLI surfaces today — `pin_tier` appears nowhere outside core — which is the only reason these are latent rather than live. Wiring `pin_tier` up without closing them first would ship the bypass.

### 3. The branch is red, and the failure is in its own code

`npx vitest run` at the head: 4,695 passed, **2 failed**, both in `packages/cli/test/status.test.ts` — the provenance identity block this branch adds.

The cause is `packages/core/src/schemas/config.ts:234`. `PlurConfigSchema` is a plain `z.object().partial()` with no `provenance` key and no `.passthrough()`, so Zod strips `provenance:` out of `config.yaml` at load. `result.config.provenance` is always `undefined`, and `_buildProvenanceField()` can never read a configured identity or mode — every engram written on this branch gains a provenance block stamped `agent:unidentified` that cannot be configured or disabled.

This is branch code from `f18cbfad` rather than the fix commit, but it makes the pull request unmergeable, and it is precisely what CI would have caught had it ever run. The two head commits, `chore: re-trigger CI` and `ci: re-trigger CI on review/1082-pinned`, are empty no-ops that did not trigger anything; this branch has never had a check run against it.

### Also worth fixing

- `packages/core/src/inject.ts:371` — hard-tier overflow is dropped with a bare `continue`. Nothing is pushed to an evicted list, nothing reaches `buildEvictionWarnings`. Soft-tier evictions produce a warning naming what was dropped; the tier whose entire selling point is guaranteed injection produces silence. Since the cap can be overrun at write time, this is reachable: a user is told their engram is guaranteed, and it then vanishes from the prompt with no signal.
- `packages/core/src/index.ts:2487` — the cap check runs before dedup. Re-learning a statement that would dedup into an existing hard-tier engram, changing nothing but `write_count`, is rejected near the cap, because the check counts that existing engram and then adds the full cost of a write that would never create a row. Pre-existing from #1082, not made worse here, but the move was the opportunity to fix the ordering.
- `packages/core/src/schemas/engram.ts:459` — `pinned_priority` is schema-constrained to 1–100, but `learn()` stamps `context.pinned_priority` verbatim and nothing Zod-validates the engram before save. An out-of-range value sorts ahead of every legitimately-prioritized engram in soft-tier eviction. The sanitizer test's own use of `9999` is a fair illustration of the value a hostile pack would pick.

### The merge conflict is semantic, not a rebase chore

`main` has since grown a complete provenance subsystem — `packages/core/src/provenance.ts`, `config.provenance` with `generate`/`path`/`include_statement`/`default_license`, and its own `_buildProvenanceField`. This branch adds a competing minimal `_buildProvenanceField` reading `config.provenance.mode` and `.identity`, a shape `main` does not have. The two occupy the same config key incompatibly, which is what makes the state `CONFLICTING` rather than merely stale, and the branch's failing status tests are written against the model `main` has already replaced.

`main` carries no `pinned_tier` at all, so none of #1082 has landed by another route — this branch is the only home for the feature, 112 commits behind, with its parent pull request closed. Deciding which provenance model survives is a prerequisite to rebasing, not a follow-up to it.

### What I reviewed

The fix commit in full, plus the parts of the #1082 commit it depends on: `index.ts` `learn`/`learnRouted`, `inject.ts` tier logic, `packs.ts`, the config and engram schemas, and the status command. I did not review the claw/hermes documentation changes, the `tensions.ts` `MeasuredUnder` helper, or the 112 commits of divergence from `main`.

PR #1206, opened 2026-09-16
feat(release): --tweet-only, a correct reply template, and a tagged headline

plur9:
All three came out of re-posting the 0.20.0 announcement after it was deleted.

## `--tweet-only`

There was no way to post a release tweet without a full release run — and **a
release cannot be re-run to re-post**: its versions are already on npm and step
3.7 refuses, correctly. So a deleted or reworded announcement had no path back
short of posting by hand.

```sh
./scripts/release.sh 0.20.0 --tweet-only
```

Generates from the CHANGELOG, checks the 270-char budget, posts, exits. It
short-circuits at line 345 — **before step 0 and the version bump** — so it
cannot touch files, git, or npm.

Step 9's poster is **extracted into `post_tweet_pair()`** and called from both
places rather than copied. A second copy would drift, and the copy that drifts
is the one nobody exercises until a release night.

## The reply template was wrong

It read *"Claude Code / Cursor / Windsurf"*, then OpenClaw and Hermes.

- **Windsurf has no integration in this repo.** `grep -rli windsurf packages/*/src`
  finds nothing — it is generic MCP. Listing it beside harnesses that ship hooks
  implies a parity that does not exist.
- **Codex and Antigravity were missing**, and both have real hook integrations
  (`hook-codex-*`, `hook-agy-*`).
- **opencode was missing** — the headline of this very release — so the reply
  never carried `plur init --opencode`.

Now:

```
Claude Code / Cursor / Codex / Antigravity:
npm update -g @plur-ai/mcp @plur-ai/cli

opencode:
npm i -g @plur-ai/opencode && plur init --opencode

OpenClaw:
openclaw plugins install @plur-ai/claw

Hermes:
pip install --upgrade plur-hermes
```

## Headline

`opencode agents get persistent memory!` → `@opencode agents now remember.`
Tags the project rather than naming it flatly, and is shorter. The tweet is
derived from the `## 0.20.0` section's opening, so the CHANGELOG is where the
wording lives — the two stay in step by construction.

## Verified

`bash -n` clean. `--preview-tweet` still works (246/270), so the extraction did
not break the normal path. Posted live with `--tweet-only`:

- https://x.com/plur_ai/status/2100339432628826342
- https://x.com/plur_ai/status/2100339434482753655

No package versions touched; 0.20.0 is already published and unchanged.


miles-on-nightshift (dismissed):
CI all-pass. post_tweet_pair() extraction is the right call — single implementation eliminates the drift risk that --tweet-only would otherwise create. --tweet-only flag semantics are correct: publishes nothing, touches no files, guards correctly against re-running a version step 3.7 would refuse. CHANGELOG headline change is editorial (v0.20.0 already shipped). LGTM.

crtahlin (changes requested):
Requesting changes — the reply-template fix this pull request is named for is not in the commit, and the new flag's help text tells the operator it makes no network calls.

The `--tweet-only` mechanics are sound. It short-circuits at line 349, ahead of Step 0 and every bump, commit, tag, push and publish step, so it genuinely writes no files and publishes no packages. The `post_tweet_pair()` extraction is behaviour-preserving: a whitespace-normalised diff of the node payload against `origin/main` is identical, and the call site is a one-for-one substitution.

### Blocking

- **`scripts/release.sh:239` — the reply template is unchanged.** `REPLY` on this branch is byte-identical to `main`: still "Claude Code / Cursor / Windsurf", then OpenClaw and Hermes. There is no Codex line, no Antigravity, no opencode. The title, the branch name `fix/tweet-only-and-reply-template`, and an entire section of the description say this was fixed and quote the replacement text. It is not in commit `2c700f36` and not on `main`. Reproducing `generate_tweet` in isolation against this branch's CHANGELOG gives a 246-character main tweet, matching the description, and the 181-character *old* reply. Merging as-is closes the pull request with a third of its stated content undelivered, and the next release still posts the reply this pull request calls wrong.

- **`scripts/release.sh:33-34` — the help text for `--tweet-only` says it makes no network calls.** The lines "No file mutations, no network. Use to iterate CHANGELOG copy until the tweet fits 280 chars." were `--preview-tweet`'s continuation. The `--tweet-only` entry was inserted above them at lines 28-32, so they now read as its description. The one new flag that posts publicly to X is documented as safe to iterate copy with, and `--preview-tweet` loses its explanation entirely. In a script whose whole design is to gate before irreversible action, that is the inversion that causes a wrong public post. Moving those two lines back above the `--tweet-only` block fixes it.

### Should fix

- `scripts/release.sh:349` — `--tweet-only` short-circuits ahead of `--preview-tweet`, `--dry-run` and `--skip-tweet`. Exercising an isolated copy of the parse loop, `--tweet-only --dry-run`, `--tweet-only --skip-tweet` and `--tweet-only --preview-tweet` each set both variables, and in every case the live post happens and the other flag is ignored. Before this change the *safest* mode short-circuited first; now the most destructive one does, and `--skip-tweet` — which means "do not post to X" — is silently overridden. Either reject the combination at parse time or honour `DRY_RUN` inside the block.
- `scripts/release.sh:359-362` — `--tweet-only` posts without showing what it posts. Step 9 echoes `$TWEET` and `$REPLY` before calling `post_tweet_pair`; this path prints only a character count. It is the only path that publishes to a public account with no display of the copy, and it is exactly the path used while someone is reworking that copy.
- `scripts/release.sh:349` — nothing checks that `$VERSION` was actually released. A CHANGELOG section is the only precondition, so a typo or an in-progress section announces a release that does not exist and links to a tag that was never pushed. A `git rev-parse -q --verify v$VERSION` would cost one line.

### Notes

- The `# Usage:` header at line 3 and the "Step order" block at lines 57-79 were not updated for `--tweet-only`, though the runtime usage echo at line 165 was.
- `CHANGELOG.md:5` — "@opencode agents now remember." is interpolated verbatim into the tweet, so it becomes a real X mention of whichever account owns `@opencode`. Worth confirming that handle belongs to the project before it is baked into every regeneration of this announcement.
- Re-running `--tweet-only` for the same version posts a duplicate pair; X's duplicate-content rejection is the only guard, and it surfaces as a failed run rather than a clear message.
- **The seven green checks tell you nothing about this diff.** Nothing in the test suite or any workflow reads `scripts/release.sh` — every reference is prose in a comment, `ci.yml` runs the unrelated `smoke-release.sh`, and there is no shellcheck or `bash -n` step anywhere. Filed as #1214.

Worth keeping: credentials are handled correctly — read from `process.env` inside node, never through argv, never echoed, and the failure path prints the API response and status rather than signing material. `--tweet-only` also takes no argument, so it cannot swallow a following flag, and `--claw --tweet-only` fails loudly; the B1 flag-swallowing class from the 0.20.0 audit does not reappear.

Tests: PASS — full suite after install and build, 401 files / 5713 tests, 1 expected fail, exit 0. `scripts/release.sh` was never executed in any mode; `bash -n` is clean, and the parse loop and `generate_tweet` were exercised as isolated copies.

plur9:
@crtahlin — review points addressed in `0d0bc52f`. Both blocking points were right.

### Blocking

- **The reply template was not in the commit.** It is now. `REPLY` lists Claude Code / Cursor / Codex / Antigravity (all through `@plur-ai/cli`), opencode (`npm i -g @plur-ai/opencode && plur init --opencode`), then OpenClaw and Hermes. Windsurf is gone. I checked each line against the tree first: `@plur-ai/opencode` is published (0.1.1), `init --opencode` exists (`packages/cli/src/commands/init.ts:1122`), the Codex and Antigravity hook commands exist, and nothing under `packages/*/src` mentions Windsurf.
- **The `--tweet-only` help text said "no network".** Fixed. That text is back under `--preview-tweet`, and `--tweet-only` now opens with "POSTS PUBLICLY to X".

### Should fix — all three done

- **Flag precedence.** `--tweet-only` now refuses to run with any other flag, at parse time and before anything else happens: `--dry-run`, `--skip-tweet`, `--preview-tweet`, `--trust-ci`, `--no-website`, `--claw`, `--dsh` or `--opencode`. The error points to `--preview-tweet` for seeing the copy without posting.
- **Posting without showing the copy.** It now prints the main tweet and the reply, each with its character count, before `post_tweet_pair`, the same way step 9 does.
- **No check that the version was released.** It now runs `git ls-remote --exit-code --tags origin refs/tags/v$VERSION`. That checks origin rather than a local tag, because the tweet links to the pushed tag, and an aborted release can leave a local-only tag. It reports "not a tag on origin" separately from "could not reach origin".

### Notes

- The `# Usage:` header and the step-order block now cover `--tweet-only`.
- The reply grows from 181 to 254 characters, and it had no length check anywhere even though it posts after the main tweet. It now gets the same 270 budget in `--tweet-only`, `--preview-tweet` and step 3.5.
- A duplicate post rejected by X now gets its own message instead of a bare 403.
- `@opencode` is **not resolved in this push**. Whether that handle belongs to the project still needs confirming, so the CHANGELOG line is unchanged for now.
- #1214 (no CI step reads `release.sh`) stays separate.

### Verified

- `bash -n` is clean. `shellcheck -S warning` reports only the existing SC1090 on the env `source`.
- I ran a copy of the script with `post_tweet_pair` stubbed out and no credentials loaded:
  - `--tweet-only` combined with `--dry-run`, `--skip-tweet`, `--preview-tweet` or `--claw 0.5.0` → refused, exit 1.
  - `9.9.9 --tweet-only` → "not a tag on origin", exit 1.
  - `0.20.0 --tweet-only` → prints the 246-char tweet and the 254-char reply, then reaches the stub.
- The real `0.20.0 --preview-tweet` → exit 0, and both the tweet and the reply are within budget.
- Nothing was posted.

Issue #1127, opened 2026-09-04
forget(scope:) direct path reports false absence when the store is unreachable


crtahlin:
Follow-up from [plur-ai/plur#1114](https://github.com/plur-ai/plur/pull/1114).

`forget()`'s explicit-scope shortcut — the `targetScope && targetScope !== 'primary'` branch in `packages/core/src/index.ts` — calls `driver.getById(serverId)` and throws `Engram "<id>" not found in store "<scope>"` when it returns null.

`RemoteStore.getById` swallows transport errors and returns null, so an expired token or an unreachable server is reported as absence. That is the same defect [#907](https://github.com/plur-ai/plur/issues/907) and [#1109](https://github.com/plur-ai/plur/issues/1109) removed from the walking path.

It matters more now, because [#1114](https://github.com/plur-ai/plur/pull/1114)'s new error explicitly advises the caller to retry with `scope: "<scope>"`, which routes them straight into this path. During the outage that triggered the advice, following it produces exactly the false-absence claim the change exists to eliminate.

Apply the same tri-state `probeById` resolution here and throw a "could not reach" error instead of "not found".

PR #1209, opened 2026-09-17
fix(core): defer namespaced-id throw in forget() walk until completion (#1126)

### LINKED ISSUE #1126: Sequence the namespaced-id forget throw against the scope-unaware bareEngramId change
Follow-up from [plur-ai/plur#1114](https://github.com/plur-ai/plur/pull/1114). Affects [#1122](https://github.com/plur-ai/plur/pull/1122).

[#1114](https://github.com/plur-ai/plur/pull/1114) adds a hard throw in the `forget()` remote walk, gated on `id !== serverId`. That gate reads as "`_stripRemotePrefix` stripped this store's prefix, therefore this store is the intended target."

[#1122](https://github.com/plur-ai/plur/pull/1122) changes `_stripRemotePrefix`'s fallback from `return id` to `return bareEngramId(id)`, which strips any two-to-four character uppercase prefix without consulting the store's scope.

With both landed, the gate is true for **every** configured store whenever the caller identifier is namespaced. The walk then throws `Cannot reach "<scope>" to retire …` on the first unreachable store, even when that store never held the engram — and a retirement that would have succeeded on a later, reachable store fails outright.

Resolution options:

- Keep `_stripRemotePrefix` scope-aware and give the recall identifier surfacing in #1122 its own bare-form helper. This is the preferred shape: the scope-aware inverse already exists and should not be weakened.
- Or replace the gate with a scope-aware predicate, for example `namespaceEngramId(serverId, entry.scope) === id`.

Add a test with two configured stores, one unreachable, where the namespaced identifier belongs to the reachable one.

Separately, the throw exits the walk immediately rather than after it. `storePrefix` is a lossy three-character derivation, so two configured scopes can share a prefix. Deferring the throw until the walk completes, and firing it only when nothing was retired, keeps the actionable message, preserves multi-store completion, and blunts this interaction.

miles-on-nightshift:
## Problem

`#1114` added a hard throw in the `forget()` remote walk gated on `id !== serverId`. That gate reads as "`_stripRemotePrefix` stripped this store's prefix, therefore this store is the intended target."

`storePrefix()` is a lossy 3-char derivation — two distinct scopes produce the same prefix. For example:

```
storePrefix('group:test')  === 'GTE'
storePrefix('group:tempo') === 'GTE'
```

Both stores accept `ENG-GTE-...` as their own namespace. So when the walk encounters the unreachable store first in config order, it throws `Cannot reach "group:test" to retire "..."` — never reaching `group:tempo`, which holds the engram and is reachable.

This is the ordering defect: an early-exit throw on the wrong store aborts a retirement that would have succeeded.

## Fix

Replace the immediate `throw` with a deferred error (`pendingUnreachableError`):
- Record the error and `continue`
- Fire it only after the walk completes without a retirement
- If a subsequent store retires the engram, the deferred error is discarded silently

After the walk, `pendingUnreachableError` surfaces **above** `"Engram not found"` (cannot reach is more actionable than false absence) and **below** `refusedBy` (a refused DELETE proves presence).

Also updates the inline comment that claimed `"Throw immediately"`.

## Test

`packages/core/test/forget-namespaced-walk-order.test.ts` — three scenarios:

1. Two stores sharing prefix `GTE`, unreachable first in walk order. With the bug: throws on the first iteration. With the fix: continues and retires from `group:tempo`. ✓
2. Reverse order (reachable owner first): retires successfully, deferred error is never fired. ✓  
3. Both stores unreachable: `"Cannot reach"` surfaces at walk-end — not `"Engram not found"`. ✓

The existing `forget(namespaced-id) with unreachable remote throws actionable error` test (single store, `remote-routing.test.ts`) still passes — the deferred error fires correctly when there's only one prefix-matching store.

## References

Closes #1126. Pairs with #1127 (A9) — A9 fixes the direct-scope path unreachable defect; this PR fixes the walk-path ordering defect. A9 was settled first because the error message the walk path surfaces changes once A9 lands.

## Test plan
- [ ] New test file passes: 3/3 ✓
- [ ] `packages/core/test/remote-routing.test.ts` all pass ✓ 
- [ ] `packages/core/test/walk-cannot-tell.test.ts` all pass ✓
- [ ] `packages/core/test/forget-direct-scope-unreachable.test.ts` all pass ✓
- [ ] Full suite green (188 tests across 12 test files run in worktree)


crtahlin (approved):
## Approved, with three follow-ups filed

The change is correct and the deferral is safe on the axis that matters. The only retirement inside the walk is followed immediately by `return`, so there is no mutate-then-throw and no partial-success rollback question — what changes is *when* the error surfaces, not which engram is deleted, except in the prefix-collision case this pull request is explicitly about. [#1114](https://github.com/plur-ai/plur/pull/1114)'s guarantee is preserved: a store that was never reached still never gets reported as "not found", the same message just fires at walk end instead of mid-walk.

The core test earns its place. It asserts *which* engram was retired rather than merely that nothing threw — exactly one DELETE against the reachable host carrying the bare id, and zero DELETEs to the unreachable one. I reverted the deferral in a worktree and re-ran it: it goes red with `promise rejected "Cannot reach \"group:test\"..." instead of resolving`. The other two cases pass before and after by design, which is the right shape for them — the reversed-order case is a regression guard, and the both-unreachable case guards the new deferred fire and would fail with "Engram not found" if the throw were dropped rather than deferred.

Rewriting the superseded [#1114](https://github.com/plur-ai/plur/pull/1114) comment rather than leaving it to contradict the code is also right, and the new lossy-prefix reasoning makes the function legible to the next reader.

### Three gaps, none blocking

**The ordering claim is untrue** — [#1217](https://github.com/plur-ai/plur/issues/1217). `if (pendingUnreachableError) throw` sits directly *above* the `refusedBy` check, while the comment above it and this pull request's description both say it "Sits below `refusedBy` because a refused DELETE proves presence". Reproduced with a two-store setup — unreachable `group:test` first, `group:tempo` owning the engram and refusing the DELETE with 403 — and the thrown message is `Cannot reach "group:test" ...`. The user is told to retry a connection when the engram exists and the token simply lacks delete rights. The read-only-store throw inside the loop already outranks the deferred error too, so the current ordering is inconsistent with itself as well as with its comment.

**The recorded error is discarded silently** — [#1218](https://github.com/plur-ai/plur/issues/1218). When a later store does retire, the namespaced branch neither pushes to `unreachedStores` nor calls `logger.warning`, unlike the bare-id branch immediately below it. The caller sees a clean success and is never told that another prefix-matching store could not be reached — which is precisely the case where the retirement may have hit the wrong store. `pendingUnreachableError` is also overwritten rather than accumulated, so with two unreachable stores only the last is named.

**The gate is still prefix-based, not scope-aware** — [#1216](https://github.com/plur-ai/plur/issues/1216). Continuing past an unreachable prefix-matching store lets the walk retire a *different* store's engram whenever that store shares the lossy three-character prefix and holds the same bare id. Bare ids are minted per store, so a same-day, same-sequence collision is plausible rather than theoretical, and `remote-store.ts:758-760` names this exact failure. The walk already has the fetched engram in `found`, which carries `scope`, so it could confirm `found.scope === entry.scope` before issuing the DELETE. That is the complete fix; this change is the blunting that [#1126](https://github.com/plur-ai/plur/issues/1126) also asks for.

### On closing #1126

Defensible. The main body's preferred resolution — keep `_stripRemotePrefix` scope-aware, give recall its own bare-form helper — is already satisfied on `main`: `_stripRemotePrefix` still returns `id` unchanged on a prefix miss, and [#1122](https://github.com/plur-ai/plur/pull/1122) landed `bareEngramId` only in the MCP recall surface. This implements the issue's closing paragraph and adds the two-store test it asks for. [#1216](https://github.com/plur-ai/plur/issues/1216) is filed so the scope-aware gate is tracked before the issue closes rather than after.

### Tests

`packages/core`: 3,567 passed, 36 skipped, 1 expected-fail. Targeted runs of `forget-namespaced-walk-order`, `remote-routing` and `walk-cannot-tell`: 43/43 green. The one `injection-dedup` failure is an invocation artifact, not a defect here — that test resolves `packages/core/dist/index.js` from `process.cwd()` and passes 21/21 from the repository root. `packages/mcp` was not run: the change does not reach it, and no `Cannot reach` assertion exists anywhere under `packages/mcp`.

Issue #1207, opened 2026-09-17
opencode cannot reach PLUR Enterprise: the plugin drops the .plur.yaml remote fields that #1196 made safe to adopt


plur9:
Split out of #1198, which spread the trust-gated remote leg to Codex and
Antigravity. `@plur-ai/opencode` was not revisited, and the timing says why:

| Time (2026-09-16) | Commit |
|---|---|
| 15:31 | `96507aa4` the opencode plugin (#1195) |
| 15:56 | `6c6a9bf4` gate `remote_url`/`remote_token` on directory trust (#1196) |
| 17:18 | `ec06b224` Codex and Antigravity reach Enterprise, gated on trust (#1198) |

The plugin merged 25 minutes before the gate it needed existed. Its
`scope.ts` says as much, and defers rather than refuses:

> Deliberately narrow: `readProjectConfig` also returns `remote_url` /
> `remote_token` / `remote_scopes` (honored by the CLI's `hook-inject`), and
> this function's return type does not have those fields at all — the plugin
> never reads them. See `packages/core/src/project-config.ts` for the wider
> exposure that leaves open (tracked separately, not an opencode-plugin
> concern).

What it was deferring to shipped half an hour later.

## Symptom

An Enterprise customer opens a repo whose `.plur.yaml` carries the documented
`plur init-remote` output — `scope`, `remote_url`, `remote_token` — in
opencode. The plugin adopts `scope` (trust-gated, D2) and drops the remote
fields. `injectHybrid` is then called with a scope but no `remote_project`,
and `_remoteRecallHosts` (`core/src/index.ts:4495`) dials a host only when a
**registered store** in `~/.plur/config.yaml` shares the session scope's org
segment. A project declaring a remote endpoint registers nothing, so nothing
is dialed.

Team memory silently never arrives. Explicit `plur_*` MCP tool calls still
reach Enterprise, so it reads as half-working rather than broken — the same
failure mode #1198 described, in the one adapter #1198 did not cover.

## Why this is not an opencode restriction

Nothing in the host blocks it. `chat.message` already awaits `injectHybrid`,
whose remote leg dials whenever a trusted scope implicates a registered store
— so remote calls from inside the plugin are a live path today, with network
access, async hooks and the latency budget all already exercised. Unlike
Cursor (#1200), there is no bounded-hook problem to solve first.

What is missing is reading two fields and passing them through.

## Current workaround

Register each Enterprise scope as a store, then trust the directory:

[...]

PR #1210, opened 2026-09-17
ops(release): open the 0.20.1 section, and keep the enterprise gate

plur9:
Release prep for 0.20.1 + `@plur-ai/opencode` 0.1.1 (the #1207 patch, merged in #1208).

## Why the heading rename is required

`release.sh` never rewrites it. Both the tweet builder (line 206) and the manifest gate (line 715) `awk` for a literal `## <version>` section, and Step 3.55 refuses a heading still marked `(unreleased)`. With `## Unreleased` the release aborts with "CHANGELOG.md has no section for 0.20.1".

The section already declares `(#1207, #1208)`, which is what the manifest gate compares against the squash subject.

## The enterprise gate

`packages/opencode/test/e2e-enterprise.manual.mjs` — the live run from the #1207 verification, kept rather than discarded. `e2e.manual.mjs` proves local recall reaches the model; this proves the `.plur.yaml` remote path delivers **team** memory. Two things make a pass mean something:

- the local store is left with **zero engrams and zero configured stores**, asserted before the session, so a pass cannot come from local memory
- the work repo's `.plur.yaml` is written by the real `plur init-remote`, trust grant included, so it exercises the onboarding path a customer follows

It asserts on the injected block via a read-only observer, not on the model's prose, and then independently on `cache/remote-health.json` — a real per-host `POST /api/v1/recall` outcome. The token is read from **stdin**, never argv or env, because the agent under test has a bash tool and the sibling gate observed one run `env`.

## Verification

**Enterprise gate, against `plur.datafund.io`:**

\`\`\`
local store: 0 engrams, 0 stores
[plur:opencode] project remote: https://plur.datafund.io
[plur:opencode] recall for ses_…: 7 engrams
PASS: every expected remote phrase is in the injected block (NOT Docker, git pull)
PASS: https://plur.datafund.io dialed, last_state=ok
\`\`\`

**Official gate** (`node packages/opencode/test/e2e.manual.mjs`): all three assertions pass — recall reached the model, injection path live, transcript accretion 0 across 3 turns (4 invocations, `openai/gpt-5.5`).

**Unit:** `pnpm --filter @plur-ai/opencode test` — 80 passed.

## Two server behaviours found, both by design

- An engram written through `POST /api/v1/engrams` lands in a policy review hold (`commitment: draft`, `review_state: policy`) and is excluded from recall until approved; `PATCH` refuses with "Cannot update a held engram." The gate's docs now warn about it, since it is the trap for anyone seeding a fixture instead of asserting on live engrams.
- A statement containing a hostname draws `review_flag: sensitive_content`, which also removes it from recall.

Both fixtures I seeded while finding this out are retired (`ENG-2026-09-17-012`, `-013`).

## Release command

\`\`\`sh
./scripts/release.sh 0.20.1 --opencode 0.1.1
\`\`\`

Core has to ship because the shared trust gate now lives there and the plugin does not bundle it. No version files are touched here — `release.sh` owns every bump.


miles-on-nightshift (approved):
CI all-pass. The 0.20.1 CHANGELOG rename is required for release.sh (awk gate at line 715 matches exact '## <version>'). e2e-enterprise.manual.mjs: token on stdin is the correct hardening (prevents env-read by the agent under test), zero-local-store assertion is the right discriminating condition. README docs are clear on how to supply the correct phrases. LGTM.

miles-on-nightshift:
merge-shepherd: auto-merged (low-risk class: docs/tests/CI/chore; CI green; independent review approved).

Issue #1005, opened 2026-08-24
Enumerate and close every engram write path that bypasses statement sanitization


crtahlin:
Follow-up from the batch review of #942 and #953.

#953's stated justification is that `Plur.learn()` is "the single chokepoint every write path shares, so no path can bypass it". That claim is false as written, and the correction matters because it is the whole argument for fixing #940 there rather than somewhere else.

This issue is to enumerate every path that writes an engram, establish which ones reach a sanitizer, and close the ones that do not.

## Paths confirmed to bypass `learn()`

**1. `learnRouted()` on its remote route — `packages/core/src/index.ts:2978`**

When `_resolveRemoteStoreForScope(scope)` returns a driver, the raw statement goes into `_buildEngramShape(statement, ...)` and is POSTed to the remote store. On failure it is written locally via `_appendEngram`. `learn()` is never entered.

This is the important one. The command-line interface (`packages/cli/src/commands/learn.ts:134`) and the Python software development kit (which shells out to that same interface) both go through `learnRouted`, and a code comment at `index.ts:3111` calls it "the PRIMARY production write path". Those are two of the four paths #952 names as needing the fix.

It is also the highest-impact variant of the attack, because a forged entry on a shared team store reaches other people's system prompts.

Note #942 does cover this, because it sanitizes at the tool-call boundary *before* `learnRouted` is called. So the two open fixes have different coverage, and neither is a superset of the other.

**2. `plur_session_end` — `packages/mcp/src/tools.ts:3090`**

Calls `await plur.learn(statement, { type })` on each item of `engram_suggestions` with no `sanitizeStatement()`. It would be covered by #953 but not by #942, which contradicts #942's claim that its fix "prevents it for every MCP client".

**3. The dedup UPDATE and MERGE paths — `packages/core/src/learn-async.ts:228` and `:263`**

`updated.statement = statement` and `merged.statement = ${merged.statement} ${statement}` both take the raw incoming statement, and neither goes through `learn()`. Reachable from `Plur.learnAsync` and `Plur.learnBatch`.

**4. Pack install and secondary stores** — tracked separately in #1004, since the fix is architecturally different.

## Enumeration of the tool-call write surface, as of #942's head

- `plur_learn` — sanitized
- `plur_learn_batch` — sanitized
- `plur_session_end` — **not sanitized**
- `plur_ingest`, `plur_episode_to_engram` — route through `Plur.learn()`, so covered only if #953 merges
[...]

PR #1212, opened 2026-09-18
fix(mcp): sanitise plur_session_end, the last unsanitised tool-call write

miles-on-nightshift:
## What

Applies `sanitizeStatement()` to every `plur_session_end` engram suggestion before it is passed to `plur.learn()`.

`plur_learn` and `plur_learn_batch` were sanitised by #942. `plur_session_end` was identified as the remaining gap in the same review, but the broader PR was closed in favour of #1108. This is the standalone re-landing of that one change, as promised in #942's closing comment.

## Why this path matters

`sanitizeStatement` does two things that `core`'s `learn()` does not:

1. Cuts at tool-call markers (`</statement>`, `<parameter name=`) — not just line terminators.
2. `plur_session_end` is the write path most likely to carry those markers: the agent is transcribing its own session, and the closure of a tool call is exactly where envelope artifacts show up.

This is not hypothetical — a separate audit of a live store found 52 engrams with leaked tool-call markup in statement and rationale fields.

## Relation to #942

[#942](https://github.com/plur-ai/plur/issues/942) fixed the MCP write surface for `plur_learn` and `plur_learn_batch`. Its closing comment included:

> The `plur_session_end` marker cut will be re-landed as a standalone change.

This is that standalone change. It was split out because the broader sanitisation work in #942 had review concerns unrelated to this specific path.

## Test

`packages/mcp/test/tools.test.ts`: new test `sanitises engram_suggestions written by plur_session_end (#940)` — calls `plur_session_end` with a statement carrying a forged engram boundary (`\n[ENG-FORGED-001]`), asserts the boundary is absent from the stored engram, and carries a vacuity guard so a silent write failure cannot make the assertion pass.

MCP suite: **468 passed, 0 failed**.

## Diff

Two files changed, both in `@plur-ai/mcp`:
- `packages/mcp/src/tools.ts` — wrap statement with `sanitizeStatement()` at the `plur.learn()` call in the session_end handler
- `packages/mcp/test/tools.test.ts` — pin the invariant


crtahlin (dismissed):
Approved. The fix is correct; the test that ships with it is not, and that is worth fixing before merge even though it does not block.

`sanitizeStatement()` now runs on every engram suggestion `plur_session_end` hands to `learn()`, bringing it in line with `plur_learn` and `plur_learn_batch`. It is idempotent and leaves clean statements untouched.

### Should fix

- `packages/mcp/test/tools.test.ts:438-458` — **the new test does not bite.** I reverted the `sanitizeStatement()` call at `packages/mcp/src/tools.ts:3557` and the test still passed. Its payload (`'prefer snake_case in Python\n[ENG-FORGED-001] ...'`) only exercises core's `collapseLineTerminators`, which `_validateLearnInput` calls unconditionally on every `learn()` at `packages/core/src/index.ts:2862` — so `not.toMatch(/\n\[/)` holds with or without this change. The behaviour this pull request actually adds, the cut at `</statement>` and `<parameter name=`, has no test on the session-end path.

  A payload that does bite: `'probe marker case uses camelCase</statement>\n<parameter name="statement">duplicated body</parameter>'` with `not.toContain('</statement>')`. I confirmed that one fails on reverted code and passes on this branch. It is also closer to the corruption the docstring at `packages/mcp/src/tools.ts:505-512` describes.

### Notes

- `packages/mcp/src/tools.ts:3557` — the marker cut exists only in `packages/mcp/src/tools.ts`. Every non-MCP write path still passes raw statements: opencode's auto-harvester (`packages/opencode/src/learn.ts:34`, `:87`), dsh (`packages/dsh/src/learn.ts:144`), claw (`packages/claw/src/context-engine.ts:377`, `packages/claw/src/cli.ts:39`) and the CLI (`packages/cli/src/commands/learn.ts:215`). opencode's harvester is the same shape as `session_end` — an agent transcribing its own output with `claim_class: 'inferred'` — so it can carry the same markup. #1005 already tracks that sweep; `packages/core/src/sanitize.ts` makes the argument for core being the right home.
- The framing is a little stronger than the code warrants. The forged entry boundary from #940 is already defended twice — at write time by `collapseLineTerminators`, and at render time by `entrySafe` in `packages/core/src/inject.ts:970`. What this uniquely adds is removal of duplicated tool-call markup, which is corpus hygiene rather than an injection defence. The inline comment is honest about core also sanitising; the title is not.
- A suggestion consisting entirely of markup now sanitises to empty and is rejected by core with "statement must be a non-empty string", which is misleading since the caller did pass a non-empty string. It surfaces in `engrams_failed` with the raw statement, so nothing is lost silently, and `plur_learn` behaves the same way — consistent rather than new.
- `packages/mcp/src/tools.ts:3510-3527` — the pre-write loop checks `statement.length === 0` before sanitising, so the empty-after-sanitise case escapes the "validate everything before writing anything" guarantee its own comment states, and fails mid-loop instead. Harmless in practice; sanitising inside the validation loop would restore the invariant.
- Two paths were not checked against the "last unsanitised tool-call write" claim: `plur_ingest` (`tools.ts:1966`) and `plur_capture` (`tools.ts:1900`). Filed as #1213.

Worth keeping: the vacuity guard in the new test — asserting the engram was written at all before asserting a negative about its content — is exactly the right instinct, and it is the only assertion in that test that can currently fail.

Tests: PASS — `packages/mcp` 40 files / 468 tests and `packages/core` 230 files / 3757 tests, both green after install and build. Verified by revert that the shipped fix does real work on a marker payload, and that the new test does not.

plur9:
@crtahlin — your "should fix" on the test is done in `06498cbd`. This is test-only; `src/` is untouched.

The payload is now yours: `…camelCase</statement>\n<parameter name="statement">duplicated body</parameter>`. The test asserts that the stored statement contains neither marker, and that it equals exactly the text before the cut. The vacuity guard stays.

Verified by revert. With the `sanitizeStatement()` call at `tools.ts:3557` removed, the test fails with `expected 'probe marker case uses camelCase</sta…' not to contain '</statement>'`. With the call restored, it passes. The full `packages/mcp` suite passes: 40 files, 468 tests.

This push dismissed your approval (`dismiss_stale_reviews`), so a re-approval is needed. Your other notes are not part of this push:
- sanitising inside the validation loop;
- the "last unsanitised" framing in the title;
- the non-MCP write paths (#1005);
- `plur_ingest` and `plur_capture` (#1213).

miles-on-nightshift:
## CTO review — 2026-09-22

Cannot self-approve as miles-on-nightshift. Documenting findings for crtahlin or Gregor to action.

**Finding: APPROVE.** Fix is correct and in the right place.

**Verified:**
- The sanitiseStatement wrapper is added at the right call site — the engram_suggestions loop in plur_session_end. core's learn() collapses line terminators on every write, but does not strip </statement> or <parameter name= markers; sanitizeStatement does.
- Vacuity guard present: the test first asserts the engram was written before asserting its content. Removing the wrapper would cause the expect().not.toContain() assertions to fail.
- R4: reverting the one-line change leaves the </statement> marker in the written statement, failing the test. Confirmed by inspection.
- Sibling sweep: the PR title calls this "the last unsanitised tool-call write"; the existing test at line 435 already covers plur_learn. If the prior sweep was complete, this closes the last gap.

crtahlin (approved):
Approved. The new test bites, and it bites for the right reason.

My earlier review approved the fix but found its regression test worthless: reverting `sanitizeStatement()` left it green, because the payload only exercised core's existing `collapseLineTerminators` rather than the marker cut this change adds. `06498cbd` replaces that payload, and I verified the replacement two ways rather than one.

**Forward mutation.** Reverting only the `sanitizeStatement()` call at `packages/mcp/src/tools.ts:3557` turns the test RED: `expected 'probe marker case uses camelCase</sta…' not to contain '</statement>'`. The received value is `probe marker case uses camelCase</statement> <parameter name="statement">duplicated body</parameter>` — note the embedded newline has already become a space, which proves `collapseLineTerminators` ran and the marker survived it anyway.

**Converse mutation.** Neutralising `collapseLineTerminators` in `packages/core/src/sanitize.ts` and rebuilding core leaves the test GREEN. So the assertion depends on the MCP-side marker cut alone, which is exactly the property the previous payload lacked.

Both mutations were reverted and the worktree verified identical to `06498cbd` before removal.

### One thing to fix at merge time

`06498cbd` carries a `Co-Authored-By:` AI trailer. `CONTRIBUTING.md:183` forbids AI attribution in commits, pull requests and issues outright. GitHub builds a squash message from the commit messages, so a default squash-merge puts that trailer on `main` — **replace the squash message by hand before merging.** Nothing in CI catches this today, which is filed separately as [#1224](https://github.com/plur-ai/plur/issues/1224).

### Notes

- `packages/mcp/src/tools.ts:3557` — `sanitizeStatement` truncates silently and still reports success. By probe: `a legitimate note about <parameter name= syntax in docs` is stored as `a legitimate note about` with `engrams_created: 1` and no warning. Pre-existing behaviour inherited from [#942](https://github.com/plur-ai/plur/pull/942), but this change extends it to the highest-volume write path. Filed as [#1223](https://github.com/plur-ai/plur/issues/1223).
- `packages/mcp/src/tools.ts:3510-3527` — unchanged here and still true: the pre-write loop checks `statement.length === 0` before sanitising, so a statement that is empty only *after* sanitising escapes the loop's own stated "validate everything before writing anything" guarantee and fails mid-loop with a message blaming the caller for passing an empty string. Covered by [#1223](https://github.com/plur-ai/plur/issues/1223).
- Neither [#1213](https://github.com/plur-ai/plur/issues/1213) nor [#1005](https://github.com/plur-ai/plur/issues/1005) is changed by this commit. One addition worth attaching to #1213: this handler's own `plur.capture(summary)` at `tools.ts:3537` passes the raw summary, the same episode-summary gap on a different line.
- `packages/mcp/test/tools.test.ts:452` — the cut lands on `</statement>` because it appears first, so `<parameter name=` is never exercised as the leading marker on this path. Low value to add separately, since `sanitizeStatement` takes the minimum index across both markers in one pass.
- The title "the last unsanitised tool-call write" is still stronger than the code warrants while #1213's two paths are open, but the renamed test is accurate about what it proves.

The comment at `tools.test.ts:441-448` explaining *why* the previous payload could not catch a regression is the right thing to leave behind — it records the reasoning at the point where a future edit would weaken it again. Keeping the vacuity guard through the rewrite, and adding the exact-equality assertion at `:466` so an over-cutting sanitiser would also fail, are both improvements on what I asked for.

Tests: PASS — `packages/mcp`, 468 tests across 40 files green in a throwaway worktree; CI green on Node 20, 22, 24 and 26.

Issue #681, opened 2026-07-24
domain_hint follow-ups: batch gate parity + document the domain convention in CLAUDE.md


crtahlin:
Two non-blocking follow-ups from the review of [#677](https://github.com/plur-ai/plur/pull/677) (missing-domain nudge, part of [#671](https://github.com/plur-ai/plur/issues/671)).

## 1. Batch gate isn't at full parity with the single-item gate

The single `plur_learn` and CLI paths gate the `domain_hint` on all four conditions, including **the write did not auto-route** (`_routed` / `wasRouted`). The batch path (`packages/mcp/src/tools.ts`, `plur_learn_batch`) computes `noDomainCount` from *no domain + no explicit scope* only — it omits the per-item routed check:

```ts
const noDomainCount = raw.filter(e =>
  !(typeof e.domain === 'string' && e.domain.length > 0) &&
  !(typeof e.scope === 'string' && e.scope.length > 0)).length
```

A no-domain / no-scope item that nonetheless **auto-routed** (possible: 3 matching tags reach exactly 0.5 per the ranker math) would be counted in the aggregate hint even though the single-item path would have stayed silent for it. Advisory-only and rare (no-domain items rarely clear the threshold), so benign — but for exact parity, filter batch items by their result's `_routed` before counting.

**Acceptance:** batch `noDomainCount` excludes items whose result carries `structured_data._routed`; add a test with a batch item that auto-routes without a domain and assert it is not counted.

## 2. Document the domain convention the hint points to

The MCP hint text says *"see the domain convention in CLAUDE.md"*, but CLAUDE.md (line 226) currently documents only the `scope` field — the `plur.<team>.<area>` domain shape is not written there. The hint is still actionable (it gives the shape inline, e.g. `plur.engineering.mcp`), but the pointer is currently dangling.

This is acceptance-criterion 1 of #671 ("a documented domain convention exists, aligned with the scopes' covers[]"). Either add the convention section to CLAUDE.md (engineering→`plur.engineering.*`, comms→`plur.comms.*`, research→`plur.research.*`, leadership→`plur.leadership.*`, org→`plur.company.*`/`plur.org.*`) or soften the hint's pointer.

**Acceptance:** CLAUDE.md documents the `plur.<team>.<area>` domain convention; the hint's reference resolves to it.

Refs: #677, #671.

plur9:
[...]

PR #1219, opened 2026-09-21
fix(core,cli,mcp): never auto-route an unscoped write into a shared scope

### LINKED ISSUE #1115: Writes with no explicit scope silently auto-route to a shared/remote store — require a scope, or never auto-route to a shared scope

crtahlin:
Closes #1115.

## Problem

An unscoped write was auto-routed wherever a registered scope's `covers` matched its domain prefix — shared scopes included. A personal engram could land in a team store and be pushed to that store's remote, announced only by an `info` field in the tool result. Once the copy was remote, local cleanup could not undo it.

The decision was also narrower than it looked. `_resolveUnscopedScope` routed a clean forward domain-prefix match deterministically, bypassing the confidence gate and ignoring tags and statement entirely, while `plur_suggest_scope` ranked with all three channels. Consulting the suggestion tool and then writing without a scope could give two different answers.

This is why raising `scope_routing.match_threshold` did not help: the forward-domain path returns before the threshold is consulted, so the only setting that looked like a fix gated the wrong branch.

## Fix

`decideAutoRoute` in `scope-routing.ts` is now the one decision, used by the write path and by a new `Plur.previewAutoRoute`, which `plur_suggest_scope` reports as `would_route`. The two cannot drift apart again, because there is only one decision left to drift.

Eligibility is unchanged: a forward domain match routes deterministically, everything weaker stays gated by `match_threshold`. What changed is that **a shared candidate is refused** — on both eligibility branches, so the deterministic path that actually leaked is covered, not only the scored one.

The refusal does not end the search, so the next eligible personal candidate still wins and routing among `local` / `global` / `user:*` / `agent:*` is untouched, where a wrong guess costs nothing a rescope cannot fix. Explicit scopes are untouched.

`scope_routing.allow_shared_auto_route` (default `false`) restores the previous behaviour for an install that genuinely wants covers-driven team routing — deliberately and in writing, rather than by accident.

## The refusal is not silent

It travels as `_routeRefused` beside the existing `_routed` marker and surfaces in `plur_learn` as a `warning` naming the scope it declined and how to choose it deliberately — rather than another `info` field of the kind the issue reports as easy to miss in a long session.

## Tests

The existing routing suites targeted `group:` scopes throughout, which encoded the reported bug as correct behaviour. Those cases are re-pointed at personal scopes so they keep testing the ranker; left on shared scopes their negative cases would have passed because of the refusal rather than the mechanism they were written for.

Two demotion cases whose premise was "auto routed into a shared scope" now reach the guard the only way still possible, an explicit shared scope, and the guard itself is unchanged.

`routing-fuzz`'s independent oracle runs every case against both a shared and a personal target: against shared nothing may route and a refusal marker must be present; against personal the original oracle still holds. Without that second target its no-under-route invariant would have gone vacuous.

Eight new cases cover the contract, including that `previewAutoRoute` and the write path agree.

**Verified on this branch, rebased onto `main` @ e26901c1:** full suite 5671 passed, 1 expected fail, 0 test failures. The 7 failing test files are all `@plur-ai/opencode` failing to resolve `@plur-ai/core`, which reproduces unchanged on `main` — a local dependency-install gap, not this change. Disabling the refusal turns exactly five of the new cases red, including the fuzz oracle and the preview-agreement case.

## Known gaps, tracked separately

- `plur_learn_batch` and the CLI report nothing at all when a write is auto-routed — `_routed` is read only to suppress a domain hint. Follow-up on this branch.
- The write carries no indication on the wire of whether its scope was chosen by the caller or by the router, so a server cannot distinguish a deliberate team write from a routed one. Filed separately; it is what a server-side mirror of this issue needs.


crtahlin:
## Pushed fbf4907f — the two surfaces that reported nothing

The PR description listed these as follow-ups on this branch. They are now in it.

**`plur_learn_batch` echoed neither decision.** `_routed` was read only to decide whether to show a domain hint, and `_routeRefused` was read nowhere — so a batch write that landed in a store the caller never named, or was declined from a shared one, reached the caller with no signal of either. The gap was already visible in the suite: the #681 parity test has to read `structured_data._routed` directly, and its comment records that a first draft asserted on the response and "silently measured nothing". Each result now carries `routed` / `route_refused`.

A key on item 34 of 50 is not a signal, so refusals are summarised once at the top level too. `warning` was a single string owned by the failure path; the two reasons to warn are now composed rather than one replacing the other, which a test pins.

**The CLI had the same shape** — both markers read only to suppress a hint, so a write command could not say where the engram went. The refusal goes through `outputText` beside the existing demotion warning, because it changes what the user should do next and must survive `--quiet` for the same reason.

## One thing for whoever cuts the next release

The CHANGELOG entry for this fix was written before 0.20.1 was cut, and the rebase left it under a heading that has now shipped — which would claim the fix is in a release that does not contain it. I moved it to an `## Unreleased` heading rather than pick a version number. `scripts/release.sh` extracts sections with `awk '$0 ~ "^## "v{p=1} /^## [0-9]/{...}'`, so a non-numeric heading is inert; I confirmed the 0.20.1 extraction is unchanged. Rename it when you open the next version section.

## Tests

Four MCP cases and two CLI cases. Verified against the unpatched sources: three of the four MCP cases and both CLI cases fail without the change. The fourth covers `plur_learn`'s own refusal response, which had no test at all — it passes either way and is coverage, not a regression guard.

The CLI's human-readable branch is TTY-only (`shouldOutputJson` returns `!isTTY()` and there is no `--no-json`), so it is left to inspection exactly as the neighbouring demotion warning is.

Full suite: **5677 passed, 1 expected fail, 0 test failures.** The seven failing test files are all `@plur-ai/opencode` failing to resolve `@plur-ai/core`; that reproduces unchanged on `main` and is a local dependency-install gap, not this branch.

miles-on-nightshift (approved):
## Approved

Reviewed , ,  (MCP),  (CLI), and all affected test suites.

### What I verified

**The decision is now one function.**  in  is called by  (write path) and by  (suggestion surface). The drift #1115 reports — consulting the suggestion tool and landing somewhere else — is closed structurally, not by synchronising two independent implementations.

**The shared-scope refusal is on both eligibility branches.** The deterministic forward-domain bypass was the path that actually leaked; the threshold-gated path was secondary. The guard in `decideAutoRoute` applies in the `for` loop before the return, so both branches hit it regardless of how the candidate was ranked. Confirmed by tracing the loop: `coverContainsDomain` candidates cleared first, then the confidence gate — shared refusal runs after both.

**The refusal does not stop the search.** `refusedShared` is recorded and the loop continues. A personal `user:*` scope declaring matching `covers` still wins. The fuzz suite covers this: `mkStore(PERSONAL_SCOPE)` with the same covers succeeds where `mkStore(SHARED_SCOPE)` is refused.

**Test quality.** The routing-fuzz independent oracle now runs two passes per case — shared (always refused) and personal (original oracle) — so the no-under-route invariant is not vacuous after the fix. The guard-fuzz demotion case was correctly moved to an explicit scope (since auto-route into shared no longer happens), preserving the demotion test without encoding the bug. The MCP batch suite covers the multi-warning case — failure + refusal — confirming neither displaces the other.

**Sibling sweep checked.** , , CLI , and  all report both routing outcomes.  carries  on the wire (#1221). / operate on an existing engram's scope and do not call ; they are correctly excluded.

**One observation, not blocking:**  reads  from the engram. The stamping site ( or ) is in the middle section of  that was not in the diff I read — CI passes, so it must be there. Worth a quick grep when you land this to confirm the read and the write are in the same field name.

miles-on-nightshift (approved):
## Approved

Reviewed scope-routing.ts, index.ts, tools.ts (MCP), learn.ts (CLI), and all affected test suites.

### What I verified

**The decision is now one function.** decideAutoRoute in scope-routing.ts is called by _resolveUnscopedScope (write path) and by previewAutoRoute (suggestion surface). The drift #1115 reports — consulting the suggestion tool and landing somewhere else — is closed structurally, not by synchronising two independent implementations.

**The shared-scope refusal is on both eligibility branches.** The deterministic forward-domain bypass was the path that actually leaked; the threshold-gated path was secondary. The guard in decideAutoRoute applies in the for loop before the return, so both branches hit it regardless of how the candidate was ranked. Confirmed by tracing the loop: coverContainsDomain candidates cleared first, then the confidence gate — shared refusal runs after both.

**The refusal does not stop the search.** refusedShared is recorded and the loop continues. A personal user:* scope declaring matching covers still wins. The fuzz suite covers this: mkStore(PERSONAL_SCOPE) with the same covers succeeds where mkStore(SHARED_SCOPE) is refused.

**Test quality.** The routing-fuzz independent oracle now runs two passes per case — shared (always refused) and personal (original oracle) — so the no-under-route invariant is not vacuous after the fix. The guard-fuzz demotion case was correctly moved to an explicit scope (since auto-route into shared no longer happens), preserving the demotion test without encoding the bug. The MCP batch suite covers the multi-warning case — failure + refusal — confirming neither displaces the other.

**Sibling sweep checked.** plur_learn, plur_learn_batch, CLI learn.ts, and plur_suggest_scope all report both routing outcomes. remote-store.ts carries scope_source on the wire (#1221). setPinned/setPinnedAsync operate on an existing engram's scope and do not call _resolveUnscopedScope; they are correctly excluded.

**One observation, not blocking:** remote-store.ts reads e.structured_data?._scopeSource from the engram. The stamping site (_buildEngramShape or learnRouted) is in the middle section of index.ts that was not visible in the diff — CI passes, so it is there. Worth a quick grep when you land this to confirm the read and write use the same field name.

